PR #385, opened 2020-08-13
Add parameters to Git LFS

raychu86:
## Motivation

This PR is motivated by #341 

The `inner_snark_pk.params`, `outer_snark_pk.params`, and `posw_snark_pk.params` are now stored via Git LFS. When you git clone the repository or checkout a branch, it will fetch the large parameter files from LFS. 

This is currently a Draft PR because this requires a bit more testing and making sure that Git LFS works well with the already existing `pre-push` git hook. Additionally, the logic for remote parameter fetching from S3 has not been removed.

The CI tests fails because running the tests currently does not fetch the parameters from LFS (need to add a command to properly fetch the files). Alternatively, we may need to use the remote logic as a failsafe/backup.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,3 @@
+*inner_snark_pk.params filter=lfs diff=lfs merge=lfs -text
+*outer_snark_pk.params filter=lfs diff=lfs merge=lfs -text
+*posw_snark_pk.params filter=lfs diff=lfs merge=lfs -text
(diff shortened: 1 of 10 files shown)


howardwu:
After evaluating this further, using Git LFS requires each user to run specific commands when building from source to build LFS, whereas AWS is a simple download. As such, we will hold off on using it for now.

Issue #413, opened 2020-08-23
[Bug] Node stopped syncing after an error.


kobigurk:
## 🐛 Bug Report

It seems like there was a panic in the message handler, after which the node stopped outputting any log messages, which I assume means it stopped syncing.

These are the last messages:
```
thread 'tokio-runtime-worker' panicked at 'message handler errored: RecvError(())', network/src/server/server.rs:164:27
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
[2020-08-22T02:47:19Z INFO  snarkos_network::message::channel] Message "ping", Sent to V4(52.42.131.71:4131)
[2020-08-22T05:56:57Z ERROR jsonrpc_http_server] Error serving connection: Error(Parse(Method))


[2020-08-22T15:52:15Z ERROR jsonrpc_http_server] Error serving connection: Error(Parse(Method))
```

Restarting the node continued the sync.

## Your Environment

- Commit 2b2f6f937e3f7da42f4b942bda9bffd8174983ad
- Rust 1.45.2
- Ubuntu 20.04

PR #417, opened 2020-08-26
Update connection and sync logic.

raychu86:
## Motivation

This PR adds bug fixes for handling locks and adds a simple check for pending blocks during sync.


howardwu:
#413

PR #424, opened 2020-08-30
feat(polycommit): support an upper bound on hiding bounds

kobigurk:
This is done so we can use an SRS that has a limited amount of elements of the form `alpha*tau^i`.

### DIFF
--- a/marlin/src/lib.rs
+++ b/marlin/src/lib.rs
@@ -136,8 +136,11 @@ impl<F: PrimeField, PC: PolynomialCommitment<F>, D: Digest> Marlin<F, PC, D> {
         }
 
         let coeff_support = AHPForR1CS::get_degree_bounds::<C>(&index.index_info);
+        // Marlin only needs degree 2 random polynomials
+        let supported_hiding_bound = 2;
         let (committer_key, verifier_key) =
-            PC::trim(&srs, index.max_degree(), Some(&coeff_support)).map_err(Error::from_pc_err)?;
+            PC::trim(&srs, index.max_degree(), supported_hiding_bound, Some(&coeff_support))
+                .map_err(Error::from_pc_err)?;
 
         let commit_time = start_timer!(|| "Commit to index polynomials");
         let (index_comms, index_comm_rands): (_, _) =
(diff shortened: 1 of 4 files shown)


marlin/src/lib.rs
```diff
@@ -136,8 +136,11 @@ impl<F: PrimeField, PC: PolynomialCommitment<F>, D: Digest> Marlin<F, PC, D> {
         }
 
         let coeff_support = AHPForR1CS::get_degree_bounds::<C>(&index.index_info);
+        // Marlin only needs degree 2 random polynomials
+        let supported_hiding_bound = 2;
```
Pratyush, line 140: degree of the random polynomial is 2, but the hiding bound is 1; the translation should be handled at the PC layer, no?

polycommit/src/marlin_pc/mod.rs
```diff
@@ -204,7 +202,7 @@ impl<E: PairingEngine> PolynomialCommitment<E::Fr> for MarlinKZG10<E> {
         let powers = pp.powers_of_g[..=supported_degree].to_vec();
         // We want to support making up to supported_degree queries to committed
         // polynomials.
-        let powers_of_gamma_g = pp.powers_of_gamma_g[..=(supported_degree + 1)].to_vec();
+        let powers_of_gamma_g = pp.powers_of_gamma_g[..=supported_hiding_bound].to_vec();
```
Pratyush, line 205: To align with the other comment: 
```suggestion
        let powers_of_gamma_g = pp.powers_of_gamma_g[..=(supported_hiding_bound + 1)].to_vec();
```

polycommit/src/marlin_pc/mod.rs
```diff
@@ -204,7 +202,7 @@ impl<E: PairingEngine> PolynomialCommitment<E::Fr> for MarlinKZG10<E> {
         let powers = pp.powers_of_g[..=supported_degree].to_vec();
         // We want to support making up to supported_degree queries to committed
```
Pratyush, line 203:
```suggestion
        // We want to support making up to `supported_hiding_bound` queries to committed
```

marlin/src/lib.rs
kobigurk: Good point

Pratyush:
ok this LGTM!

PR #425, opened 2020-08-30
feat(polycommit): introduces SonicKZG10 to polycommit

kobigurk:
SonicKZG10 uses extra log(n) elements in G2 to lower the proof size and prover time by about 10%.


Pratyush:
Is this mostly a port of the upstream poly-commit?

kobigurk:
Yep, pretty much. Just the hiding bounds parts has been introduced. So it
doesn't require a deep review :)

PR #465, opened 2020-09-20
[Tracking] snarkOS networking upgrade - Server, Blocks, Peers, Inbound, Outbound

howardwu:
## Motivation

snarkOS networking upgrade - Server, Blocks, Peers, Inbound, Outbound


metrics/src/prometheus.rs
```diff
@@ -0,0 +1,89 @@
...
+        .register(Box::new(RPC_REQUESTS.clone()))
+        .expect("RPC_REQUESTS to be registered");
+}
+
+pub async fn metrics_handler() -> Result<impl Reply, Rejection> {
+    let encoder = prometheus::TextEncoder::new();
+
+    // Encode our metrics into a response.
+    let mut res = {
+        let mut buffer = Vec::new();
```
ljedrz, line 55: `buffer` could be created in the main block of `metrics_handler`, removing the need for extra allocations incurred when it's created anew and re-populated in the `res_prom` block; it suffices that it is `clear()`ed between the 2 blocks (and it doesn't need to be cleared the second time in `res_prom`)

metrics/src/prometheus.rs
```diff
@@ -0,0 +1,89 @@
...
+        res
+    };
+
+    // Encode Prometheus' metrics into a response.
+    let res_prom = {
+        let mut buffer = Vec::new();
+        if let Err(e) = encoder.encode(&prometheus::gather(), &mut buffer) {
+            eprintln!("could not encode prometheus metrics: {}", e);
+        };
+        let res_prom = match String::from_utf8(buffer.clone()) {
```
ljedrz, line 76: this `clone()` is not needed; the `buffer` gets consumed here and doesn't need to be used later

network/src/connection_manager.rs
```diff
@@ -0,0 +1,404 @@
...
+    },
+    RequestManager,
+};
+use snarkos_consensus::MerkleTreeLedger;
+
+use chrono::Utc;
+use std::{collections::HashMap, net::SocketAddr, sync::Arc};
+use tokio::sync::RwLock;
+
+#[derive(Clone)]
```
ljedrz, line 36: it might be undesirable for `ConnectionManager` to derive `Clone`, as it contains a `HashMap` and `Vec` that will get deep-cloned, incurring extra allocations; cloned instances could also become out-of-sync wrt. those two fields; it might be desirable, just making sure

network/src/connection_manager.rs
```diff
@@ -0,0 +1,404 @@
...
+
+    /// Returns the local address of the node.
+    #[inline]
+    pub async fn get_local_address(&self) -> SocketAddr {
+        // Acquire the peer book reader.
+        let peer_book = self.peer_book.read().await;
+        // Get the local address of the node.
+        let local_address = peer_book.local_address().clone();
+        // Drop the peer book reader.
+        drop(peer_book);
```
ljedrz, line 134: locks don't need to be explicitly dropped if it's the end of the block they've been acquired in - they are dropped automatically; the same applies to a few other spots later in the diff

network/src/connection_manager.rs
```diff
@@ -0,0 +1,404 @@
...
+            self.connect_to_bootnodes().await;
+        }
+
+        // TODO (howardwu): Unify `Ping` and `Version`requests.
+        //  This is a remnant and these currently do not need to be distinct.
+
+        // Broadcast a `Ping` request to each connected peer.
+        self.broadcast_ping_requests().await;
+        // Broadcast a `Version` request to each connected peer.
+        self.broadcast_version_requests().await;
```
ljedrz, line 245: some of these methods can probably be run concurrently (as opposed to halting the current task with subsequent `await`s) with `tokio::join`; in addition, if any of them are expected to be executed in tandem, like e.g. `broadcast_ping_requests` and `broadcast_version_requests`, they could reuse some of the results of their internal calls, e.g. to `get_all_connected`; this could be achieved with merging them into a single method or by passing the common, previously obtained values to both

network/src/connection_manager.rs
```diff
@@ -0,0 +1,404 @@
...
+    }
+
+    /// Broadcasts a `Version` message to all connected peers.
+    #[inline]
+    async fn broadcast_version_requests(&self) {
+        // Get the local address of the node.
+        let local_address = self.get_local_address().await;
+
+        // Broadcast a `Version` message to each connected peer for a periodic sync.
+        if self.get_num_connected().await > 0 {
```
ljedrz, line 271: this check feels unnecessary, since the loop below won't do anything if `get_all_connected` is empty (implying the result of `get_num_connected` being `0`)

network/src/external/protocol/ping/ping_pong_worker.rs
```diff
@@ -34,43 +34,38 @@ pub enum PingState {
...
     state: PingState,
-    channel: Arc<Channel>,
     nonce: u64,
 }
 
-impl PingProtocol {
+impl PingPongWorker {
     /// Send the initial ping message to a peer.
-    pub async fn send(channel: Arc<Channel>) -> Result<Self, PingProtocolError> {
+    pub async fn send(channel: &Arc<Channel>) -> Result<Self, PingProtocolError> {
```
ljedrz, line 44: `Arc<T>` implements `Deref<Target = T>`, so if it doesn't need to be cloned, just `&T` (in this case `&Channel`) can be put as an argument

network/src/internal/message_handler.rs
```diff
@@ -295,12 +304,26 @@ impl Server {
...
+        // and include the peer if it is new.
+        let mut peer_book = self.context.peer_book.write().await;
+        if !peer_book.is_connected(&channel.address) {
+            peer_book.found_peer(&channel.address);
+        }
 
-        channel.write(&Peers::new(connected_peers)).await?;
+        // Broadcast the sanitized list of connected peers back to requesting peer.
+        let mut peers = HashMap::new();
+        for (remote_address, peer_info) in peer_book.get_all_connected().iter() {
```
ljedrz, line 318: if it is expected that there can be many simultaneously connected peers, the call to `get_all_connected` could be done earlier, the returned value cloned, and the `peer_book` lock manually `drop()`ped right afterwards in order not to keep it open for too long

network/src/internal/message_handler.rs
```diff
@@ -309,58 +332,79 @@ impl Server {
...
-            } else if peer_book.connected_contains(addr) {
-                peer_book.update_connected(addr.clone(), time.clone());
-            } else {
-                peer_book.update_gossiped(addr.clone(), time.clone());
+            }
+            // Inform the peer book that we found a peer.
+            // The peer book will determine if we have seen the peer before,
+            // and include the peer if it is new.
+            else if !peer_book.is_connected(&channel.address) {
+                peer_book.found_peer(&channel.address);
```
ljedrz, line 361: this has already been done at lines 340-342; also, it doesn't seem to benefit from being done in a loop - `channel.address` and `peer_book` don't seem to change in the meantime; removing this would also allow the `peer_book` lock to be dropped earlier, e.g. at like 343

network/src/internal/peer_book.rs
```diff
@@ -0,0 +1,267 @@
...
+use snarkos_metrics::Metrics;
+use snarkos_models::{algorithms::LoadableMerkleParameters, objects::Transaction};
+use snarkos_storage::Ledger;
+
+use serde::{Deserialize, Serialize};
+use std::{collections::HashMap, net::SocketAddr};
+
+/// An data structure for tracking and indexing the history of
+/// all connected and disconnected peers to the node.
+#[derive(Clone, Debug, Serialize, Deserialize)]
```
ljedrz, line 28: as before, deriving `Clone` on objects containing collections is usually undesirable due to deep-cloning; if `PeerBook` is expected to be wrapped in an `Arc`, it doesn't need to be `Clone`able - the `Arc` can still be cloned

network/src/internal/peer_book.rs
```diff
@@ -0,0 +1,267 @@
...
+        }
+    }
+
+    /// Add the given address to the disconnected peers in the `PeerBook`.
+    /// Returns `true` on success. Otherwise, returns `false`.
+    #[inline]
+    pub fn found_peer(&mut self, address: &SocketAddr) -> bool {
+        if self.local_address() == *address {
+            // Case 1: The peer is our node.
+            return false;
```
ljedrz, line 199: in order to avoid a future `clippy` complaint, the `return` can be removed here

network/src/internal/peer_book.rs
```diff
@@ -0,0 +1,267 @@
...
+            _ => None,
+        }
+    }
+
+    // TODO (howardwu): Implement manual serializers and deserializers to prevent forward breakage
+    //  when the PeerBook or PeerInfo struct fields change.
+    /// Serializes and writes the `PeerBook` to storage.
+    #[inline]
+    pub fn store<T: Transaction, P: LoadableMerkleParameters>(
+        &mut self,
```
ljedrz, line 253: the mutable borrow here doesn't seem to be needed

network/src/request_manager.rs
```diff
@@ -0,0 +1,379 @@
...
+    /// A counter for the number of send requests that succeeded.
+    send_success_count: Arc<AtomicU64>,
+    /// A counter for the number of send requests that failed.
+    send_failure_count: Arc<AtomicU64>,
+}
+
+impl RequestManager {
+    /// Creates a new instance of a `ConnectionManager`.
+    #[inline]
+    pub fn new() -> Self {
```
ljedrz, line 46: `RequestManager` should instead just `#[derive(Default)]`, which would create an object with the same values as listed below

network/src/request_manager.rs
```diff
@@ -0,0 +1,379 @@
...
+            };
+            // Drop the handshake write lock.
+            drop(handshakes);
+
+            // Acquire the pending addresses write lock.
+            let mut pending_peers = pending_addresses.write().await;
+            // Remove the remote address from the pending addresses.
+            pending_peers.remove(&remote_address);
+            // Drop the pending addresses write lock.
+            drop(pending_peers);
```
ljedrz, line 188: this `drop` is automatic

network/src/server.rs
```diff
@@ -72,37 +77,60 @@ impl Server {
...
         }
     }
 
+    /// Returns the default bootnode addresses of the network.
+    pub fn get_bootnodes(&self) -> Vec<SocketAddr> {
+        // Initialize the vector to be returned.
+        let mut bootnode_addresses = Vec::with_capacity(self.context.bootnodes.len());
+        // Iterate through and parse the list of bootnode addresses.
+        for bootnode in self.context.bootnodes.iter() {
+            if let Ok(bootnode_address) = bootnode.parse::<SocketAddr>() {
```
ljedrz, line 90: I'm assuming the address was already validated when it was added to `context.bootnodes`, but there's no harm in pattern-matching here :+1:

PR #472, opened 2020-09-23
toolkit wasm pack release build  bug fix

sarawut11:
Toolkit wasm release build has following issue. This PR fixes it.

[INFO]: 🎯  Checking for the Wasm target...
[INFO]: 🌀  Compiling to Wasm...
    Finished release [optimized] target(s) in 0.10s
[INFO]: ⬇️  Installing wasm-bindgen...
[INFO]: Optimizing wasm binaries with `wasm-opt`...
[wasm-validator error in module] unexpected true: Exported global cannot be mutable, on 
global$0
(module
 (type $i32_i32_=>_i32 (func (param i32 i32) (result i32)))
 (type $i32_=>_none (func (param i32)))
 (type $i32_i32_=>_none (func (param i32 i32)))
 (type $i32_i32_i32_=>_none (func (param i32 i32 i32)))
 (type $i32_i32_i32_=>_i32 (func (param i32 i32 i32) (result i32)))
 (type $i32_i32_i32_i32_=>_none (func (param i32 i32 i32 i32)))
 (type $i32_=>_i32 (func (param i32) (result i32)))
 (type $i32_i64_i64_i64_i64_i64_i64_i64_i64_=>_none (func (param i32 i64 i64 i64 i64 i64 i64 i64 i64)))
 (type $i32_i32_i32_i32_i32_=>_none (func (param i32 i32 i32 i32 i32)))
 (type $i32_=>_i64 (func (param i32) (result i64)))
 (type $none_=>_none (func))
...

### DIFF
--- a/toolkit/Cargo.toml
+++ b/toolkit/Cargo.toml
@@ -35,3 +35,6 @@ wasm-bindgen = { version = "0.2.68" }
 criterion = { version = "0.3.1" }
 rand_chacha = { version = "0.2", default-features = false }
 wasm-bindgen-test = { version = "0.3.18" }
+
+[package.metadata.wasm-pack.profile.release]
+wasm-opt = false


kobigurk:
I'm not sure that removing `wasm-opt` is the best choice. @kellpossible any idea?

kellpossible:
From memory `wasm-opt` strips the debug info and other stuff, and can make the wasm binary a lot smaller. Would be worth measuring the difference before deciding to make this change.
.

kobigurk:
> From memory `wasm-opt` strips the debug info and other stuff, and can make the wasm binary a lot smaller for release. Would be worth measuring the difference before deciding to make this change.
> .

Any idea what could be the cause of the error @sarawut11 saw?

PR #473, opened 2020-09-24
Clippy compliance, part 1

ljedrz:
This is the first round of adjustments suggested by `clippy`, paving the way towards enabling its lints in the CI or githooks. All the specific changes are confined to individual commits.

Some of the suggestions were silenced with `#[allow(clippy::x)]`, with `x` being:
- `module_inception` (not very important and adjusting would incur a lot of diff "traffic")
- `many_single_char_names` (fine in complex calculations, especially when following scientific names)
- `too_many_arguments` (better to treat those individually; they will be easy to find with the new attribute)
- `suspicious_arithmetic_impl` / `suspicious_op_assign_impl` (ditto)

Cc https://github.com/AleoHQ/snarkOS/issues/145

### DIFF
--- a/algorithms/src/commitment_tree/mod.rs
+++ b/algorithms/src/commitment_tree/mod.rs
@@ -14,6 +14,8 @@
 // You should have received a copy of the GNU General Public License
 // along with the snarkOS library. If not, see <https://www.gnu.org/licenses/>.
 
+#![allow(clippy::module_inception)]
+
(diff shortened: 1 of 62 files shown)


raychu86:
This is great @ljedrz! Removing a great number of unnecessary clones and conversions is definitely important and also makes the code much easier to read!

howardwu (approved):
LGTM

Issue #471, opened 2020-09-23
[Feature] Toolkit Signatures using Account View Key and Account Address.


raychu86:
## 🚀 Feature

The current toolkit signatures currently use the `Account Private Key` to sign and `Signature Public Key` to verify Schnorr signatures. 

We need to add to functionality to sign messages with the `Account View Key` and verify the signatures with the `Account Address`. 

## Motivation

The motivation of this feature is to allow users to sign messages that can be verified with a user's public address.

## Implementation

The `Account View Key` and `Account Address` are a `Group EncrytionScheme` private and public key-pair. The `GroupEncryption` parameters share the same sized `generator_powers` field, however the `GroupEncryption` scheme is missing a salt field. 

A possible solution is to just add an additional salt field that will only be used for message signing. The `GroupEncryption` scheme will then also implement the `SignatureScheme` and call into the Schnorr signature algorithms directly. This will not require parameter generation (address derivation will remain unchanged), because we are just adding an additional 32 bytes of randomness to the end of the current `EncryptionScheme` parameters.

ljedrz:
I think this issue needs to be transferred to `AleoHQ/aleo`, but I'm not sure if that repo is configured to obtain issue transfers from snarkOS - at least I don't have that option under `Transfer issue`.

PR #476, opened 2020-09-25
Update signature model in toolkit

raychu86:
## Motivation

This PR is motivated by issue #471 

This PR updates the signature model in `snarkos_toolkit` to be baked into the actual `Account`, `ViewKey` and `Address` structs instead of standalone Signature objects. Additionally, now there are two different types of signatures.

1. Schnorr Signature where:
  The Account Private Key `sk_sig` is the signature private key.
  The Account Private Key `pk_sig` is the signature public key.

2. Schnorr Signature where:
  The Account View Key is the signature private key.
  The Account Address is the signature public key.

## Test Plan

Both the native rust and the wasm_bindgen tests have been updated in `snarkos_toolkit` to reflect the new changes.

## Related PRs

This PR is built off of PR #475.


raychu86:
Migrated into snarkVM and the Aleo SDK.

Issue #145, opened 2020-05-24
Setup clippy


howardwu:

ljedrz:
We already have `clippy` in the CI, so this issue can be closed.

PR #477, opened 2020-09-25
Clippy compliance, part 2

ljedrz:
Builds on https://github.com/AleoHQ/snarkOS/pull/473; can be merged in sequence or be the only one merged. While this PR makes the simple `cargo clippy` pass already, there's still a PR for `cargo clippy --workspace --examples --tests` left to go.

There are a few new lints handled this time, and as before, I tried to put them all in dedicated commits; the exception is the second-to-last `clippy: miscellaneous` commit, where the kinds of lints already handled in the previous PR were included (and perhaps a few cases from this run slipped through) - I could split them further, but since most of them are quite straightforward (redundant conversions, unnecessary `clone()`s etc.), I decided to let them be.

As before, this initiative serves to allow the repo to enable `clippy` in the CI or as a git hook; even though a lot of the lints are just readability improvements, a fair share have performance implications and even logic/security ones, as demonstrated by https://github.com/AleoHQ/snarkOS/commit/d1cdc2d14d809e29b46bcf85602eecd2e522277d.

Even though these changes are very prone to bitrot due to 145 files being changed, rebasing them shouldn't be too difficult, as most individual changes are limited to 1~4 lines; the same goes for potentially rebasing other changes after these changes go in - I've made sure to limit the diff to minimum by using `#[allow(clippy::x)]` in the more "cosmetic" cases.

Some range-/iterator-related changes and type aliases have received names chosen on a best-effort basis, but I'll be happy to adjust any and all of them if need be.

Cc #145

### DIFF
--- a/algorithms/src/commitment_tree/mod.rs
+++ b/algorithms/src/commitment_tree/mod.rs
@@ -14,6 +14,8 @@
 // You should have received a copy of the GNU General Public License
 // along with the snarkOS library. If not, see <https://www.gnu.org/licenses/>.
 
+#![allow(clippy::module_inception)]
+
 pub mod commitment_path;
 pub use commitment_path::*;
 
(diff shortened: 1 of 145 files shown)


howardwu (approved):
LGTM

PR #478, opened 2020-09-28
Clippy compliance, part 3 (final)

ljedrz:
Builds on https://github.com/AleoHQ/snarkOS/pull/477 (which builds on https://github.com/AleoHQ/snarkOS/pull/473); can be merged in sequence or be the only one merged.

Most of the changes this time around were similar to the ones in the previous PRs, with 2 notable exceptions which were put in separate commits:
- changing a common closure in `gadgets::utilities::boolean` to a free function (prompted by `clippy`, but useful on its own)
- two `clippy::eval_order_dependence` fixes

This PR concludes the `clippy` compliance series, making `cargo clippy --workspace --examples --tests` pass without warnings, allowing the repo to be configured so that either a CI run or a git hook includes a `clippy` pass.

Cc #145

### DIFF
--- a/errors/src/storage/storage.rs
+++ b/errors/src/storage/storage.rs
@@ -20,11 +20,7 @@ use crate::{
     parameters::ParametersError,
 };
 
-use bincode;
 use std::fmt::Debug;
-
-#[cfg(feature = "librocksdb")]
-use rocksdb;
 
 #[derive(Debug, Error)]
 pub enum StorageError {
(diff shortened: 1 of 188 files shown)


errors/src/storage/storage.rs
```diff
@@ -20,12 +20,8 @@ use crate::{
     parameters::ParametersError,
 };
 
-use bincode;
 use std::fmt::Debug;
 
-#[cfg(feature = "librocksdb")]
```
howardwu, line 26: I believe this is needed for our WASM compilations to work.

howardwu (approved):
LGTM

Want to check on `feature = "librocksdb"` before merging

errors/src/storage/storage.rs
ljedrz: `rocksdb` is the name of the dependency, so it is automatically available with this feature enabled - it's the same principle as with `bincode` not requiring an explicit `use` import

PR #481, opened 2020-09-29
Make cargo bench operational

ljedrz:
A tiny fix needed for `cargo +nightly bench --workspace` to work; I recommend adding a
```
cargo +nightly bench --workspace --no-run
```
line to one of the CI runs so that the benches are not accidentally broken in the future. This only compiles the benches without running them, so the build time impact should be acceptable.


howardwu (approved):
LGTM

PR #483, opened 2020-09-30
Use criterion in curve benchmarks

ljedrz:
Builds on https://github.com/AleoHQ/snarkOS/pull/481, only the second commit is new.

Changes the benchmarking setup in `curves::benches` from `bencher` to `criterion`, which is already widely used in `benchmarks` and provides superior functionalities.

Rationale: there is room for improvement in terms of memory handling across the codebase, and having good benchmarks will aid future changes in that field.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2499,6 +2499,7 @@ dependencies = [
 name = "snarkos-curves"
 version = "1.1.4"
 dependencies = [
+ "criterion",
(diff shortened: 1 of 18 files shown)


howardwu (approved):

howardwu:
LGTM, amazing updates to benchmarks, thanks @ljedrz !

PR #492, opened 2020-10-01
More vector pre-allocations

ljedrz:
A sweep of mostly straightforward memory-handling tweaks targeting `Vec::new` and `vec![]` that are followed by `.push()` loops; pre-allocating the expected vector capacity allows to perform fewer re-allocations during runtime, leading to potential performance and memory use (possibly reducing over-allocations) improvements.
notes:
- in some places the final vector length could be greater than the specified value; in those cases the minimal/conservative scenario is assumed, possibly including an extra `1` or `2` elements that might be pushed in the related conditionals
- in cases where the loop contains early returns (e.g. via `?`), the "happy path" (i.e. no errors) is assumed
- capacity calculations where the calculation was already present in some form (e.g. in a loop range) are preserved verbatim; new ones, especially subtractions, are using the `saturating` methods in order not to cause overflow panics
- a few explicit types were removed, as they could be inferred by the compiler

Some of the more noteworthy related adjustments are:
- adding a `len` method to `BitIterator` allowing to pre-allocate in `push()`-loops using it
- deriving `Default` for more objects, limiting the number of `vec![]` groups in the code (improving future memory-handling tweaks and readability)
- optimizing the was `MessageHandler::serialize` performs its writes

### DIFF
--- a/dpc/src/base_dpc/record/record_serializer.rs
+++ b/dpc/src/base_dpc/record/record_serializer.rs
@@ -104,7 +104,7 @@ impl<C: BaseDPCComponents, P: MontgomeryModelParameters + TEModelParameters, G: 
         // reserve the MSB of the data field element's valid bitsize and set the bit to 1.
         assert_eq!(Self::PAYLOAD_ELEMENT_BITSIZE, Self::DATA_ELEMENT_BITSIZE - 1);
 
-        // This element needs to be represented in the contraint field; its bits and the number of elements
+        // This element needs to be represented in the constraint field; its bits and the number of elements
         // are calculated early, so that the storage vectors can be pre-allocated.
         let payload = record.payload();
         let payload_bits = bytes_to_bits(&to_bytes![payload]?);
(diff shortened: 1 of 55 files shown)


dpc/src/base_dpc/record/record_serializer.rs
```diff
@@ -104,10 +104,16 @@ impl<C: BaseDPCComponents, P: MontgomeryModelParameters + TEModelParameters, G:
         // reserve the MSB of the data field element's valid bitsize and set the bit to 1.
         assert_eq!(Self::PAYLOAD_ELEMENT_BITSIZE, Self::DATA_ELEMENT_BITSIZE - 1);
 
+        // This element needs to be represented in the contraint field; its bits and the number of elements
```
howardwu, line 107:
```suggestion
        // This element needs to be represented in the constraint field; its bits and the number of elements
```

howardwu (approved):
LGTM

howardwu (approved):

PR #493, opened 2020-10-02
Fewer Vec return values

ljedrz:
Another round of memory-handling improvements aimed at improving performance across the board. This time the changes are mostly targeting functions that return vectors, and the PR is best read in a commit-by-commit fashion, as a single change can span several files.

Another positive side effect of these changes is, in some of the cases (where slicing is replaced with `Iterator` methods), removing the need for bound checking, which has additional potential for improving runtime performance.


howardwu (approved):
LGTM

howardwu (approved):
LGTM

PR #494, opened 2020-10-05
More efficient iterator utilization

ljedrz:
Another round of memory-handling improvements aimed at improving performance across the board. This time the changes are mostly targeting iterators being `collect()`ed, and the PR is best read in a commit-by-commit fashion, as a single change can span several files.

A positive side effect of some of these changes is reducing indirection when using iterators, making the code easier to reason about and potentially a bit faster, too.


howardwu (approved):
LGTM

howardwu (approved):

PR #508, opened 2020-10-26
DPC tweaks

ljedrz:
This PR targets allocations in the `snarkos-dpc` crate; this is achieved with, among others:
- removing redundant (re)allocations
- reordering code where possible
- using `std::borrow::Cow`
- reducing the scope of allocations (allocating later, using `drop` on collections in some long functions)
- reusing already-allocated vectors
- utilizing iterators more often
- passing arguments by value where the function requires ownership (in order to be able to consume some objects instead of cloning all references)

The last two commits, while having a small (positive) impact on object sizes, are mostly aimed at improving code readability, as I found no use for the `Option` wrappers around all the members of `InnerCircuit` and `OuterCircuit`. They can be easily dropped in case this is a planned future feature.

I've investigated some benchmarking options, but those aren't easy - they either require a very long setup (and are themselves very long, going into dozens of minutes), or are unfit to demonstrate some of the improvements in this PR (e.g. when an object is created only once, reducing the number of allocations in the actual code thanks to argument consumption, but requiring extra clones for looped benchmarking).

There are still some possible DPC-related improvements that I'm investigating, but due to the size of this PR, I'm already marking it as ready for review and will build upon it if need be (and eventually file a new one).


ljedrz:
Rebased.

dpc/src/base_dpc/outer_circuit/outer_circuit_gadget.rs
```diff
@@ -193,30 +193,6 @@ where
     let ledger_digest_fe = ToConstraintField::<C::InnerField>::to_field_elements(ledger_digest)
         .map_err(|_| SynthesisError::AssignmentMissing)?;
 
-    let mut serial_numbers_fe = Vec::with_capacity(old_serial_numbers.len());
```
howardwu, line 196: Can we revert these changes?

Changing this gadget will invalidate the current outer-circuit-pk and -vk parameters used by testnet1.

I will confirm this has been caught by CI

howardwu (changes requested):
Pending outer circuit gadget logical changes, this PR is ready for approval

dpc/src/base_dpc/outer_circuit/outer_circuit_gadget.rs
howardwu: As we skip DPC tests in CI, this compatibility break isn't being checked. Running locally now

howardwu (approved):
Ran locally to confirm gadget changes did not affect parameters. LGTM!

PR #509, opened 2020-10-28
A range of perf improvements

ljedrz:
A host of perf improvements concentrating on the slow bits in the longest tests; a few of them were easily benchmarkable and thus got some `criterion` benches (they're in the first commit and can be removed if desired); the results on my machine were:
- `txids_to_roots`: ~6% improvement
- `merkle_root`: ~14% improvement

These changes are mostly concentrated on avoidable allocations, including:
- using (stack-bound) arrays instead of `Vec`s for some hashes
- removing child objects that seem to just be duplicates not need to exist at all
- better `Vec` initialization
- removing redundant conversions
- passing more arguments by value (i.e. consuming them)

Since this PR is still moderate in length, it might get extended until it's reviewed/merged, but it can be merged at any point in time.

### DIFF
--- a/marlin/src/lib.rs
+++ b/marlin/src/lib.rs
@@ -357,8 +357,7 @@ impl<F: PrimeField, PC: PolynomialCommitment<F>, D: Digest> Marlin<F, PC, D> {
             .into_iter()
             .chain(AHPForR1CS::prover_first_round_degree_bounds(&index_info))
             .chain(AHPForR1CS::prover_second_round_degree_bounds(&index_info))
-            .chain(AHPForR1CS::prover_third_round_degree_bounds(&index_info))
-            .collect::<Vec<_>>();
+            .chain(AHPForR1CS::prover_third_round_degree_bounds(&index_info));
 
         // Gather commitments in one vector.
         let commitments = index_vk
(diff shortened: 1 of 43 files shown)


marlin/src/lib.rs
```diff
@@ -127,7 +127,7 @@ impl<F: PrimeField, PC: PolynomialCommitment<F>, D: Digest> Marlin<F, PC, D> {
     pub fn index<'a, C: ConstraintSynthesizer<F>>(
         srs: &UniversalSRS<F, PC>,
         c: &C,
-    ) -> Result<(IndexProverKey<'a, F, PC, C>, IndexVerifierKey<F, PC, C>), Error<PC::Error>> {
```
howardwu, line 130: We should revert the changes to this file and `marlin/src/snark.rs`.

The `index` method should return a tuple of the indexed `ProvingKey` and indexed `VerificationKey`. These are uniquely indexed for a given `(SRS, Circuit)` tuple.

Alternatively, we may consider constructing two methods, such as `index_prover` and `index_verifier` or equivalent naming convention, to ensure we continue to provide complete functionality.

ljedrz: I reverted those changes for now; it seemed to me that the verification key was just being duplicated there, but I was worried the resulting API change might be undesirable - yeah, we can revisit this later on.

howardwu (approved):

PR #510, opened 2020-11-05
perf(parameters): parallelize some of the SNARK parameter checks

ljedrz:
While working on improving the slowest tests, I noticed that the `Parameters` take a very long time to load; I was initially suspecting the deserialization work itself, but later determined that the slowest operation involved was actually a post-deserialization check, namely `AffineCurve::is_in_correct_subgroup_assuming_on_curve`.

In order to reduce the time taken by those checks, `rayon` is used in order to perform them in parallel _after_ the deserialization process has concluded. I first considered doing it directly within the deserialization methods of the affected collections, but it resulted in a rather large impact on the codebase; verifying that all the values had been checked also proved to be a challenge. That's why in the end a `static PROCESSING_SNARK_PARAMS: AtomicBool` was used in order to determine if the `Parameters` are being processed and postpone their checks so they can be parallelized later on. In order to future-proof this optimization, a `static SNARK_PARAMS_AFFINE_COUNT: AtomicU64` was introduced as well; it doesn't affect the performance and ensures that any future changes to the affected objects don't result in any object validation accidentally being omitted.

The parameter loading times (for the 5 tests that are currently using them) before and after this change are as follows:
```
before: 101 + 160 + 161 + 155 + 104 = 681s = 11m 21s
after:  32  + 41  + 41  + 41  + 37  = 192s = 03m 12s
```

These wins will improve SNARK parameter loading times in non-test scenarios as well.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2791,6 +2791,7 @@ dependencies = [
  "blake2",
  "rand",
  "rand_xorshift",
+ "serial_test",
  "snarkos-algorithms",
  "snarkos-curves",
  "snarkos-errors",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2791,7 +2791,6 @@ dependencies = [
  "blake2",
  "rand",
  "rand_xorshift",
- "serial_test",
  "snarkos-algorithms",
  "snarkos-curves",
  "snarkos-errors",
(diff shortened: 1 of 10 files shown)


ljedrz:
I wouldn't expect the code optimized here to be run by multiple node instances at the same time, but I can introduce a safeguard for it to make the optimization resistant to it.

howardwu (approved):

PR #536, opened 2020-11-26
TestConstraintSystem interning

ljedrz:
Builds on https://github.com/AleoHQ/snarkOS/pull/533 (first commit).

Introduces aggressive interning in the `TestConstraintSystem`; the interned objects are currently:
- namespace path segments
- constraints (triplets of `LinearCombination`s)
- `Field`s belonging to `LinearCombination`s, `inputs` and `aux`

While there's still room for improvement, with these changes I am able to successfully compile snippets from https://github.com/AleoHQ/leo/issues/450 on a 32GB RAM machine (same as the user who reported the issue); the RAM peak in `massif` profiling for a single `2i32 ** 2i32` operation is now reduced from 5.8GB to just 2.2GB, which is a decrease of around 62%.

Cc https://github.com/AleoHQ/leo/issues/418, https://github.com/AleoHQ/leo/issues/450

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1075,6 +1075,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d36fab90f82edc3c747f9d438e06cf0a491055896f2a279638bb5beed6c40177"
 
 [[package]]
+name = "hashbrown"
+version = "0.9.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d7afe4a420e3fe79967a00898cc1f4db7c8a49a9333a29f8a4bd76a253d5cd04"
+
+[[package]]
 name = "hermit-abi"
 version = "0.1.14"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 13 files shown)


ljedrz:
Performance note: while these changes show regressions in most arithmetic microbenchmarks in `models/benches`, this is due to the fact that their every iteration is performed with a random identifier (essentially a completely new namespace), causing the interning setup to do its job too well; the `const` benchmarks, on the other hand, show improvements.

ljedrz:
Update: rebased and added 5 new commits improving the earlier results (5.8GB > 2.2GB for a single `2i32 ** 2i32` op) to 1.2GB (a decrease of **80%**), slightly simplifying the code (removing `Rc<Refcell<T>>` where not needed) and also improving the performance by greatly reducing the number of allocations (only some arithmetic microbenchmarks regress now, and it's still only due to their inherent "uninternability") and string operations (optimized assertion).

The memory use in repeated `pow` operations is still linear, but it's already a great improvement over the initial conditions.

ljedrz:
Superseded by https://github.com/AleoHQ/snarkOS/pull/541.

PR #541, opened 2020-12-02
TestConstraintsystem interning and cleanups

ljedrz:
Supersedes https://github.com/AleoHQ/snarkOS/pull/536; contains an improved interning setup, as well as a cleanup of objects upon popping namespaces, more helper objects and methods, and documentation.

The main changes are:

1. Introduce aggressive, `indexmap`-based interning in the `TestConstraintSystem`; the interned objects are the `F: Field`s, namespace path segments and, to some extent, named objects. Another "non-intrusive" change is the introduction of a custom `OptionalVec` object that's used in order to maintain the lists of constraints and input variables, keeping their indices stable and reusing the underlying memory when doing cleanups. These changes alone have reduced the memory use of the problematic user-reported cases in Leo by over 80%.

2. Keep track of named objects associated with the namespaces they belong to and remove them when it is popped. This change is crucial in order for the memory use to remain more or less constant; even with aggressive object interning, the main source of memory blowup are the variable indices, which are currently unique for **every** object, including all the intermediate ones. The cleanups are currently disabled for test purposes, so that all the constraints and variables can still be verified.

The results of changes can best be seen using profiling; I used `valgrind`'s `massif` to compare a few scenarios.

current heap use for a `2i32 ** 2i32` operation repeated 4 times (due to `massif`'s overhead I was risking OOM with more):
massif_leo_2i32_pow_times_4_current

heap use for a `2i32 ** 2i32` operation repeated 4 times with interning, but **no cleanups**; much better (and even twice as fast, too), but still linear:
massif_leo_2i32_pow_times_4_interned

heap use for a `2i32 ** 2i32` operation repeated **8** times with interning **and** cleanups (this PR); using less than 800KiB and constant:
massif_leo_2i32_pow_times_8_full

These wins are applicable not only to `pow` operations - they are just the ones that create most intermediate objects; **all** the objects associated with `TestConstraintSystem::{alloc, alloc_input, enforce}` now use less memory and no longer cause linear memory build-up when performed multiple times.

fixes https://github.com/AleoHQ/leo/issues/418, https://github.com/AleoHQ/leo/issues/421, https://github.com/AleoHQ/leo/issues/428, https://github.com/AleoHQ/leo/issues/450

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1075,6 +1075,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d36fab90f82edc3c747f9d438e06cf0a491055896f2a279638bb5beed6c40177"
 
 [[package]]
+name = "hashbrown"
+version = "0.9.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d7afe4a420e3fe79967a00898cc1f4db7c8a49a9333a29f8a4bd76a253d5cd04"
+
+[[package]]
 name = "hermit-abi"
 version = "0.1.14"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1256,11 +1262,12 @@ dependencies = [
 
 [[package]]
 name = "indexmap"
-version = "1.4.0"
+version = "1.6.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c398b2b113b55809ceb9ee3e753fcbac793f1956663f3c36549c1346015c2afe"
+checksum = "55e2e4c765aa53a0424761bf9f41aa7a6ac1efa87238f59560640e27fca028f2"
 dependencies = [
  "autocfg",
+ "hashbrown",
 ]
 
 [[package]]
@@ -2688,6 +2695,7 @@ dependencies = [
  "bincode",
  "criterion",
  "derivative",
+ "indexmap",
  "itertools",
  "rand",
  "rand_xorshift",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -945,6 +945,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "fxhash"
+version = "0.2.1"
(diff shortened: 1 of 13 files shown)


ljedrz:
As proof that other cases reported in the related issues are now solved too, I'll post their `massif` profiling results shortly as well.

ljedrz:
heap use for a single `2i28 ** 2i28` operation with this PR (OOMs without it: https://github.com/AleoHQ/leo/issues/418):
massif_leo_2i128_pow_once

ljedrz:
heap profile for the no longer stack-overflowing https://github.com/AleoHQ/leo/issues/421:
massif_leo_issue_421

ljedrz:
the profile for a slightly adjusted (so that there's no type inference error caused by recent Leo changes) `function that gets called multiple times` case from https://github.com/AleoHQ/leo/issues/450:
massif_leo_issue_450_function_calls

<details><summary>the adjusted test case</summary>
<p>

```
function main () {
    let x = 2i32;
    oom(x);
    oom(x);
}

function oom(x: i32) {
    let y1 = x ** 2i32;
    let y2 = x ** 2i32;
    let y3 = x ** 2i32;
    let y4 = x ** 2i32;
}
```

</p>
</details>

ljedrz:
heap profile for the no longer stack-overflowing AleoHQ/leo#428:
massif_leo_issue_428

ljedrz:
The profile for the `for loop` case in https://github.com/AleoHQ/leo/issues/450:
massif_leo_for

ljedrz:
the profile for a `for` loop with a `2u32 ** 2u32` operation performed 48 times:

ljedrz:
the profile for a for loop with a `2i16 ** 2i16` operation performed 128 times:
massif_leo_for_i16

ljedrz:
the profile for a for loop with a `2u16 ** 2u16` operation performed 256 times:
massif_leo_for_u16

howardwu (approved):

PR #542, opened 2020-12-03
TestConstraintSystem interning improvements

ljedrz:
Builds on https://github.com/AleoHQ/snarkOS/pull/541, I'll rebase it once that one is merged.

After having accumulated a number of profiling results, I took a closer look at them and found room for further improvement; the most important changes are:
- `compute_path` now performs an extra lookup in order to avoid a potentially unnecessary `String` allocation
- `InternedPath`, instead of being a list of `InternedPathSegments`, now consists of its parent namespace index and the final segment; this removes a lot of allocations and improves lookup time in non-test scenarios; a side-effect of this change is that an additional map, `interned_full_paths`, needs to be maintained for test purposes
- the "holes" in the `OptionalVec` object are now contained in a plain `Vec` (as it doesn't matter which hole is being filled, as long as all are filled in before resizing)
- the `get`, `set`, and `intern_path` methods are built only for test purposes

A comparison of a memory profile of a relatively complex operation (`2i128 ** 2i128`) between https://github.com/AleoHQ/snarkOS/pull/541 and this PR shows its added performance improvements.

without these changes:
massif_leo_2i128_pow_once

with these changes (the runtime is decreased):
massif_leo_2i128_pow_once_cont

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1075,6 +1075,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d36fab90f82edc3c747f9d438e06cf0a491055896f2a279638bb5beed6c40177"
 
 [[package]]
+name = "hashbrown"
+version = "0.9.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d7afe4a420e3fe79967a00898cc1f4db7c8a49a9333a29f8a4bd76a253d5cd04"
+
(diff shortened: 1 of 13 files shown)


ljedrz:
I just updated the PR with a small fix for building arithmetic gadget benchmarks; speaking of which, with this PR they show up to over **70%** improvements across the board.

howardwu (approved):

PR #550, opened 2020-12-08
Networking upgrade (rebased)

niklaslong:
This rebases the `fix/peers` branch onto `master` and supersedes #465. 
This is opened as a draft to track the continued work on the snarkOS network upgrade. 

State at time of opening: 
- network crate tests are green (unit)
- minor compiler warnings (no errors)
- integration tests are failing and not up to date.


ljedrz:
@howardwu We are now at a stage where we consider the `network` crate to be stable enough to become the basis for further `snarkOS` work. We've performed extensive stress-tests and concluded that while there are still some things left to iron out and improve, it can be done as follow-ups in order to unblock other work and allow further changes to be reviewed individually.

The `clippy` CI failure is handled in https://github.com/AleoHQ/snarkOS/pull/582, which should be good to go as soon as https://github.com/AleoHQ/snarkVM/pull/13 is merged.

howardwu (approved):

PR #582, opened 2021-01-15
Point to SnarkVM derives crate

niklaslong:
This PR is meant as a follow up to AleoHQ/SnarkVm#13 and updates the `metrics` crate to use the `derives` crate in SnarkVM as a dependency. 

Note, this will need to be updated with the new commit hash once the SnarkVM PR is merged, and the build is expected to fail until this is the case.

Update: this also deletes the `derives` crate from SnarkOS in favour of the SnarkVM version.


ljedrz:
LGTM :+1:

howardwu (approved):

PR #584, opened 2021-01-17
A few syncing tweaks

ljedrz:
- adjust syncing intervals to values that are more suitable to real-world scenarios
- fix the sync interval sanity check
- don't even spawn syncing tasks in bootstrapper mode
- add peer & block sync intervals to the config


niklaslong (commented):
LGTM 👌

howardwu (approved):

PR #589, opened 2021-01-18
Extend connection on-failure dropping rules, fix an error name

ljedrz:
There are more read-related errors that should cause a connection to be dropped; include them and fix the name of one of them.


niklaslong (commented):
👍

howardwu (approved):

howardwu:
Nice catch, glad we're streaming and not steaming :)

PR #594, opened 2021-01-20
Decouple consensus from the network node

ljedrz:
Builds on https://github.com/AleoHQ/snarkOS/pull/590, only the top commit is new.

Decoupling the consensus bits from the network node is a big step towards modularization, and enables faster network-only tests as well as larger-scale tests related to the network (e.g. peering).

There are 2 main points to achieving this goal:
1. Isolate all the consensus-related objects in `Environment` and put them in an `Option`
2. Remove the block height from the `Version` message (in order to enable handshakes without the consensus layer)

This change set does that and includes a handful of related drive-bys:
- remove the discrete `Verack` object (it was no longer necessary after the big refactoring)
- delegate config sanity checks to a dedicated `Config::check` method
- introduce `ConsensusSetup` and `TestSetup` objects for test purposes

Open point: should there be a new dedicated message type dedicated to periodic broadcasts? `Version`/`Verack` should only be used during the handshake, and they can become obsolete once a more advanced handshake setup is introduced.


niklaslong:
After reading through the changes, it occured to me boostraper nodes probably don't need to have a `consensus` at all and should be purely involved in connecting peers.

ljedrz:
Rebased.

howardwu (approved):

PR #601, opened 2021-01-25
Introduce Ping and Pong messages, improve last_seen handling and peering

ljedrz:
Individual commits describe the specific changes; the notable ones are:
- introduce `Ping` and `Pong` messages and calculate RTT for peers; the `Ping` contains the `BlockHeight`, replacing the `Version` message in this functionality
- remove the timestamp from `Version` (unused)
- use an `AtomicU32` to store `BlockHeight`, add a related type alias
- introduce a `PeerQuality` member of `PeerInfo`: it contains objects related to RTT calculation and the number of peer-related failures
- remove the timestamps from `Peers` messages (unused and susceptible to manipulation)
- limit the number of disconnected peers being connected to at once to avoid breaching the peer limit

### DIFF
--- a/network/documentation/network_messages/ping.md
+++ b/network/documentation/network_messages/ping.md
@@ -6,6 +6,6 @@ A ping protocol request for a `Pong`.
 
 ### Payload
 
-| Parameter | Type   |            Description            |
-|:---------:|--------|:---------------------------------:|
-| `nonce`   | number | A unique ping protocol identifier |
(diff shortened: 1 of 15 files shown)


ljedrz:
Note: the `last_seen` field is not necessarily needed with a ping-pong functionality in place - it could probably be removed altogether, but it has been improved and kept operational for now.

howardwu (approved):

PR #604, opened 2021-01-25
Introduce noise-powered secure handshakes and packet encryption

ljedrz:
Replaces the current handshake setup with a stateful [noise](https://noiseprotocol.org/) implementation handling nonces and providing (en/de)cryption of messages. The choice of the handshake pattern (`Noise_XXpsk3_25519_ChaChaPoly_SHA256`) was somewhat arbitrary (though `XX` is the usual "good default" pick as far as I'm aware) and can be changed; as for the cryptographic primitives, I picked ones that are not conflicting with existing dependencies, but they are also subject to change if desired.

In addition, `tokio` is updated to `1.0` across the board, enabling benchmarking of network functionalities; a small send-receive benchmark was added.

Note: the `nightly` image needs to be updated to a post-`1.48` one in order for the `ci/circleci: rust_nightly` run to succeed.

~Still a work in progress, since:~
- ~more tests are needed~
- ~the handling of `connecting`/`connected` peers needs to change now (after https://github.com/AleoHQ/snarkOS/pull/601 is merged)~
- ~some commented-out functions need to be fully removed (possible after https://github.com/AleoHQ/snarkOS/pull/601 is merged)~

PR #608, opened 2021-01-28
Replace periodic block sync with an on-demand one, request multiple hashes at once

ljedrz:
Changing the `Block` sync logic from periodic to on-demand should improve the sync-related performance of the network, especially when all the peers are aligned with one another, and bundling the hashes in sync requests will reduce network traffic.

These changes shouldn't be in conflict with https://github.com/AleoHQ/snarkOS/pull/604, but this PR should preferably be merged only after the `noise` changes are in.


ljedrz:
note: limiting the number of hashes that can be sent at once within a `GetBlocks` request will come in a follow up that also contains a relevant test

ljedrz:
Update: with this PR, multiple miners can be run without encountering OOM issues, and synchronization between 2 miners has been successfully tested. Miner nodes were the trickiest scenario due to heavy resource use and the fact that consensus-related operations are synchronous (as opposed to the asynchronous network).

This PR should now be prioritized before #604.

ljedrz:
Rebased.

PR #617, opened 2021-02-08
Decouple consensus from peering logic 

niklaslong:
This PR decouples the consensus from the node and peering logic. 

Some notable changes were undertaken, mainly to move the network crate to a good state for further decoupling work. This changeset also paves way to functionalities like truly asynchronous handling of actions like disconnects and RTT calculation.

Non-exhaustively:
- Renamed `Server` -> `Node`
- `Consensus` is now seperate from the `Node`/`Environment` and holds the chain sync related logic. 
- The `PeerBook` is now directly stored on the `Node`, the `Peers` struct has been removed as its purpose dissolved with the decoupling of the consensus.
- Similarly to `Peers`, the `Blocks` and `Transactions` structs were removed, their logic is implemented on `Consensus` at the time being, though this may change in future.
- Removed some unused functions and some unneded `Arc`s. Note a few more of these can be removed once decoupled from the miner logic.


network/src/consensus/transactions.rs
```diff
@@ -90,11 +75,11 @@ impl Transactions {
...
             let insertion = {
-                let parameters = self.environment.dpc_parameters();
-                let storage = self.environment.storage().read();
-                let consensus = self.environment.consensus_parameters();
+                let parameters = &self.dpc_parameters();
+                let storage = self.storage().read();
+                let consensus = &self.consensus_parameters();
 
-                if !consensus.verify_transaction(parameters, &tx, &storage)? {
+                if !consensus.verify_transaction(&parameters, &tx, &storage)? {
```
ljedrz, line 82: note: eventually we'll want to have methods like this not require `consensus_parameters` or `storage` to be passed as separate parameters (as they're already available inside `Consensus`; the same applies to the main RPC and miner objects)

network/src/environment.rs
```diff
@@ -15,70 +15,12 @@
...
-            last_block_sync: Arc::new(RwLock::new(Instant::now())),
-            transaction_sync_interval,
-        }
-    }
-}
+use std::{self, net::SocketAddr, time::Duration};
 
 /// A core data structure containing the networking parameters for this node.
 #[derive(Clone)]
 pub struct Environment {
```
ljedrz, line 23: note: since this is starting to look more and more like the `Node`'s configuration, we will probably eventually just rename it to `NodeConfig` or (re)use the config object used while creating the `Node`

network/src/lib.rs
```diff
@@ -72,92 +70,97 @@ pub(crate) type Sender = tokio::sync::mpsc::Sender<Message>;
 pub(crate) type Receiver = tokio::sync::mpsc::Receiver<Message>;
 
 /// A core data structure for operating the networking stack of this node.
+// TODO: remove inner Arcs.
```
ljedrz, line 73:
```suggestion
// TODO: remove inner Arcs once the Node itself is passed around in an Arc or contains an inner object wrapped in an Arc (causing all the Node's contents to be Arced too)
```

network/src/peers/peers.rs
```diff
@@ -140,99 +85,12 @@ impl Peers {
             self.broadcast_pings().await;
 
             // Store the peer book to storage.
-            self.save_peer_book_to_storage()?;
+            // self.save_peer_book_to_storage()?;
```
ljedrz, line 88: this is not the best place for the update, but we should keep it as-is during the refactoring; we can remove it from here once we implement a clean node shutdown

snarkos/main.rs
```diff
@@ -141,20 +127,34 @@ async fn start_server(config: Config) -> anyhow::Result<()> {
...
+    // Construct the node instance. Note this does not start the network services.
     // This is done early on, so that the local address can be discovered
     // before any other object (miner, RPC) needs to use it.
-    let mut server = Server::new(environment.clone()).await?;
+    let mut node = Node::new(environment.clone()).await?;
+
+    // Construct the consensus instance and set it on the node instance.
+    let consensus = Consensus::new(
+        node.clone(),
+        Arc::new(RwLock::new(storage)),
```
ljedrz, line 138: every time I encounter this, I'm double-checking whether this is the first time we've used the `storage` object; we should probably re-bind it, wrapping it in an `Arc<RwLock<T>>`, as soon as it's been used to create `memory_pool` to avoid any confusion; this should be left to a follow-up, however, as I'm seeing potential for further related improvements

snarkos/miner.rs
```diff
@@ -60,13 +60,10 @@ impl MinerInstance {
...
                 info!("Starting to mine the next block");
+                let consensus = self.node.consensus();
 
                 let (block, _coinbase_records) = match miner
-                    .mine_block(
-                        self.environment.dpc_parameters(),
-                        self.environment.storage(),
-                        self.environment.memory_pool(),
-                    )
+                    .mine_block(consensus.dpc_parameters(), consensus.storage(), consensus.memory_pool())
```
ljedrz, line 66: note: another one in the list of methods to simplify later on

network/src/peers/peers.rs
niklaslong: fixed

ljedrz:
LGTM :+1:

snarkos/main.rs
ljedrz: done in https://github.com/AleoHQ/snarkOS/pull/618

niklaslong:
Note: CI is failing on nightly because of the same issue reported in a [previous PR ](https://github.com/AleoHQ/snarkOS/pull/600#issuecomment-767446337). In this case it fails while compiling rocksdb/libsodium, possible the most memory intensive part of the build.

howardwu (approved):

PR #620, opened 2021-02-09
Specialized bootstrappers

ljedrz:
As it currently stands, bootstrapper nodes can participate in almost all the activities of regular nodes; this PR introduces changes that make them more specialized, so they can fulfill their primary purpose (peering) as well as possible:

- disregard all messages that aren't `GetPeers` or `Internal`
- don't broadcast anything
- don't start the consensus layer
- disconnect from peers as soon as they are provided with `Peers`
- disconnect from peers after 10s regardless of whether they had asked for peers or not
- obtain the list of peers to provide to nodes at random from the list of disconnected peers
- disallow mining in bootstrapper mode

Those changes make the bootstrappers much faster, more reliable, and easy/practical to run at many different locations due to their reduced resource use. In addition, since they now share all their known peers (as opposed to just the connected ones) at random, they reduce the probability of "islands" forming in the network and causing potential forking to occur.

This doesn't mean that any technical/"in-house" miner nodes can't be listed as bootstrappers for all the peers to try to connect to (which might be desirable in testnets) - they should just not be launched in bootstrapper node themselves, and there should be at least 1 node dedicated only to providing new peers with addresses of other peers (running in bootstrapper mode).

In addition, this PR makes the choice of peers to connect to from the list of disconnected peers (in case of an insufficient number of peers) random, in order to avoid trying to connect to unavailable peers.

### DIFF
--- a/snarkos/config.rs
+++ b/snarkos/config.rs
@@ -21,8 +21,6 @@ use crate::{
     update::UpdateCLI,
 };
 
-use snarkos_network::errors::NetworkError;
-
 use clap::ArgMatches;
 use dirs::home_dir;
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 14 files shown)


ljedrz:
The `nightly` failure appears to be the same as reported in https://github.com/AleoHQ/snarkOS/pull/617#issuecomment-775878785.

snarkos/config.rs
```diff
@@ -323,15 +321,19 @@ impl Config {
...
         // Check that the minimum and maximum number of peers is valid.
         if self.p2p.min_peers == 0 || self.p2p.max_peers == 0 {
-            return Err(NetworkError::PeerCountInvalid);
+            return Err(CliError::PeerCountInvalid);
         }
 
         // Check that the sync interval is a reasonable number of seconds.
         if !(2..=300).contains(&self.p2p.peer_sync_interval) || !(2..=300).contains(&self.p2p.block_sync_interval) {
-            return Err(NetworkError::SyncIntervalInvalid);
+            return Err(CliError::SyncIntervalInvalid);
```
niklaslong, line 332: Note: we could parse instead of validating for all these intervals in future?

niklaslong (commented):
LGTM 👍

snarkos/config.rs
ljedrz: We'll soon need to revise the synchronization intervals we're using right now, possibly with some other config cleanups, good point :+1:.

ljedrz:
Rebased.

ljedrz:
The CI errors were caused by a dependency; an issue had already been registered: https://github.com/bheisler/TinyTemplate/issues/16.

ljedrz:
Ran `cargo update`, which fixed the issue in the dependency; there was also a similar `doc` attribute error in the repo, which is now fixed as well.

PR #621, opened 2021-02-09
Fix blockheight for consensus-less nodes 

niklaslong:
Return a blockheight of `0` when the node doesn't have a consensus. This is useful for testing purposes and doesn't impact consensus-less bootnodes as they don't send `Ping` messages in the first place.


ljedrz:
Superseded by https://github.com/AleoHQ/snarkOS/pull/623/commits/9d672ab2539f59c041b32c9706aa250d76256edd in https://github.com/AleoHQ/snarkOS/pull/623.

PR #624, opened 2021-02-11
Feature-gate capnpc use in build.rs

ljedrz:
This PR takes advantage of the fact that the auto-generated file based on the Cap'n Proto schema (`snarkos_network/src/message/payload_capnp.rs`) is being committed, and puts the `capnpc` dependency and its use in `build.rs` behind a new `compile_capnp_schema` feature that's disabled by default.

This removes the need to download and use `capnproto` unless one wants to alter the schema (after which just the changes to the schema and the updated auto-generated file need to be committed).


niklaslong (commented):
LGTM other than the missing newline 👍

howardwu (approved):

PR #625, opened 2021-02-11
Remove unused dependencies

ljedrz:
A maintenance pass pruning unused dependencies.

~In addition, the workspace-level `Cargo.toml` is simplified to only contain member information and the binary dependencies were moved to a new `Cargo.toml` under `snarkos`.~ Due to multiple possible targets, in the end this change was removed in order to keep the setup simple for the users.

These changes will result in an additional reduction to the size of `Cargo.lock` once https://github.com/AleoHQ/snarkVM/pull/42 is merged and its new version is used.


ljedrz:
With these changes in place the `nightly` build no longer seems to OOM.

howardwu (approved):

PR #626, opened 2021-02-11
Update the code and CI to Rust stable 1.50

ljedrz:
Since Rust has only just been updated to 1.50, the CI might not have the fresh image yet; if that happens, the PR's CI runs can be restarted a bit later on.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -43,7 +43,7 @@ commands:
 jobs:
   rust_stable:
     docker:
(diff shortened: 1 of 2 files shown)


ljedrz:
Hmm, weird; CircleCI suggests that [1.50 is already available](https://circleci.com/docs/2.0/circleci-images/#rust), but it doesn't seem to load.

howardwu (approved):

PR #636, opened 2021-02-25
Use ring as the noise resolver

ljedrz:
Builds on https://github.com/AleoHQ/snarkOS/pull/635, the last commit is new.

The previous `libsodium` pick was dictated by dependency incompatibilities, but now it's possible to use the more Rust-native [ring](https://crates.io/crates/ring) crate; it should make compilation times faster, and hopefully reduce the occurrence of CI build issues.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1632,17 +1632,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "libsodium-sys"
-version = "0.2.6"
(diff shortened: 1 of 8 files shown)


ljedrz:
Unfortunately it seems that `ring` doesn't currently handle `CryptoResolver::resolve_dh`, which is a no-go for the XX handshake.

PR #652, opened 2021-03-03
Network topology testing

niklaslong:
This PR implements some basic topology testing and paves the way for more complex scenarios and measurements. This branch is based on #620. 

Included: 
- a `topology` module to aid in establishing topologies for testing
- tests verifying basic topologies converge nicely into meshes
- network density and degree centrality measurements
- tweaks to the peering limit when initialising connections

The tests aim for a minimum network density proportional to the min peers and a degree centrality delta which is covered by the min-max peer delta.


niklaslong:
Failing for the same reason outlined in #620.

niklaslong:
And rebased onto #620 post-dependency-fix

network/src/lib.rs
```diff
@@ -181,6 +192,11 @@ impl Node {
     async fn process_incoming_messages(&self, receiver: &mut Receiver) -> Result<(), NetworkError> {
         let Message { direction, payload } = receiver.recv().await.ok_or(NetworkError::ReceiverFailedToParse)?;
 
+        if self.environment.is_bootnode() && !(payload == Payload::GetPeers || direction == Direction::Internal) {
+            // the bootstrapper nodes should ignore inbound messages other than GetPeers
+            return Ok(());
```
ljedrz, line 197: We might want to introduce a new `NetworkError::TooManyConnections`  or `::ConnectionLimitBreach` in order to indicate that this operation failed; it doesn't matter during regular runtime, but will for e.g. a RPC to connect to a specific address or in tests.

network/src/peers/peers.rs
```diff
@@ -344,4 +387,18 @@ impl Node {
             self.peer_book.write().add_peer(peer_address);
         }
     }
+
+    fn can_connect(&self) -> bool {
+        let peer_book = self.peer_book.read();
```
ljedrz, line 392: it's extremely unlikely to matter, but just due to the fact that there's a potential call to `warn!` which, as all `stdout` operations, is quite slow, I'd limit the scope of this lock by introducing a "synthetic" one that returns a pair with `num_connected` and `num_connecting`.

network/tests/topology.rs
```diff
@@ -0,0 +1,301 @@
...
+        ..Default::default()
+    };
+    let mut nodes = test_nodes(N, setup).await;
+    connect_nodes(&mut nodes, Topology::Line).await;
+    start_nodes(&nodes).await;
+
+    // First and Last nodes should have 1 connected peer.
+    wait_until!(
+        5,
+        nodes.first().unwrap().peer_book.read().number_of_connected_peers() == 1
```
ljedrz, line 69: We should probably move some of these `PeerBook` methods to `Node` in a follow-up soon.

network/tests/topology.rs
```diff
@@ -0,0 +1,301 @@
...
+    // Set the sleep interval to 200ms to avoid lock issues.
+    // Density measurement here is proportional to the min peers: if every node in the topology
+    // only connected to the min node count, the total number of connections would be roughly 10
+    // percent of the total possible. With 50 nodes and min at 5 connections each this works out to
+    // be 125/1225 ≈ 0.1.
+    wait_until!(15, network_density(&nodes) >= 0.1, 200);
+
+    // Make sure the node with the largest degree centrality and smallest degree centrality don't
+    // have a delta greater than the max-min peer interval allows for. This check also provides
+    // some insight into the proper homogenous meshing of the network.
```
ljedrz, line 135:
```suggestion
    // some insight into whether the network is meshed in a homogeneous manner.
```

network/tests/topology.rs
```diff
@@ -0,0 +1,301 @@
...
+        nodes.push(node);
+    }
+
+    nodes
+}
+
+async fn start_nodes(nodes: &[Node]) {
+    for node in nodes {
+        // Nodes are started with a slight delay to avoid having peering intervals in phase (this
+        // is the hypothetical real-world worst case scenario).
```
ljedrz, line 49:
```suggestion
        // is the hypothetical worst case scenario).
```
I'd consider it _extremely_ unlikely in real-world scenarios, especially since we're talking about possibly many nodes; adding one in tests makes them more likely to reflect actual conditions.

network/tests/topology.rs
```diff
@@ -0,0 +1,301 @@
...
+        peer_sync_interval: 5,
+        min_peers: MIN_PEERS,
+        max_peers: MAX_PEERS,
+        ..Default::default()
+    };
+    let mut nodes = test_nodes(N, setup).await;
+    connect_nodes(&mut nodes, Topology::Mesh).await;
+    start_nodes(&nodes).await;
+
+    // Set the sleep interval to 200ms to avoid lock issues.
```
ljedrz, line 126:
```suggestion
    // Set the sleep interval to 200ms to avoid locking issues.
```
We should also add (can be in a follow-up) tests that would check the worst-case peering conditions so that we're sure that those issues result only in a single update slowdown / skip as opposed to anything more problematic.

network/tests/topology.rs
```diff
@@ -0,0 +1,301 @@
...
+    // Density measurement here is proportional to the min peers: if every node in the topology
```
ljedrz, line 127:
```suggestion
    // Density measurement here is proportional to the min peers: if every node in the network
```

ljedrz:
LGTM :ok_hand:; left some minor remarks and suggestions.

network/src/lib.rs
ljedrz: (this was of course directed at `initialize_connection`, which got amended now :ok_hand:)

ljedrz (approved):

howardwu (approved):

PR #658, opened 2021-03-08
Use the Storage trait

ljedrz:
This is a draft of changes applicable after https://github.com/AleoHQ/snarkVM/pull/97 is merged, most notable of which are:
- `Node`, `Consensus` and `RpcImpl` become generic over `Storage`
- `snarkos-storage` gains 2 new features, `mem_storage` and `rocksdb_storage`, the latter being the default
- the `RpcImpl` object is refactored a bit, loses the lock over `MerkleTreeStorage` (unnecessary) and the `new_secondary_storage_instance` method (it is somewhat incompatible with the potentially in-memory-only storage options and the RPC instance already runs in secondary storage mode by default)
- `MemDb` and `RocksDb` objects that implement the `Storage` trait are introduced in `snarkos-storage`
- `rocksdb` is now an optional dependency (but still used by default)
- `storage/tests/mod.rs` is removed, as those tests are pretty basic/synthetic and incompatible with generic storage backends

In order to limit the immediate impact of this big set of changes, the RPC implementation, at least the one in `snarkos::main`, currently expects the `RocksDB` backend to be used. This limitation can be lifted later on.


ljedrz:
Note: I introduced a `network` feature to `snarkos-testing`, so that snarkVM can depend on it without having to pull all the network-specific dependencies; it is also the reason why `MinerInstance` was moved to `snarkos-network/consensus`, which is a much more fitting right place for it.

howardwu (approved):

PR #671, opened 2021-03-24
Fuzz test slightly corrupted messages

niklaslong:
This collection of fuzz tests aims to test messages with a small percentage of their bytes randomly replaced and is an extension of currently implemented fuzz tests. 

Testing slightly corrupted but otherwise valid messages is more likely to expose issues with deserialisation than completely random payloads. These tests target both pre and post-handshake nodes.

### DIFF
--- a/network/tests/fuzzing.rs
+++ b/network/tests/fuzzing.rs
@@ -56,7 +56,7 @@ async fn fuzzing_zeroes_pre_handshake() {
     let mut stream = TcpStream::connect(node_addr).await.unwrap();
     wait_until!(1, node.peer_book.read().number_of_connecting_peers() == 1);
 
-    let _ = stream.write_all(&vec![0u8; 64]).await;
+    let _ = stream.write_all(&[0u8; 64]).await;
     wait_until!(1, node.peer_book.read().number_of_connecting_peers() == 0);
 }
 
@@ -70,7 +70,7 @@ async fn fuzzing_zeroes_post_handshake() {
     let (node, fake_node) = handshaken_node_and_peer(node_setup).await;
     wait_until!(1, node.peer_book.read().number_of_connected_peers() == 1);
 
-    fake_node.write_bytes(&vec![0u8; 64]).await;
+    fake_node.write_bytes(&[0u8; 64]).await;
     wait_until!(1, node.peer_book.read().number_of_connected_peers() == 0);
 }
 
--- a/network/tests/fuzzing.rs
+++ b/network/tests/fuzzing.rs
@@ -25,7 +25,7 @@ use tokio::{io::AsyncWriteExt, net::TcpStream};
 
 use std::net::SocketAddr;
 
-pub const ITERATIONS: usize = 10000;
+pub const ITERATIONS: usize = 5000;
 pub const CORRUPTION_PROBABILITY: f64 = 0.1;
 
 fn corrupt_bytes(serialized: &[u8]) -> Vec<u8> {
--- a/network/tests/fuzzing.rs
+++ b/network/tests/fuzzing.rs
@@ -176,6 +176,36 @@ async fn fuzzing_corrupted_version_pre_handshake() {
 }
 
 #[tokio::test(flavor = "multi_thread")]
+async fn fuzzing_corrupted_version_post_handshake() {
+    // tracing_subscriber::fmt::init();
+
+    let (node1, mut node2) = spawn_2_fake_nodes().await;
+
+    tokio::spawn(async move {
+        loop {
+            let _ = node2.read_payload().await;
+        }
+    });
+
+    let mut rng = thread_rng();
+
+    for _ in 0..ITERATIONS {
+        let version = Version::serialize(&Version::new(1, 4141)).unwrap();
+
+        // Replace a random percentage of random bytes at random indices in the serialised message.
+        let corrupted_version: Vec<u8> = version
+            .into_iter()
+            .map(|byte| if rng.gen_bool(0.1) { rng.gen() } else { byte })
+            .collect();
+
+        let header = MessageHeader::from(corrupted_version.len());
+
+        node1.write_bytes(&header.as_bytes()).await;
+        node1.write_bytes(&corrupted_version).await;
+    }
+}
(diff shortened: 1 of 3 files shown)


network/tests/fuzzing.rs
```diff
@@ -137,3 +141,425 @@ async fn fuzzing_post_handshake() {
...
+    let mut rng = thread_rng();
+
+    for _ in 0..ITERATIONS {
+        let mut stream = TcpStream::connect(node_addr).await.unwrap();
+        let version = Version::serialize(&Version::new(1u64, stream.local_addr().unwrap().port())).unwrap();
+
+        // Replace a random percentage of random bytes at random indices in the serialised message.
+        let corrupted_version: Vec<u8> = version
+            .into_iter()
+            .map(|byte| if rng.gen_bool(0.1) { rng.gen() } else { byte })
```
ljedrz, line 166: it would be nice if the alteration probability was a configurable `const` like `ITERATIONS`

network/tests/fuzzing.rs
```diff
@@ -137,3 +141,425 @@ async fn fuzzing_post_handshake() {
...
+
+        let header = MessageHeader::from(corrupted_version.len());
+
+        node1.write_bytes(&header.as_bytes()).await;
+        node1.write_bytes(&corrupted_version).await;
+    }
+}
+
+#[tokio::test(flavor = "multi_thread")]
+async fn fuzzing_corrupted_empty_payloads_pre_handshake() {
```
ljedrz, line 209: I understand the the reason for fuzzing messages that are usually immediately rejected in pre-handshake communication is the potential for breakage when they are deserialized (i.e. so it can even be determined that they should be rejected); I'd leave a short comment so that this rationale is clear

network/tests/fuzzing.rs
```diff
@@ -137,3 +141,425 @@ async fn fuzzing_post_handshake() {
...
+            let _ = stream.write_all(&header.as_bytes()).await;
+            let _ = stream.write_all(&corrupted_payload).await;
+        }
+    }
+
+    assert_eq!(node.peer_book.read().number_of_connected_peers(), 0);
+}
+
+#[tokio::test(flavor = "multi_thread")]
+async fn fuzzing_corrupted_empty_payloads_post_handshake() {
```
ljedrz, line 242: I'd adjust the `empty_payloads` function names to `no_payloads` or `messages_without_payloads` so they are not confused with messages that do carry payloads but which are empty for test purposes

network/tests/fuzzing.rs
```diff
@@ -137,3 +141,425 @@ async fn fuzzing_post_handshake() {
...
+
+    let mut rng = thread_rng();
+
+    for payload in &[Payload::GetMemoryPool, Payload::GetPeers, Payload::Pong] {
+        for _ in 0..ITERATIONS {
+            let serialized = Payload::serialize(payload).unwrap();
+            let corrupted_payload: Vec<u8> = serialized
+                .into_iter()
+                .map(|byte| if rng.gen_bool(0.1) { rng.gen() } else { byte })
+                .collect();
```
ljedrz, line 261: this is good material for a common helper function that could be used for all the tests - it will deduplicate the code a bit

network/tests/fuzzing.rs
```diff
@@ -137,3 +141,425 @@ async fn fuzzing_post_handshake() {
...
+        consensus_setup: None,
+        ..Default::default()
+    };
+
+    let node = test_node(node_setup).await;
+    let node_addr = node.local_address().unwrap();
+
+    let mut rng = thread_rng();
+
+    let blob: Vec<u8> = (0u8..255).collect();
```
ljedrz, line 285: it would be a good idea to fully randomize the contents of blobs so that there's a chance that some combination reveals a bug in _their_ deserialization

network/tests/fuzzing.rs
```diff
@@ -137,3 +141,425 @@ async fn fuzzing_post_handshake() {
...
+            let header = MessageHeader::from(corrupted_payload.len());
+
+            node1.write_bytes(&header.as_bytes()).await;
+            node1.write_bytes(&corrupted_payload).await;
+        }
+    }
+}
+
+#[tokio::test(flavor = "multi_thread")]
+async fn fuzzing_currupted_peers_pre_handshake() {
```
ljedrz, line 424: couldn't `ping` and `peers` be joined with the `payloads_with_blobs` tests? they could be renamed `payloads_with_bodies` so that it still work, and it would deduplicate the code

network/tests/fuzzing.rs
```diff
@@ -137,3 +141,425 @@ async fn fuzzing_post_handshake() {
         node1.write_bytes(&random_bytes).await;
     }
 }
+
+#[tokio::test(flavor = "multi_thread")]
```
ljedrz, line 145: was `multi_thread` chosen so that these tests are faster or that there is more strain on the test node, so that these fuzz tests serve as stress tests too? as for the speed, I'm not sure if multi-threaded executors can improve the runtime, as the node is likely to immediately reject most messages; if, on the other hand, the runs you've done locally indicate that it is indeed a performance improvement, I'd make a note about it in a comment

network/tests/fuzzing.rs
```diff
@@ -137,3 +141,425 @@ async fn fuzzing_post_handshake() {
...
+    let blob: Vec<u8> = (0u8..255).collect();
+
+    for payload in &[
+        Payload::Block(blob.clone()),
+        Payload::MemoryPool(vec![blob.clone(); 10]),
+        Payload::SyncBlock(blob.clone()),
+        Payload::Transaction(blob),
+    ] {
+        for _ in 0..ITERATIONS {
+            let serialized = Payload::serialize(payload).unwrap();
```
ljedrz, line 334: for the best performance I'd created the serialized payloads before the loop starts, then loop over _them_ instead of the pre-serialized ones, so that only the `corrupted_payload` gets recreated each time

ljedrz (approved):
LGTM with a few minor comments :+1:

howardwu (approved):

PR #682, opened 2021-04-08
Resource-oriented refactoring

ljedrz:
We've already had some task accounting, but some of it was still not in place due to the objects involved not being sufficiently accessible. This PR refactors some of them so that the following goals can be achieved:
- complete `task` accounting
- truly asynchronous connection handling (especially valuable for disconnects)
- a `Node::shut_down` method and task cleanup performed on `Drop`
- more cohesive object handling with less room for them becoming disjoint (most notably `Environment`)

Notable changes:
- `Environment, Inbound, Outbound, PeerInfo` are no longer `Clone`
- `Node` now only contains an `Arc<InnerNode>` (transparent in terms of ergonomics)
- `Direction::Internal` no longer exists, and neither do related internal messages
- `Consensus`-related methods no longer require the list of peers to be passed to them (it's available from within them)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2627,6 +2627,7 @@ dependencies = [
  "derivative",
  "hex",
  "log",
+ "once_cell",
  "parking_lot",
  "peak_alloc",
  "rand 0.8.3",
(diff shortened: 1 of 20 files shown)


niklaslong (approved):
Looks great! 👍

howardwu (approved):

PR #683, opened 2021-04-08
Revise the Node and Environment objects

ljedrz:
This PR renames `Environment` to `Config` and renames its module to `config`, moves `Node` into its own module, and moves `name` and `local_address` members to `Node`, making `Config` an object where only the pre-configured settings are kept.

### DIFF
--- a/network/src/config.rs
+++ b/network/src/config.rs
@@ -24,7 +24,7 @@ use std::{
 };
 
 /// A core data structure containing the pre-configured parameters for the node.
-pub struct Environment {
+pub struct Config {
     /// The pre-configured desired address of this node.
     pub desired_address: Option<SocketAddr>,
     /// The minimum number of peers required to maintain connections with.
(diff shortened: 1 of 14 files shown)


niklaslong (approved):
LGTM 👍

howardwu:
Just to confirm, does this mean we no longer need prometheus? If I read this correctly, we remove the counters right?

ljedrz:
@howardwu these specific counters are currently not being incremented anywhere, so they are essentially dead code. I'll be happy to revise the Prometheus instrumentation as next order of business and adjust accordingly, perhaps including isolation of a common container object dedicated to all statistics related to the node. It is definitely practical to keep track of the number of received messages and their size, so they are not going away, regardless of the "consumer" (prometheus/RPC).

ljedrz:
Rebased.

ljedrz:
the nightly `clippy` error is unrelated to the PR, same as in https://github.com/AleoHQ/snarkOS/pull/684.

howardwu (approved):

PR #684, opened 2021-04-09
Dedicate an OS thread to mining

ljedrz:
Spawning a miner currently only spawns an asynchronous task, but mining is a synchronous process, which means it can cause blocking to the async executor (even though it's multi-threaded). This PR uses an actual OS thread instead, which frees the async executor so it can carry out other functionalities, notably anything network-related.

Filing it as a draft for now, as it'll need to be rebased after the pending refactoring PRs are merged.

### DIFF
--- a/benchmarks/syncing/syncing.rs
+++ b/benchmarks/syncing/syncing.rs
@@ -23,8 +23,13 @@ use snarkos_testing::network::{blocks::*, handshaken_node_and_peer, TestSetup};
 fn providing_sync_blocks(c: &mut Criterion) {
     let rt = tokio::runtime::Runtime::new().unwrap();
 
+    let test_setup = TestSetup {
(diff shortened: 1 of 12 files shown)


ljedrz:
`clippy` seems to be failing due to some `clang-sys` issue with the `nightly` compiler.

ljedrz:
Rebased; the PR now builds on https://github.com/AleoHQ/snarkOS/pull/694.

ljedrz:
Update: I've added a few commits that greatly speed up the first very sync attempt; in addition, I've reduced the number of fuzz test iterations, though the CI sometimes still glitches on tests; I'll investigate this shortly, but it's not related to this PR.

howardwu (approved):

howardwu (approved):

PR #694, opened 2021-04-15
Connect to peers faster, delay mining

ljedrz:
This short PR adjusts the node's start-up process by starting its network services before mining, connecting to peers without an initial delay, and delaying mining by a few seconds.

### DIFF
--- a/network/src/node.rs
+++ b/network/src/node.rs
@@ -159,12 +159,12 @@ impl<S: Storage + Send + Sync + 'static> Node<S> {
         let peer_sync_interval = self.config.peer_sync_interval();
         let peering_task = task::spawn(async move {
             loop {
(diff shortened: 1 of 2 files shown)


howardwu (approved):

ljedrz:
Rebased.

damons:
Should mining be delayed until all blocks have been sync'd up to some threshold of consensus/longest chain at least as reported by the bootnode?

ljedrz:
@damons in the event the node doesn't want to use the default list of bootstrappers or they are down, it could mean that the node would not start mining at all; there's pros and cons for both approaches, with the delay being the safest approach. With the dedicated mining thread, the bootstrappers should respond fast, and the node should start syncing very soon, so it should be sufficient; they can also be configured to provide their height even faster.

PR #698, opened 2021-04-19
Some clippy lints

ljedrz:
This PR handles some `clippy` lints that sometimes show up in GitHub diff views, most notably `upper_case_acronyms`, which is now allowed in the crates where it is triggered.


ljedrz:
Rebased.

ljedrz:
This one will be superseded by a newer PR.

PR #705, opened 2021-04-23
Handshake timeouts, async connection attempts

ljedrz:
This PR imposes a 5s deadline on all handshakes, breaking the connection in case it's not complete within that limit. In addition, it makes multiple connection attempts (connecting to bootstrappers and disconnected peers) happen asynchronously, so that they can be completed faster.

2 tests for this new behavior are added to the network handshake tests.


niklaslong (approved):
LGTM 👍

PR #715, opened 2021-04-27
Upgrade networking infrastructure - updates syncing, peers, bootstrap nodes

howardwu:
### Bootstrap node changes
- Update to support unlimited inbound and outbound connection requests
- Update to not arbitrarily drop peers after a given amount of time
- Removes non-bootstrap requirement for sync layer to be enabled
- Bootstrap node no longer forces disconnect upon sending `GetPeers`
- Bootstrap nodes will continually broadcast `GetPeers` requests to connected peers
- Bootstrap nodes will process all inbound peers instead of a bounded set
- Bootstrap nodes will attempt to connect to all disconnected peers (instead of the minimum number of peers)

### Sync handler changes
- Updates selection algorithm to pick bootstrap node as sync node first (in declared order)
  - Falls back to last seen peer if no bootstrap node is connected to
- Remove requirement of `!is_syncing` for responding to a`GetSync`, `GetBlock`, and `GetMemoryPool` message
  - A corner case exists where the longest chain node (from your node's perspective) is syncing to an even longer chain node. In this case, we need to allow for downstream syncing.
- Introduces new sync blocks logic to run on periodic interval rather than upon receiving a `Ping` message
- Decrease block sync count per batch to 64
- Decrease block sync interval to 4 seconds
- Increase peer sync interval to 8 seconds
- Increase mempool sync interval to 12 seconds

### Peer handler changes
- Moves check of minimum connected peers into `connect_to_disconnected_peers` and `broadcast_getpeers_requests`, no longer checks this for `bootnodes`
- Fixes overcounting bug in determining `count` for disconnected peers case
- All nodes only send connected peers when responding to `GetPeers`
- Introduces handshake timeout for bootnodes of 30 seconds

### Peer book changes
- Introduces `block_height` to `PeerQuality`, which is set every time upon receiving a `Ping` message
- Removes saving the peer book to storage
- Update `PeerQuality::remaining_sync_blocks` to `AtomicU32`

### Fuzzing test changes
- Decrease fuzzing iterations to 500 (for faster CI times)

### Config changes
- For now, override the `config.toml` file each time we start snarkOS

### CLI changes
- Fix verbosity bug, now supports running `cargo run -- --verbose 3` for `trace` logging

### CI changes
- Increase CI runtime to 60 minutes

### DIFF
--- a/network/src/inbound/inbound.rs
+++ b/network/src/inbound/inbound.rs
@@ -176,21 +176,24 @@ impl<S: Storage + Send + Sync + 'static> Node<S> {
                 }
             };
 
-            // Handle Ping/Pong messages immediately in order not to skew latency calculation.
-            match &message.payload {
-                Payload::Ping(..) => {
-                    self.outbound
-                        .send_request(Message::new(Direction::Outbound(reader.addr), Payload::Pong))
-                        .await;
+            // Route the message to the inbound handler of this node.
+            {
+                // Handle Ping/Pong messages immediately in order not to skew latency calculation.
+                match &message.payload {
+                    Payload::Ping(..) => {
+                        self.outbound
+                            .send_request(Message::new(Direction::Outbound(reader.addr), Payload::Pong))
+                            .await;
+                    }
+                    Payload::Pong => {
+                        self.peer_book.received_pong(reader.addr);
+                    }
+                    _ => {}
                 }
-                Payload::Pong => {
-                    self.peer_book.received_pong(reader.addr);
-                }
-                _ => {}
-            }
 
-            // Messages are queued in a single tokio MPSC receiver.
-            self.inbound.route(message).await
+                // Messages are queued in a single tokio MPSC receiver.
+                self.inbound.route(message).await
+            }
         }
     }
 
@@ -224,7 +227,7 @@ impl<S: Storage + Send + Sync + 'static> Node<S> {
                 if let Some(ref sync) = self.sync() {
                     sync.received_block(source, block, false).await?;
 
-                    // update the peer and possibly finish the sync process
+                    // Update the peer and possibly finish the sync process.
                     if self.peer_book.got_sync_block(source) {
                         sync.finished_syncing_blocks();
                     } else {
--- a/network/src/inbound/inbound.rs
+++ b/network/src/inbound/inbound.rs
@@ -110,7 +110,11 @@ impl<S: Storage + Send + Sync + 'static> Node<S> {
                                 // Update the remote address to be the peer's listening address.
                                 let remote_address = channel.addr;
                                 // Save the channel under the provided remote address.
-                                node.inbound.channels.write().await.insert(remote_address, Arc::new(channel));
+                                node.inbound
+                                    .channels
+                                    .write()
+                                    .await
+                                    .insert(remote_address, Arc::new(channel));
 
                                 let node_clone = node.clone();
                                 let peer_listening_task = tokio::spawn(async move {
--- a/network/src/inbound/inbound.rs
+++ b/network/src/inbound/inbound.rs
@@ -86,13 +86,14 @@ impl<S: Storage + Send + Sync + 'static> Node<S> {
         let listener_handle = task::spawn(async move {
             info!("Listening for nodes at {}", listener_address);
 
+            let bootnodes = node.config.bootnodes();
+
             loop {
                 match listener.accept().await {
                     Ok((stream, remote_address)) => {
                         info!("Got a connection request from {}", remote_address);
 
                         // Wait a maximum timeout limit for a connection request.
-                        let bootnodes = node.config.bootnodes();
                         let timeout = match bootnodes.contains(&remote_address) {
                             true => Duration::from_secs(crate::HANDSHAKE_BOOTNODE_TIMEOUT_SECS as u64),
                             false => Duration::from_secs(crate::HANDSHAKE_PEER_TIMEOUT_SECS as u64),
--- a/network/src/inbound/inbound.rs
+++ b/network/src/inbound/inbound.rs
@@ -158,7 +158,7 @@ impl<S: Storage + Send + Sync + 'static> Node<S> {
                 Ok(message) => message,
                 Err(error) => {
                     // Log the failure and increment the failure count.
-                    error!("A network error occurred while listening for messages: {}", error);
+                    error!("Unable to read message from {}: {}", reader.addr, error);
                     failure_count += 1;
 
                     // Increment the fatal count if the error is a fatal error.
@@ -173,7 +173,7 @@ impl<S: Storage + Send + Sync + 'static> Node<S> {
                     match disconnect_from_peer {
                         true => {
                             // TODO (howardwu): Implement a handler so the node does not lose state of undetected disconnects.
-                            warn!("Disconnecting from an unreliable peer");
+                            warn!("Disconnecting from {} (unreliable)", reader.addr);
                             let _ = self.disconnect_from_peer(reader.addr);
                             // The error has been handled and reported, we may now safely break.
                             break;
--- a/network/src/inbound/inbound.rs
+++ b/network/src/inbound/inbound.rs
@@ -247,16 +247,12 @@ impl<S: Storage + Send + Sync + 'static> Node<S> {
             }
             Payload::GetBlocks(hashes) => {
                 if let Some(ref sync) = self.sync() {
-                    if !sync.is_syncing_blocks() {
-                        sync.received_get_blocks(source, hashes).await?;
-                    }
+                    sync.received_get_blocks(source, hashes).await?;
                 }
             }
             Payload::GetMemoryPool => {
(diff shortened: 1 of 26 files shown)


network/src/inbound/inbound.rs
```diff
@@ -192,32 +187,31 @@ impl<S: Storage + Send + Sync + 'static> Node<S> {
...
-                    self.peer_book.received_pong(reader.addr);
+            // Route the message to the inbound handler of this node.
+            {
+                // Handle Ping/Pong messages immediately in order not to skew latency calculation.
+                match &message.payload {
+                    Payload::Ping(block_height) => {
+                        self.outbound
+                            .send_request(Message::new(Direction::Outbound(reader.addr), Payload::Pong))
+                            .await;
+                        self.peer_book.received_ping(reader.addr, *block_height);
```
ljedrz, line 198: this operation should still be queued with all other requests (i.e. handled at `process_incoming_messages`) in order to avoid being maliciously spammed with `Ping` messages; `received_pong` is guarded against this with `quality.expecting_pong`.

network/src/inbound/inbound.rs
```diff
@@ -100,51 +80,59 @@ impl<S: Storage + Send + Sync + 'static> Node<S> {
...
         let listener_handle = task::spawn(async move {
+            info!("Listening for nodes at {}", listener_address);
+
             loop {
                 match listener.accept().await {
                     Ok((stream, remote_address)) => {
                         info!("Got a connection request from {}", remote_address);
 
+                        // Wait a maximum timeout limit for a connection request.
+                        let bootnodes = node.config.bootnodes();
```
ljedrz, line 95: this assignment can be moved before the body of the `loop`, as that list doesn't change during runtime

network/src/node.rs
```diff
@@ -119,106 +143,168 @@ impl<S: Storage + Send + Sync + 'static> Node<S> {
...
-
-                        if !consensus.is_syncing_blocks() {
-                            info!("Updating transactions");
-
-                            // select last seen node as block sync node
-                            let sync_node = self_clone.peer_book.last_seen();
-                            consensus.update_transactions(sync_node).await;
+            self.register_task(sync_mempool_task);
+
+            let node_clone = self.clone();
```
ljedrz, line 263: the `Sync` object already contains a reference to `Node`, so it doesn't need to be additionally cloned here

network/src/node.rs
```diff
@@ -119,106 +143,168 @@ impl<S: Storage + Send + Sync + 'static> Node<S> {
...
+            let sync_clone = Arc::clone(sync);
+            let block_sync_interval = sync_clone.block_sync_interval();
+            let sync_block_task = task::spawn(async move {
+                loop {
+                    sleep(block_sync_interval).await;
```
ljedrz, line 268: this delay being put here as opposed to the end of the body of the `loop` imposes an initial delay for the block sync; might be better to put it at the end instead

network/src/peers/peer_info.rs
```diff
@@ -39,6 +40,8 @@ pub enum PeerStatus {
 
 #[derive(Debug, Default)]
 pub struct PeerQuality {
+    /// The current block height of this peer.
+    pub block_height: RwLock<BlockHeight>,
```
ljedrz, line 44: since `BlockHeight` is a `u32`, this should just be an `AtomicU32` for better performance

network/src/peers/peers.rs
```diff
@@ -54,83 +59,67 @@ impl<S: Storage + Send + Sync + 'static> Node<S> {
...
+
         // Drop peers whose RTT is too high or have too many failures.
         for (addr, peer_quality) in self
             .peer_book
             .connected_peers()
             .iter()
             .map(|(addr, info)| (*addr, &info.quality))
         {
+            // Skip this check if the peer is a bootnode.
+            if bootnodes.contains(&addr) {
```
ljedrz, line 73: nit: this could be a `filter` in the iterator above

ljedrz (approved):
LGTM with a few comments :+1:

network/src/peers/peer_info.rs
```diff
@@ -49,7 +52,7 @@ pub struct PeerQuality {
     pub rtt_ms: AtomicU64,
     /// The number of failures associated with the peer; grounds for dismissal.
     pub failures: AtomicU32,
-    /// The number of remaining blocs to sync with.
+    /// The number of remaining blocks to sync with.
     pub remaining_sync_blocks: AtomicU16,
```
ljedrz, line 56: note: changing this to `AtomicU32` will fix https://github.com/AleoHQ/snarkOS/issues/716

network/src/inbound/inbound.rs
howardwu: Agreed, will do!

network/src/inbound/inbound.rs
howardwu: Good catch, thanks

network/src/peers/peers.rs
```diff
@@ -356,79 +369,88 @@ impl<S: Storage + Send + Sync + 'static> Node<S> {
...
         }
 
-        self.outbound.channels.write().remove(&remote_address);
-
-        let result = self.peer_book.set_disconnected(remote_address);
-        debug!("Set {} as disconnected", remote_address);
-        result
-        // TODO (howardwu): Attempt to blindly send disconnect message to peer.
+        // Remove the peer from the channel.
+        if self.outbound.channels.write().remove(&remote_address).is_some() {
```
ljedrz, line 429: limiting the disconnect to this is fine _unless_ it's disconnecting from a peer that is not `connected` but still `connecting`, otherwise there will be lingering `connecting` entries

network/src/node.rs
howardwu: I think this would require changing the function signature from:
```rust
pub async fn start_services(&self)
```
to
```rust
pub async fn start_services(static self)
```

howardwu: I believe this would require `Node` to have a static lifetime in the method

network/src/node.rs
howardwu: Good catch, thanks!

network/src/peers/peers.rs
howardwu: agreed, thanks!

network/src/node.rs
howardwu:
```bash
   --> network/src/node.rs:177:33
    |
177 |     pub async fn start_services(&self) {
    |                                 ^^^^^
    |                                 |
    |                                 this data with an anonymous lifetime `'_`...
    |                                 ...is captured here...
...
265 |             let sync_block_task = task::spawn(async move {
    |                                   ----------- ...and is required to live as long as `'static` here
```

ljedrz: the node would be accessed via `sync_clone.node()` inside the task; that being said, it's not any blocker

niklaslong (approved):
Looks good with what was mentioned above 👍

network/src/peers/peers.rs
howardwu: Discussed offline, updated

network/src/node.rs
howardwu: Ah good catch, I was using `sync`

howardwu: Fixed

network/src/peers/peers.rs
```diff
@@ -286,22 +286,44 @@ impl<S: Storage + Send + Sync + 'static> Node<S> {
...
+        // Check if this node server is below the permitted number of connected peers.
+        let min_peers = self.config.minimum_number_of_connected_peers() as usize;
+        if number_of_peers >= min_peers {
+            return;
+        }
+
+        // Set the number of peers to attempt a connection to.
+        let count = min_peers - number_of_peers;
+
+        trace!("Connecting to {} disconnected peers", count);
```
ljedrz, line 310: `count` should be replaced with `std::cmp::min(count, self.peer_book.disconnected_peers().len())` in order not to display a number that's higher than the number of known disconnected peers if `min_peers` is higher

network/src/peers/peers.rs
```diff
@@ -286,22 +286,44 @@ impl<S: Storage + Send + Sync + 'static> Node<S> {
...
 
         // Iterate through a selection of random peers and attempt to connect.
         let random_peers = self
             .peer_book
             .disconnected_peers()
             .iter()
             .map(|(k, _)| k)
+            .filter(|peer| **peer != own_address)
```
ljedrz, line 318: this filter should also exclude the bootnodes, which we might already actively be trying to connect to

network/src/peers/peers.rs
howardwu: Resolved in post networking

network/src/peers/peers.rs
howardwu: Resolved in post networking

network/src/peers/peer_info.rs
howardwu: Resolved in post networking

PR #719, opened 2021-04-28
Lockless connection writes, fix a peering error

ljedrz:
This PR changes the connection write architecture so that writes are queued in per-connection `channel`s and processed in per-connection tasks. This allows us to remove the locks inside the `ConnWriter` object.

In addition, a logic bug in `update_peers` that caused connections to be dropped is fixed and some constants are tweaked.

Re-enabling the temporarily-disabled fuzz tests required the following changes:
- making fatal errors cause disconnects again (those errors indicate that the connection is unrecoverable)
- removing `async` from `Peers::disconnect_from_peer`

Fixes https://github.com/AleoHQ/snarkOS/issues/717.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -53,7 +53,7 @@ jobs:
           name: Build and run tests
           no_output_timeout: 60m
           command: >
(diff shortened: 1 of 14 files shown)


niklaslong (approved):

ljedrz:
The soon-to-time-out `nightly` circleCI run (and hopefully all the future runs as well) should be fixed with https://github.com/AleoHQ/snarkOS/pull/720.

PR #722, opened 2021-04-29
Make a lot more methods non-async, tweak peer timestamps and sync logic

ljedrz:
After https://github.com/AleoHQ/snarkOS/pull/719, many methods can become non-async, as they are no longer using any asynchronous calls; in addition, changing the `Inbound` channel's sending logic to `Sender::try_send` instead of `::send` allows it to become non-async too. The slight downside is that overflowing messages will not get delivered (but there's the sync and propagation logic to fix that), but the benefit is that the node is not susceptible to being "flooded" with messages waiting to be processed; that channel has a fairly large capacity too.

In addition, in order to avoid situations where a node would ask for a block sync while still processing queued sync blocks (which can now happen when a peer disconnects just after having sent a batch), this PR leverages the block sync expiry mechanism more, so that a disconnect from a peer that has just provided sync blocks no longer automatically changes the node's state to `Idle`.

Finally, there's a small performance tweak that omits a peer's timestamp update if their message is a sync block, avoiding updating it many times when processing a batch of sync blocks.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -316,6 +316,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "circular-queue"
+version = "0.2.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1d34327ead1c743a10db339de35fb58957564b99d248a67985c55638b22c59b5"
+dependencies = [
+ "version_check",
+]
+
+[[package]]
 name = "clang-sys"
 version = "1.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2558,7 +2567,9 @@ dependencies = [
  "capnp",
  "capnpc",
  "chrono",
+ "circular-queue",
  "derivative",
+ "fxhash",
  "hex",
  "log",
  "once_cell",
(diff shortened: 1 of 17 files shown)


ljedrz:
note: while that sync expiry tweak should help the situation, I'm still occasionally getting batches with older blocks

ljedrz:
Based on testnet observations, it seems likely that the `Block` messages might be propagated too much in the network; even though the original sender is excluded from reverse propagation, it can still get back it from a peer that has received it from another common peer.

Therefore, I introduced a quick and simple cache for recently processed `Block` messages. It uses a small circular buffer containing fast hashes of entries, and new unique entries push the oldest ones out. Since `Block`s aren't mined too often, a small cache should be sufficient.

ljedrz:
Since the outbound channel is still often full, I increased their depth to 4x their previous size, i.e. 4096.

ljedrz:
Deferred, since https://github.com/AleoHQ/snarkOS/pull/725 needs `async` in some of the methods that would lose it in this PR.

ljedrz:
The changes in this PR have already been split out into other PRs, so it can be closed now.

PR #725, opened 2021-04-29
Lockless

Protryon:
Overview of changes:
* Remove high dependence on sync & async locks
* Replace node management with an actor model to improve consistency & encapsulation


network/src/drop_join.rs
```diff
@@ -0,0 +1,86 @@
...
+
+impl<X: Send + Sync + 'static> DropJoinable for tokio::task::JoinHandle<X> {
+    fn destroy(self) {
+        self.abort();
+    }
+}
+
+impl<X: Send + Sync + 'static> DropJoinable for std::thread::JoinHandle<X> {
+    fn destroy(self) {
+        // todo: this could be blocking in an async thread??
```
ljedrz, line 31: it can be; it would probably be a good idea to launch a `tokio::task::spawn_blocking` over this operation

benchmarks/syncing/syncing.rs
```diff
@@ -36,7 +36,8 @@ fn providing_sync_blocks(c: &mut Criterion) {
 
     let blocks = TestBlocks::load(NUM_BLOCKS);
     for block in &blocks.0 {
-        provider.expect_sync().consensus.receive_block(&block).unwrap();
+        rt.block_on(provider.expect_sync().consensus.receive_block(&block))
```
ljedrz, line 39: would it make sense to wrap the loop in a `task::spawn_blocked`, so that it is processed in one of the threads dedicated to blocking in its entirety?

network/src/drop_join.rs
Protryon: agreed, done

benchmarks/syncing/syncing.rs
Protryon: does it matter here? its a prelude for a benchmark run with a controlled runtime.

network/src/outbound/connection_writer.rs
```diff
@@ -73,12 +73,6 @@ impl ConnWriter {
             self.writer.write_all(&self.buffer[..encrypted_len]).await?;
         }
 
-        // If message is a `SyncBlock` message, log it as a trace.
-        match payload {
-            Payload::SyncBlock(_) => trace!("Sent a '{}' message to {}", payload, self.addr),
-            _ => debug!("Sent a '{}' message to {}", payload, self.addr),
```
ljedrz, line 79: I'm assuming we're avoiding the log here due to this call being "hot"? However, we should still log this someplace, possibly when the message is sent to the connection's `Sender`.

Protryon: no this was a rebase failure, will readd

benchmarks/syncing/syncing.rs
ljedrz: ah yes, didn't notice it was benchmark code :ok_hand:

ljedrz:
The tests seem to have run into the issue with the current RPC server implementation. It is likely that https://github.com/AleoHQ/snarkOS/pull/759 will be needed in order for this PR to work.

ljedrz:
Alternatively, you could pass the `Handle` to the `tokio` runtime to the `Miner` as it used to be like before the unused `async` cleanups (https://github.com/AleoHQ/snarkOS/pull/755/commits/b5e2154cb6d859e971542198407fad1120e22d42) and use it directly.

network/src/config.rs
```diff
@@ -32,7 +33,7 @@ pub struct Config {
     /// The maximum number of peers permitted to maintain connections with.
     maximum_number_of_connected_peers: u16,
     /// The default bootnodes of the network.
-    pub bootnodes: RwLock<Vec<SocketAddr>>,
```
ljedrz, line 35: note for the future: I don't think that list is ever modified during runtime

network/src/peers/peer_book.rs
```diff
@@ -70,11 +71,11 @@ impl From<SerializedPeerBook> for PeerBook {
 #[derive(Debug, Default)]
 pub struct PeerBook {
     /// The map of the addresses currently being handshaken with.
-    connecting_peers: RwLock<HashSet<SocketAddr>>,
+    connecting_peers: MpmcMap<SocketAddr, Instant>,
```
ljedrz, line 74: we no longer need an `Instant` to check for any timeouts: the handshake timeout mechanism takes care of this
```suggestion
    connecting_peers: MpmcMap<SocketAddr, ()>,
```

network/src/peers/peer_book.rs
```diff
@@ -206,40 +207,40 @@ impl PeerBook {
...
     /// and adds the given address to the disconnected peers in this `PeerBook`.
     ///
-    pub fn set_disconnected(&self, address: SocketAddr) -> bool {
+    pub async fn set_disconnected(&self, address: SocketAddr) -> bool {
         // Case 1 - The given address is a connecting peer, attempt to disconnect.
-        if self.connecting_peers.write().remove(&address) {
+        if self.connecting_peers.remove(address).await.is_some() {
             metrics::decrement_gauge!(stats::CONNECTIONS_CONNECTING, 1.0);
-            return false;
+            return true;
```
ljedrz, line 231:
```suggestion
            return false;
```
this should still be `false`, as it's only used to indicate whether the disconnected peer had been `connected`, and not `connecting`

network/src/peers/peer_info.rs
```diff
@@ -91,8 +108,7 @@ pub struct PeerInfo {
     /// The bool indicates whether it's abortable - otherwise it
     /// needs to be awaited instead.
     #[serde(skip)]
-    #[allow(clippy::type_complexity)]
-    pub tasks: Arc<Mutex<Vec<(task::JoinHandle<()>, bool)>>>,
+    tasks: DropJoin<task::JoinHandle<()>>,
```
ljedrz, line 111: we need to be able to `.await` the outbound connection tasks instead of just `.abort()ing` them, otherwise we will not be able to accurately measure the capacity of outbound channels over time - can we add such a functionality to `DropJoin`?

network/src/peers/peer_info.rs
```diff
@@ -183,21 +207,10 @@ impl PeerInfo {
         self.quality.expecting_pong.store(false, Ordering::SeqCst);
         self.quality.remaining_sync_blocks.store(0, Ordering::SeqCst);
 
-        for (handle, abortable) in self.tasks.lock().drain(..).rev() {
-            if abortable {
-                handle.abort();
-            } else {
-                task::spawn(async move {
-                    // An arbitrary amount of time to allow the task to shut down cleanly.
-                    if tokio::time::timeout(Duration::from_secs(5), handle).await.is_err() {
```
ljedrz, line 192: we need to maintain this functionality

network/src/peers/peers.rs
```diff
@@ -425,39 +427,18 @@ impl<S: Storage + Send + Sync + 'static> Node<S> {
...
 
-    /// TODO (howardwu): Implement manual serializers and deserializers to prevent forward breakage
-    ///  when the PeerBook or PeerInfo struct fields change.
-    ///
-    /// Stores the current peer book to the given storage object.
-    ///
-    /// This function serializes the peer book into a byte vector for storage.
-    ///
-    #[inline]
-    pub(crate) fn save_peer_book_to_storage(&self) -> Result<(), NetworkError> {
```
ljedrz, line 436: I think it's safe to say we won't want to use this function in this shape (since we don't want to keep the `PeerBook` in the storage belonging to `Sync`), so we might as well remove it for now :+1:

rpc/src/rpc_impl.rs
```diff
@@ -266,7 +265,8 @@ impl<S: Storage + Send + core::marker::Sync + 'static> RpcFunctions for RpcImpl<
                     transaction,
                 };
 
-                if let Ok(inserted) = self.memory_pool()?.lock().insert(&storage, entry) {
+                // this block_on will halt the tokio worker until insert completion -- can cause problems if not in a multi-threaded environment (tests)
+                if let Ok(inserted) = futures::executor::block_on(self.memory_pool()?.insert(&storage, entry)) {
```
ljedrz, line 269: do we have a test for this call? I'm worried it might panic with the current RPC server impl - we should test it first (could be with a regular node) to make sure it doesn't happen

ljedrz: ok, there is a test, but it uses the internal testing machinery of `jsonrpc`; we should also test it on a regular node

storage/src/mem.rs
```diff
@@ -17,9 +17,7 @@
 use crate::NUM_COLS;
 use snarkvm_objects::{DatabaseTransaction, Op, Storage, StorageError};
 
-use parking_lot::RwLock;
```
ljedrz, line 20: why not just use `parking_lot`? we are going to have it as an indirect dependency regardless, so is there any benefit to using the `std` rendition?

testing/examples/test_blocks.rs
```diff
@@ -49,13 +49,13 @@ fn mine_block<S: Storage>(
...
     // add it to the chain
-    miner.consensus.receive_block(&block)?;
+    miner.consensus.receive_block(&block).await?;
 
     let new_block_height = miner.consensus.ledger.get_current_block_height();
     assert_eq!(old_block_height + 1, new_block_height);
 
     // Duplicate blocks dont do anything
-    miner.consensus.receive_block(&block)?;
+    miner.consensus.receive_block(&block).await.ok(); // throws a duplicate error -- seemingly intentional
```
ljedrz, line 58: it's a very old test, so perhaps it's an old API where duplicate blocks hadn't returned an error yet

testing/src/network/topology.rs
```diff
@@ -56,7 +56,9 @@ async fn line(nodes: &mut Vec<Node<LedgerStorage>>) {
     // Start each node with the previous as a bootnode.
     for node in nodes {
         if let Some(addr) = prev_node {
-            node.config.bootnodes.write().push(addr);
+            let mut bootnodes = (&**node.config.bootnodes.load()).clone();
+            bootnodes.push(addr);
+            node.config.bootnodes.store(Arc::new(bootnodes));
```
ljedrz, line 61: ok, so I guess this would be that one use we have for runtime modification of bootnodes

ljedrz:
ok, I did another pass; @niklaslong can you take a look as well to make sure I didn't miss anything? it's a rather large set of changes, so the more reviewers, the better

network/src/peers/peer_book.rs
```diff
@@ -70,11 +71,11 @@ impl From<SerializedPeerBook> for PeerBook {
 #[derive(Debug, Default)]
 pub struct PeerBook {
     /// The map of the addresses currently being handshaken with.
-    connecting_peers: RwLock<HashSet<SocketAddr>>,
+    connecting_peers: MpmcMap<SocketAddr, Instant>,
     /// The map of connected peers to their metadata.
-    connected_peers: RwLock<HashMap<SocketAddr, PeerInfo>>,
+    connected_peers: MpmcMap<SocketAddr, PeerInfo>,
```
ljedrz, line 76: I'm wondering if we shouldn't keep locks over these 3 collections, or at least change them at a later time; today I noticed that due to the async nature of connections we sometimes attempt to set a peer as `connected` more than once, which is fixed in https://github.com/AleoHQ/snarkOS/pull/779 via synchronous checks for the peer's status

ljedrz: After running a node on this branch for a while I've noticed that the mismatch between the numbers of connected/disconnected peers has grown compared to the current setup, so I'd propose to postpone changing these 3 particular instances of `RwLock` to `MpmcMap` until later, preferably after https://github.com/AleoHQ/snarkOS/pull/779.

ljedrz: The same happens to the calculation of `connecting` peers (and outside of metrics), which can break connection logic.

rpc/src/rpc_impl.rs
ljedrz: I just checked with a different RPC (`disconnect_protected`), and using `block_on` makes it hang, presumably due to the RPC server's `tokio 0.2` executor being single-threaded. We can't reliably use `await` in our RPC (unless via a direct `Handle`) until https://github.com/AleoHQ/snarkOS/pull/759 goes in.

ljedrz:
After having tested these changes for a while, and having detected some possible issues mentioned in the review comments, I've come to the conclusion that in order to avoid introducing tricky bugs, we should be introducing these changes gradually.

I propose that the snarkVM update is done first, including the related `MerkleTree` changes; we process blocks in a synchronous fashion, so there should be little that could go wrong here. The `ArcSwap` and `AtomicInstant` changes could go in next, and in the end, once the improved connection accounting and custom RPC server are in, we will be at a better spot to start replacing `RwLock<HashMap<T, U>>`s with `MpmcMap<T, U>`.

network/src/config.rs
Protryon: just for tests I think

storage/src/mem.rs
Protryon: I guess. This MemDb seems unused everywhere, as far as I can tell. Will change.

rpc/src/rpc_impl.rs
Protryon: yes. the test was what prompted this comment. would be good to test in staging definitely. the right solution is to not use synchronous jsonrpc.

network/src/peers/peer_info.rs
Protryon: Then we don't even need `DropJoin` for the tasks where this is applicable. If the futures are spawned in tokio tasks, then tokio will continue to poll them directly.

Protryon: Is the concern here a race over the metrics reporting after reading?

network/src/peers/peer_info.rs
Protryon: what is the purpose of this functionality? i would expect an immediate termination to be the expected behavior. is there a concern with not flushing packets to network fully, or something to that effect?

network/src/peers/peer_info.rs
ljedrz: The issue is that when outbound messages get queued (and bump the related metric showing how many messages are currently queued to be sent), but the corresponding `Receiver` doesn't `recv` them (e.g. when the owning task gets `abort()`ed abruptly), we are stuck with lingering entries that over time suggest that we constantly have outbound messages waiting to be sent, while in reality we are sending them immediately.

network/src/peers/peer_info.rs
ljedrz: This is the same thing as described in https://github.com/AleoHQ/snarkOS/pull/725#discussion_r637777974.

network/src/peers/peer/cipher.rs
```diff
@@ -0,0 +1,85 @@
...
+pub struct Cipher {
+    state: TransportState,
+    buffer: Box<[u8]>,
+    noise_buffer: Box<[u8]>,
+}
+
+impl Cipher {
+    pub fn new(state: TransportState, buffer: Box<[u8]>, noise_buffer: Box<[u8]>) -> Self {
+        assert_eq!(buffer.len(), crate::MAX_MESSAGE_SIZE);
+        assert_eq!(noise_buffer.len(), crate::MAX_MESSAGE_SIZE);
```
ljedrz, line 33: the `noise` buffer doesn't need to be larger than `65535B`, as that's the max message limit in the specification

network/src/peers/peer/cipher.rs
```diff
@@ -0,0 +1,85 @@
...
+
+pub struct Cipher {
+    state: TransportState,
+    buffer: Box<[u8]>,
+    noise_buffer: Box<[u8]>,
+}
+
+impl Cipher {
+    pub fn new(state: TransportState, buffer: Box<[u8]>, noise_buffer: Box<[u8]>) -> Self {
+        assert_eq!(buffer.len(), crate::MAX_MESSAGE_SIZE);
```
ljedrz, line 32: the read buffer can be greater than the protocol's max message size for performance reasons

network/src/peers/peer/connector.rs
```diff
@@ -0,0 +1,76 @@
...
+
+use super::PeerAction;
+use crate::{stats, NetworkError, Node, Peer, PeerEvent, PeerEventData, PeerHandle, Version};
+
+use super::network::PeerNetwork;
+
+impl Peer {
+    async fn inner_connect(&mut self, our_version: Version) -> Result<PeerNetwork, NetworkError> {
+        metrics::increment_gauge!(stats::CONNECTIONS_CONNECTING, 1.0);
+        defer::defer(|| metrics::decrement_gauge!(stats::CONNECTIONS_CONNECTING, 1.0));
```
ljedrz, line 28: without binding the return value to something, this closure will be executed right here, meaning we'll never see a rise in `CONNECTIONS_CONNECTING`

network/src/peers/peer/handshake.rs
```diff
@@ -0,0 +1,281 @@
...
+        crate::HANDSHAKE_PATTERN
+            .parse()
+            .expect("Invalid noise handshake pattern!"),
+        Box::new(snow::resolvers::SodiumResolver),
+    );
+    let static_key = builder.generate_keypair()?.private;
+    let noise_builder = builder.local_private_key(&static_key).psk(3, crate::HANDSHAKE_PSK);
+    let mut noise = noise_builder.build_responder()?;
+    let mut buffer: Box<[u8]> = vec![0u8; crate::MAX_MESSAGE_SIZE].into();
+    let mut noise_buffer: Box<[u8]> = vec![0u8; crate::MAX_MESSAGE_SIZE].into();
```
ljedrz, line 56: Same note as in https://github.com/AleoHQ/snarkOS/pull/725/files#r637784386

network/src/peers/peer/connector.rs
Protryon: good point

network/src/peers/peer/cipher.rs
Protryon: how does it help with performance? and its always allocated to be that size currently

ljedrz: it might have a positive impact on performance in the future (since we could then read bytes from the stream faster, queuing them in memory), but at this point in time the important thing is that the `buffer` is at least as big as `crate::MAX_MESSAGE_SIZE`, i.e. `assert!(buffer.len() >= crate::MAX_MESSAGE_SIZE);` would be a better pick here

Protryon: that won't have a performance improvement here. we could use a buffered read stream, yeah, but right now its read exact length packets, and because we don't do (a lot) of small reads, buffered read stream is probably negligible. this buffer is just for incoming packets, not for a general read buffer.

network/src/peers/peer/connector.rs
```diff
@@ -22,13 +28,23 @@ use crate::{stats, NetworkError, Node, Peer, PeerEvent, PeerEventData, PeerHandl
...
+
 impl Peer {
     async fn inner_connect(&mut self, our_version: Version) -> Result<PeerNetwork, NetworkError> {
         metrics::increment_gauge!(stats::CONNECTIONS_CONNECTING, 1.0);
         let _x = defer::defer(|| metrics::decrement_gauge!(stats::CONNECTIONS_CONNECTING, 1.0));
 
-        let stream = TcpStream::connect(self.address).await?;
-        self.inner_handshake_initiator(stream, our_version).await
+        let tcp_stream;
+        select! {
```
ljedrz, line 39: `tokio::time::timeout` is probably a simpler pick here

consensus/src/memory_pool.rs
```diff
@@ -117,9 +127,10 @@ impl<T: Transaction> MemoryPool<T> {
 
         let mut holding_serial_numbers = vec![];
         let mut holding_commitments = vec![];
-        let mut holding_memos = Vec::with_capacity(self.transactions.len());
+        let mut holding_memos = Vec::with_capacity(self.transactions.inner().len());
```
ljedrz, line 130:
```suggestion
        let mut holding_memos = Vec::with_capacity(self.transactions.len());
```

network/src/inbound/inbound.rs
```diff
@@ -77,36 +73,19 @@ impl<S: Storage + Send + Sync + 'static> Node<S> {
             loop {
                 match listener.accept().await {
                     Ok((stream, remote_address)) => {
-                        info!("Got a connection request from {}", remote_address);
-
                         if !node_clone.can_connect() {
-                            metrics::increment_counter!(stats::CONNECTIONS_ALL_REJECTED);
```
ljedrz, line 83: why is this counter removed? not accepting a connection due to connection limits is something we'd like to know about, as it could indicate that the limit is too low

network/src/inbound/inbound.rs
```diff
@@ -192,8 +119,6 @@ impl<S: Storage + Send + Sync + 'static> Node<S> {
             unreachable!("All messages processed sent to the inbound receiver are Inbound");
         };
 
-        self.peer_book.register_message(source);
```
ljedrz, line 195: why is this removed? it updates the number of messages received from the specific peer and its last activity timestamp

network/src/inbound/inbound.rs
```diff
@@ -262,32 +189,32 @@ impl<S: Storage + Send + Sync + 'static> Node<S> {
                     if sync.is_empty() {
                         // An empty `Sync` is unexpected, as `GetSync` requests are only
                         // sent to peers that declare a greater block height.
-                        self.peer_book.register_failure(source);
```
ljedrz, line 265: if we no longer use this method and instead register peer failures using some alternative means, the method should be removed

network/src/inbound/inbound.rs
```diff
@@ -262,32 +189,32 @@ impl<S: Storage + Send + Sync + 'static> Node<S> {
...
                         warn!("{} doesn't have sync blocks to share", source);
-                    } else if self.peer_book.expecting_sync_blocks(source, sync.len()) {
-                        trace!("Received {} sync block hashes from {}", sync.len(), source);
-                        self.received_sync(source, sync);
+                        if let Some(peer) = self.peer_book.get_peer_handle(source) {
+                            peer.soft_fail().await;
+                        }
+                    } else {
```
ljedrz, line 196: as `clippy` suggests, this `else { if .. }` block can be collapsed

network/src/node.rs
```diff
@@ -55,60 +54,34 @@ pub struct InnerNode<S: Storage> {
...
-    tasks: Mutex<Vec<task::JoinHandle<()>>>,
+    tasks: DropJoin<task::JoinHandle<()>>,
     /// The threads spawned by the node.
-    threads: Mutex<Vec<thread::JoinHandle<()>>>,
+    threads: DropJoin<thread::JoinHandle<()>>,
     /// An indicator of whether the node is shutting down.
     shutting_down: AtomicBool,
 }
 
-impl<S: Storage> Drop for InnerNode<S> {
```
ljedrz, line 74: why is this `Drop` impl removed? I assume it's because the strong reference count to `InnerNode` has increased and there is no chance for it to be triggered? this might have been the case for some time now

network/src/peers/peer/cipher.rs
```diff
@@ -0,0 +1,105 @@
...
+
+pub struct Cipher {
+    state: TransportState,
+    buffer: Box<[u8]>,
+    noise_buffer: Box<[u8]>,
+}
+
+impl Cipher {
+    pub fn new(state: TransportState, buffer: Box<[u8]>, noise_buffer: Box<[u8]>) -> Self {
+        assert_eq!(buffer.len(), crate::MAX_MESSAGE_SIZE + 4096);
```
ljedrz, line 34: what's the surplus size for?

network/src/peers/peer/connector.rs
```diff
@@ -0,0 +1,93 @@
...
+use futures::{select, FutureExt};
+use snarkvm_objects::Storage;
+use tokio::{net::TcpStream, sync::mpsc};
+
+use super::PeerAction;
+use crate::{stats, NetworkError, Node, Peer, PeerEvent, PeerEventData, PeerHandle, Version};
+
+use super::network::PeerNetwork;
+
+const CONNECTION_TIMEOUT_SECS: u64 = 3;
```
ljedrz, line 31: we might want to keep it with the other crate-level consts for easier tweaking, and I think we already have one for it there

network/src/peers/peer/connector.rs
```diff
@@ -0,0 +1,93 @@
...
+
+impl Peer {
+    async fn inner_connect(&mut self, our_version: Version) -> Result<PeerNetwork, NetworkError> {
```
ljedrz, line 34: this name could be more informative, e.g. `perform_handshake`

network/src/peers/peer/connector.rs
```diff
@@ -0,0 +1,93 @@
...
+                        .await
+                        .ok();
+                    if let Err(e) = self.run(node, network, receiver).await {
+                        self.quality.failures += 1;
+                        error!(
+                            "unrecoverable failure communicating to outbound peer '{}': '{:?}'",
+                            self.address, e
+                        );
+                    }
+                    metrics::decrement_gauge!(stats::CONNECTIONS_CONNECTED, 1.0);
```
ljedrz, line 79: shouldn't this decrement belong to the `if let Err(e)` block above?

network/src/peers/peer_book.rs
```diff
@@ -14,503 +14,263 @@
...
-    #[inline]
-    pub fn is_connecting(&self, address: SocketAddr) -> bool {
-        self.connecting_peers.read().contains(&address)
+impl PeerBook {
+    pub fn spawn() -> Self {
+        let (sender, receiver) = mpsc::channel(256);
+        let peers = PeerBook {
+            disconnected_peers: Default::default(),
+            connected_peers: Default::default(),
+            pending_connections: Arc::new(AtomicU32::default()),
```
ljedrz, line 84:
```suggestion
            pending_connections: Default::default(),
```

network/src/peers/peer/handshake.rs
```diff
@@ -0,0 +1,288 @@
...
+pub struct HandshakeData {
+    pub version: Version,
+    pub noise: TransportState,
+    pub buffer: Box<[u8]>,
+    pub noise_buffer: Box<[u8]>,
+}
+
+async fn responder_handshake<W: AsyncWrite + Unpin, R: AsyncRead + Unpin>(
+    remote_address: SocketAddr,
+    own_version: &Version,
```
ljedrz, line 42: if I recall correctly, `Version` is now `Copy`, so no need to borrow it

network/src/inbound/inbound.rs
ljedrz: ok, I see it's removed, the diff was just collapsed

network/src/inbound/inbound.rs
ljedrz: I see it's been moved into the peer code; the diff was just collapsed

network/src/peers/peer/network.rs
```diff
@@ -0,0 +1,72 @@
...
+    net::tcp::{OwnedReadHalf, OwnedWriteHalf},
+};
+
+use crate::{NetworkError, Payload};
+
+use super::cipher::Cipher;
+
+// used in integration tests
+#[doc(hidden)]
+pub struct PeerNetwork {
```
ljedrz, line 28: this name isn't great, as it's just a single connection; maybe a `PeerStream` or a `PeerConnection` would be better here?

network/src/peers/peer/outbound_handler.rs
```diff
@@ -0,0 +1,162 @@
...
+}
+
+#[derive(Clone, Debug)]
+pub struct PeerHandle {
+    pub(super) sender: mpsc::Sender<PeerAction>,
+}
+
+impl PeerHandle {
+    pub async fn load(&self) -> Option<Peer> {
+        metrics::increment_gauge!(stats::QUEUES_OUTBOUND, 1.0);
```
ljedrz, line 45: the handling of this stat doesn't seem correct; it should only be incremented when a message is queued to be sent to a peer (and decremented when the message has been pushed to the stream)

ljedrz: if we are interested in peer-handling queues to also be measured, they need to have their own new metric

network/src/peers/peer/outbound_handler.rs
```diff
@@ -0,0 +1,162 @@
...
+        self.sender.send(PeerAction::ExpectingSyncBlocks(amount)).await.ok();
+    }
+
+    pub async fn soft_fail(&self) {
+        metrics::increment_gauge!(stats::QUEUES_OUTBOUND, 1.0);
+        self.sender.send(PeerAction::SoftFail).await.ok();
+    }
+}
+
+pub(super) enum PeerResponse {
```
ljedrz, line 92: this name is a bit misleading, as it might suggest a network response; maybe a `PeerHandlerResponse` or something similar would work better

network/src/peers/peer/outbound_handler.rs
```diff
@@ -0,0 +1,162 @@
...
+    ) -> Result<PeerResponse, NetworkError> {
+        metrics::decrement_gauge!(stats::QUEUES_OUTBOUND, 1.0);
+        match message {
+            PeerAction::Disconnect => Ok(PeerResponse::Disconnect),
+            PeerAction::Send(message) => {
+                if matches!(message, Payload::Ping(_)) {
+                    self.quality.expecting_pong = true;
+                    self.quality.last_ping_sent = Some(Instant::now());
+                }
+                network.write_payload(&message).await?;
```
ljedrz, line 111: only this operation should decrement `QUEUES_OUTBOUND`

network/src/peers/peer/peer.rs
```diff
@@ -0,0 +1,146 @@
...
+        Self {
+            address,
+            status: PeerStatus::Disconnected,
+            quality: Default::default(),
+            is_bootnode,
+            network_failures: 0,
+        }
+    }
+
+    pub fn judge(&self) -> bool {
```
ljedrz, line 65: this name could be more informative, e.g. `is_faulty` or `should_be_kicked`; `judge` on its own doesn't suggest much

network/src/peers/peer/peer.rs
```diff
@@ -0,0 +1,146 @@
...
+        } else {
+            Self::peer_handshake_timeout()
+        }
+    }
+
+    pub fn peer_handshake_timeout() -> Duration {
+        Duration::from_secs(crate::HANDSHAKE_PEER_TIMEOUT_SECS as u64)
+    }
+
+    pub(super) async fn run<S: Storage + Send + Sync + 'static>(
```
ljedrz, line 81: this name could be more informative, e.g. `spawn_message_handlers` or `handle_network_messages`

network/src/peers/peer/peer.rs
```diff
@@ -0,0 +1,146 @@
...
+                        break;
+                    }
+                    let message = message.unwrap();
+                    match self.process_message(&mut network, message).await? {
+                        PeerResponse::Disconnect => break,
+                        PeerResponse::None => (),
+                    }
+                },
+                data = read_receiver.recv().fuse() => {
+                    if data.is_none() {
```
ljedrz, line 115: I think a single `match` would be more readable here, we can match on `Some(Ok(data))`, `Some(Err(e))` and `None` at once

network/src/peers/peer/peer_quality.rs
```diff
@@ -0,0 +1,77 @@
...
+        let last_seen = self.last_seen;
+        if let Some(last_seen) = last_seen {
+            now - last_seen > chrono::Duration::seconds(crate::MAX_PEER_INACTIVITY_SECS.into())
+        } else {
+            // in the peer book, but never been connected to before
+            false
+        }
+    }
+
+    pub fn see(&mut self) {
```
ljedrz, line 56: this name could be better, e.g. `register_activity` or `update_timestamp`

network/src/peers/peer/receiver.rs
```diff
@@ -0,0 +1,91 @@
...
+        stream: TcpStream,
+        our_version: Version,
+    ) -> Result<(Peer, PeerNetwork), NetworkError> {
+        metrics::increment_gauge!(stats::CONNECTIONS_CONNECTING, 1.0);
+        let _x = defer::defer(|| metrics::decrement_gauge!(stats::CONNECTIONS_CONNECTING, 1.0));
+
+        Peer::inner_handshake_responder(remote_address, stream, our_version).await
+    }
+
+    pub fn receive<S: Storage + Send + Sync + 'static>(
```
ljedrz, line 39: this name could be improved, e.g. `receive_connection`

network/src/peers/peer/receiver.rs
```diff
@@ -0,0 +1,91 @@
...
+use snarkvm_objects::Storage;
+use tokio::{net::TcpStream, sync::mpsc};
+
+use super::PeerAction;
+use crate::{stats, NetworkError, Node, Peer, PeerEvent, PeerEventData, PeerHandle, PeerStatus, Version};
+
+use super::network::PeerNetwork;
+
+impl Peer {
+    async fn inner_receive(
```
ljedrz, line 28: this name could be better, e.g. `respond_to_handshake`

network/src/peers/peer/receiver.rs
```diff
@@ -0,0 +1,91 @@
...
+        tokio::spawn(async move {
+            let (mut peer, network) = match Peer::inner_receive(remote_address, stream, node.version()).await {
+                Err(e) => {
+                    error!(
+                        "failed to receive incoming connection from peer '{}': '{:?}'",
+                        remote_address, e
+                    );
+                    event_target
+                        .send(PeerEvent {
+                            address: SocketAddr::new(IpAddr::V4(Ipv4Addr::new(0, 0, 0, 0)), 0),
```
ljedrz, line 55:
```suggestion
                            address: SocketAddr::new(IpAddr::V4(Ipv4Addr::UNSPECIFIED), 0),
```

ljedrz: that being said, maybe it would be better for `PeerEvent.address` to be an `Option<SocketAddr>` instead?

network/src/peers/peer/receiver.rs
```diff
@@ -0,0 +1,91 @@
...
+                        })
+                        .await
+                        .ok();
+                    return;
+                }
+                Ok(x) => x,
+            };
+
+            peer.set_connected();
+            metrics::increment_gauge!(stats::CONNECTIONS_CONNECTED, 1.0);
```
ljedrz, line 66: maybe that gauge could be incremented within `Peer::set_connected`? it would make sense, same as for `Peer::set_disconnected` to decrement it (and to increment `CONNECTIONS_DISCONNECTED`)

network/src/peers/peer_book.rs
```diff
@@ -14,503 +14,263 @@
...
+}
+
+impl PeerBookRef {
+    // gets terminated when sender is dropped from PeerBook
+    async fn handle_peer_events(self, mut receiver: mpsc::Receiver<PeerEvent>) {
+        while let Some(event) = receiver.recv().await {
+            match event.data {
+                PeerEventData::Connected(handle) => {
+                    self.pending_connections.fetch_sub(1, Ordering::SeqCst);
+                    if let Some(old_peer) = self.connected_peers.insert(event.address, handle).await {
```
ljedrz, line 57: can we even run into a double-connection like this? we should at least `WARN` when this happens, as it's undesirable

network/src/peers/peer_book.rs
```diff
@@ -14,503 +14,263 @@
...
-        peer_info.set_connected();
-
-        // Add the address into the connected peers.
-        if self.connected_peers.write().insert(listener, peer_info).is_none() {
-            metrics::increment_gauge!(stats::CONNECTIONS_CONNECTED, 1.0);
-        } else {
-            error!("{} is already a connected peer!", listener);
+    /// concurrently iterates over peers
+    async fn for_each_peer<F: Future<Output = ()>, FN: Fn(PeerHandle) -> F>(&self, func: FN) {
+        let mut futures = vec![];
```
ljedrz, line 182:
```suggestion
        let mut futures = Vec::with_capacity(self.connected_peers.len());
```

network/src/peers/peer_book.rs
```diff
@@ -14,503 +14,263 @@
...
+        futures::future::join_all(futures).await.into_iter().flatten().collect()
+    }
 
-        // Case 2 - The given address is a connected peer, attempt to disconnect.
-        if let Some(mut peer_info) = self.connected_peers.write().remove(&address) {
-            // Update the peer info to disconnected.
-            peer_info.set_disconnected();
-
-            metrics::decrement_gauge!(stats::CONNECTIONS_CONNECTED, 1.0);
+    pub async fn judgement(&self) {
```
ljedrz, line 201: this name is not informative enough; it could be e.g. `check_quality_of_peers` or `evaluate_peers`

network/src/peers/peer_book.rs
```diff
@@ -14,503 +14,263 @@
...
-            }
-        } else {
-            // shouldn't occur, but just in case
-            warn!("Received a Pong from an unknown peer: {}!", source);
-        }
+    pub async fn last_seen(&self) -> Option<SocketAddr> {
+        self.connected_peers_snapshot()
+            .await
+            .into_iter()
+            .max_by(|a, b| a.quality.last_seen.cmp(&b.quality.last_seen))
```
ljedrz, line 252:
```suggestion
            .max_by_key(|peer| peer.quality.last_seen)
```

ljedrz:
After another pass and having done some local testing, I have the following observations:
- this PR should have the miner-, mempool- and peer-related changes split into individual smaller PRs, as it should be possible to isolate them
- there are some issues with metrics related to connection handling and queue utilization, and it's visible in tests; right now they are a bit scattered around, but instead there should be only a single spot where they are incremented and decremented so that we don't miss anything
- with this setup I think we can attempt to connect to a single peer multiple times, which we should avoid - there shouldn't be a need to replace an existing connection with a new one

network/src/node.rs
Protryon: `connected_peers` no longer needs aborting due to channel drop = close. i think i discounted the other two, but yeah this is unreachable anyways, even for tests.

network/src/peers/peer/cipher.rs
Protryon: so we can fit the 4-byte header (we seemed to limit length of packet not including the header). did some extra because someone asked for it IIRC, or just a clean page size. don't recall. if you care, we can probably tune it. might have been related to noise padding bug or something like that.

network/src/peers/peer/connector.rs
Protryon: is it a crate-level const? it's only used for connections, it make sense to keep it local to the connector to me. open to further discussion.

network/src/peers/peer/connector.rs
Protryon: the intent is to abstract the tcp connection and handshake. this function also does more than just initiate the handshake, it starts a TCP connection it completes the handshake. it literally `connect`s.

network/src/peers/peer/connector.rs
Protryon: no, this will always fire on connection termination, regardless of error/failure state.

network/src/peers/peer/handshake.rs
Protryon: i see i actually made that change (to make it `Copy`). that was a mistake, it should be clone, since it may contain strings and such in the future. will update

network/src/peers/peer/network.rs
Protryon: the intent is more akin to `PeerNetworkHandle` or a `PeerIOHandle` imo. might not be a stream or a connection -- it's intent is to abstract the read/write/IO stuff. will rename

network/src/peers/peer/outbound_handler.rs
Protryon: messages and other actions are in the same queue now, it makes sense to me to keep it in one metric for now, since anything in that queue will have perf implications and possible IO. open to adding more metrics. do you know if `metrics` crate supports tagged metrics? would be super useful here.

network/src/peers/peer/outbound_handler.rs
Protryon: i disagree, this abstraction layer is separate from IO entirely, it's adjacent to `PeerAction`, and is local anyways.

network/src/peers/peer/peer.rs
Protryon: changed to `judge_bad`

network/src/peers/peer/peer.rs
Protryon: that breaks abstraction. the external world sees a peer `run` and they don't need to know that the peer handles network messages or spawns message handlers.

network/src/peers/peer/peer_quality.rs
Protryon: personal taste, disagree

network/src/peers/peer/receiver.rs
Protryon: abstraction, disagree

network/src/peers/peer/receiver.rs
Protryon: adds a lot of bloat to use option, will use `UNSPECIFIED`

network/src/peers/peer_book.rs
Protryon: it can happen in some (mostly) synthetic cases like two peers send outbound connections to each other at the exact same time. will add a log

Protryon: also hung connections we haven't timed out yet, but the peer reconnects

network/src/peers/peer_book.rs
Protryon: unneeded microoptimization, but sure, ill add it

network/src/peers/peer_book.rs
Protryon: disagree, readability

network/src/peers/peer_book.rs
ljedrz: the new setup does plenty of allocations, and if we can avoid reallocating with zero extra cost (like here), we should

network/src/peers/peer/outbound_handler.rs
ljedrz: prometheus can utilize labels: https://prometheus.io/docs/practices/naming/; heads up: the structure of the names is important, otherwise tools like grafana don't recognize them properly

network/src/peers/peer/outbound_handler.rs
ljedrz: it's just a matter of increased clarity - we do call other nodes "peers" after all, so if someone was looking at this part of the code without being aware of the whole picture, they might be confused

network/src/peers/peer/peer.rs
ljedrz: exposing internal methods to the outside world would break abstraction, but this is again just a matter of increasing readability - someone who is browsing the code but has not digested all of it yet might see `peer.run(...)` and not be able to tell what it does within the context where it's called, being forced to jump to its source

network/src/peers/peer/receiver.rs
Protryon: it gets managed by code lifecycle here, which is relatively failproof and easy to read. `set_connected` is currently kinda useless since peers internal state isnt used for anything right now AFAIK

network/src/peers/peer/cipher.rs
howardwu: If we intend to keep this padding, `MAX_MESSAGE_SIZE` should be defined as an equation, e.g.:
```rust
/// Message size composed of size header and message.
const MAX_MESSAGE_SIZE: usize = 12345 + 4096;
```

network/src/peers/peer/connector.rs
howardwu: This should be moved to where all of the other magical constants are set

network/src/peers/peer/connector.rs
```diff
@@ -0,0 +1,107 @@
...
+
+use super::PeerAction;
+use crate::{stats, NetworkError, Node, Peer, PeerEvent, PeerEventData, PeerHandle, Version};
+
+use super::network::PeerIOHandle;
+
+const CONNECTION_TIMEOUT_SECS: u64 = 3;
+
+impl Peer {
+    async fn inner_connect(&mut self, our_version: Version) -> Result<PeerIOHandle, NetworkError> {
```
howardwu, line 34: Can we add some comments defining the difference between an inner_connect and a connect?

Also please move all private functions to the end of the impl section

network/src/peers/peer/outbound_handler.rs
```diff
@@ -0,0 +1,164 @@
...
+
+impl PeerHandle {
+    pub async fn load(&self) -> Option<Peer> {
+        metrics::increment_gauge!(stats::QUEUES_OUTBOUND, 1.0);
+        let (sender, receiver) = oneshot::channel();
+        self.sender.send(PeerAction::Get(sender)).await.ok()?;
+        receiver.await.ok()
+    }
+
+    pub async fn judge_bad(&self) {
```
howardwu, line 49: What does judge_bad mean? Can you please use a more clear name

network/src/peers/peer/outbound_handler.rs
```diff
@@ -0,0 +1,164 @@
...
+        metrics::increment_gauge!(stats::QUEUES_OUTBOUND, 1.0);
+        self.sender.send(PeerAction::GotSyncBlock).await.ok();
+    }
+
+    pub async fn expecting_sync_blocks(&self, amount: u32) {
+        metrics::increment_gauge!(stats::QUEUES_OUTBOUND, 1.0);
+        self.sender.send(PeerAction::ExpectingSyncBlocks(amount)).await.ok();
+    }
+
+    pub async fn soft_fail(&self) {
```
howardwu, line 80: What defines a soft_fail?

network/src/peers/peer/outbound_handler.rs
```diff
@@ -0,0 +1,164 @@
...
+        self.sender.send(PeerAction::QualityJudgement).await.ok();
+    }
+
+    /// returns true if disconnected, false if not connected anymore
+    pub async fn disconnect(&self) -> bool {
+        metrics::increment_gauge!(stats::QUEUES_OUTBOUND, 1.0);
+        self.sender.send(PeerAction::Disconnect).await.is_ok()
+    }
+
+    pub async fn send_payload(&self, payload: Payload) {
```
howardwu, line 60: Is there a difference between send_payload and send_message?

network/src/peers/peer/outbound_handler.rs
```diff
@@ -0,0 +1,164 @@
...
+        metrics::increment_gauge!(stats::QUEUES_OUTBOUND, 1.0);
+        self.sender.send(PeerAction::Send(payload)).await.ok();
+    }
+
+    pub async fn cancel_sync(&self) {
+        metrics::increment_gauge!(stats::QUEUES_OUTBOUND, 1.0);
+        self.sender.send(PeerAction::CancelSync).await.ok();
+    }
+
+    pub async fn got_sync_block(&self) {
```
howardwu, line 70: This should be renamed to `received_sync_block` to follow existing convention

network/src/peers/peer/mod.rs
```diff
@@ -0,0 +1,37 @@
...
+
+pub use outbound_handler::*;
+pub use peer::*;
+pub use peer_quality::*;
+// used in integration tests
+#[doc(hidden)]
+pub use cipher::Cipher;
+#[doc(hidden)]
+pub use network::*;
+pub use peer_events::*;
```
howardwu, line 37: Please cleanup this import declaration

howardwu:
There's a lot going on in this PR. Please update the README.md with an architecture design doc for us to review.

network/src/peers/peer/connector.rs
Protryon: this is a fairly common pattern to be able to use try operators (`?`), since Rust only lets us use them at a function-return level scope. there isn't a difference other than one handles errors more freely (versus `connect` which has very verbose error handling)

network/src/peers/peer/outbound_handler.rs
Protryon: `judge_bad` means to judge if a peer is bad. can you be more specific what your looking for? a description of what is judging perhaps?

network/src/peers/peer/outbound_handler.rs
Protryon: was avoiding a namespace conflict with `fail` i think -- will rename to `fail`. it's just a failure being dispatched from outside the peer

network/src/peers/peer/outbound_handler.rs
Protryon: where do you see `send_message`? this file only has `send_payload`. we don't use `Message` structs because there is no routing or directionality left unclear when this code is called (it's outbound to a specific peer)

network/src/peers/peer/outbound_handler.rs
Protryon: thats fine; but what convention? is there a standard somewhere? just historical? please give more context

network/src/peers/peer/mod.rs
Protryon: ill order them up i suppose, is there anything your looking for in particular here?

PR #728, opened 2021-05-03
Peer handling improvements

ljedrz:
This PR improves the peer-handling logic by introducing the following changes:
- disconnect from completely inactive peers according to a `MAX_PEER_INACTIVITY_SECS` const
- enforce `max-peers` for bootnodes (they should just be started with high `min-peers` and `max-peers` values)
- don't limit the number of entries saved from peer lists
- use the node's `id` (formerly `name`) to avoid self-connections

In addition, there are 2 more `ignore`d tests that check connection cleanups. It doesn't seem that any sockets "leak", but it's possible that in real-life scenarios some disconnects are not visible to the node, which should be solved with the new check for inactivity.

note: the last of these changes changes the cap'n proto serialization logic, but it's backward-compatible.

should fix https://github.com/AleoHQ/snarkOS/issues/726

### DIFF
--- a/network/src/inbound/inbound.rs
+++ b/network/src/inbound/inbound.rs
@@ -78,7 +78,7 @@ impl<S: Storage + Send + Sync + 'static> Node<S> {
             (listener_address, listener)
         };
         self.set_local_address(listener_address);
-        info!("Initializing listener for node ({:x})", self.name);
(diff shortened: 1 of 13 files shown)


niklaslong (approved):
👍

ljedrz:
Rebased.

PR #734, opened 2021-05-07
Allow concurrent connection requests

ljedrz:
The inbound connection requests are currently processed in sequence; under normal conditions this is fine and avoids a situation where a node would accept too many connections at once, but it is susceptible to connection request spam attacks, even with the handshake timeout mechanism we have in place.

This PR makes inbound connection processing concurrent, by spawning tasks dedicated to handshake handling as soon as a connection is accepted. In order to avoid connecting to too many peers and having to rely only on the peer cleanup mechanism, the `max_peer` limit is enforced, and the node registers all rejected requests in its `Stats`.

There's also a dedicated test case that verifies this new behavior.

Cc https://github.com/AleoHQ/snarkOS/issues/730 (since limiting inbound connections will have a positive impact on the `Too many open files` issues and reduces the possibility of OOMs).

### DIFF
--- a/network/src/inbound/inbound.rs
+++ b/network/src/inbound/inbound.rs
@@ -95,7 +95,6 @@ impl<S: Storage + Send + Sync + 'static> Node<S> {
                                 .connections
                                 .all_rejected
                                 .fetch_add(1, Ordering::Relaxed);
-                            warn!("Maximum number of connections reached; rejecting {}", remote_address);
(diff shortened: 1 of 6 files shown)


ljedrz:
note: most of the diff comes from the extra indentation; the extent of actual changes is quite small

niklaslong (approved):

PR #735, opened 2021-05-10
A few tweaks

ljedrz:
This PR introduces a few low-impact tweaks, the most notable of which are:
- set the node's state to `Syncing` as soon as a sync provider is chosen; the previous behavior was that a node would only set its state to `Syncing` once it has started receiving actual `SyncBlocks`, but low block sync interval could cause a `GetSync` request to be sent twice
- be more strict about non-critical peer failures; reduce the penalty disconnect threshold from `> 10` to `>= 3`

Cc https://github.com/AleoHQ/snarkOS/issues/730.


niklaslong (approved):
LGTM

PR #759, opened 2021-05-18
Temporary custom rpc server impl

ljedrz:
We are currently using a deprecated crate for our JSON-RPC server implementation, which is very sub-optimal, as it carries plenty of old dependencies (mostly due to it depending on `tokio 0.2`), makes us unable to `await` in the RPC methods, and makes reasoning about resources very tricky.

This PR introduces an alternative JSON-RPC server implementation which is compatible with all the existing calls; it's built on `hyper` (like the current version) and is fully compatible with the rest of our setup.

Once the successor to the `jsonrpc-*` crates (`jsonrpsee`) is ready, we'll be able to use it instead. But for now, a custom implementation is a far better option than a mismatch in async executors.

note: as a small drive-by, the `disconnect` RPC is slightly refactored so that its implementation is aligned with all the other protected calls.

Cc https://github.com/AleoHQ/snarkOS/issues/754

### DIFF
--- a/rpc/src/custom_rpc_server.rs
+++ b/rpc/src/custom_rpc_server.rs
@@ -1,0 +1,335 @@
+// Copyright (C) 2019-2021 Aleo Systems Inc.
+// This file is part of the snarkOS library.
+
+// The snarkOS library is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
+// The snarkOS library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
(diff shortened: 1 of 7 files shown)


rpc/src/custom_rpc_server.rs
```diff
@@ -0,0 +1,335 @@
...
+                .map_err(convert_crate_err);
+            serde_json::to_value(resp).unwrap_or_default()
+        }
+        "decoderawtransaction" => {
+            let resp = rpc
+                .decode_raw_transaction(params[0].as_str().unwrap_or("").into())
+                .map_err(convert_crate_err);
+            serde_json::to_value(resp).unwrap_or_default()
+        }
+        "sendtransaction" => {
```
niklaslong, line 192: Some naming is inconsistent (in the spec). 
```suggestion
        "sendrawtransaction" => {
```

niklaslong (approved):
LGTM 👍 
I made a small comment regarding naming consistency in the spec.

rpc/src/custom_rpc_server.rs
ljedrz: That's fair; let's leave this open for later in case we want to alter the name of the endpoint so that it matches the naming convention of the other `raw` ones.

PR #774, opened 2021-05-20
Increment the block height metric

ljedrz:
`Sync::received_block` is triggered by both the node's miner and on receipt of new and valid `Block`/`SyncBlock`; `Block` entries are passed on to `::propagate_block`, while `SyncBlock`s aren't, so they need to trigger the metric update separately.

Fixes https://github.com/AleoHQ/snarkOS/issues/773.


niklaslong (approved):

niklaslong:
This is updating correctly now 👍 🎉

PR #779, opened 2021-05-21
Improve the accounting of connected peers

ljedrz:
This PR deduplicates common post-handshake connection maintenance logic into a common, synchronous `Node::set_connected` method which should be able to detect double connection attempts with much greater accuracy. In addition, extra `ERROR` logs are added to indicate if there is a gap in the related logic.

Fixes https://github.com/AleoHQ/snarkOS/issues/778.

### DIFF
--- a/network/src/peers/peer_book.rs
+++ b/network/src/peers/peer_book.rs
@@ -214,9 +214,11 @@ impl PeerBook {
         peer_info.set_connected();
 
         // Add the address into the connected peers.
-        self.connected_peers.write().insert(listener, peer_info);
-
-        metrics::increment_gauge!(stats::CONNECTIONS_CONNECTED, 1.0);
+        if self.connected_peers.write().insert(listener, peer_info).is_none() {
+            metrics::increment_gauge!(stats::CONNECTIONS_CONNECTED, 1.0);
+        } else {
+            error!("{} had already been a connected peer!", listener);
+        }
     }
 
     ///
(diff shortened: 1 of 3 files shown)


network/src/peers/peer_book.rs
```diff
@@ -214,9 +214,11 @@ impl PeerBook {
         peer_info.set_connected();
 
         // Add the address into the connected peers.
-        self.connected_peers.write().insert(listener, peer_info);
-
-        metrics::increment_gauge!(stats::CONNECTIONS_CONNECTED, 1.0);
+        if self.connected_peers.write().insert(listener, peer_info).is_none() {
+            metrics::increment_gauge!(stats::CONNECTIONS_CONNECTED, 1.0);
+        } else {
+            error!("{} had already been a connected peer!", listener);
```
niklaslong, line 220:
```suggestion
            error!("{} is already a connected peer!", listener);
```

network/src/peers/peer_book.rs
```diff
@@ -265,31 +269,22 @@ impl PeerBook {
...
-                .connected_peers()
-                .get(&address)
-                .cloned()
-                .ok_or(NetworkError::PeerBookMissingPeer);
+            return self.connected_peers().get(&address).cloned();
         }
 
         // Check if the address is a known disconnected peer.
-        if self.is_disconnected(address) {
+        if !only_if_connected && self.is_disconnected(address) {
```
niklaslong, line 282: Isn't this check already covered by `self.is_connected`?

niklaslong (commented):
Left a few comments, otherwise looks good 👍 
I'll run a node on this branch for a while to check Grafana/Prometheus.

network/src/peers/peer_book.rs
ljedrz: in this particular case yes, but I'd like to keep that extra param for the future, in case either option is desired

ljedrz:
There seems to be an outage of CircleCI services; the tests will need to be re-run at a later time.

ljedrz:
Confirmed to be working quite well:
image
note: the branch that produced the results above also contained https://github.com/AleoHQ/snarkOS/pull/777

niklaslong (approved):

Issue #780, opened 2021-05-22
[Bug] Unable to get block locator hashes from storage


ljedrz:
## 🐛 Bug Report

The node might report the following error:
```
ERROR: Unable to get block locator hashes from storage
```
Which makes it unable to perform a block sync with its peers.

raychu86:
Likely an issue with the line `block_locator_hashes.push(self.get_block_hash(index)?);` in the `get_block_locator_hashes` method. Somehow the entry in the `COL_BLOCK_LOCATOR` for the given block number is missing.

damons:
My miner has hit this.

ljedrz:
>Somehow the entry in the COL_BLOCK_LOCATOR for the given block number is missing.
Yes, because the `Ledger::decommit_latest_block` method currently doesn't update the `KEY_BEST_BLOCK_NUMBER` value.

ljedrz:
`KEY_BEST_BLOCK_NUMBER` is not typically used to obtain the block height, but it's used to set it up during node's startup.

ljedrz:
This issue doesn't appear to occur anymore; after encountering many forks and having done thorough testing around it, I think it can safely be closed.

PR #781, opened 2021-05-24
Don't omit the first 2 operations in decommit_latest_block

ljedrz:
This PR fixes a logic error found while investigating https://github.com/AleoHQ/snarkOS/issues/780. When a node detects that it's on a non-canon fork and decides to decommit one or more of its latest blocks, it currently doesn't include 2 operations, as the database transaction is erroneously started twice in the `Ledger::decommit_latest_block` method.

Cc https://github.com/AleoHQ/snarkOS/issues/780; additional fixes might be needed.

### DIFF
--- a/storage/src/objects/dpc_state.rs
+++ b/storage/src/objects/dpc_state.rs
@@ -129,15 +129,23 @@ impl<T: Transaction, P: LoadableMerkleParameters, S: Storage> Ledger<T, P, S> {
     }
 
     /// Rebuild the stored merkle tree with the current stored commitments
-    pub fn update_merkle_tree(&self) -> Result<(), StorageError> {
+    pub fn update_merkle_tree(&self, new_best_block_number: u32) -> Result<(), StorageError> {
         self.rebuild_merkle_tree(vec![])?;
+        let new_digest = self.cm_merkle_tree.read().root();
 
-        let update_current_digest = DatabaseTransaction(vec![Op::Insert {
+        let mut database_transaction = DatabaseTransaction::new();
+
+        database_transaction.push(Op::Insert {
+            col: COL_DIGEST,
+            key: to_bytes![new_digest]?.to_vec(),
+            value: new_best_block_number.to_le_bytes().to_vec(),
+        });
+        database_transaction.push(Op::Insert {
             col: COL_META,
             key: KEY_CURR_DIGEST.as_bytes().to_vec(),
-            value: to_bytes![self.cm_merkle_tree.read().root()]?.to_vec(),
-        }]);
+            value: to_bytes![new_digest]?.to_vec(),
+        });
 
-        self.storage.batch(update_current_digest)
+        self.storage.batch(database_transaction)
     }
 }
--- a/consensus/tests/consensus_sidechain.rs
+++ b/consensus/tests/consensus_sidechain.rs
@@ -157,4 +157,41 @@ mod consensus_sidechain {
 
         assert_eq!(old_block_height, new_block_height);
     }
+
+    #[test]
+    fn decommit() {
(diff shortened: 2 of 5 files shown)


Protryon (approved):

ljedrz:
It seems that 2 related tests need to be adjusted; I'm on it.

ljedrz:
Actually, it seems that it was a merkle tree digest update that was missing; still, I'm not too knowledgeable about that code, so it would be best if @howardwu or @raychu86 could take a closer look here.

raychu86:
The error from #780 shouldn't have to deal directly with Merkle tree updates. 

On a high level, the commitment merkle tree is composed of the commitments of the canon blocks in the longest valid chain. So, when a block is committed/decommitted, the merkle tree must be reconstructed with the new valid commitments.

The issue from #780 seems more like a missing entry in the `COL_BLOCK_LOCATORS` table.

storage/src/objects/dpc_state.rs
```diff
@@ -128,15 +128,23 @@ impl<T: Transaction, P: LoadableMerkleParameters, S: Storage> Ledger<T, P, S> {
...
     /// Rebuild the stored merkle tree with the current stored commitments
-    pub fn update_merkle_tree(&self) -> Result<(), StorageError> {
+    pub fn update_merkle_tree(&self, new_best_block_number: u32) -> Result<(), StorageError> {
         self.rebuild_merkle_tree(vec![])?;
+        let new_digest = self.cm_merkle_tree.read().root();
 
-        let update_current_digest = DatabaseTransaction(vec![Op::Insert {
+        let mut database_transaction = DatabaseTransaction::new();
+
+        database_transaction.push(Op::Insert {
```
raychu86, line 137: Not entirely sure if this operation is necessary. Because we only need to insert into `COL_DIGEST` when a new block is committed and remove when a block is decommitted.

ljedrz: Without this additional operation, the `consensus_sidechain::{fork_out_of_order, new_out_of_order}` fail due to a `Ledger digest is invalid.` error ([CI logs](https://app.circleci.com/pipelines/github/AleoHQ/snarkOS/2606/workflows/a21ff086-f4c7-4bb9-beca-a2b180f67363/jobs/6439)).

ljedrz:
Rebased and added an extra commit.

ljedrz:
Added a commit with a block storage validator tool.

storage/src/objects/dpc_state.rs
ljedrz: Unless one of the 2 currently inactive operations:
```
database_transaction.push(Op::Delete {
    col: COL_DIGEST,
    key: self.current_digest()?,
});
```
should just be deleted - then this change is not needed.

ljedrz:
Rebased and squashed the block validator commits together.

ljedrz:
I am currently able to bring a broken database to a state where it appears to be coherent in terms of block relationships, but it seems we might have some logic mismatch in canon block checks, as the `CanonChain` path (suggesting a valid canonical parent) in `Ledger::receive_block` can encounter a parent verification error later down the road in `Ledger::process_block`.

ljedrz:
Moved the block validator tool to its own PR, https://github.com/AleoHQ/snarkOS/pull/797.

ljedrz:
@howardwu this PR is good to go now and the newly added test does fail with the current `staging`; the only open question is whether the `COL_DIGEST` operation should be added (i.e. it was earlier omitted by mistake) or removed entirely - other than that, this change may fix #780 entirely.

ljedrz:
Some CircleCI workflows have failed due to some CircleCI incident; I'll rerun them a bit later.

howardwu (approved):

PR #784, opened 2021-05-25
Fix testnet1 bootnode IPs

howardwu:
## Motivation

Testnet1 bootnode IPs were being commented out in a previous PR, this has been fixed.


ljedrz:
note: some of these bootnodes seem to be disconnecting from my node pretty often with an EOF, which suggests that they might be registered under 2 (or more) network interfaces; I checked them with the `getnodeinfo` RPC and noticed that their `listening_addr` has the `0.0.0.0` IP, which could be causing this - in order to avoid it, all the bootnodes should be launched with a specified `--ip X` parameter, where `X` is not `0.0.0.0`, but the corresponding bootnode address in the `TESTNET_BOOTNODES` list.

howardwu:
That's great to know, is there an issue with our self-discovery? I'm certain we had logic to identify and set the local address

PR #786, opened 2021-05-25
Remove the Result wrapper from RPC responses

ljedrz:
Fixes https://github.com/AleoHQ/snarkOS/issues/785 and uses a different crate with a more convenient API for RPC objects.

In addition, this PR avoids a potential panic that had been applicable to the previous implementation as well.


niklaslong (approved):
LGTM 👍

PR #787, opened 2021-05-25
Register the number of received RPC requests

ljedrz:
This PR adds a new metric registering the number of received RPC requests.

### DIFF
--- a/rpc/Cargo.toml
+++ b/rpc/Cargo.toml
@@ -81,6 +81,9 @@ version = "17"
 [dependencies.jsonrpc-derive]
 version = "17"
 
+[dependencies.metrics]
+version = "0.15"
+
 [dependencies.parking_lot]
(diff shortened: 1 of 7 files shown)


rpc/Cargo.toml
```diff
@@ -81,6 +81,9 @@ version = "17"
 [dependencies.jsonrpc-derive]
 version = "17"
 
+[dependencies.metrics]
+version = "0.15"
```
niklaslong, line 85: latest is `16.0`?

niklaslong (approved):
Small comment regarding `metrics` version, other than that it looks good.

rpc/Cargo.toml
ljedrz: yep, but we'll need to perform a workspace-wide update regardless, and it's better to do that in its own PR

Issue #789, opened 2021-05-25
[Feature] CI builds for each individual crate in the workspace


howardwu:
## 🚀 Feature

Implement CI build checks for every crate in the workspace

PR #790, opened 2021-05-26
Update config.yml with new CI for crates, update snarkVM to `0.3.1`

howardwu:
## Motivation

- Update config.yml with new CI for crates
- Update snarkVM to `0.3.1`


howardwu:
#789

PR #796, opened 2021-05-26
Add a protected shutdown RPC

ljedrz:
The node doesn't currently have a "clean" way of getting shut down. This RPC endpoint provides us with one.

### DIFF
--- a/network/src/node.rs
+++ b/network/src/node.rs
@@ -329,6 +329,8 @@ impl<S: Storage + Send + core::marker::Sync + 'static> Node<S> {
         for handle in self.tasks.lock().drain(..).rev() {
             handle.abort();
         }
+
+        info!("The node can be shut down now");
     }
 
     pub fn register_task(&self, handle: task::JoinHandle<()>) {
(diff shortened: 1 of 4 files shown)


ljedrz:
Note: once this PR is merged, the bootnodes that are updated to it should always be started with custom `rpc-username` and `rpc-password` params; we might also want to randomize at least the password, as suggested by an existing `TODO` in `config.rs`.

damons:
Are RPC calls made over insecure sockets in clear text?

ljedrz:
It's currently simple HTTP authentication, but as long as we don't plan to expose RPC to the outside world, it's fine.

howardwu:
Should we wait for better security enforcements/protocols on RPC protected endpoints before merging this in? It seems to be a significant dos concern

ljedrz:
We can postpone this PR for any amount of time we need, as it's small and self-contained. In the meantime, if we need this functionality for any ad-hoc test, we can just cherry-pick the only commit it contains, and I can rebase whenever necessary :+1:.

PR #797, opened 2021-05-28
Introduce a block storage validator tool

ljedrz:
Split out from https://github.com/AleoHQ/snarkOS/pull/781 to keep that one focused on bugfixes only.

Introduces a new method, `Ledger::validate`, that currently verifies that all the blocks are stored in a coherent way, i.e. it starts from the current height, checks if the best block's header and number are registered, and goes to its parent and performs the same check on it, including one that verifies that the previously checked block is one of its children, and so on up to the genesis block or the given block number limit.

Cc https://github.com/AleoHQ/snarkOS/issues/780.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2838,6 +2838,7 @@ dependencies = [
  "snarkvm-parameters",
  "snarkvm-utilities",
  "thiserror",
+ "tracing",
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


ljedrz:
The `snarkos-network` failure is transient and unrelated.

Protryon (approved):

howardwu (approved):

PR #804, opened 2021-05-31
Increase the default peer counts

ljedrz:
Since there are no apparent performance/stability issues when a node is connected to 100 peers, it should be quite safe to increase the default `min-peers` and `max-peers` values. This PR changes them from `7` and `25` to `20` and `50`, which should improve the quality of the network.


ljedrz:
CircleCI is still encountering issues; will restart later.

howardwu (approved):

PR #809, opened 2021-05-31
Fix duplicate sync blocks

ljedrz:
This PR should already solve the problem of duplicate sync blocks being received, but it doesn't fully address the underlying issue (i.e. the possibility of a node to have forked far enough for the sync providers to not recognize any of the received locator hashes) yet.

There are 2 changes:
1. don't ask for blocks with hashes that already belong to the node's canon blocks
2. when obtaining locator hashes, instead of the current behavior (get at most 20 last hashes and then go down quadratically), the node:
- divides its block height by 16 (this value was picked due to it producing a similar number of locator hashes as the current approach)
- picks hash indices quadratically (but starting with 4 instead of 1) until the step is greater than the result of the initial division
- once the quadratic step is greater than the result of the initial division, divide the number of remaining hashes by 16 again
- pick hash indices in a proportional fashion (stepping by the result of the new division) until above index/height 0
- provide the genesis block hash

Compare the current values returned for a height of 269k (around the current height of the testnet):
```
269000
/* follows with the next top 19 block indices */
268999
268998
268997
268996
268995
268994
268993
268992
268991
268990
268989
268988
268987
268986
268985
268984
268983
268982
268981
/* starting to go quadratic */
268979
268975
268967
268951
268919
268855
268727
268471
267959
266935
264887
260791
252599
236215
203447
137911 // a step of ~65k
6839 // a step of ~131k
0
```

with the new behavior:
```
269000
/* immediately goes quadratic, starting at 4 */
268996
268988
268972
268940
268876
268748
268492
267980
266956
264908
260812
252620
236236
/ *switching to a proportional divisor */
221472
206708
191944
177180
162416
147652
132888
118124
103360
88596
73832
59068
44304
29540
14776
12
0
```

The value of the proportional divisor could be higher and could depend on the height; this would make the size of the requests greater too, though, so it might be a good idea to keep it on the small side at least initially, and possibly increasing it once the node detects that it has a hard time obtaining unique sync hashes from peers.

Feedback is most welcome.

Cc https://github.com/AleoHQ/snarkOS/issues/783.


ljedrz:
After having done plenty of testing with this change, I can say that due to having more fine-grained height intervals it at least greatly decreases the probability of receiving duplicate initial blocks.

niklaslong (commented):
Seems good, the only concern I have is the one you've already raised with regards to the growth of the hash list length over time.

ljedrz:
> the growth of the hash list length over time

The current approach actually already produces a longer list (not by a lot, but noticeably) than the proposed alternative.

niklaslong (approved):

ljedrz:
Rebased; at this point it's limited to making the locator hashes more granular, in order to not collide with any sync changes from `lockless` that got merged recently.

ljedrz:
Hmm, that `snarkos-consensus` failure is weird, as syncing is not involved in those tests; I'll look into it locally.

ljedrz:
There was a bug in how the indices were calculated for very low heights; it's fixed now.

storage/src/objects/block_header.rs
```diff
@@ -64,29 +64,35 @@ impl<T: TransactionScheme, P: LoadableMerkleParameters, S: Storage> Ledger<T, P,
         // Start from the latest block and work backwards
         let mut index = self.get_current_block_height();
 
-        // Update the step size with each iteration
-        let mut step = 1;
+        // The number of "chunks" into which the height is to be divided after the quadratic
+        // step used initially grows too high.
+        let divisor = std::cmp::min(16u32, std::cmp::max(index, 1));
```
ljedrz, line 69: the desired divisor is `16`, but for block heights lower than that, it's the block height or `1` for a block height of `0`

niklaslong (approved):

ljedrz:
Rebased.

ljedrz:
For the reference, this is how the difference would look like in the current testnet (height ~360k):

```
old behavior (38 indices):
360000
359999
359998
359997
359996
359995
359994
359993
359992
359991
359990
359989
359988
359987
359986
359985
359984
359983
359982
359981
359979
359975
359967
359951
359919
359855
359727
359471
358959
357935
355887
351791
343599
327215
294447
228911
97839
0

new behavior (31 indices):
360000
359996
359988
359972
359940
359876
359748
359492
358980
357956
355908
351812
343620
327236
306784
286332
265880
245428
224976
204524
184072
163620
143168
122716
102264
81812
61360
40908
20456
4
0
```

and the divisor can easily be tweaked if desired (and possibly increased ad-hoc when the previous sync was unsuccessful), giving the following results for a value of e.g. `24`:
```
new behavior (38 indices):
360000
359996
359988
359972
359940
359876
359748
359492
358980
357956
355908
351812
343620
329303
314986
300669
286352
272035
257718
243401
229084
214767
200450
186133
171816
157499
143182
128865
114548
100231
85914
71597
57280
42963
28646
14329
12
0
```

ljedrz:
In addition, 3 last commits introduce cleanups to some related functions.

ljedrz:
It occurred to me that it would be a good idea to always send the desired number of locator hashes, so that we can make better assumptions about their quality; in the last commit I adjusted the new algorithm to always produce the maximum specified number of block locator hashes, and the comparison between the 3 algorithms can be seen [here](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=aa4458e2b70e4242de39c18c7e4f8ec5).

raychu86:
The initial implementation puts more emphasis on the first 10 blocks because ideally, most forks are corrected within a few blocks. The new approach certainly gives more information as a whole, but generally skips them and starts indexing with a larger initial step. 

IMO the first few blocks are the most important (in a stable network) and we shouldn't be skipping over them.

ljedrz:
@raychu86 the initial implementation actually goes over the top **20** blocks first, which seems like plenty in stable conditions; I could tweak the new approach so that it includes the top **N** blocks first, with **10** sounding like a reasonable number.

ljedrz:
I updated the PR to always provide the top 10 block hashes; the distances then go quadratically, and finally proportionally, like before; an updated playground link: https://play.rust-lang.org/?version=stable&mode=release&edition=2018&gist=9308e7305106f71ca9a7388a79fd11b3.

howardwu:
One minor nit: can we keep with the quadratic growth starting from `2` (rather than `4`) as this PR defines.

ljedrz:
Done.

howardwu (approved):

PR #820, opened 2021-06-02
Update snarkVM to 0.5.3

ljedrz:


ljedrz:
@howardwu the `snarkos-toolkit` and `WASM Tests` jobs will fail due to a feature mismatch; `toolkit` would compile if its `snarkvm-dpc` dependency didn't exclude `default-features`, but since changing it might have WASM implications, I'll leave that one to you, in case any new feature needs to be isolated.

parameters/examples/create_genesis_block.rs
```diff
@@ -50,7 +50,7 @@ pub fn generate<C: BaseDPCComponents>() -> Result<Vec<u8>, TransactionError> {
         difficulty_target: 0x07FF_FFFF_FFFF_FFFF_u64,
         nonce: 0,
         pedersen_merkle_root_hash: PedersenMerkleRootHash([0u8; 32]),
-        proof: ProofOfSuccinctWork::default(),
+        proof: ProofOfSuccinctWork([0; 972]),
```
ljedrz, line 53: The `Default` impl is no longer available, so I just hardcoded the value it used to provide.

howardwu: This is correct, thank you for calling that out.

howardwu:
@ljedrz can you file a PR in snarkVM with the feature flag change to `*/Cargo.toml`?

ljedrz:
Rebased.

ljedrz:
@howardwu how would you like the features to look like? I was able to compile `snarkos-toolkit` on its own by introducing the following new feature to `snarkvm-dpc`:
```
wasm = [
  "full",
  "snarkvm-algorithms/wasm",
  "snarkvm-gadgets/full",
  "snarkvm-parameters/remote",
]
```
and by adding `features = [ "wasm" ]` to `[dependencies.snarkvm-dpc]` in `snarkos-toolkit`, but I'm not sure if this is the way it should be done.

toolkit/Cargo.toml
```diff
@@ -26,20 +26,16 @@ path = "benches/account.rs"
...
 features = [ "wasm" ]
 
 [dependencies.snarkvm-dpc]
-version = "0.4.0"
-default-features = false
-
-[dependencies.snarkvm-objects]
-version = "0.4.0"
+version = "0.5.1"
 default-features = false
```
howardwu, line 35:
```suggestion
default-features = false
features = [ "wasm" ]
```

howardwu:
@ljedrz can you update all snarkVM imports to 0.5.2?

ljedrz:
I updated the TOML files, but while `snarkos-toolkit` was fixed, the WASM tests still failed; I found the cause and filed a fix in https://github.com/AleoHQ/snarkVM/pull/192.

ljedrz:
Updated to snarkVM 0.3.

ljedrz:
The tests are green @howardwu, the only failure is some newly activated `clippy` lints on nightly.

howardwu (approved):

PR #823, opened 2021-06-03
Parallel Syncing

Protryon:


network/src/sync/blocks.rs
```diff
@@ -89,10 +106,11 @@ impl<S: Storage + Send + std::marker::Sync + 'static> Node<S> {
         };
 
         info!(
-            "Received block from {} of epoch {} with hash {:?}",
+            "Received block from {} of epoch {} with hash {:?} (current head {})",
             remote_address,
             block_struct.header.time,
-            hex::encode(block_struct.header.get_hash().0)
+            hex::encode(block_struct.header.get_hash().0),
```
ljedrz, line 112:
```suggestion
            block_struct.header.get_hash()
```
it's not something from this PR, but this alternative, simpler form should suffice, paired with `{}` instead of `{:?}` in the format literal

network/src/sync/master.rs
```diff
@@ -0,0 +1,304 @@
...
+}
+
+struct SyncBlock {
+    address: SocketAddr,
+    block: Vec<u8>,
+}
+
+impl<S: Storage + Send + Sync + 'static> SyncMaster<S> {
+    pub fn new(node: Node<S>) -> (Self, mpsc::Sender<SyncInbound>) {
+        let (sender, receiver) = mpsc::channel(256);
```
ljedrz, line 43:
```suggestion
        let (sender, receiver) = mpsc::channel(crate::MAX_BLOCK_SYNC_COUNT);
```

network/src/sync/master.rs
```diff
@@ -0,0 +1,304 @@
...
+                },
+                _ = timeout => {
+                    break;
+                }
+            }
+        }
+    }
+
+    async fn receive_sync_hashes(&mut self, max_message_count: usize) -> HashMap<SocketAddr, Vec<BlockHeaderHash>> {
+        const TIMEOUT: u64 = 15;
```
ljedrz, line 124: I'd make this into a crate-level `const` so it's easier to keep track of

network/src/sync/master.rs
Protryon: We can receive more than 64 blocks since some peers will send us different blocks. Also doesn't hurt to have buffer (basically free)

network/src/sync/master.rs
Protryon: its specific to this function and shouldn't be reused.

network/src/sync/master.rs
```diff
@@ -0,0 +1,304 @@
...
+
+    async fn find_sync_nodes(&mut self) -> Vec<Peer> {
+        let our_block_height = self.node.expect_sync().current_block_height();
+        let mut interesting_peers = vec![];
+        for node in self.node.peer_book.connected_peers_snapshot().await {
+            if !node.judge() && node.quality.block_height > our_block_height + 1 {
+                interesting_peers.push(node);
+            }
+        }
+        interesting_peers.sort_by(|x, y| y.quality.block_height.cmp(&x.quality.block_height));
```
ljedrz, line 59: do we have a method to extract the peer's height? a `sort_by_key` would be more concise here

ljedrz: perhaps paired with a `cmp::Reverse` too

Protryon: perhaps, this is more readable though imo

network/src/sync/master.rs
```diff
@@ -0,0 +1,304 @@
...
+
+        // trim nodes close to us if any are > 10 blocks ahead
```
ljedrz, line 61: why 10 specifically? maybe we should base the cutoff value on the size of a sync batch instead?

network/src/sync/master.rs
```diff
@@ -0,0 +1,304 @@
...
+        }
+
+        interesting_peers
+    }
+
+    async fn block_locator_hashes(&mut self) -> Vec<BlockHeaderHash> {
+        match self.node.expect_sync().storage().get_block_locator_hashes() {
+            Ok(block_locator_hashes) => block_locator_hashes,
+            _ => {
+                error!("Unable to get block locator hashes from storage");
```
ljedrz, line 76: not related to this PR specifically, but I'm wondering if we shouldn't shut down or even panic if this happens, as we're unlikely to recover

network/src/sync/master.rs
```diff
@@ -0,0 +1,304 @@
...
+            }
+        }
+    }
+
+    async fn send_sync_messages(&mut self) -> usize {
+        let sync_nodes = self.find_sync_nodes().await;
+
+        info!("requested block information from {} peers", sync_nodes.len());
+        let block_locator_hashes = self.block_locator_hashes().await;
+        let mut sent = 0usize;
```
ljedrz, line 87: isn't `sent` equal to` future_set.len()`? we could just calculate that after the loop for brevity

network/src/sync/master.rs
Protryon: it's arbitrary, the intent is to avoid syncing from same-level peers if we don't need to since it's inefficient. if the number is too large, then it will be more prone to syncing from one peer when it shouldn't.

network/src/sync/master.rs
Protryon: agreed, but lets do this in another pr

network/src/sync/master.rs
```diff
@@ -0,0 +1,304 @@
...
+        let end = Instant::now() + Duration::from_secs(timeout_sec);
+        loop {
+            let timeout = tokio::time::sleep_until(end).fuse();
+            pin_mut!(timeout);
+            select! {
+                msg = self.incoming.recv().fuse() => {
+                    if msg.is_none() {
+                        break;
+                    }
+                    if handler(msg.unwrap()) {
```
ljedrz, line 112: this condition is somewhat unclear - it seems that `handler` returning `true` indicates an error - this should be documented, or the name of the closure adjusted so it's more apparent

Protryon: it's not an error, but a terminator. `true` to return. it's an internal method, so i didn't see a need for documenting this, but i'll add some for clarity.

network/src/sync/master.rs
howardwu: 10 is too arbitrary and industry standard defaults to 1. This number has been discussed at length in other communities, can we please get a better justification for 10?

Protryon: We can change it to one/remove this functionality if it's controversial. Was a premature optimization on my part.

PR #824, opened 2021-06-03
Various Improvements to Parallel Syncing

Protryon:


network/src/errors/network.rs
```diff
@@ -63,6 +63,22 @@ impl NetworkError {
             _ => false,
         }
     }
+
+    pub fn is_trivial(&self) -> bool {
```
ljedrz, line 67: there's some overlap with `is_fatal` here; maybe `is_connection_breaking` would be a more obvious name here? also, `BrokenPipe` would probably fit there too

network/src/peers/peer/peer.rs
```diff
@@ -47,23 +48,47 @@ pub struct Peer {
...
             status: PeerStatus::Disconnected,
             quality: Default::default(),
             is_bootnode,
-            network_failures: 0,
         }
     }
 
-    pub fn judge(&self) -> bool {
-        self.quality.rtt_ms > 1500 || self.quality.failures >= 3 || self.quality.is_inactive(chrono::Utc::now())
+    pub fn judge(&mut self) -> bool {
```
ljedrz, line 66: this name could be more clear, as it returning a `bool` doesn't immediately suggest if this is a positive or negative judgement

network/src/errors/network.rs
Protryon: broken pipe would be an implementation issue/error, and i haven't seen any. on board to rename though.

PR #847, opened 2021-06-07
Increase tokio stack size

Protryon:


ljedrz (approved):
LGTM; 2 notes:
- this change is desirable, but the inbound queue can get full with it, which is a different issue
- `tokio` stack size should later be decreased, as it shouldn't need to be that high

howardwu (approved):

PR #854, opened 2021-06-08
Fix nightly clippy errors

ljedrz:
This PR fixes the nightly `clippy` errors and some warnings as a drive-by; not all the warnings could be fixed yet,as the `#[rpc]` attribute seems to conflict with the new way of including external docs.

### DIFF
--- a/consensus/src/consensus.rs
+++ b/consensus/src/consensus.rs
@@ -149,7 +149,7 @@ impl<S: Storage> Consensus<S> {
             // 1) The block is a genesis block, or
             // 2) The block is unknown and does not correspond with the canon chain.
             if crate::is_genesis(&block.header) && self.ledger.is_empty() {
-                self.process_block(&block)?;
+                self.process_block(block)?;
             } else {
                 self.ledger.insert_only(block)?;
             }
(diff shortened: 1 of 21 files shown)


ljedrz:
Interesting; so we're at a spot where `nightly` clippy is giving us errors, but fixing them results in errors in `stable`. We'll need to address this once a new version of Rust is released.

Protryon:
We should just not target nightly, especially for formatting, but in general. It's not stable -- literally. Don't know why we do now.

ljedrz:
It's good to be aware of what the near future will bring, but I agree - `beta` might be a better CI target as a "heads up" if we want to keep one.

ljedrz:
Keeping this PR open as WIP for further dedicated discussion; the first commit should be salvageable, but the PR shouldn't be merged in its current state.

howardwu:
Do we know when the change will land in `stable`? Given these types of breaking changes have (at least in past experience) been fairly infrequent, perhaps we could wait for the feature in stable and update it then?

ljedrz:
The change is currently scheduled for `1.54`, but there has been some active discussion around it, so I'm not sure if it's going to land then. In any case, perhaps we could target `beta` instead of `nightly`? We would still know what is going to break in the future, but with more certainty and CI stability.

ljedrz:
Superseded by https://github.com/AleoHQ/snarkOS/pull/868.

PR #855, opened 2021-06-09
Revert the removal of the check for duplicate blocks

ljedrz:
As requested by @howardwu; even with the stack overflow issue being fixed with https://github.com/AleoHQ/snarkOS/pull/847, the node can still be slowed down significantly if it repeatedly enters the path that applies after the check for duplicate block entries that was commented out in https://github.com/AleoHQ/snarkOS/pull/781 (as it could interfere with post-forking syncing). Since we can still receive batches of duplicate sync blocks under certain conditions, it is safer to restore the previous behavior until a long-term fix is found.

### DIFF
--- a/storage/src/objects/block_path.rs
+++ b/storage/src/objects/block_path.rs
@@ -44,14 +44,10 @@ impl<T: TransactionScheme, P: LoadableMerkleParameters, S: Storage> Ledger<T, P,
     pub fn get_block_path(&self, block_header: &BlockHeader) -> Result<BlockPath, StorageError> {
         let block_hash = block_header.get_hash();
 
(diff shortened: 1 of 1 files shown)


ljedrz:
Filed https://github.com/AleoHQ/snarkOS/issues/856 to track the aforementioned slowness.

ljedrz:
Actually, testing this change in tandem with the new test from https://github.com/AleoHQ/snarkOS/pull/858 hasn't triggered any failures in my initial tests, so it's entirely possible that we can just bring this check back without risking any post-fork sync issues :+1:.

niklaslong (approved):

Protryon (approved):

PR #864, opened 2021-06-15
Add a transaction storage verifier + fixer

ljedrz:
This PR is the fist step towards fixing https://github.com/AleoHQ/snarkOS/issues/861 and https://github.com/AleoHQ/snarkOS/issues/863; it extends the functionality of the existing block storage validator to transactions and their components, and is able to fix the storage of nodes affected by the aforementioned issues (most importantly https://github.com/AleoHQ/snarkOS/issues/861, which can completely halt the block sync process).

While I have already successfully tested it with one database (the node using it was initially stuck, and running the validator allowed it to fully sync with the chain) and it should be complete (i.e. ready for review) already, I'll still conduct a few additional tests with other databases to ensure that it has no bugs - it just currently takes a while (some performance improvements should be possible, but it's follow-up material).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2994,6 +2994,7 @@ dependencies = [
  "snarkvm-parameters",
  "snarkvm-utilities",
  "thiserror",
+ "tokio",
  "tracing",
  "tracing-subscriber",
 ]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2995,6 +2995,7 @@ dependencies = [
  "snarkvm-utilities",
  "thiserror",
  "tracing",
+ "tracing-subscriber",
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


ljedrz:
It appears that the fork tests introduced recently catch an issue with the storage; I think I'll disable them temporarily to keep this PR small and bring them back soon once I've fixed the root cause.

niklaslong (approved):
LGTM 👍

ljedrz:
> It appears that the fork tests introduced recently catch an issue with the storage; I think I'll disable their storage checks temporarily to keep this PR small and bring them back soon once I've fixed the root cause.

Sadly, it seems that the test genesis block contained some transaction components that had not been accounted for; I included that check now and the 2 tests that failed before now pass.

ljedrz:
Rebased so I can build on top of it more easily while it's pending.

Protryon (approved):

PR #865, opened 2021-06-15
Fix native-cpu flag on release builds

Protryon:
This PR fixes the `noconfig` feature flag to disable native CPU targetting in release builds.


ljedrz (approved):
LGTM; the `.github/workflows/release.yml` file builds `snarkOS` with `--features noconfig`, and we want that one to not use these `rustflags` :+1:

howardwu (approved):

howardwu:
LGTM, this is the actual desired behavior. Thank you for fixing.

Note: We need to do the same on the `snarkVM` and `leo` repos.

PR #868, opened 2021-06-16
Remove nightly CI, temporarily remove broken external doc macros

Protryon:
In a recent iteration of Nightly rust, they removed support for `external_docs` feature. It was replaced with a nightly-stabilized format `extended_key_value_attributes`, but that format cannot be guarded using another cfg macro as it fails in stable at parse time.

See [rust#83366](https://github.com/rust-lang/rust/pull/83366)

`extended_key_value_attributes` will reach stable channel in Rust 1.54 it appears. We can readd inlined documentation then.

Additionally, this PR removes nightly CI (but notably not nightly format). This avoids us from breaking due to depending on nightly.


ljedrz (approved):
LGTM :ok_hand:

howardwu (approved):

PR #871, opened 2021-06-16
Fix syncing which involves previously known blocks

ljedrz:
This is slightly affected by https://github.com/AleoHQ/snarkOS/issues/856, but the node doesn't get blocked anymore. Also, since we're no longer requesting duplicate blocks (this PR changes this restriction to canon blocks, so that previously known blocks can become canon), that issue is way less severe than before, and takes a while only when there are a lot of pre-existing blocks to fast-forward to.

The test added by this PR fails without the `fix` commit.

Fixes https://github.com/AleoHQ/snarkOS/issues/870.

Cc https://github.com/AleoHQ/snarkOS/issues/822

### DIFF
--- a/consensus/src/consensus.rs
+++ b/consensus/src/consensus.rs
@@ -167,8 +167,13 @@ impl<S: Storage> Consensus<S> {
                     // Attempt to fast forward the block state if the node already stores
                     // the children of the new canon block.
                     let child_path = self.ledger.longest_child_path(block.header.get_hash())?;
-                    for child_block_hash in child_path {
+                    for child_block_hash in child_path.into_iter().skip(1) {
                         let new_block = self.ledger.get_block(&child_block_hash)?;
+
+                        debug!(
+                            "Fast-forwarding using the block's children. Height {}",
+                            self.ledger.get_current_block_height() + 1
+                        );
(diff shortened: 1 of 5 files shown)


ljedrz:
I forgot the fast-forwarding; it also appears that we currently always attempt to fast-forward with the new canon block - I fixed it.

consensus/src/consensus.rs
```diff
@@ -167,8 +167,13 @@ impl<S: Storage> Consensus<S> {
                     // Attempt to fast forward the block state if the node already stores
                     // the children of the new canon block.
                     let child_path = self.ledger.longest_child_path(block.header.get_hash())?;
-                    for child_block_hash in child_path {
+                    for child_block_hash in child_path.into_iter().skip(1) {
                         let new_block = self.ledger.get_block(&child_block_hash)?;
+
+                        debug!(
+                            "Fast-forwarding using the block's children. Height {}",
```
howardwu, line 174: nit: sorry to nit this, but given it's a `debug` message, could you make `block's children` more clear? I'm not quite sure people will grok that

niklaslong (approved):

Protryon (approved):

PR #875, opened 2021-06-17
A range of tweaks and fixes around storage and its verifier

ljedrz:
This PR contains a set of fixes and adjustments detected or driven by the storage verifier.

The list of changes, in commit order:
- a new test of a ledger that forks many times back and forth, plus an adjustment to the `TestBlocks` API
- remove a duplicate database operation (within a single method / database transaction)
- reuse the `block_header_hash` variable (a refactoring)
- more fine-grained fix options for the storage validator
- actually check for duplicates among serial numbers (as opposed to transaction ids)
- make `Ledger::get_transaction` safer (by using `get` instead of indexing)
- make the `decoderawtransaction` and `gettransactioninfo` RPCs not fail if block number can't be found
- update the transaction's location when committing it

Fixes https://github.com/AleoHQ/snarkOS/issues/863, https://github.com/AleoHQ/snarkOS/issues/872.

### DIFF
--- a/benchmarks/syncing/syncing.rs
+++ b/benchmarks/syncing/syncing.rs
@@ -37,7 +37,7 @@ fn providing_sync_blocks(c: &mut Criterion) {
 
     const NUM_BLOCKS: usize = 10;
 
-    let blocks = TestBlocks::load(NUM_BLOCKS, "test_blocks_100_1");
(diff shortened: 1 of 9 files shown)


ljedrz:
The RPC fix got bundled in, as I fixed it while looking at the storage of transactions, and couldn't use the RPC that I needed.

niklaslong (approved):

Protryon (approved):

PR #878, opened 2021-06-17
Network crawler v2

### LINKED ISSUE #802: [Feature] Protocol (as opposed to rpc) based network crawler
## 🚀 Feature

A network based crawler would be more reliable than RPC (which relies on nodes running on the default port). It could also  be used for in-depth network analysis (graph exposed through RPC and behind a feature flag) and network shaping when coupled with the already implemented centrality measurements.

### LINKED ISSUE #818: [Feature] Improve the availability of bootnodes
Due to all the nodes maintaining persistent connections, the default list of bootnodes that all the nodes attempt to connect to at first can quickly exhaust their connection limits, regardless of how generous they are. There are several ways in which we can improve this situation:
- making the bootnodes "thin" and their connections short-lived, i.e. removing their sync capabilities and instead making them laser-focused on just providing lists of peers; that would make them capable of handling huge numbers of connections, making us able to have fewer bootnodes in total; such bootnodes could also serve as network crawlers (see https://github.com/AleoHQ/snarkOS/issues/802) and actively improve the "shape" of the network by providing "smart" peer lists
- making the connections of bootnodes short-lived (though not as short as in the first point), e.g. disconnecting from peers after e.g. 5 minutes; that time should allow the new peers to learn about other potential peers and possibly form long-lived connections with them instead of the bootnodes
- reduce the number of bootnodes that the nodes attempt to connect to in a single attempt; this could make the initial wait time a bit longer, but would be likely to reduce the strain on all the bootnodes by giving nodes the chance to learn about other peers and connect to them instead of all the bootnodes; this would work best if paired with https://github.com/AleoHQ/snarkOS/issues/817

niklaslong:
Introduces a network-based peer crawler. This has been implemented as a consensus-less bootnode (note: this removes the sync layer from _all_ bootnodes) so it can operate with higher peer counts.  Closes #802 and closes #818.

The current implementation will leave 20% of the min-max peer delta as free capacity for incoming connections, so they will remain performant as bootnodes. Other connections are discarded on each peer update.

The centrality measurements have been moved out of the tests and refactored to work with `KnownNetwork`. These could be exposed via rpc and/or used to enhance the network graph.


network/src/node.rs
```diff
@@ -114,20 +116,29 @@ impl<S: Storage + core::marker::Sync + Send + 'static> Node<S> {
...
             threads: Default::default(),
             shutting_down: Default::default(),
             master_dispatch: RwLock::new(None),
-        })))
+        }));
+
+        if node.config.is_bootnode() {
+            node.network_topology
+                .set(NetworkTopology::default())
+                .expect("network topology was set more than once");
```
ljedrz, line 138: since this is in `Node::new`, we might want to use `unwrap()` with a comment or mention that it's impossible in the `expect` message

network/src/peers/peers.rs
```diff
@@ -48,27 +47,31 @@ impl<S: Storage + Send + Sync + 'static> Node<S> {
...
+        // Calculate the peer counts to disconnect and connect based on the node type and current
+        // peer counts.
+        let (number_to_disconnect, number_to_connect) = match self.config.is_bootnode() {
+            true => (
+                // Bootnodes disconnect down to the min peer count, this to free up room for
+                // the next crawled peers...
+                active_peer_count.saturating_sub(min_peers),
+                // ...then they connect to peers up to min peers below the max, so as to keep room for
+                // new incoming connections.
+                max_peers.saturating_sub(2 * min_peers),
```
ljedrz, line 64: this is twice the number specified in the comment; also, shouldn't it depend on `active_peer_count` too?

network/src/peers/peers.rs
```diff
@@ -165,44 +182,37 @@ impl<S: Storage + Send + Sync + 'static> Node<S> {
...
-                .iter()
-                .filter(|peer| **peer != own_address && !bootnodes.contains(peer))
-                .copied()
-                .choose_multiple(&mut rand::thread_rng(), count)
+            let mut candidates = self.peer_book.disconnected_peers_snapshot().await;
+
+            candidates.retain(|peer| peer.address != own_address && !bootnodes.contains(&peer.address));
+
+            if self.config.is_bootnode() {
+                // Bootnodes choose peers they haven't dialed in a while.
```
ljedrz, line 204: this is a good idea, but it must be paired with a mechanism for discarding or skipping inaccessible peers from the list of disconnected peers, otherwise the bootnodes might only try to connect to old peers that might never become online

network/src/peers/peers.rs
```diff
@@ -287,17 +297,24 @@ impl<S: Storage + Send + Sync + 'static> Node<S> {
...
-                .add_peer(peer_address, self.config.bootnodes().contains(&peer_address))
+                .add_peer(*peer_address, self.config.bootnodes().contains(&peer_address))
                 .await;
         }
+
+        if let Some(topology) = self.network_topology.get() {
+            // If this node is tracking the network topology, record the connections. This can
+            // then be used to construct the graph and query peer info from the peerbook.
+
+            topology.update(source, peers);
```
ljedrz, line 316: I'm wondering if for added safety/performance we shouldn't wrap this call in a `task::spawn_blocking` :thinking:

network/src/topology.rs
```diff
@@ -0,0 +1,492 @@
...
+    net::SocketAddr,
+    ops::Sub,
+};
+
+use chrono::{DateTime, Utc};
+use nalgebra::{DMatrix, DVector, SymmetricEigen};
+use parking_lot::RwLock;
+
+// Purges connections that haven't been seen within this time (in hours).
+const STALE_CUTOFF: i64 = 4;
```
ljedrz, line 38:
```suggestion
const STALE_PEER_CUTOFF_TIME: i64 = 4;
```
or something similar that's a bit more informative

network/src/topology.rs
```diff
@@ -0,0 +1,492 @@
...
+
+/// A connection between two peers.
+///
+/// Implements `partialEq` and `Hash` so that the `source`-`target` order has no impact on equality
```
ljedrz, line 42:
```suggestion
/// Implements `partialEq` and `Hash` manually so that the `source`-`target` order has no impact on equality
```

network/src/topology.rs
```diff
@@ -0,0 +1,492 @@
...
+impl Hash for Connection {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        let (a, b) = (self.source, self.target);
+
+        match a.cmp(&b) {
+            Ordering::Greater => {
+                b.hash(state);
+                a.hash(state);
+            }
+
```
ljedrz, line 73:
```suggestion
```

network/src/topology.rs
```diff
@@ -0,0 +1,492 @@
...
+        a == d && b == c || a == c && b == d
+    }
+}
+
+impl Hash for Connection {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        let (a, b) = (self.source, self.target);
+
+        match a.cmp(&b) {
+            Ordering::Greater => {
```
ljedrz, line 69: This is to ensure that the hash is the same regardless of whether it's a connection `a > b` or `b > a`, right? I'd add a short comment to make this more clear

network/src/topology.rs
```diff
@@ -0,0 +1,492 @@
...
+            source,
+            target,
+            last_seen: Utc::now(),
+        }
+    }
+}
+
+/// Keeps track of crawled peers and their connections.
+#[derive(Debug, Default)]
+pub struct NetworkTopology {
```
ljedrz, line 94:
```suggestion
pub struct KnownNetwork {
```

Technically, this object doesn't contain information on the topology of the network (it can be derived from it), but the network itself.

network/src/topology.rs
```diff
@@ -0,0 +1,492 @@
...
+
+        // Find which connections need to be removed.
+        //
+        // With sets: a - b = removed connections (if and only if one of the two addrs is the
+        // source), otherwise it's a connection which doesn't include the source and shouldn't be
+        // removed. We also keep connections seen within the last few hours as peerlists are capped
+        // in size and omitted connections don't necessarily mean they don't exist anymore.
+        let connections_to_remove: HashSet<Connection> = self
+            .connections
+            .read()
```
ljedrz, line 118: we might need to hold this lock open for the duration of the entire `update` operation if we decide to spawn blocking tasks to perform this operation :thinking:

ljedrz: alternatively, we could force some synchronicity by having a single long-running blocking task that runs `update` based on input from a queue

network/src/topology.rs
```diff
@@ -0,0 +1,492 @@
...
+            })
+            .copied()
+            .collect();
+
+        // Only retain connections that aren't removed.
+        self.connections
+            .write()
+            .retain(|connection| !connections_to_remove.contains(&connection));
+
+        // Scope the write lock.
```
ljedrz, line 132: I'm guessing it's for future-proofing? (since the lock gets dropped right after this scope anyways)

network/src/topology.rs
```diff
@@ -0,0 +1,492 @@
...
+    }
+
+    /// Returns a snapshot of all the connections.
+    pub fn connections(&self) -> HashSet<Connection> {
+        self.connections.read().clone()
+    }
+
+    /// Returns `true` if the topology has tracked any connections, `false` otherwise.
+    pub fn has_connections(&self) -> bool {
+        self.connections.read().len() > 0
```
ljedrz, line 155:
```suggestion
        !self.connections.read().is_empty()
```

network/src/topology.rs
```diff
@@ -0,0 +1,492 @@
...
+        .eigenvalues
+        .iter()
+        .zip(eigen.eigenvectors.column_iter())
+        .map(|(value, vector)| (*value, vector.clone_owned()))
+        .collect();
+
+    // Sort eigenvalue-vector pairs in descending order.
+    pairs.sort_unstable_by(|(a, _), (b, _)| {
+        if ascending {
+            a.partial_cmp(b).unwrap()
```
ljedrz, line 398: are we certain we can't encounter any `f64::NAN` values that would panic here?

network/src/topology.rs
```diff
@@ -0,0 +1,492 @@
...
+        let a = "12.34.56.78:9000".parse().unwrap();
+        let b = "98.76.54.32:1000".parse().unwrap();
+
+        assert_eq!(Connection::new(a, b), Connection::new(b, a));
+        assert_eq!(Connection::new(a, b), Connection::new(a, b));
+    }
+
+    #[test]
+    fn connections_update() {
+        let a = "11.11.11.11:1000".parse().unwrap();
```
ljedrz, line 423: let's pacify `clippy` by using names like `addr_a` etc. here, perhaps?

ljedrz (commented):
This is very cool; I've left some comments and now I'm going to run some tests with this branch.

ljedrz:
This PR introduces changes to the way the bootnodes behave (most notably that they would no longer carry the sync layer, which I support 100%), which need to be a part of the PR's description.

Also, due to the adjustments to how the connections are handled, this PR mitigates https://github.com/AleoHQ/snarkOS/issues/818, if I'm not mistaken.

network/src/topology.rs
niklaslong: Indeed 👍

network/src/peers/peers.rs
niklaslong: This is already somewhat the case with the introduction of `judge_bad` in lockless. Before rebasing/rebuilding, I was tracking `is_routable` as a property set in case of connection failure, left this out for now so it could be discussed.

network/src/topology.rs
niklaslong: I'd be surprised but I'll check `nalgebra`.

niklaslong: The algorithm should be safe so long as the matrix is symmetric (guaranteed by the type). `nalgebra` `unwrap`s the `SymmetricEigen::new`, there's nothing else that could result in an invalid value at that point. I'll thoroughly test this computation before exposing it via rpc (likely a follow up) just to be sure but so far haven't encountered any issues.

network/src/topology.rs
```diff
@@ -90,14 +95,32 @@ impl Connection {
...
 
+impl Default for KnownNetwork {
+    fn default() -> Self {
+        // Buffer size of 1000 messages seems reasonable to begin with.
+        let (tx, rx) = mpsc::channel(1000);
+
+        Self {
+            sender: tx,
+            receiver: Mutex::new(rx),
+            connections: RwLock::new(HashSet::new()),
```
ljedrz, line 113:
```suggestion
            connections: Default::default(),
```

network/src/peers/peers.rs
niklaslong: I've modified this in [77250a0](https://github.com/AleoHQ/snarkOS/pull/878/commits/77250a0d88231ce4bdcb67dac65b25f7a3067be2) to use fill up to 80% of the node's capacity for crawling.

network/src/peers/peers.rs
ljedrz: Right, `judge_bad_offline` indeed mitigates this issue :+1:.

ljedrz (approved):
LGTM :ok_hand:

niklaslong:
rebased

niklaslong:
The failing test on CI is fixed by https://github.com/AleoHQ/snarkOS/pull/884/commits/6b976a51eac09c773e9072e25b32697700a0fdaa.

snarkos/main.rs
```diff
@@ -119,7 +119,7 @@ async fn start_server(config: Config) -> anyhow::Result<()> {
     info!("Storage finished loading");
 
     // Enable the sync layer.
-    {
+    if !node.config.is_bootnode() {
```
howardwu, line 122: Bootnodes need the mempool enabled for the time being, this will break broadcast transaction on Aleo Explorer otherwise

howardwu:
```suggestion
{
```

niklaslong: Reverted 👍

niklaslong:
I'll test this branch thoroughly with the consensus layer re-enabled.

howardwu (approved):

PR #882, opened 2021-06-21
Safer block decommits

ljedrz:
This PR makes the block decommit process safer by performing all the related database operations within a single transaction. This should have the side effect of making it slightly faster as well.

Note: while it is technically possible to perform multiple decommits in a single batch, we don't need to optimize for it, as it is not the usual path.


niklaslong (approved):

ljedrz:
Deferring until the storage validator verifies merkle tree digests to be extra safe.

ljedrz:
This will be included in https://github.com/AleoHQ/snarkOS/pull/972.

Protryon:
Superseded by #972

PR #884, opened 2021-06-22
Parallelize the storage validator, enable fixing stored digests

ljedrz:
This PR uses `rayon` to make the storage validator faster, by parallelizing the checks it performs. In addition, it extends the functionality of the validator to check for both missing and superfluous stored digests.

There's also a fix for a missing digest that the genesis block's transactions can point to - it's now included when the genesis block is stored.

Closes https://github.com/AleoHQ/snarkOS/issues/883.


ljedrz:
The GHA clippy error is spurious - for some reason that job runs an older commit.

niklaslong (approved):

ljedrz:
Rebased.

ljedrz:
>The GHA clippy error is spurious - for some reason that job runs an older commit.

It might have been some GitHub glitch, as the diff was not showing the erroneous line and the tests were passing; it broke "properly" after the rebase, pushed a fixing commit.

storage/src/objects/insert_commit.rs
```diff
@@ -283,6 +283,15 @@ impl<T: TransactionScheme, P: LoadableMerkleParameters, S: Storage> Ledger<T, P,
             value: block_header_hash.0.to_vec(),
         });
 
+        // If it's the genesis block, store its initial applicable digest.
+        if is_genesis {
+            database_transaction.push(Op::Insert {
+                col: COL_DIGEST,
+                key: to_bytes![self.current_digest()?]?.to_vec(),
+                value: 0u32.to_le_bytes().to_vec(),
```
ljedrz, line 291: note: this means we'll now have 2 digests that point to block `0` if there are any transactions in the genesis block.

howardwu (approved):

PR #904, opened 2021-07-04
Configure Dependabot to open PRs on staging branch

howardwu:
## Motivation

Configure Dependabot to open PRs on staging branch

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,7 +5,7 @@ updates:
   schedule:
     interval: daily
     time: "10:00"
-  target_branch: "staging"
+  target-branch: "staging"
   open-pull-requests-limit: 10
   ignore:
   - dependency-name: bech32
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,6 +5,7 @@ updates:
   schedule:
     interval: daily
     time: "10:00"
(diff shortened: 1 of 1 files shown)


howardwu:
@ljedrz @niklaslong it appears there's an unrelated failure affecting this PR's CI

weikengchen (approved):

.github/dependabot.yml
```diff
@@ -5,6 +5,7 @@ updates:
   schedule:
     interval: daily
     time: "10:00"
+  target_branch: "staging"
```
ljedrz, line 8:
```suggestion
  target-branch: "staging"
```

`dependabot`'s docs suggest that this should be `target-branch` instead: https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/configuration-options-for-dependency-updates#target-branch

ljedrz (changes requested):
I think we need to change the key from an underscore to a dash, as I commented on the diff; other than that it's a desirable change :+1:.

niklaslong:
Re-running the CI jobs fixes both failures in `snarkos-testing` and `rust-stable`.

ljedrz (approved):

niklaslong (approved):

PR #905, opened 2021-07-05
Update snarkVM to 0.6, mpmc-map to 0.2, run cargo update

ljedrz:
The first 2 commits update the TOML files, and the last one is a combined LOCK update after a `cargo update` run.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -57,9 +57,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.40"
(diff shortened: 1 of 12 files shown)


ljedrz:
That went much better locally :slightly_smiling_face:; then again, `snarkVM` 0.7 is coming soon - I'm leaving this PR open in case the `mpmc-map` commit is reused, but changing it to a draft.

ljedrz:
Superseded by https://github.com/AleoHQ/snarkOS/pull/937.

PR #912, opened 2021-07-05
chore: clippy lints

niklaslong:
Fixes a few needless borrow lints.

In addition some of the lints are caused by false positives: rust-lang/rust-clippy#7422, these should be fixed in the next `nightly` release.


ljedrz (approved):
Humans will never beat `clippy` at this lint :slightly_smiling_face:.

LGTM :ok_hand:

PR #914, opened 2021-07-05
feat: expose centrality measurements via rpc

niklaslong:
Exposes the network centrality measurements through the `getnetworkgraph` rpc endpoint. Note, the computation of said measurements only occures at the time of the rpc call. Though the computation hasn't yet been benchmarked, it's reasonably fast for current testnet sizes (i.e. ~`1000` nodes).

### DIFF
--- a/rpc/src/rpc_types.rs
+++ b/rpc/src/rpc_types.rs
@@ -19,7 +19,7 @@
 use chrono::{DateTime, Utc};
 use jsonrpc_core::Metadata;
 use serde::{Deserialize, Serialize};
-use std::{collections::HashSet, net::SocketAddr};
+use std::net::SocketAddr;
 
 /// Defines the authentication format for accessing private endpoints on the RPC server
 #[derive(Clone, Debug, Eq, PartialEq, Serialize, Deserialize)]
--- a/rpc/src/rpc_types.rs
+++ b/rpc/src/rpc_types.rs
@@ -283,19 +283,42 @@ pub struct TransactionRecipient {
     pub amount: u64,
 }
 
-#[derive(Clone, Debug, Eq, PartialEq, Serialize, Deserialize)]
+/// The crawled known network and measurements.
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
 pub struct NetworkGraph {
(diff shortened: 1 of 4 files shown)


rpc/src/rpc_types.rs
```diff
@@ -283,19 +283,42 @@ pub struct TransactionRecipient {
     pub amount: u64,
 }
 
-#[derive(Clone, Debug, Eq, PartialEq, Serialize, Deserialize)]
+/// The crawled known network and measurements.
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
 pub struct NetworkGraph {
-    pub vertices: HashSet<Vertice>,
-    pub edges: HashSet<Edge>,
```
niklaslong, line 289: This change won't impact the uniqueness of the collections' values, since they are constructed in `NetworkMetrics` with `HashSet`s. Using `Vec` here instead, removes the requirement for the `Vertices` and `Edges` to implement `Hash` and `Eq`.

rpc/src/rpc_types.rs
```diff
@@ -283,19 +283,42 @@ pub struct TransactionRecipient {
...
+    /// Known connections.
+    pub edges: Vec<Edge>,
 }
 
-#[derive(Clone, Debug, Eq, PartialEq, Serialize, Deserialize, Hash)]
+/// Metadata and measurements pertaining to a node in the graph of the known network.
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
 pub struct Vertice {
     pub addr: SocketAddr,
     pub is_bootnode: bool,
```
ljedrz, line 312: this is something I missed before: this doesn't necessarily determine whether the node is actually a bootnode or not, only whether it's a known/curated bootnode, as it's based on our own list of bootnodes; this distinction might not matter right now, but it might in the future, so I'm just pointing it out :detective:

niklaslong: Indeed, we don't actually currently communicate that information through the network protocol.

ljedrz (approved):
LGTM :ok_hand:

niklaslong:
rebased

PR #920, opened 2021-07-06
Fix outbound metrics

ljedrz:
This PR fixes/restores the following metrics:
- `outbound::ALL_SUCCESSES`
- `outbound::ALL_FAILURES`
- `queues::OUTBOUND`

Cc @sadroeck
Fixes https://github.com/AleoHQ/snarkOS/issues/869


consensus/src/consensus.rs
```diff
@@ -243,6 +247,9 @@ impl<S: Storage> Consensus<S> {
         // 2. Insert/canonize block
         self.ledger.insert_and_commit(block)?;
 
+        // Increment the current block height metric
+        metrics::increment_gauge!(BLOCK_HEIGHT, 1.0);
```
Protryon, line 251: can we just always report the absolute value to avoid divergence?

ljedrz: in other words, we'd just always set the gauge to the value of the atomic? I didn't see any divergence, but yeah, I don't see why not, just to be extra safe.

ljedrz: the same would apply to the `queues::OUTBOUND` value; since this is spread between 3 PRs building on top of one another, I'd prefer to do it as a follow-up to avoid conflicting with myself

network/src/peers/peer/outbound_handler.rs
```diff
@@ -38,49 +44,44 @@ pub(super) enum PeerAction {
 #[derive(Clone, Debug)]
 pub struct PeerHandle {
     pub(super) sender: mpsc::Sender<PeerAction>,
+    pub queued_outbound_message_count: Arc<AtomicUsize>,
```
Protryon, line 47: i dont like delta metrics because they can get out of sync. i don't see a way to solve this with tokio channels though... is this metric useful? it seemed useful when we had one channel to rule them all, but it increasingly looks to me like maybe we want a payload emission at app layer -> payload emitted on network latency metric? would also be way easier to implement and not error prone.

consensus/src/consensus.rs
Protryon: yeah, it's just future proofing so we don't get the inevitable bug of some PR having it diverge as a side effect. thanks.

network/src/peers/peer/outbound_handler.rs
Protryon: i think this metric would be more useful for identifying actual latency problems because it tracks each payload individually -- however if there is disagreement, we should just make a wrapped `Sender/Receiver` type that can more tightly couple the metric into the channel (and be used in several places)

ljedrz: when running in bootnode mode and with a few hundred connections, we are actually constantly at a combined depth of all the outbound queues greater than 1k items, which might indicate that we're slowly reaching practical connection limits (at least when syncing at the same time); we need to know if we're able to push out all the messages quickly, otherwise we might be growing memory or become swamped and reply slowly.

> it seemed useful when we had one channel to rule them all

we've always had individual per-peer outbound channels (or at least ever since we've had metrics), so the same principle applies

ljedrz: that being said, and even though I saw no divergence with a lot of connection changes (meaning this is an improvement to what we currently have), I agree that the `queues::OUTBOUND` setup proposed in this PR is not exactly trivial, and I'm open to alternatives that will allow us to monitor how many items we have waiting to be sent to peers

Protryon: does `metrics` crate/prometheus have tagged metrics? would probably make the data more useful as well, so we can see concrete metrics on a per-connection basis. latency metrics also sound like they capture this rather well. i'm open to depth as well, and we can just pass this through for now, but the usage of out-of-band metrics like this is not advised.

ljedrz: > does metrics crate/prometheus have tagged metrics?

I think it's done via labels: https://prometheus.io/docs/practices/naming/#labels.

> we can see concrete metrics on a per-connection basis

an individual breakdown could be interesting indeed, as some outliers might be causing the combined values to become "weird"

> i'm open to depth as well, and we can just pass this through for now

I'll file a new GitHub issue related to the aforementioned "wholesale" value updates to avoid divergence and one for prometheus labels so we can handle those as follow-ups; as for now, since these PRs do appear to fix (or restore) the values, I propose to merge them as-is to keep the changes small and self-contained

niklaslong (approved):
LGTM other than the points raised by @Protryon 👍

ljedrz:
I opened issues for the aforementioned improvements :+1:.

ljedrz:
Rebased.

consensus/src/consensus.rs
unordered-set: I'm trying to fix this in https://github.com/AleoHQ/snarkOS/pull/965/files

PR #923, opened 2021-07-06
Only catch up the secondary storage if the block height is behind the primary one

ljedrz:
While investigating https://github.com/AleoHQ/snarkOS/issues/900, I ruled out primary storage coherence issues and was not able to reproduce (using the script by @e-nikolov) the bug when the secondary storage instance normally used for RPC purposes was disabled (i.e. only the primary instance was used).

While the ultimate solution would be not to use the secondary storage altogether (which might not be viable yet, as it hasn't yet been determined if it wouldn't lead to degraded performance), I was able to greatly reduce the number of times `Ledger::catch_up_secondary` is called by limiting it to the times when the secondary storage's block height is actually behind the primary one. With this change in place, I haven't been able to reproduce the issue.

Fixes https://github.com/AleoHQ/snarkOS/issues/900.

### DIFF
--- a/rpc/src/rpc_impl.rs
+++ b/rpc/src/rpc_impl.rs
@@ -111,7 +111,8 @@ impl<S: Storage + Send + core::marker::Sync + 'static> RpcFunctions for RpcImpl<
 
         let storage = &self.storage;
 
-        storage.catch_up_secondary(false)?;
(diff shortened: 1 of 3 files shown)


niklaslong (approved):
LGTM 👍

PR #926, opened 2021-07-07
Refactor the forking tiebreaker to use aggregated difficulty, instead of chain length

julesdesmit:
## Motivation

This change should reduce the length of orphan chains, and reduce the number of nodes that sync onto orphan chains and get stuck there. As an orphaned chain uses less computing power than the main one, its difficulty will drastically drop as it is extended. After a while, and with a bit of luck, this can cause an orphaned miner to keep going on its own route and always stay ahead of the main chain in terms of raw numbers. However, his aggregated difficulty will diminish severely as time goes on, due to the low computing power used. Thus, this change should make sure orphaned miners are brought back to the main chain quicker than before, and it should also ensure that syncing nodes will re-org to the main chain if they start going down an orphaned block path.

## Test Plan

The fix has been tested on my personal node by syncing from scratch 3 times. All 3 times, the node successfully managed to sync its way to the tip of the chain without getting stuck on an orphan chain. The node was observed, from time to time, to be syncing its way into an orphan chain, but then forking back to the main chain, after noticing a side path with a difficulty many times higher than what it was currently on - indicating that it is now able to successfully identify the chain on which the most work is done.

The next step will be to roll this out on a wider scale. As this merges on staging, there will be a few nodes already using this fix, which will give us further indication of the feasibility of this fix.

### DIFF
--- a/consensus/src/consensus.rs
+++ b/consensus/src/consensus.rs
@@ -393,9 +393,11 @@ impl<S: Storage> Consensus<S> {
         let mut aggregate_difficulty = 0u128;
 
         for i in 0..path_size {
-            let block = self.ledger.get_block_from_block_number(current_block_height - i)?;
+            let block_header = self
+                .ledger
+                .get_block_header(&self.ledger.get_block_hash(current_block_height - i)?)?;
 
-            aggregate_difficulty += block.header.difficulty_target as u128;
+            aggregate_difficulty += block_header.difficulty_target as u128;
         }
 
         Ok(aggregate_difficulty)
--- a/consensus/src/consensus.rs
+++ b/consensus/src/consensus.rs
@@ -187,8 +187,6 @@ impl<S: Storage> Consensus<S> {
                     // perform a fork to the side chain.
                     let canon_difficulty =
                         self.get_canon_difficulty_from_height(side_chain_path.shared_block_number)?;
-                    println!("{}", canon_difficulty);
-                    println!("{}", side_chain_path.aggregate_difficulty);
 
                     if side_chain_path.aggregate_difficulty > canon_difficulty {
                         debug!(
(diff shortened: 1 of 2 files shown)


ljedrz:
I changed the PR's base from `master` to `staging`, which is the default branch to file PRs against :+1:.

consensus/src/consensus.rs
```diff
@@ -191,11 +183,14 @@ impl<S: Storage> Consensus<S> {
...
+                    // If the side chain is now heavier than the canon chain,
                     // perform a fork to the side chain.
-                    if side_chain_path.new_block_number > self.ledger.get_current_block_height() {
+                    let canon_difficulty =
+                        self.get_canon_difficulty_from_height(side_chain_path.shared_block_number)?;
+
+                    if side_chain_path.aggregate_difficulty > canon_difficulty {
                         debug!(
-                            "Determined side chain is longer than canon chain by {} blocks",
+                            "Determined side chain is heavier than canon chain by {} blocks",
```
ljedrz, line 193: nit: some percentage or other value might be more informative than the height difference now

julesdesmit: Good point :) I'll adjust it!

storage/src/objects/block_path.rs
```diff
@@ -64,21 +68,31 @@ impl<T: TransactionScheme, P: LoadableMerkleParameters, S: Storage> Ledger<T, P,
                 // This is a canon parent
                 Ok(block_num) => {
                     // Add the children from the latest block
-
                     let longest_path = self.longest_child_path(block_hash)?;
 
+                    // Add all the difficulty targets associated with the longest_path,
+                    // skipping the first element (which is the hash associated to
+                    // `block_header`).
+                    for hash in longest_path[1..].iter() {
```
ljedrz, line 76:
```suggestion
                    for hash in longest_path.iter().skip(1) {
```

ljedrz (approved):
Left a small suggestion, LGTM otherwise :ok_hand:.

julesdesmit:
Amended!

niklaslong (approved):

PR #927, opened 2021-07-07
Only provide routable addresses in peer lists

### LINKED ISSUE #817: [Feature] Don't provide unroutable addresses in peer lists
The current peer-sharing mechanism only checks that we don't provide a peer with its own address among the addresses we're connected to; we can do better than this, though, including:
- excluding bootnodes (peers are generally going to know the same ones)
- excluding addresses that connected to us (as opposed to us being the initiator; potentially unroutable)
- filtering for addresses that we would verify to be routable

This could be related to https://github.com/AleoHQ/snarkOS/issues/802, as the network crawler would know a lot of peers and be able to produce great peer lists.

niklaslong:
This PR introduces changes to track if peers are routable. This allows nodes to only provide dialable listeners in peer lists, thus lowering the overall incidence of connection errors. 

I would recommend these changes be thoroughly tested on testnet as they do have some non-negligeable implications. Firstly, sharing peers in this manner means the crawler will only ever encounter one side of a connection. With a large network this shouldn't cause too many issues, especially as the mesh is dense but the it could result in nodes being unnacounted for. Similarly, peers in clusters could find themselves "isolated" from the rest of the network. 

Fixes #817.

### DIFF
--- a/network/src/peers/peers.rs
+++ b/network/src/peers/peers.rs
@@ -290,49 +290,45 @@ impl<S: Storage + Send + Sync + 'static> Node<S> {
 
     pub(crate) async fn send_peers(&self, remote_address: SocketAddr) {
         // Broadcast the sanitized list of connected peers back to the requesting peer.
-        //
-        // Order of tried sets for bootnode peer lists:
-        //
-        // 1. routable connected peers (only set shared by regular nodes)
-        // 2. connected peers
-        // 3. disconnected peers
 
         use crate::Peer;
         use rand::prelude::SliceRandom;
 
         let connected_peers = self.peer_book.connected_peers_snapshot().await;
 
-        let filter = |peer: &Peer| peer.address != remote_address && !self.config.bootnodes().contains(&peer.address);
-        let strict_filter = |peer: &Peer| filter(peer) && peer.is_routable.unwrap_or(false);
+        let basic_filter =
+            |peer: &Peer| peer.address != remote_address && !self.config.bootnodes().contains(&peer.address);
+        let strict_filter = |peer: &Peer| basic_filter(peer) && peer.is_routable.unwrap_or(false);
 
-        // 1.
+        // Strictly filter the connected peers by only including the routable addresses.
         let strictly_filtered_peers: Vec<SocketAddr> = connected_peers
             .iter()
             .filter(|peer| strict_filter(peer))
             .map(|peer| peer.address)
             .collect();
 
-        let peers = match (self.config.is_bootnode(), strictly_filtered_peers.is_empty()) {
-            (false, _) | (true, false) => strictly_filtered_peers,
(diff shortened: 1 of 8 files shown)


niklaslong:
CI is failing with a network error during setup, will retry a little later.

ljedrz (approved):
Looks reasonable to me; while a node might mark another one as non-routable in case of a connection failure, that other node can still connect to it, as it only affects suitability for belonging to a peer list; also, regular nodes will most likely oscillate around their `min-peers` and not reach their `max-peers` boundary.

ljedrz:
The CI runs likely errored due to the current CircleCI outages; let's re-run the failed ones tomorrow.

niklaslong:
rebased

niklaslong:
rebased post #944

network/src/peers/peers.rs
```diff
@@ -290,12 +290,56 @@ impl<S: Storage + Send + Sync + 'static> Node<S> {
...
+        // 1. routable connected peers (only set shared by regular nodes)
+        // 2. connected peers
+        // 3. disconnected peers
+
+        use crate::Peer;
+        use rand::prelude::SliceRandom;
+
+        let connected_peers = self.peer_book.connected_peers_snapshot().await;
+
+        let filter = |peer: &Peer| peer.address != remote_address && !self.config.bootnodes().contains(&peer.address);
```
ljedrz, line 305: suggestion: there's a `strict_filter`, so this one could be a `basic_filter`

network/src/peers/peers.rs
```diff
@@ -290,12 +290,56 @@ impl<S: Storage + Send + Sync + 'static> Node<S> {
...
     pub(crate) async fn send_peers(&self, remote_address: SocketAddr) {
         // Broadcast the sanitized list of connected peers back to the requesting peer.
-        let peers = self
-            .peer_book
-            .connected_peers()
-            .into_iter()
-            .filter(|&addr| addr != remote_address)
-            .choose_multiple(&mut rand::thread_rng(), crate::SHARED_PEER_COUNT);
+        //
+        // Order of tried sets for bootnode peer lists:
```
ljedrz, line 294: > bootnode peer lists

the logic below applies to all peers - only the second branch is bootnode-specific

network/src/peers/peers.rs
```diff
@@ -290,12 +290,56 @@ impl<S: Storage + Send + Sync + 'static> Node<S> {
...
+        //
+        // 1. routable connected peers (only set shared by regular nodes)
```
ljedrz, line 296:
```suggestion
        // 1. routable connected peers
```

ljedrz: alternatively `(shared by non-bootnodes)` with `1` annotated with `(shared by bootnodes)`

PR #946, opened 2021-07-15
Dont store transaction locations of non-canon blocks

ljedrz:
While https://github.com/AleoHQ/snarkOS/pull/875 improved the state of stored transaction locations, it missed one possible scenario: a node can receive a side-chain block containing a transaction that had already been imported as part of a canon block; even though that side block wouldn't become canon, its insertion would still overwrite that transaction's canon location with the side chain's one, just because it was received later.

This PR removes the storage of transaction location on block insertions, leaving it only for block commits.

Cc https://github.com/AleoHQ/snarkOS/issues/863.


ljedrz:
AFAIK this is also included in https://github.com/AleoHQ/snarkOS/pull/972.

Protryon:
Superseded by #972

PR #951, opened 2021-07-15
feat: exit after storage validation

ljedrz:
This PR makes the node shut down after the storage validator or storage trim tools are used.


niklaslong (approved):
LGTM 👍

PR #952, opened 2021-07-16
Fix path to config in the banner

unordered-set:
## Motivation

When I originally saw the banner, it was misleading, and I've updated the wrong config file.

## Test Plan

This is just a string change, so I've checked that it is still building.

## Related PRs

Well, actually I did some extra things:
1) added some constants for the config file name
2) added more constants about config overwriting behaviour

but overall it looked pretty ugly and probably not worth it. Please, let me know if you want me to update the pull request with that versions.

### DIFF
--- a/snarkos/display.rs
+++ b/snarkos/display.rs
@@ -52,7 +52,7 @@ pub fn render_welcome(config: &Config) -> String {
             }
             Err(_) => {
                 output +=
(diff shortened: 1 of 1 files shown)


ljedrz (approved):
Good catch, thanks!

ljedrz:
Hmm, why isn't the CI status being read? I'll reopen the PR, hopefully it'll get unstuck.

unordered-set:
I am trying to reopen it again

unordered-set:
@ljedrz, do you mind starting that manual checks please?

PR #957, opened 2021-07-19
feat: distinguish crawler nodes and bootnodes

niklaslong:
Considering bootnodes currently still implement a sync layer and are run in decent numbers on the testnet, this PR introduces a command line flag to distinguish between crawler nodes and bootnodes. Bootnodes generally disconnect down to 80% of their peering capacity to ensure high availability to nodes joining the network and don't connect to any peers on their own (other than their bootnode list if they have `0` connections). 

Merging these changes would also pave the way for crawler nodes to operate in support of a cluster of bootnodes, providing them with peers they could then pass on to new peers joining the network. 

This new distinction would also allow us to remove the sync layer from either the bootnodes or the crawlers (whichever isn't used by aleo.network).

### DIFF
--- a/network/src/peers/peers.rs
+++ b/network/src/peers/peers.rs
@@ -56,32 +56,44 @@ impl<S: Storage + Send + Sync + 'static> Node<S> {
 
         // Calculate the peer counts to disconnect and connect based on the node type and current
         // peer counts.
-        let (number_to_disconnect, number_to_connect) = match self.config.is_bootnode() {
-            true => {
-                // Bootnodes disconnect down to the min peer count, this to free up room for
-                // the next crawled peers...
-                let number_to_disconnect = active_peer_count.saturating_sub(min_peers);
-                // ...then they connect to disconnected peers leaving 20% of their capacity open
-                // incoming connections.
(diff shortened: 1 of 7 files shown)


network/src/peers/peers.rs
```diff
@@ -56,32 +56,44 @@ impl<S: Storage + Send + Sync + 'static> Node<S> {
...
+            )
         };
 
         if number_to_disconnect != 0 {
             let mut current_peers = self.peer_book.connected_peers_snapshot().await;
 
-            // Bootnodes will disconnect from random peers...
-            if !self.config.is_bootnode() {
+            // Bootnodes and crawlers will disconnect from random peers...
+            if !self.config.is_bootnode() && !self.config.is_crawler() {
```
niklaslong, line 96: Follow-up: we may want bootnodes to disconnect as soon as the peerlists have been sent (or from oldest first, making bootnodes FIFO).

ljedrz (commented):
I'm in favor of this change, though it slightly increases the complexity of the overall setup, so I'm leaving this PR open to further discussion - it needs wider support.

ljedrz (approved):
LGTM :ok_hand:

PR #958, opened 2021-07-20
feat: seedable initial peers that aren't marked as bootnodes 

niklaslong:
Refactors the topology setup helpers to connect started nodes as opposed to seeding their bootnodes. This is useful for peering tests; this PR also fixes tests broken by the previous peering changes in #927.


network/src/peers/peers.rs
```diff
@@ -142,42 +143,33 @@ impl<S: Storage + Send + Sync + 'static> Node<S> {
...
     ///
-    /// This function attempts to reconnect this node server with any bootnode peer
-    /// that this node may have failed to connect to.
-    ///
-    /// This function filters out any bootnode peers the node server is
+    /// This function filters out any peers the node server is
     /// either connnecting to or already connected to.
     ///
-    async fn connect_to_bootnodes(&self) {
+    async fn connect_to_peers(&self, peers: Arc<Vec<SocketAddr>>) {
```
ljedrz, line 151:
```suggestion
    async fn connect_to_addresses(&self, addrs: &[SocketAddr]) {
```
nit: we already have a `Peers` object, so it's better not to mix them up; also, it would be preferable to pass a slice as an argument to this function if it's possible (I see `copied` being used, so it should work)

network/tests/topology.rs
```diff
@@ -190,87 +190,30 @@ async fn star_converges_to_mesh() {
...
     wait_until!(
         15,
         degree_centrality_delta(&nodes) <= (MAX_PEERS - MIN_PEERS).into(),
         200
     );
 }
 
-#[tokio::test(flavor = "multi_thread")]
-#[ignore]
-async fn binary_star_contact() {
```
ljedrz, line 207: why is this test removed? was it flaky or are other tests sufficient to test mesh formation?

network/src/peers/peers.rs
```diff
@@ -93,10 +93,11 @@ impl<S: Storage + Send + Sync + 'static> Node<S> {
             }
         }
 
-        // Attempt to connect to the default bootnodes of the network if the node has no active
+        // Attempt to connect to the default bootnodes and initial peers of the network if the node has no active
         // connections.
         if self.peer_book.get_active_peer_count() == 0 {
-            self.connect_to_bootnodes().await;
+            self.connect_to_peers(self.config.bootnodes()).await;
+            self.connect_to_peers(self.config.initial_peers()).await;
```
ljedrz, line 100: since the initial peers shouldn't be used as bootnodes, but just as starting peers, they should only be connected to _only_ once, i.e. outside of `Node::update_peers` but somewhere else instead; if this wasn't a test-first functionality, the right place would be `main.rs`, but the test nodes should perhaps call this on their own once they're started?

ljedrz (changes requested):
These changes are desirable for better testing opportunities; I raised some points that need to be addressed, but otherwise it looks good :+1:.

network/tests/topology.rs
niklaslong: It was flaky, at least in its current state. I was planning on exploring cases similar in complexity once this is merged, I  may well re-introduce something similar to it. This was mostly to clean things up in preparation for the next set of changes. Happy to leave it in as an ignored case for now, if that's preferable.

ljedrz: yeah, you can just add a comment why it's currently ignored

network/src/peers/peers.rs
niklaslong: Agreed 👍

network/tests/topology.rs
niklaslong: I've reintroduced the test, with a few tweaks it's actually more stable than it previously was.

network/src/peers/peers.rs
```diff
@@ -93,10 +93,10 @@ impl<S: Storage + Send + Sync + 'static> Node<S> {
             }
         }
 
-        // Attempt to connect to the default bootnodes of the network if the node has no active
+        // Attempt to connect to the default bootnodes and initial peers of the network if the node has no active
```
ljedrz, line 96:
```suggestion
        // Attempt to connect to the default bootnodes of the network if the node has no active
```
this is now obsolete, no?

network/src/peers/peers.rs
```diff
@@ -142,42 +142,33 @@ impl<S: Storage + Send + Sync + 'static> Node<S> {
...
-            .filter(|peer| **peer != own_address)
-            .copied()
-        {
+        for node_addr in addrs.iter().filter(|addr| **addr != own_address).copied() {
             let node = self.clone();
-            if node.peer_book.is_connected(bootnode_address) {
-                return;
+
+            if node.peer_book.is_connected(node_addr) {
+                break;
```
ljedrz, line 158: nothing wrong with this PR here, but I see room for a small improvement: this check could be moved above the `node` binding, so that we avoid its clone; sure, it's just an `Arc`, but still, it's not free and avoidable

network/tests/topology.rs
```diff
@@ -115,15 +102,14 @@ async fn spawn_nodes_in_a_mesh() {
...
-    start_nodes(&nodes).await;
+    connect_nodes(&mut nodes, Topology::Mesh).await;
 
     // Set the sleep interval to 200ms to avoid locking issues.
     // Density measurement here is proportional to the min peers: if every node in the network
     // only connected to the min node count, the total number of connections would be roughly 10
-    // percent of the total possible. With 50 nodes and min at 5 connections each this works out to
-    // be 125/1225 ≈ 0.1.
-    wait_until!(15, network_density(&nodes) >= 0.1, 200);
+    // percent of the total possible. With 25 nodes and min at 5 connections each this works out to
```
ljedrz, line 110: this change is done so that the CI has an easier time with the test, right?

niklaslong: It's more accurate (i.e. restoring the original intent of the test), the peer count was lowered without lowering the MIN_PEERS which changed the density expectation. Forgot to add the expected density calculation to these tests.

ljedrz (approved):
LGTM :ok_hand:

network/src/peers/peers.rs
```diff
@@ -142,42 +142,33 @@ impl<S: Storage + Send + Sync + 'static> Node<S> {
...
-            .iter()
-            .filter(|peer| **peer != own_address)
-            .copied()
-        {
-            let node = self.clone();
-            if node.peer_book.is_connected(bootnode_address) {
-                return;
+        for node_addr in addrs.iter().filter(|addr| **addr != own_address).copied() {
+            if self.peer_book.is_connected(node_addr) {
+                break;
```
ljedrz, line 156: this PR doesn't change this, but I just realized that this should be a part of the `filter` predicate as opposed to a `break`, no? we don't want to disregard the rest of the list as soon as we encounter an active connection after all

niklaslong: This used to be a `return`. I agree though, this should be checked during filtering.

PR #959, opened 2021-07-20
Faster storage validator

ljedrz:
This PR improves the performance of the storage validator; it introduces the following improvements:
- remove the locks used by the validator and `#[inline]` 2 small, widely-used database methods (a tiny improvement)
- use the `COL_BLOCK_TRANSACTIONS` column to obtain block's transactions in bulk as opposed to obtaining them individually via `Ledger::get_transaction_bytes` (~40% speed improvement for a database with a modest number of glitches, should be much more for a glitchy one)

There is still room for improvement by introducing partial transaction deserialization, but I don't know if it's viable yet; since the PR is already fairly large, I decided to open it, and further improvements - if any - will build on top of it.

### DIFF
--- a/storage/src/validator.rs
+++ b/storage/src/validator.rs
@@ -27,16 +27,14 @@ use snarkvm_algorithms::traits::LoadableMerkleParameters;
 use snarkvm_dpc::{Block, BlockHeaderHash, DatabaseTransaction, LedgerScheme, Op, Storage, TransactionScheme};
 use snarkvm_utilities::{to_bytes, ToBytes};
 
-use parking_lot::Mutex;
 use rayon::prelude::*;
+use tokio::{sync::mpsc, task};
 use tracing::*;
 
 use std::{
     collections::HashSet,
-    sync::{
-        atomic::{AtomicBool, Ordering},
-        Arc,
-    },
+    mem,
+    sync::atomic::{AtomicBool, Ordering},
 };
 
 macro_rules! check_for_superfluous_tx_components {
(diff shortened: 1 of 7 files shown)


howardwu:
@ljedrz looks good. Can you check that the `MemDb` implementation is receiving the requisite updates as well?

ljedrz:
The in-memory implementation can't (or rather doesn't need to) be validated this way, so the `MemDb` impl doesn't need to be adjusted individually. That being said, fixing any logic-related bugs detected by the validator will automatically improve all the storage backends, so it's not being left out.

ljedrz:
Sadly, partial deserialization doesn't seem to significantly improve the performance of the storage validator, and it makes it considerably more complex. However, the extra profiling led to https://github.com/AleoHQ/snarkVM/pull/287, which provides a small speed boost and makes the related code more idiomatic.

storage/src/validator.rs
```diff
@@ -427,12 +436,12 @@ impl<T: TransactionScheme + Send + Sync, P: LoadableMerkleParameters, S: Storage
             }
             component_sender.send(ValidatorAction::RegisterTxComponent(COL_DIGEST, tx_digest)).unwrap();
 
-            let tx_memo = tx.memorandum();
-            if !self.contains_memo(tx_memo) {
+            let tx_memo = to_bytes![tx.memorandum()].unwrap();
```
niklaslong, line 439: I'm assuming this can't fail?

niklaslong (approved):
LGTM, just left one question. 👍

storage/src/validator.rs
ljedrz: correct: https://github.com/AleoHQ/snarkVM/blob/v0.5.4/utilities/src/bytes.rs#L129-L148

PR #965, opened 2021-07-22
Fix metrics

### LINKED ISSUE #964: [Bug] inconsistency between block height in RPC and prometheus
## 🐛 Bug Report

This bug affects my monitoring a lot. So, when I'm checking block high like that:

```
curl -s --data-binary '{"jsonrpc": "2.0", "id":"1", "method": "getnodestats" }' -H 'content-type:application/json' http://localhost:3030/ | jq '.result.misc.block_height'
344264
```

it is 344264.

When I'm checking like that:

```
curl localhost:9000 | grep block_height
# TYPE snarkos_misc_block_height_total counter
snarkos_misc_block_height_total 1167399
```

## Expected Behavior
I am expecting these values to match

## Your Environment
snarkOS Version : 1.3.10

unordered-set:
## Motivation

Fix #964 

## Test Plan

I have tested it locally (by introducing a new metrics), it works:

```
$ curl 127.0.0.1:9000 | grep block_height | grep -v TYPE
snarkos_misc_chain_block_height_total 344264
snarkos_misc_block_height_total 356358
$ curl -s --data-binary '{"jsonrpc": "2.0", "id":"documentation", "method": "getnodestats", "params": [] }' -H 'content-type: application/json' http://localhost:3030/ | jq '.[].misc?.
block_height?';
344264
```

### DIFF
--- a/metrics/src/names.rs
+++ b/metrics/src/names.rs
@@ -63,6 +63,7 @@ pub mod queues {
 pub mod misc {
     pub const BLOCK_HEIGHT: &str = "snarkos_misc_block_height_total";
     pub const BLOCKS_MINED: &str = "snarkos_misc_blocks_mined_total";
+    pub const CHAIN_BLOCK_HEIGHT: &str = "snarkos_misc_chain_block_height_total";
     pub const DUPLICATE_BLOCKS: &str = "snarkos_misc_duplicate_blocks_total";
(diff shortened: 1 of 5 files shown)


ljedrz:
@unordered-set changing the target branch to `staging` (which is the desired target branch for PRs) results in a conflict, further suggesting that the linked issue has been fixed there.

ljedrz:
As discussed in the dedicated issue, this fix is no longer applicable to `staging` (where it's been resolved).

PR #969, opened 2021-07-22
feat: include the orphan rate in the sample grafana dashboard

ljedrz:
This PR adds the orphan block count to the list of fields available in the dashboard, and it performs minor adjustments to some of the existing fields.


niklaslong (approved):
LGTM 👍

PR #973, opened 2021-07-23
fix: erronous check for tracking the known network

niklaslong:
Follow-up to #957, re-enables tracking the known network for crawlers.


ljedrz (approved):
LGTM :ok_hand:

PR #977, opened 2021-07-23
Release v1.3.11

howardwu:
## Motivation

Release v1.3.11


ljedrz (approved):
This is a good point for a release, and the next one should probably be after the snarkVM update :+1:.

unordered-set:
Could you please also update tags/v1.3.11 ? Thank you!

PR #978, opened 2021-07-24
[Ignore] groth16 private branch

howardwu:
[Ignore] groth16 private branch

Its snarkVM cousin may be found here https://github.com/AleoHQ/snarkVM/pull/268

## Deltas

- Removes requirement from `Ledger` and `Storage` that memos must be unique
- Switches node server port to `14131`
- Removes `snarkOS-parameters` module
- Removes hardcoded transaction size in `snarkvm-consensus/src/memory_pool.rs`
- Updates to use `Parameters::inner_circuit_id()` instead of computing the ID
- Removes storage of memory pool (now truly a memory pool) (previously needed this function for testnet1)

## Development

Currently trialing:
```bash
rm -rf ~/.snarkOS  && cargo run --release -- --is-miner --miner-address aleo1shhq355tyaptcej65tkrweej5wth7wqg5hcqg3hf8as5s9rtrypqs8vy3u
```


ljedrz:
@howardwu do we still need to keep this branch open? The same applies to https://github.com/AleoHQ/snarkOS/pull/1139.

howardwu:
Yes, we do.

PR #997, opened 2021-07-28
Make network crawlers disregard their connection count when requesting peer lists

ljedrz:
When the crawler functionality was separated from the bootnodes, one spot was omitted where a bootnode-specific behavior was desirable for network crawlers; this PR accounts for crawlers there now, and provides a new method, `Config::is_regular_node`, which makes handling bootnode+crawler-specific (or regular-node-only) behavior easier.

In addition, the bootnodes and crawlers no longer periodically disconnect from semi-random peers, but from the oldest ones, in order to cycle their peers better. A few comments were also tweaked for better readability.

@howardwu the new release hasn't been tagged or pushed to [crates.io](https://crates.io) yet, correct? If not, then it would be a good idea to include this change in it.

### DIFF
--- a/network/src/peers/peers.rs
+++ b/network/src/peers/peers.rs
@@ -60,8 +60,8 @@ impl<S: Storage + Send + Sync + 'static> Node<S> {
             // Crawlers disconnect down to the min peer count, this to free up room for
             // the next crawled peers...
             let number_to_disconnect = active_peer_count.saturating_sub(min_peers);
-            // ...then they connect to disconnected peers leaving 20% of their capacity open
-            // incoming connections.
+            // ...then they connect to disconnected peers leaving 20% of their capacity open to
+            // potential incoming connections.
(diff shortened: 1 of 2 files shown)


network/src/peers/peers.rs
```diff
@@ -247,12 +250,13 @@ impl<S: Storage + Send + Sync + 'static> Node<S> {
 
     /// Broadcasts a `GetPeers` message to all connected peers to request for more peers.
     async fn broadcast_getpeers_requests(&self) {
-        // Check that this node is not a bootnode.
-        if !self.config.is_bootnode() {
+        // If the node is not a bootnode or a crawler, check if the request for peers is needed
+        // based on the number of active connections.
+        if self.config.is_regular_node() {
```
ljedrz, line 255: this is the check that accidentally omitted (and effectively included) crawlers when they were separated from bootnodes

howardwu (approved):

PR #999, opened 2021-07-28
Make the outbound metric not grow with time

ljedrz:
In some disconnect-intensive scenarios (e.g. a network crawler node), the queued outbound message count metric would gradually retain some entries with time. This PR "tightens" the way this metric is modified, and appears to solve the issue when tested with a crawler over an extended period of time.

### DIFF
--- a/network/src/peers/peer/outbound_handler.rs
+++ b/network/src/peers/peer/outbound_handler.rs
@@ -64,9 +64,10 @@ impl PeerHandle {
     }
 
     pub async fn send_payload(&self, payload: Payload) {
-        metrics::increment_gauge!(OUTBOUND, 1.0);
-        self.queued_outbound_message_count.fetch_add(1, Ordering::SeqCst);
-        self.sender.send(PeerAction::Send(payload)).await.ok();
+        if self.sender.send(PeerAction::Send(payload)).await.is_ok() {
+            self.queued_outbound_message_count.fetch_add(1, Ordering::SeqCst);
+            metrics::increment_gauge!(OUTBOUND, 1.0);
+        }
     }
 
     pub async fn cancel_sync(&self) {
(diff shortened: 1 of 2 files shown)


network/src/peers/peer/outbound_handler.rs
```diff
@@ -64,9 +64,10 @@ impl PeerHandle {
     }
 
     pub async fn send_payload(&self, payload: Payload) {
-        metrics::increment_gauge!(OUTBOUND, 1.0);
-        self.queued_outbound_message_count.fetch_add(1, Ordering::SeqCst);
-        self.sender.send(PeerAction::Send(payload)).await.ok();
+        if self.sender.send(PeerAction::Send(payload)).await.is_ok() {
+            self.queued_outbound_message_count.fetch_add(1, Ordering::SeqCst);
+            metrics::increment_gauge!(OUTBOUND, 1.0);
+        }
```
ljedrz, line 70: this increment now happens only if the `send` operation succeeds

network/src/peers/peer/outbound_handler.rs
```diff
@@ -110,8 +111,9 @@ impl Peer {
                 })?;
 
                 metrics::increment_counter!(metrics::outbound::ALL_SUCCESSES);
-                metrics::decrement_gauge!(OUTBOUND, 1.0);
+
                 self.queued_outbound_message_count.fetch_sub(1, Ordering::SeqCst);
+                metrics::decrement_gauge!(OUTBOUND, 1.0);
```
ljedrz, line 116: this ordering change is for better readability, so that it's more clear that this metric is tightly bound to an atomic

network/src/peers/peer_book.rs
```diff
@@ -65,6 +65,10 @@ impl PeerBookRef {
                 }
                 PeerEventData::Disconnect(peer, status) => {
                     self.connected_peers.remove(peer.address).await;
+
+                    let queued_outbound_message_count = peer.queued_outbound_message_count.swap(0, Ordering::SeqCst);
+                    metrics::decrement_gauge!(snarkos_metrics::queues::OUTBOUND, queued_outbound_message_count as f64);
```
ljedrz, line 70: the fact this decrement was missing is most likely the primary cause of the issue

howardwu (approved):

PR #1002, opened 2021-07-29
Network crawler tweaks

ljedrz:
This PR introduces a few crawler-related improvements:
- some code deduplication
- removal of a lock over a crawler-related `Receiver`
- removal of one call to `RwLock::write` (taking advantage of a nearby scope instead)
- providing a value for `node_count` even before any connections between peers are identified

### DIFF
--- a/network/src/node.rs
+++ b/network/src/node.rs
@@ -199,11 +199,9 @@ impl<S: Storage + Send + core::marker::Sync + 'static> Node<S> {
             let node_clone = self.clone();
 
             let known_network_task = task::spawn(async move {
+                let known_network = node_clone.known_network().unwrap();
                 loop {
-                    // Should always be present since we check for it before this block.
-                    if let Some(known_network) = node_clone.known_network() {
-                        known_network.update().await
-                    }
+                    known_network.update().await
                 }
             });
             self.register_task(known_network_task);
--- a/network/src/node.rs
+++ b/network/src/node.rs
@@ -199,6 +199,7 @@ impl<S: Storage + Send + core::marker::Sync + 'static> Node<S> {
             let node_clone = self.clone();
 
             let known_network_task = task::spawn(async move {
+                // Safe since we check the presence of `known_network`.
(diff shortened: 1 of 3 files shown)


rpc/src/rpc_impl.rs
```diff
@@ -421,8 +425,14 @@ impl<S: Storage + Send + core::marker::Sync + 'static> RpcFunctions for RpcImpl<
             .map(|(height, members)| PotentialFork { height, members })
             .collect();
 
+        let node_count = if network_metrics.node_count == 0 {
+            known_network.nodes().len()
+        } else {
+            network_metrics.node_count
```
niklaslong, line 431: These should have the same value, since the `nodes` collection is built from the connections in both cases?

network/src/node.rs
```diff
@@ -199,11 +199,11 @@ impl<S: Storage + Send + core::marker::Sync + 'static> Node<S> {
             let node_clone = self.clone();
 
             let known_network_task = task::spawn(async move {
-                loop {
-                    // Should always be present since we check for it before this block.
-                    if let Some(known_network) = node_clone.known_network() {
-                        known_network.update().await
-                    }
+                let known_network = node_clone.known_network().unwrap();
```
niklaslong, line 202:
```suggestion
                // Safe since we check the presence of `known_network`. 
                let known_network = node_clone.known_network().unwrap();
```

rpc/src/rpc_impl.rs
ljedrz: not always; until the node gets any lists of peers, it has no `connections`, but it's very likely to have some `nodes` due to the calls to `update_height` caused by post-handshake pings.

network/src/node.rs
ljedrz: sure, why not; also, I'd just have gone with `if let Some` earlier on, but we're spawning a task here, and `known_network()` returns a reference, so it wasn't viable

niklaslong (approved):

PR #1014, opened 2021-08-04
Network metrics benchmark

niklaslong:
A benchmark for the network metrics computation. Thankfully it's fairly fast for high connection counts (800k takes ~`1.6s` on my machine), since the adjacency matrices are fairly sparse, even for dense meshes.


ljedrz (approved):
LGTM :ok_hand:

PR #1022, opened 2021-08-09
Update the node cleanup test

niklaslong:
This PR updates the test checking leaks in the node's connection handling.

Results: 
```
---- heap use summary ----

before node setup:     56kB
after node setup:      105kB
after 32 connections:  123kB
after 256 connections: 123kB

average use: 122kB
maximum use: 25501kB
growth:      0B, 0.00%
```

### DIFF
--- a/network/src/peers/peer_book.rs
+++ b/network/src/peers/peer_book.rs
@@ -266,6 +266,10 @@ impl PeerBook {
         debug!("Added {} to the peer book", address);
     }
 
+    pub async fn remove_peer(&self, address: SocketAddr) -> Option<Peer> {
+        self.disconnected_peers.remove(address).await
+    }
+
     ///
     /// Returns the `SocketAddr` of the last seen peer to be used as a sync node, or `None`.
     ///
--- a/network/src/peers/peer_book.rs
+++ b/network/src/peers/peer_book.rs
@@ -266,7 +266,10 @@ impl PeerBook {
         debug!("Added {} to the peer book", address);
     }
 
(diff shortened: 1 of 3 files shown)


ljedrz:
Cc https://github.com/AleoHQ/snarkOS/issues/1015

network/src/peers/peer_book.rs
```diff
@@ -266,6 +266,10 @@ impl PeerBook {
         debug!("Added {} to the peer book", address);
     }
 
+    pub async fn remove_peer(&self, address: SocketAddr) -> Option<Peer> {
```
ljedrz, line 269: We might want to comment on the fact that this function is only used in tests; otherwise it would make sense to first check the collection of connected peers and perform a disconnect first (which is not needed for test purposes).

Alternatively, the function could be renamed to `remove_disconnected` peer, which would make its purpose clear as well.

network/tests/cleanup.rs
```diff
@@ -77,52 +77,79 @@ async fn check_node_cleanup() {
     #[global_allocator]
     static PEAK_ALLOC: PeakAlloc = PeakAlloc;
 
+    const NUM_CONNS: usize = 256;
```
ljedrz, line 80: I'd just add a short comment here explaining that it's the connection-related leaks this test is checking for. Also, since the connection count is lower now, does the test take a long time to complete? Otherwise it might be a good idea to un-`#[ignore]` it.

ljedrz (approved):
LGTM with some minor suggestions :+1:

Protryon (approved):

PR #1024, opened 2021-08-09
Increase block cache size

ljedrz:
This PR extends the size of the `Block` dedup cache from `64` to `16 * 1024`; a new, faster hash is used, which improves the performance of cache operations by `33%` (the bigger cache with a faster hash is actually faster than the current small cache). A related benchmark was included in the workspace benches.

In addition, `Block` cache hits will now affect the `BLOCKS` and `DUPLICATE_BLOCKS` metrics.

Cc https://github.com/AleoHQ/snarkOS/issues/1023.


ljedrz:
note: this PR doesn't fully resolve the https://github.com/AleoHQ/snarkOS/issues/1023 issue, but it appears to mitigate it to some extent

Protryon (approved):

Protryon:
I'm all for perf. improvements, but I would expect increasing the cache size to have minimal impact since the rebroadcast block is likely to be the latest block or very close to it. Thoughts?

niklaslong (approved):

ljedrz:
>I would expect increasing the cache size to have minimal impact since the rebroadcast block is likely to be the latest block or very close to it. Thoughts?

I was being conservative with its size here just to be extra safe, as it didn't impact performance at all; but agreed, once we're seeing a perf improvement in the testnet wrt queue utilization, this size could be reduced.

PR #1026, opened 2021-08-10
Add fork points to sync

### LINKED ISSUE #979: [Feature] Send fork points in block locator hashes
To alleviate issues with syncing with forks around, we could see places we have forked > 64 blocks, and include those critical points as hashes in block locator hashes.

Protryon:
Fixes #979.

This PR adds the tip of forks near canon tip to block locator hashes. This avoids the issues outlined in #979.

Future work may include doing more than just tip of forks, but this is sufficient in current testing and has minimal impact for an initial rollout.

Tested to tip canon tip twice and over a few days in local testing, no issues.

### DIFF
--- a/consensus/src/consensus/inner/mod.rs
+++ b/consensus/src/consensus/inner/mod.rs
@@ -25,7 +25,17 @@ use crate::{
     MemoryPool,
 };
 use anyhow::*;
-use snarkos_storage::{BlockFilter, BlockOrder, BlockStatus, Digest, DynStorage, ForkDescription, SerialBlock, SerialTransaction, VMTransaction};
+use snarkos_storage::{
+    BlockFilter,
+    BlockOrder,
+    BlockStatus,
+    Digest,
+    DynStorage,
+    ForkDescription,
+    SerialBlock,
+    SerialTransaction,
+    VMTransaction,
+};
 use snarkvm_dpc::{
     testnet1::{instantiated::Components, Record as DPCRecord, TransactionKernel},
     DPCScheme,
--- a/consensus/src/consensus/inner/mod.rs
+++ b/consensus/src/consensus/inner/mod.rs
@@ -25,15 +25,7 @@ use crate::{
     MemoryPool,
 };
 use anyhow::*;
-use snarkos_storage::{
-    BlockStatus,
-    Digest,
-    DynStorage,
-    ForkDescription,
-    SerialBlock,
-    SerialTransaction,
-    VMTransaction,
-};
+use snarkos_storage::{BlockFilter, BlockOrder, BlockStatus, Digest, DynStorage, ForkDescription, SerialBlock, SerialTransaction, VMTransaction};
 use snarkvm_dpc::{
     testnet1::{instantiated::Components, Record as DPCRecord, TransactionKernel},
     DPCScheme,
@@ -68,28 +60,38 @@ struct LedgerData {
 
 impl ConsensusInner {
     /// scans uncommitted blocks for forks
-    async fn scan_forks(&mut self) -> Result<()> {
-        let blocks = self
+    async fn scan_forks(&mut self) -> Result<Vec<(Digest, Digest)>> {
+        let canon_hashes = self
             .storage
-            .get_block_hashes(None, snarkos_storage::BlockFilter::NonCanonOnly)
+            .get_block_hashes(
+                Some(crate::OLDEST_FORK_THRESHOLD as u32 + 1),
+                BlockFilter::CanonOnly(BlockOrder::Descending),
+            )
             .await?;
-        info!("scanning {} blocks for forks", blocks.len());
-        for hash in blocks {
-            let header = self.storage.get_block_header(&hash).await?;
-            let parent_state = self.storage.get_block_state(&header.previous_block_hash).await?;
-            match parent_state {
-                BlockStatus::Unknown => continue, // orphan
-                BlockStatus::Committed(_) => (),  // uncomitted child of canon, could be a fork or a hanging block
-                BlockStatus::Uncommitted => {
-                    // mid-fork or orphan chain, wait for head of fork
-                    continue;
-                }
-            };
-            let block = self.storage.get_block(&hash).await?;
 
-            self.try_commit_block(&hash, &block).await?;
+        if canon_hashes.len() < 2 {
+            // windows will panic if len < 2
+            return Ok(vec![]);
         }
-        Ok(())
+
+        let mut known_forks = vec![];
+
+        for canon_hashes in canon_hashes.windows(2) {
+            // windows will ignore last block (furthest down), so we pull one extra above
+            let target_hash = &canon_hashes[1];
+            let ignore_child_hash = &canon_hashes[0];
+            let children = self.storage.get_block_children(target_hash).await?;
+            if children.len() == 1 && &children[0] == ignore_child_hash {
+                continue;
+            }
+            for child in children {
+                if &child != ignore_child_hash {
+                    known_forks.push((target_hash.clone(), child));
+                }
+            }
+        }
+
+        Ok(known_forks)
     }
 
     fn fresh_ledger(&self, blocks: Vec<SerialBlock>) -> Result<LedgerData> {
--- a/network/src/lib.rs
+++ b/network/src/lib.rs
@@ -49,7 +49,7 @@ pub mod topology;
 pub use topology::*;
 
 /// The maximum number of block hashes that can be requested or provided in a single batch.
-pub const MAX_BLOCK_SYNC_COUNT: u32 = snarkos_storage::NUM_LOCATOR_HASHES;
+pub const MAX_BLOCK_SYNC_COUNT: u32 = snarkos_storage::NUM_LOCATOR_HASHES * 2;
 /// The maximum amount of time allowed to process a single batch of sync blocks. It should be aligned
 /// with `MAX_BLOCK_SYNC_COUNT`.
 pub const BLOCK_SYNC_EXPIRATION_SECS: u8 = 30;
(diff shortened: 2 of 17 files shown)


storage/src/key_value/agent/block.rs
```diff
@@ -223,34 +227,49 @@ impl<S: KeyValueStorage + 'static> Agent<S> {
...
+        &mut self,
+        points_of_interest: Vec<Digest>,
+        oldest_fork_threshold: usize,
+    ) -> Result<Vec<Digest>> {
         let canon = self.canon()?;
+        let target_height = canon.block_height as u32;
 
         // The number of locator hashes left to obtain; accounts for the genesis block.
-        let mut num_locator_hashes = std::cmp::min(crate::NUM_LOCATOR_HASHES - 1, canon.block_height as u32);
+        let mut num_locator_hashes = std::cmp::min(crate::NUM_LOCATOR_HASHES - 1, target_height);
```
ljedrz, line 239: if the final number of locator hashes is greater than `NUM_LOCATOR_HASHES`, the `snarkos_network::MAX_BLOCK_SYNC_COUNT` value needs to be adjusted, so that more hashes can be processed - otherwise, the hashes above `NUM_LOCATOR_HASHES` would be ignored (it was an anti-DoS measure, so the node can't be spammed with too many locator hashes.

storage/src/key_value/agent/block.rs
```diff
@@ -223,34 +227,49 @@ impl<S: KeyValueStorage + 'static> Agent<S> {
...
 
         // The output list of block locator hashes.
-        let mut block_locator_hashes = Vec::with_capacity(num_locator_hashes as usize);
+        let mut block_locator_hashes = Vec::with_capacity(num_locator_hashes as usize + points_of_interest.len());
+
+        for hash in points_of_interest {
+            trace!("block locator hash -- interesting: block# none: {}", hash);
```
ljedrz, line 245: what does `block# none` refer to?

storage/src/key_value/agent/block.rs
```diff
@@ -223,34 +227,49 @@ impl<S: KeyValueStorage + 'static> Agent<S> {
...
+            block_locator_hashes.push(hash);
```
ljedrz, line 246: should the fork hashes be provided before the top hashes? those will probably constitute the happy path

storage/src/key_value/agent/block.rs
```diff
@@ -223,34 +227,49 @@ impl<S: KeyValueStorage + 'static> Agent<S> {
...
-        let mut hash_index = canon.block_height as u32;
+        let mut hash_index = target_height;
 
         // The number of top blocks to provide locator hashes for.
         let num_top_blocks = std::cmp::min(10, num_locator_hashes);
 
         for _ in 0..num_top_blocks {
-            block_locator_hashes.push(self.get_block_hash_guarded(hash_index)?);
+            let hash = self.get_block_hash_guarded(hash_index)?;
+            trace!("block locator hash -- top: block# {}: {}", hash_index, hash);
```
ljedrz, line 257: do we still need the `trace` logs here? they could get very verbose, so perhaps it would be better to comment them out?

storage/src/key_value/agent/block.rs
```diff
@@ -262,14 +281,17 @@ impl<S: KeyValueStorage + 'static> Agent<S> {
...
+            block_locator_hashes.push(hash);
             hash_index = hash_index.saturating_sub(quadratic_step);
             quadratic_step *= 2;
         }
 
         // Update the size of the proportional step so that the hashes of the remaining blocks have the same distance
         // between one another.
-        proportional_step = hash_index / num_proportional_steps;
+        proportional_step =
+            (hash_index.min(oldest_fork_threshold as u32) / num_locator_hashes).min(crate::NUM_LOCATOR_HASHES - 1);
```
ljedrz, line 294: does this still turn out to be split into good intervals with the divisor changing from `num_proportional_steps` to `num_locator_hashes`? do you have e.g. a playground link (like [the one from the previous related PR](https://play.rust-lang.org/?version=stable&mode=release&edition=2018&gist=a294d01d9c2d247104bb94afef83a230) I could test some input values against?

storage/src/key_value/agent/block.rs
```diff
@@ -367,17 +393,30 @@ impl<S: KeyValueStorage + 'static> Agent<S> {
...
+                let mut values = self.inner.get_column(KeyValueColumn::BlockIndex)?
+                    .into_iter()
+                    .filter(|(key, value)| key.len() != 4 && value.len() == 4) // only interested in block hash keys
+                    .map(|(key, value)| (key[..].into(), u32::from_le_bytes((&value[..]).try_into().unwrap())))
+                    .collect::<Vec<(Digest, u32)>>();
+                values.sort_by(|a, b| a.1.cmp(&b.1));
+                match order {
+                    BlockOrder::Ascending => values.into_iter().map(|x| x.0).collect(),
+                    BlockOrder::Descending => values.into_iter().rev().map(|x| x.0).collect(),
+                    BlockOrder::Unordered => unimplemented!(),
```
ljedrz, line 413:
```suggestion
                    BlockOrder::Unordered => unreachable!(),
```

ljedrz:
it would be great if we could have a test for this functionality; one of the `consensus_sidechain` ones could be copied and modified to check if the fork points are returned as expected, and if they belong to the locator hash list

ljedrz (commented):
left a handful of open questions

storage/src/key_value/agent/block.rs
Protryon: blocks that are points of interest are not in canon chain (in practice) -- they therefore do not have a block#.

storage/src/key_value/agent/block.rs
Protryon: we don't want to be sidetracked by a fork -- it ultimately shouldn't matter though.

storage/src/key_value/agent/block.rs
Protryon: it's trace level, and I found them quite useful for diagnosing issues in the field -- considering we don't fetch block locator hashes more than once a sync cycle, I think it's fine

storage/src/key_value/agent/block.rs
Protryon: I don't currently have a playground link, but it's based on the idea of any gaps > `sync_block_count` will be a bad thing (we get no valid sync blocks), so let's just not have gaps. let me know if you need a more compelling argument

network/src/lib.rs
```diff
@@ -50,7 +50,7 @@ pub mod topology;
 pub use topology::*;
 
 /// The maximum number of block hashes that can be requested or provided in a single batch.
-pub const MAX_BLOCK_SYNC_COUNT: u32 = snarkos_storage::NUM_LOCATOR_HASHES;
+pub const MAX_BLOCK_SYNC_COUNT: u32 = snarkos_storage::NUM_LOCATOR_HASHES * 2;
```
ljedrz, line 53: this should be fine, as long as there are fewer than `NUM_LOCATOR_HASHES` fork points - might be a good idea to limit them to that

network/src/sync/master.rs
```diff
@@ -80,7 +80,27 @@ impl SyncMaster {
...
 
     async fn block_locator_hashes(&mut self) -> Result<Vec<Digest>> {
-        match self.node.storage.get_block_locator_hashes().await {
+        let forks_of_interest = self.node.expect_sync().consensus.scan_forks().await?;
+        let blocks_of_interest: Vec<Digest> = forks_of_interest.into_iter().map(|(_canon, fork)| fork).collect();
+        let mut tips_of_blocks_of_interest: Vec<Digest> = Vec::with_capacity(blocks_of_interest.len());
+        for block in blocks_of_interest {
+            if tips_of_blocks_of_interest.len() > crate::MAX_BLOCK_SYNC_COUNT as usize {
+                debug!("reached limit of blocks of interest in sync block locator hashes");
+                continue;
```
ljedrz, line 89:
```suggestion
                break;
```

Protryon: lol yeah, good catch. fixed.

consensus/src/consensus/inner/mod.rs
```diff
@@ -68,28 +70,38 @@ struct LedgerData {
 
 impl ConsensusInner {
     /// scans uncommitted blocks for forks
```
ljedrz, line 72:
```suggestion
    /// scans uncommitted blocks with a known path to the canon chain for forks
```

network/src/sync/master.rs
```diff
@@ -80,7 +80,27 @@ impl SyncMaster {
     }
 
     async fn block_locator_hashes(&mut self) -> Result<Vec<Digest>> {
-        match self.node.storage.get_block_locator_hashes().await {
+        let forks_of_interest = self.node.expect_sync().consensus.scan_forks().await?;
+        let blocks_of_interest: Vec<Digest> = forks_of_interest.into_iter().map(|(_canon, fork)| fork).collect();
+        let mut tips_of_blocks_of_interest: Vec<Digest> = Vec::with_capacity(blocks_of_interest.len());
```
ljedrz, line 85:
```suggestion
        let blocks_of_interest = forks_of_interest.into_iter().map(|(_canon, fork)| fork);
        let mut tips_of_blocks_of_interest: Vec<Digest> = Vec::new();
```

ljedrz (approved):
LGTM with a small suggestion :+1:

network/src/sync/master.rs
Protryon: iter needs to be mut to be called. why remove the capacity?

ljedrz: not sure which `mut` you have in mind (as the one here wasn't removed), but `blocks_of_interest` can just be an iterator; the capacity is removed, because small forks are not included, and I think it's the most usual case we'll encounter.

niklaslong (approved):

PR #1028, opened 2021-08-10
Lower invalid block error message verbosity

Protryon:
* error => warn generally for invalid blocks
* warn => debug for `InvalidBlock` errors
* used a `Digest` for `InvalidBlock` errors to get a nicer display.


ljedrz (approved):
LGTM in general, but we should probably check what the specific issue with the block is, in order to `ERROR` or `WARN` in the non-digest cases; this might require some new error variants.

PR #1030, opened 2021-08-10
Safer initial fast-forward

Protryon:
This PR doesn't error out (and therefore panic) when the longest chain of canon tip contains invalid blocks.


ljedrz (approved):
LGTM :+1:; my node no longer breaks locally on startup

PR #1031, opened 2021-08-10
Delegate Inbound

Protryon:
This PR should alleviate the high inbound queue depth. Right now all inbound requests from all peers are running through one task, and will block to receive blocks from one peer or to send sync blocks to one peer.


ljedrz (approved):
I think this has the potential to improve things :+1:

PR #1034, opened 2021-08-10
Update log level for invalid transactions

Protryon:


ljedrz (approved):
:+1:, with the same note as before, i.e. that we might want to have more error variants so that we can ignore the tolerable ones and warn on the ones that might indicate an issue.

PR #1036, opened 2021-08-11
Fix the block height metric

niklaslong:
Fixes the block height metric by moving its update next to the block commit. The value is also fully overwritten instead of incrementing by 1 (which could lead the value to being offset).


ljedrz (approved):
LGTM other than one warning:
```
warning: unused imports: `misc::*`, `self as metrics`
  --> network/src/sync/blocks.rs:27:23
   |
27 | use snarkos_metrics::{self as metrics, misc::*};
   |                       ^^^^^^^^^^^^^^^  ^^^^^^^
   |
   = note: `#[warn(unused_imports)]` on by default
```
as a drive by, you could also remove one other warning that we have in the workspace:
```
warning: unused `#[macro_use]` import
  --> rpc/src/lib.rs:27:1
   |
27 | #[macro_use]
   | ^^^^^^^^^^^^
   |
   = note: `#[warn(unused_imports)]` on by default
```

niklaslong:
Fixed, with a few fly-by clippy lints.

PR #1042, opened 2021-08-11
Adjust the storage validator to the new storage

ljedrz:
This PR re-enables the storage validator tool after the recent storage revamp.

Even though it appears to work, the PR is still a draft, as I'm still not sure if I used the best APIs, and due to the fact the validator-related tests are still disabled (as they are incompatible with the validator now).

My intention is for the validator to be as low-level as possible and to remain snappy ~(hence the `KeyValueStorage` API tweaks)~.

### DIFF
--- a/storage/src/key_value/agent/mod.rs
+++ b/storage/src/key_value/agent/mod.rs
@@ -160,6 +160,14 @@ impl<S: KeyValueStorage + Validator + 'static> Agent<S> {
         result
     }
 
+    fn store_init_digest(&mut self, digest: Digest) -> Result<()> {
+        self.inner()
+            .store(KeyValueColumn::DigestIndex, &digest[..], &0u32.to_le_bytes()[..])?;
+        self.inner()
+            .store(KeyValueColumn::DigestIndex, &0u32.to_le_bytes()[..], &digest[..])?;
+        Ok(())
+    }
+
     fn wrap<T, F: FnOnce(&mut Self) -> Result<T>>(&mut self, func: F) -> Result<T> {
         self.inner().begin()?;
         let out = func(self);
@@ -206,6 +214,7 @@ impl<S: KeyValueStorage + Validator + 'static> Agent<S> {
             Message::GetCanonBlocks(limit) => Box::new(self.get_canon_blocks(limit)),
             Message::GetBlockHashes(limit, filter) => Box::new(self.get_block_hashes(limit, filter)),
             Message::Validate(limit, fix_mode) => Box::new(self.validate(limit, fix_mode)),
+            Message::StoreInitDigest(digest) => Box::new(self.wrap(move |f| f.store_init_digest(digest))),
         }
     }
 
--- a/storage/src/key_value/agent/mod.rs
+++ b/storage/src/key_value/agent/mod.rs
@@ -17,6 +17,7 @@
 use std::{
     any::Any,
     collections::{HashMap, VecDeque},
+    mem,
 };
 
 use snarkvm_dpc::{
@@ -41,18 +42,23 @@ use crate::{
     SerialTransaction,
     TransactionLocation,
     VMRecord,
+    Validator,
 };
 
 mod block;
 mod block_commit;
 
 pub(super) struct Agent<S: KeyValueStorage + 'static> {
-    inner: S,
+    inner: Option<S>, // the option is only for the purposes of validation, which requires ownership
 }
 
-impl<S: KeyValueStorage + 'static> Agent<S> {
+impl<S: KeyValueStorage + Validator + 'static> Agent<S> {
     pub(super) fn new(inner: S) -> Self {
-        Agent { inner }
+        Agent { inner: Some(inner) }
+    }
+
+    fn inner(&mut self) -> &mut S {
+        self.inner.as_mut().unwrap() // safe, as it's always available
     }
 
     fn read_u32(value: &[u8]) -> Result<u32> {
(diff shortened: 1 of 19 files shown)


storage/src/key_value/mod.rs
```diff
@@ -33,18 +33,20 @@ use agent::Agent;
...
-    fn get<'a>(&'a mut self, column: KeyValueColumn, key: &[u8]) -> Result<Option<Value<'a>>>;
+    fn get<'a>(&'a self, column: KeyValueColumn, key: &[u8]) -> Result<Option<Value<'a>>>;
 
-    fn exists(&mut self, column: KeyValueColumn, key: &[u8]) -> Result<bool>;
+    fn exists(&self, column: KeyValueColumn, key: &[u8]) -> Result<bool>;
 
-    fn get_column_keys<'a>(&'a mut self, column: KeyValueColumn) -> Result<Vec<Value<'a>>>;
+    fn get_column_keys<'a>(&'a self, column: KeyValueColumn) -> Result<Vec<Value<'a>>>;
 
-    fn get_column<'a>(&'a mut self, column: KeyValueColumn) -> Result<Vec<(Value<'a>, Value<'a>)>>;
```
Protryon, line 42: these must be `mut` to ensure proper thread safety. please do not change the mutability of these APIs. rocksdb crate doesn't ensure thread safety. Feel free to call rocksDB directly if you need it.

Protryon (changes requested):
We can't change the KeyValueStorage to non-mut members, it violates thread safety internal to RocksDB and changes the entire paradigm of what it's supposed to mean (single-threaded, strongly consistent). Use rocksDB directly if you want to call it from multiple threads.

RocksDB colloquially works if you read from several threads (while not writing) -- but we don't currently have an interface to support this. An option might be to request a read-only interface, taking the mutable reference, and pass this into a `Box<Fn>` or extract it somehow else.

Protryon (dismissed):

Protryon:
LGTM -- in general, if we see any issues in storage (barring unmigrated/unmitigated issues from pre-refactor) it should be critically errored/reported as a bug.

ljedrz:
Ok, the PR is now in a shape good enough for merging - the most important validator-specific test and all the consensus sidechain tests now work and pass too.

Changes since the last review:
- make the storage validator apply to all `T: KeyValueStorage`
- change a duplicate lookup for `cm` to `digest` (a copy-paste error during the rewrite)
- change one instance of `Arc<dyn Storage>` to `DynStorage`
- enable storage validation for `DynStorage`
- exclude ledger digest indices in one of the storage validator checks (I think the indices hadn't been stored as values before)
- **store the pre-genesis digest** (this was also a bug before the storage refactor, fixed in https://github.com/AleoHQ/snarkOS/pull/884/commits/957338c4586ebbdff124b66bc3f7d6e57c994ba5)
- restore the storage validator tests (excluding the validator fix tests for now)

niklaslong (commented):
Works nicely for me locally (and with a fresh db) 👍

storage/src/key_value/agent/mod.rs
```diff
@@ -41,18 +42,23 @@ use crate::{
...
     VMRecord,
+    Validator,
 };
 
 mod block;
 mod block_commit;
 
 pub(super) struct Agent<S: KeyValueStorage + 'static> {
-    inner: S,
+    inner: Option<S>, // the option is only for the purposes of validation, which requires ownership
```
Protryon, line 52: can we just decompose the agent? an agent with no inner just panics anyways -- and this makes the code less clean.

storage/src/key_value/agent/mod.rs
```diff
@@ -142,13 +148,33 @@ impl<S: KeyValueStorage + 'static> Agent<S> {
         Ok(keys.into_iter().map(|(digest, _)| digest).collect())
     }
 
+    fn validate(&mut self, limit: Option<u32>, fix_mode: FixMode) -> bool {
+        let result = if let Some(inner) = mem::take(&mut self.inner) {
+            let (ret, inner) = futures::executor::block_on(inner.validate(limit, fix_mode));
```
Protryon, line 153: please be super, super careful with block_on, and try to avoid it if possible. it can create deadlocks very easily.

Protryon: i question the validator needing ownership (why not a mut ref?) -- also if it does need ownership, clone or even better, just take self on `validate` calls since we terminate execution after validating anyways.

Protryon: also why does it need to be async if presumably the only place calling it is sync?

ljedrz: this function is intended to only ever be called in some tests (and those have been working well), so we should be fine; good point, of course

storage/src/storage.rs
```diff
@@ -160,6 +163,9 @@ pub trait Storage: Send + Sync {
 
     /// Similar to `Storage::get_canon_blocks`, gets hashes of all blocks subject to `filter` and `limit` in block-number ascending order. A maintenance function, not intended for general use.
     async fn get_block_hashes(&self, limit: Option<u32>, filter: BlockFilter) -> Result<Vec<Digest>>;
+
+    /// Performs low-level storage validation; it's mostly intended for test purposes, as there is a lower level `KeyValueStorage` interface available outside of them.
+    async fn validate(&self, limit: Option<u32>, fix_mode: FixMode) -> bool;
```
Protryon, line 168: we should make the docs here generic as to future implementations, which would presumably validate themselves as well -- maybe in a strongly typed relational DB that would be a nop, but thats okay.

storage/src/key_value/agent/mod.rs
Protryon: what do you think about taking ownership then? i.e. `self` target

ljedrz: >also why does it need to be async if presumably the only place calling it is sync?
I guess I could use `try_send` in the associated method (which should remove the need for `async`), let me check :+1:

ljedrz: >what do you think about taking ownership then? i.e. self target

I could try it out

Protryon (approved):

storage/src/key_value/agent/mod.rs
ljedrz: as discussed externally, these changes were currently not viable; they might be possible with planned future changes to the storage

PR #1043, opened 2021-08-12
Fix the inbound::ALL_SUCCESSES metric

### LINKED ISSUE #1025: [Bug] The inbound::ALL_SUCCESSES metric is too low
## 🐛 Bug Report

This has been missed so far due to the fact that the `BLOCKS` metric was missing the duplicate blocks detected by the block cache (see https://github.com/AleoHQ/snarkOS/pull/1024). However, when properly accounted for, its value, combined with the value of the other `inbound` counters, results in a number greater than what `ALL_SUCCESSES` would suggest.

niklaslong:
Ping/Pong messages were forgotten in the metric as they are handled seperately.

Closes #1025.

### DIFF
--- a/network/src/peers/peer/inbound_handler.rs
+++ b/network/src/peers/peer/inbound_handler.rs
@@ -52,11 +52,13 @@ impl Peer {
                     self.fail();
                 }
                 metrics::increment_counter!(PONGS);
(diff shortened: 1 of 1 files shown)


ljedrz (approved):
LGTM :+1:

Protryon (approved):

PR #1047, opened 2021-08-16
Use `hash_hasher` for hash maps and sets in sync

### LINKED ISSUE #1035: [Feature] Consider using hash_hasher for big HashMaps/Sets of block hashes
## 🚀 Feature

The node's sync process sometimes involves gathering a lot of sync hashes, which are then collected into a regular `HashMap`; since these are already cryptographic hashes, it might be a good idea to use something like [hash_hasher](https://crates.io/crates/hash_hasher) to improve its performance.

niklaslong:
This PR improves sync performance by leveraging `hash_hasher::HashedMap` and `HashedSet`, which work under the assumption that the data being stored is already suitably hashed for use as keys. 

I also briefly explored the use of `twox_hash::Xx3hHash`, though this turned out to be slower for maps of `<SocketAddr, Digest>` than the default SipHash 1-3 used by the std library as the keys were too small to see any speed-up. 

In addition to adding benchmarks for the relevant hash map and set operations, this PR includes a fix for the sync benchmark which was affected by the storage refactor.

Closes #1035.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2741,6 +2741,7 @@ dependencies = [
  "defer",
  "derivative",
  "futures 0.3.16",
+ "hash_hasher",
  "hex",
  "itertools",
  "log",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1101,6 +1101,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "62aca2aba2d62b4a7f5b33f3712cb1b0692779a56fb510499d5c0aa594daeaf3"
 
 [[package]]
+name = "hash_hasher"
+version = "2.0.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


niklaslong:
Comparison of `std::collections::HashMap` with `hash_hasher::HashedMap` mapping `<Digest, SocketAddr>. 

Insert: 
Screenshot 2021-08-16 at 15 20 50
Lookup:
Screenshot 2021-08-16 at 15 19 54

niklaslong:
Comparison of `std::collections::HashSet` with `hash_hasher::HashedSet` storing `Digest`s.
Insert:
Screenshot 2021-08-16 at 15 22 03
Set inclusion:
Screenshot 2021-08-16 at 15 21 19

ljedrz (approved):
LGTM :+1:; one suggestion: since there is a commit that was cleanly reverted, it would be best if both commits were just dropped.

niklaslong:
> since there is a commit that was cleanly reverted, it would be best if both the commit and the revert were just dropped.
Rebased ✅

PR #1048, opened 2021-08-17
Send block height within Block and SyncBlock messages

ljedrz:
This PR adds a block height field to the `Block` and `SyncBlock` network messages, so that the recipient node can know what the height of the sender's block is if they knew it themself (which might not be the case with re-propagated blocks).

This change alters the network message format, but due to us using schema-based serialization, it is compatible with existing nodes (they just won't provide this information or be able to interpret it themselves).


niklaslong (approved):
LGTM 👍

Protryon (approved):

PR #1051, opened 2021-08-17
Miner Termination

Protryon:
This PR adds miner termination and updates snarkVM to 0.7.9.

This PR does not include aborting a coinbase transaction formation.

### DIFF
--- a/consensus/tests/miner.rs
+++ b/consensus/tests/miner.rs
@@ -15,17 +15,22 @@
 // along with the snarkOS library. If not, see <https://www.gnu.org/licenses/>.
 
 mod miner {
-    use std::{sync::{Arc, atomic::{AtomicBool, Ordering}}, time::Duration};
+    use std::{
+        sync::{
+            atomic::{AtomicBool, Ordering},
+            Arc,
+        },
+        time::Duration,
+    };
 
     use futures::executor::block_on;
-    use snarkos_consensus::{MineContext, error::ConsensusError};
+    use snarkos_consensus::{error::ConsensusError, MineContext};
     use snarkos_storage::{SerialBlockHeader, SerialTransaction};
     use snarkos_testing::{sync::*, wait_until};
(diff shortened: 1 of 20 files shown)


consensus/tests/miner.rs
```diff
@@ -71,4 +82,52 @@ mod miner {
         let parent_header = genesis().header.into();
         test_find_block(&transactions, &parent_header).await;
     }
+
+    #[tokio::test(flavor = "multi_thread")]
+    async fn terminate_on_block() {
+        tracing_subscriber::fmt::init();
```
ljedrz, line 88:
```suggestion
```
(in order not to get a panic if there are any other tests initializing `tracing_subscriber` in this crate in the future

consensus/tests/miner.rs
```diff
@@ -71,4 +82,52 @@ mod miner {
...
+        let miner_address = FIXTURE_VK.test_accounts[0].address.clone();
+        let miner = MineContext::prepare(miner_address, consensus.clone()).await.unwrap();
+
+        let (sender, mut receiver) = mpsc::channel(10);
+
+        let terminator = Arc::new(AtomicBool::new(false));
+        let terminator_clone = terminator.clone();
+        tokio::task::spawn_blocking(move || {
+            let sender_clone = sender.clone();
+            let mining = async move {
```
ljedrz, line 101: it probably works just fine, but I'm wondering if there was any reason for this `async` block to be put into the blocking task's scope, rather than before (and outside of) it?

ljedrz: ah, ok - it's just a convenience block so it can be blocked on later

ljedrz (commented):
`testing/examples/test_blocks.rs` needs to be adjusted, but it looks like the PR is nearing completion

ljedrz (approved):
LGTM :+1:

PR #1055, opened 2021-08-18
Include Pongs in the outbound msg count metric

ljedrz:
The `Pong`s require an individual call to the related `metrics` macro.

As a drive-by, there's also a small fix for a benchmark pattern.


niklaslong (approved):

network/src/peers/peer/inbound_handler.rs
```diff
@@ -60,6 +60,10 @@ impl Peer {
                 metrics::increment_counter!(PINGS);
                 metrics::increment_counter!(ALL_SUCCESSES);
```
niklaslong, line 61: Suggestion: modify the import (or comment) so that it's obvious this is inbound::ALL_SUCCESSES.

PR #1056, opened 2021-08-18
Pre-inbound queue cache, non-blocking Block deserialization

ljedrz:
This PR moves block cache handling _before_ the inbound queue and delegates `Block` deserialization to dedicated blocking tasks. Using this setup, I'm not seeing inbound queue issues locally with a syncing node with 700 active connections (which is a number which would normally cause my node to experience full inbound queue errors).

The PR also reduces the number of bytes hashed for block cache purposes to the header and halves the block cache size.

Cc https://github.com/AleoHQ/snarkOS/issues/1037

### DIFF
--- a/network/src/node.rs
+++ b/network/src/node.rs
@@ -32,7 +32,7 @@ use std::{
     thread,
 };
 use tokio::{
-    sync::{mpsc, RwLock},
+    sync::{mpsc, Mutex, RwLock},
     task,
     time::sleep,
 };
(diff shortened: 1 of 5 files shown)


ljedrz:
Attaching my queue metrics with these changes after 30m of being connected to 700 peers:
image

niklaslong (approved):

ljedrz:
2 more hours of runtime, still syncing:
image

network/src/node.rs
```diff
@@ -60,12 +60,15 @@ pub struct InnerNode {
     pub config: Config,
     /// The inbound handler of this node.
     pub inbound: Inbound,
+    /// The cache of node's inbound messages.
+    pub inbound_cache: Mutex<Cache>,
```
Protryon, line 64: Should this be a RwLock? seems like it would be better here.

Protryon (approved):

network/src/node.rs
ljedrz: I thought about this, but every cache lookup has the potential to perform a write to it (which is very likely to happen), which would require a second, mutable lock operation; FWIW, I detected no performance degradation related to the introduction of this lock.

PR #1062, opened 2021-08-20
Add a small delay before the storage validator error lock is removed

ljedrz:
Without one, the validator is sometimes a bit too quick to try to remove the `Arc` over the list of errors, which can cause a panic by the end of validation. A small delay seems to solve this.


niklaslong (approved):
Interesting issue—LGTM 👍

PR #1063, opened 2021-08-20
Node performance tweaks

ljedrz:
This PR is the result of monitoring and tweaking the performance of a node that has a lot of connections.

performance improvements:
- convert a vector of hashes just once in `SyncMaster::send_sync_messages`
- reduce the number of syscalls when calculating sync timeouts
- reduce the scope of the Block cache lock (it no longer involves hashing)
- check if the node is syncing before checking the `Block` cache (skips the cache lookup)

drive-by tweaks:
- downgrade the log with the list of prospect sync peers to `TRACE` (very verbose)
- add a missing `DEBUG` log when a `Pong` is sent
- remove an unused `const`
- remove one redundant call to `Node::finished_syncing_blocks`

### DIFF
--- a/network/src/inbound/cache.rs
+++ b/network/src/inbound/cache.rs
@@ -19,32 +19,35 @@ use crate::Payload;
 use snarkvm_dpc::block::BlockHeader;
 
 use circular_queue::CircularQueue;
+use tokio::sync::Mutex;
 use twox_hash::xxh3::hash64;
 
 pub struct Cache {
-    queue: CircularQueue<u64>,
+    queue: Mutex<CircularQueue<u64>>,
 }
 
 impl Default for Cache {
     fn default() -> Self {
         Self {
-            queue: CircularQueue::with_capacity(8 * 1024),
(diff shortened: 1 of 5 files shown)


network/src/inbound/cache.rs
```diff
@@ -19,32 +19,35 @@ use crate::Payload;
 use snarkvm_dpc::block::BlockHeader;
 
 use circular_queue::CircularQueue;
+use tokio::sync::Mutex;
 use twox_hash::xxh3::hash64;
 
 pub struct Cache {
-    queue: CircularQueue<u64>,
+    queue: Mutex<CircularQueue<u64>>,
```
Protryon, line 26: what's the point of drilling the mutex in? it makes it so we can't have a single-threaded cache, but we practically always lock anyways. I'm also depending on using this construct in a single-thread context in block change PR.

ljedrz: that one didn't result in too much difference (just the hash calculation is outside of the scope of the lock now, and the hash is super fast), so I'm ok with reverting this.

niklaslong (approved):
LGTM 👍

PR #1067, opened 2021-08-23
Extra performance metrics

ljedrz:
This PR introduces 4 new queue-related metrics and includes them in the example `grafana` dashboard; the queues that receive these new metrics are:
- the sync master queue
- the consensus agent queue
- the storage agent queue
- the peer event queue

### DIFF
--- a/metrics/grafana/dashboard.json
+++ b/metrics/grafana/dashboard.json
@@ -8,6 +8,12 @@
         "hide": true,
         "iconColor": "rgba(0, 211, 255, 1)",
         "name": "Annotations & Alerts",
+        "target": {
+          "limit": 100,
+          "matchAny": false,
+          "tags": [],
+          "type": "dashboard"
+        },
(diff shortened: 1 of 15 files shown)


metrics/grafana/dashboard.json
```diff
@@ -834,5 +873,5 @@
   "timezone": "",
   "title": "snarkOS node",
   "uid": "PAzNcaCGz",
-  "version": 53
+  "version": 55
 }
```
niklaslong, line 877: nit: missing newline

niklaslong (approved):

metrics/grafana/dashboard.json
ljedrz: this is a straight copy from the `grafana` settings window, so on the off-chance that it dislikes a trailing newline, perhaps it's better to let it be :smiley:

PR #1073, opened 2021-08-24
Introduce the internal RTT metric

### LINKED ISSUE #1050: [Feature] Measure internal RTT
It would be useful to track the rtt internal to snarkos for received message -> sent response as this is currently still an issue in bootnodes.

niklaslong:
This PR introduces the internal RTT metric (both exposed as an average via rpc and present in grafana/prometheus with percentiles) for the query/response messages: `GetPeers`, `GetSync`, `GetBlocks` and `GetMemoryPool`. 

Block processing time will be measured in a subsquent PR (tracking issue: #1076). 

Closes #1050.

### DIFF
--- a/metrics/src/lib.rs
+++ b/metrics/src/lib.rs
@@ -71,7 +71,10 @@ impl Recorder for CombinedRecorder {
 
 #[cfg(feature = "prometheus")]
 pub fn initialize() -> Option<tokio::task::JoinHandle<()>> {
-    let prometheus_builder = metrics_exporter_prometheus::PrometheusBuilder::new();
+    let prometheus_builder = metrics_exporter_prometheus::PrometheusBuilder::new().set_buckets(&[
+        0.00001, 0.000025, 0.00005, 0.000075, 0.0001, 0.0005, 0.001, 0.005, 0.01, 0.05, 0.1, 0.5, 1.0, 2.5, 5.0, 10.0,
+        30.0,
+    ]);
 
     let (prometheus_recorder, exporter) = prometheus_builder
         .build_with_exporter()
--- a/metrics/src/metric_types.rs
+++ b/metrics/src/metric_types.rs
@@ -138,6 +138,7 @@ pub struct CircularHistogram(OnceCell<RwLock<CircularQueue<f64>>>);
 #[allow(dead_code)]
 impl CircularHistogram {
     pub(crate) const fn new() -> Self {
+        // The cell allows the creation of the object in a const fn.
         Self(OnceCell::new())
     }
 
--- a/metrics/src/metric_types.rs
+++ b/metrics/src/metric_types.rs
@@ -130,7 +130,7 @@ impl Gauge {
 use once_cell::sync::OnceCell;
 use parking_lot::RwLock;
 
-const QUEUE_CAPACITY: usize = 1024;
+const QUEUE_CAPACITY: usize = 256;
 
 /// A histogram backed by a circular queue.
 pub struct CircularHistogram(OnceCell<RwLock<CircularQueue<f64>>>);
--- a/metrics/src/metric_types.rs
+++ b/metrics/src/metric_types.rs
@@ -16,6 +16,8 @@
 
 use std::sync::atomic::{AtomicU64, Ordering};
 
+use circular_queue::CircularQueue;
+
 /// Mimics a [`metrics-core`] monotonically increasing [`Counter`] type
 pub struct Counter(AtomicU64);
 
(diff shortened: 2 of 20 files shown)


metrics/src/lib.rs
```diff
@@ -71,7 +71,10 @@ impl Recorder for CombinedRecorder {
 
 #[cfg(feature = "prometheus")]
 pub fn initialize() -> Option<tokio::task::JoinHandle<()>> {
-    let prometheus_builder = metrics_exporter_prometheus::PrometheusBuilder::new();
+    let prometheus_builder = metrics_exporter_prometheus::PrometheusBuilder::new().set_buckets(&[
```
niklaslong, line 74: Note: we may need to adjust these buckets but they seem to work quite will in my local tests.

metrics/src/metric_types.rs
```diff
@@ -124,3 +126,40 @@ impl Gauge {
...
     }
 }
+
+use once_cell::sync::OnceCell;
+use parking_lot::RwLock;
+
+const QUEUE_CAPACITY: usize = 256;
+
+/// A histogram backed by a circular queue.
+pub struct CircularHistogram(OnceCell<RwLock<CircularQueue<f64>>>);
```
ljedrz, line 136: I'm assuming the `OnceCell` is here so that the RPC-related objects can be created with a `const fn`? It's probably cheap enough, but I'm wondering if this could be made simpler :thinking:; otherwise you could add a non-doc comment here on why `OnceCell` is used.

metrics/src/stats.rs
```diff
@@ -324,6 +329,35 @@ impl MiscStats {
     }
 }
 
+/// Each histogram holds the last 1024 measurements for internal RTT for the indicated message
```
ljedrz, line 332: not `256`? if I'm not mistaken, it would be better to replace the literal here with `QUEUE_CAPACITY`

network/src/peers/peer/outbound_handler.rs
```diff
@@ -100,16 +100,29 @@ impl Peer {
...
+
                 network.write_payload(&message).await.map_err(|e| {
                     metrics::increment_counter!(metrics::outbound::ALL_FAILURES);
                     e
                 })?;
 
+                if let (Some(time_received), Some(histogram)) = (time_received, histogram) {
+                    metrics::histogram!(histogram, time_received.elapsed());
+                }
+
```
ljedrz, line 125: I think it would be best to have this together with the block above, perhaps adjusted to `if let Some(histogram) = match &message`

network/src/peers/peer/peer.rs
```diff
@@ -157,14 +157,23 @@ impl Peer {
                     if data.is_none() {
                         break;
                     }
-                    let data = match data.unwrap() {
+                    let  data = match data.unwrap() {
```
ljedrz, line 160:
```suggestion
                    let data = match data.unwrap() {
```

rpc/documentation/public_endpoints/getnodestats.md
```diff
@@ -6,43 +6,47 @@ None
...
+| `internal_rtt.getmemorypool`     | f64  | The average internal RTT for GetMemoryPool messages in seconds               |
+| `misc.block_height`              | u32  | The current block height of the node                                         |
+| `misc.blocks_mined`              | u32  | The number of blocks the node has mined                                      |
+| `misc.duplicate_blocks`          | u64  | The number of duplicate blocks received                                      |
+| `misc.duplicate_sync_blocks`     | u64  | The number of duplicate sync blocks received                                 |
+| `outbound.all_successes`         | u64  | The number of successfully sent messages                                     |
+| `outbound.all_failures`          | u64  | The number of failures to send messages                                      |
+| `queues.inbound`                 | u32  | The number of messages queued in the common inbound channel                  |
+| `queues.outbound`                | u32  | The number of messages queued in the individual outbound channels            |
```
ljedrz, line 49: this reminds me; I forgot to add the description of the new queues from https://github.com/AleoHQ/snarkOS/pull/1067; perhaps you could include them here as a drive-by :innocent:?

niklaslong: Done 👍

network/src/peers/peer/peer.rs
niklaslong: Good catch, for some reasone `rustfmt` is being funky in that long `select!` body.

network/src/peers/peer/outbound_handler.rs
niklaslong: Agreed; consolidated 👍

metrics/src/stats.rs
niklaslong: Indeed, missed this when I reduced the value.

metrics/src/metric_types.rs
niklaslong: Yes—I'm open to suggestions, I've added the doc comment in the meantime.

Protryon (approved):

PR #1074, opened 2021-08-24
Do a better job avoiding mempool syncs and mining when syncing blocks

ljedrz:
This PR introduces an extra delay before the node starts syncing the mempool or mining; this delay is equal to 1 minute since the last block sync attempt. In addition, this PR adjusts the conditions in which the block sync timestamp is updated, so that it happens only if any prospect sync peers are found - this way the node will no longer register as syncing once it's caught up with the tip of the chain, **and** it will no longer mine blocks if it notices that there are peers with a longer chain and it's unable to sync with them, making potential forks shorter.

### DIFF
--- a/network/src/node.rs
+++ b/network/src/node.rs
@@ -238,7 +238,6 @@ impl Node {
                     let is_syncing_blocks = node_clone.is_syncing_blocks();
 
                     if !is_syncing_blocks {
-                        node_clone.register_block_sync_attempt();
                         if let Err(e) = node_clone.run_sync().await {
                             error!("failed sync process: {:?}", e);
                         }
--- a/network/src/node.rs
+++ b/network/src/node.rs
@@ -229,8 +229,26 @@ impl Node {
         self.register_task(state_tracking_task);
 
         if self.sync().is_some() {
-            let bootnodes = self.config.bootnodes();
+            let node_clone = self.clone();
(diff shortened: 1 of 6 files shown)


network/src/node.rs
```diff
@@ -229,13 +231,39 @@ impl Node {
...
+            let bootnodes = self.config.bootnodes();
             let node_clone = self.clone();
             let mempool_sync_interval = node_clone.expect_sync().mempool_sync_interval();
             let sync_mempool_task = task::spawn(async move {
                 loop {
                     if !node_clone.is_syncing_blocks() {
+                        // Ensure that it's not just a short "break" from syncing blocks
+                        #[cfg(not(feature = "test"))]
+                        if let Some(elapsed) = node_clone.expect_sync().time_since_last_block_sync() {
+                            if elapsed < Duration::from_secs(60) {
```
niklaslong, line 261: What was there a particular reason for choosing `60s`?

niklaslong (approved):

network/src/node.rs
ljedrz: Yes; I initially tried to make it dependent on the block sync interval, but I quickly needed to multiply it by a relatively high value in order for the delay to be sufficient; I later observed how long a single sync batch could be processed in a few test runs (it was ~30s for some problematic cases) and just decided to add an additional safety margin for slower PCs.

Protryon (approved):

PR #1075, opened 2021-08-24
Reduce the max peers breach log level 

niklaslong:
Reduces the log level for breached max peers limit. Softly breaching the limit isn't a `WARN`-worthy message, downgraded to `DEBUG`.

### DIFF
--- a/network/src/peers/peers.rs
+++ b/network/src/peers/peers.rs
@@ -380,7 +380,7 @@ impl Node {
         let max_peers = self.config.maximum_number_of_connected_peers() as usize;
 
         if num_connected > max_peers {
-            warn!(
+            info!(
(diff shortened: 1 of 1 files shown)


network/src/peers/peers.rs
```diff
@@ -380,7 +380,7 @@ impl Node {
         let max_peers = self.config.maximum_number_of_connected_peers() as usize;
 
         if num_connected > max_peers {
-            warn!(
+            info!(
```
ljedrz, line 383:
```suggestion
            debug!(
```
this change makes the log less "scary", but this information is probably a bit too verbose for regular users

ljedrz (approved):

PR #1078, opened 2021-08-24
Sqlite implementation

Protryon:
This PR adds Sqlite implementation to snarkOS.

Performance benefits due to not using snarkVM serialization at storage layer, much simpler storage implementation.
Includes minor refactor to storage to cleanup.
Adds `--sqlite` flag to snarkOS, requires `sqlite` compilation feature to enable.

Non-default/opt-in since we don't use unit tests for it yet, and needs to do some more vetting before we consider any full transfer. Primary use cases right now are testing and Aleo Explorer.


storage/src/storage/async_adapter.rs
```diff
@@ -0,0 +1,385 @@
...
+            #[cfg(feature = "test")]
+            Message::StoreItem(col, key, value) => write!(f, "StoreItem({:?}, {:?}, {:?})", col, key, value),
+            #[cfg(feature = "test")]
+            Message::DeleteItem(col, key) => write!(f, "DeleteItem({:?}, {:?})", col, key),
+        }
+    }
+}
+
+pub(super) struct Agent<S: SyncStorage + 'static> {
+    inner: S, // the option is only for the purposes of validation, which requires ownership
```
ljedrz, line 142:
```suggestion
    inner: S,
```

snarkos/main.rs
```diff
@@ -116,24 +120,44 @@ async fn start_server(config: Config) -> anyhow::Result<()> {
...
+    let storage = if config.storage.use_sqlite {
+        #[cfg(feature = "sqlite")]
+        {
+            let mut sqlite_path = path.clone();
+            sqlite_path.push("sqlite.db");
+
+            if config.storage.validate {
+                error!("validator not implemented for sqlite");
+                return Ok(());
+            }
```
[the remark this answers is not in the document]
Protryon: it works on the high level storage implementation, so should work accordingly.

storage/src/storage/async_adapter.rs
```diff
@@ -0,0 +1,385 @@
...
+    pub(super) fn new(inner: S) -> Self {
+        Agent { inner }
+    }
+
+    fn wrap<T, F: FnOnce(&mut S) -> Result<T>>(&mut self, func: F) -> Result<T> {
+        self.inner.transact(func)
+    }
+
+    fn handle_message(&mut self, message: Message) -> Box<dyn Any + Send + Sync> {
+        trace!("received message: {}", message);
```
ljedrz, line 155:
```suggestion
        trace!("received storage request: {}", message);
```
to differentiate this from the network messages

storage/src/storage/async_adapter.rs
```diff
@@ -0,0 +1,385 @@
...
+type MessageWrapper = (Message, oneshot::Sender<Box<dyn Any + Send + Sync>>);
+
+pub struct AsyncStorage {
+    sender: mpsc::Sender<MessageWrapper>,
+}
+
+impl AsyncStorage {
+    pub fn new<S: SyncStorage + Send + 'static>(inner: S) -> AsyncStorage {
+        let (sender, receiver) = mpsc::channel(256);
+        tokio::task::spawn_blocking(move || Agent::new(inner).agent(receiver));
```
ljedrz, line 223: this should ideally be a proper thread

Protryon: agreed. changing.

snarkos/main.rs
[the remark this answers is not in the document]
ljedrz: I meant that the runtime argument `--trim-storage` is not acted on with the `sqlite` feature on (only with `rocksdb`)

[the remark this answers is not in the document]
Protryon: no, we call it below this block, there is logic in non-sqlite to run validator if trim is set, though.

storage/src/storage/sqlite/sync.rs
```diff
@@ -0,0 +1,923 @@
...
+            );
+
+            CREATE TABLE IF NOT EXISTS transaction_blocks(
+                id INTEGER PRIMARY KEY,
+                transaction_id INTEGER NOT NULL REFERENCES transactions(id) ON DELETE CASCADE,
+                block_id INTEGER NOT NULL REFERENCES blocks(id) ON DELETE CASCADE,
+                block_order INTEGER NOT NULL
+            );
+            CREATE UNIQUE INDEX transaction_block_ordering ON transaction_blocks(block_id, block_order);
+            CREATE INDEX transaction_block_looukp ON transaction_blocks(transaction_id);
```
ljedrz, line 96:
```suggestion
            CREATE INDEX transaction_block_lookup ON transaction_blocks(transaction_id);
```

storage/src/storage/sqlite/sync.rs
```diff
@@ -0,0 +1,923 @@
...
+    (&row.get::<_, Vec<u8>>(index)?[..])
+        .try_into()
+        .map_err(|_| rusqlite::Error::InvalidQuery)
+}
+
+impl SqliteStorage {
+    const SCHEMA_INDEX: u32 = 2;
+
+    fn migrate(&self, from: u32) -> Result<()> {
+        if from == 0 {
```
ljedrz, line 48: what is the significance of these indices? they could use some comments

Protryon: the names should be self-describing as to their purpose (i.e. lookup transaction id). any particular confusion?

storage/src/storage/sqlite/sync.rs
```diff
@@ -0,0 +1,923 @@
...
+        }
+        Ok(())
+    }
+
+    fn insert_block(&mut self, block: &SerialBlock) -> Result<()> {
+        let hash = block.header.hash();
+        let mut block_query = self.conn.prepare_cached(
+            r"
+        INSERT INTO blocks (hash,
+            previous_block_id,
```
ljedrz, line 162:
```suggestion
        INSERT INTO blocks (
            hash,
            previous_block_id,
```
for improved readability

storage/src/storage/sqlite/sync.rs
```diff
@@ -0,0 +1,923 @@
...
+                ?,
+                ?,
+                ?
+            )
+        ",
+        )?;
+        block_query.execute::<&[&dyn ToSql]>(&[
+            &hash,
+            &block.header.previous_block_hash,
+            &block.header.previous_block_hash,
```
ljedrz, line 186: this should instead be `canon_ledger_digest`, no?

storage/src/storage/sqlite/sync.rs
```diff
@@ -0,0 +1,923 @@
...
+                ?,
+                ?,
+                ?,
+                ?
+            )
+        ",
+        )?;
+        block_query.execute::<&[&dyn ToSql]>(&[
+            &hash,
+            &block.header.previous_block_hash,
```
ljedrz, line 185: this should instead be canon_ledger_digest, no?

storage/src/storage/sqlite/sync.rs
ljedrz, line 186: should the `previous_block_hash` be inserted twice here?

storage/src/storage/sqlite/sync.rs
```diff
@@ -0,0 +1,923 @@
...
+        //     output.insert(hash.clone(), BlockStatus::Unknown);
+        // }
+        // for row in stmt.query_map(&dyn_hashes[..], |row| Ok((row.get(0)?, row.get(1)?)))? {
+        //     let (hash, canon_height): (Digest, Option<usize>) = row?;
+        //     *output.get_mut(&hash).unwrap() = match canon_height {
+        //         Some(height) => BlockStatus::Committed(height),
+        //         None => BlockStatus::Uncommitted,
+        //     };
+        // }
+        // Ok(output.into_iter().map(|x| x.1).collect())
```
ljedrz, line 364: was this a potential optimization? are you still working on it or should it be removed?

storage/src/storage/sqlite/sync.rs
Protryon: yes, we pass the parameter in twice. one is for a lookup subquery IIRC

storage/src/storage/sqlite/sync.rs
Protryon: it was a superior implementation that didn't work. forgot the specifics.

storage/src/storage/sqlite/sync.rs
```diff
@@ -0,0 +1,923 @@
...
+        INNER JOIN blocks on blocks.id = transaction_blocks.block_id
+        WHERE blocks.hash = ?
+        ORDER BY transaction_blocks.block_order ASC",
+        )?;
+        let rows = stmt.query_map([hash], |row| {
+            Ok(SerialTransaction {
+                id: read_static_blob(row, 0)?,
+                network: Network::from_id(row.get(1)?),
+                ledger_digest: row.get(2)?,
+                old_serial_numbers: vec![row.get(3)?, row.get(4)?],
```
ljedrz, line 400: [applies to new_commitments, signatures and new_records too] do we expect there to ever be more than 2 entries for these items? if not, we could use 2-elem arrays instead of vectors to avoid 4 allocations for each transaction

storage/src/storage/sqlite/sync.rs
```diff
@@ -0,0 +1,923 @@
...
+        })?;
+        Ok(SerialBlock {
+            header,
+            transactions: rows.collect::<rusqlite::Result<_>>()?,
+        })
+    }
+
+    fn commit_block(&mut self, hash: &Digest, ledger_digest: &Digest) -> Result<BlockStatus> {
+        let canon = self.canon()?;
+        // let block = self.get_block(hash)?;
```
ljedrz, line 420:
```suggestion

```

storage/src/storage/sqlite/sync.rs
Protryon: we don't need to hyperoptimize vectors so much. that being said -- i highly doubt we will ever have any other than 2, unless we change our transaction-record model (2 in, 2 out). it's nice to keep the floor open to whatever snarkvm throws at us though, sqlite doesn't have variable length arrays generally (i think there is an extension), so i just took some liberty here to make it work.

storage/src/storage/sqlite/sync.rs
```diff
@@ -0,0 +1,923 @@
...
+                },
+            )
+            .map_err(Into::into)
+    }
+
+    fn store_init_digest(&mut self, digest: &Digest) -> Result<()> {
+        let mut block_query = self.conn.prepare_cached(
+            r"
+        INSERT INTO blocks (hash,
+            previous_block_hash,
```
ljedrz, line 587:
```suggestion
        INSERT INTO blocks (
            hash,
            previous_block_hash,
```

storage/src/storage/sqlite/sync.rs
ljedrz: I meant `from`, `SCHEMA_INDEX` and how they relate to the migration

storage/src/storage/sqlite/sync.rs
```diff
@@ -0,0 +1,923 @@
...
+                        transaction_proof: row.get(14)?,
+                        memorandum: row.get(15)?,
+                        inner_circuit_id: row.get(16)?,
+                    })
+                },
+            )
+            .map_err(Into::into)
+    }
+
+    fn store_init_digest(&mut self, digest: &Digest) -> Result<()> {
```
ljedrz, line 583: I'm not sure how to solve this properly without storing the digests separately; on one hand, we do want the genesis transactions to have the digest available, but on the other, this is a hack; I think we should now instead special-case this digest (in case it's ever queried for) instead of adding a dummy record; this should also simplify things a bit, as this method could then be removed

ljedrz: part of the reason for searching for this particular digest was ensuring that the relationship between all the stored values is enforced in the key-value store model, but the relational model no longer requires this sort of validation

storage/src/storage/sqlite/sync.rs
```diff
@@ -0,0 +1,923 @@
...
+                        is_dummy: row.get(1)?,
+                        value: AleoAmount(row.get(2)?),
+                        payload: row.get(3)?,
+                        birth_program_id: row.get(4)?,
+                        death_program_id: row.get(5)?,
+                        serial_number_nonce: row.get(6)?,
+                        commitment: row.get(7)?,
+                        commitment_randomness: row.get(8)?,
+                        serial_number_nonce_randomness: None,
+                        position: None,
```
ljedrz, line 675: are these fields currently unused (i.e. prepared for `testnet2`)?

storage/src/storage/sqlite/sync.rs
```diff
@@ -0,0 +1,923 @@
...
+    }
+
+    fn get_commitments(&mut self) -> Result<Vec<Digest>> {
+        let mut stmt = self.conn.prepare_cached(
+            r"
+        SELECT
+        transactions.new_commitment1,
+        transactions.new_commitment2
+        FROM transactions
+        INNER JOIN transaction_blocks ON transaction_blocks.transaction_id = transactions.id
```
ljedrz, line 724: do we even still need a `transaction_blocks` table now? couldn't we just store `block_hash` and `block_position` in `transactions`, together with all the other transaction info?

ljedrz: I don't think we even need to store `block_position` now that we no longer store transactions as per-block blobs containing multiple of them

storage/src/storage/sqlite/sync.rs
```diff
@@ -0,0 +1,923 @@
...
+    }
+
+    fn reset_ledger(
+        &mut self,
+        _commitments: Vec<Digest>,
+        _serial_numbers: Vec<Digest>,
+        _memos: Vec<Digest>,
+        _digests: Vec<Digest>,
+    ) -> Result<()> {
+        unimplemented!()
```
ljedrz, line 802: do we still intend to implement this, or is it not needed for `sqlite`?

storage/src/storage/sqlite/sync.rs
```diff
@@ -0,0 +1,923 @@
...
+    }
+
+    fn get_canon_blocks(&mut self, limit: Option<u32>) -> Result<Vec<SerialBlock>> {
+        let digests = self.get_block_hashes(limit, BlockFilter::CanonOnly(BlockOrder::Unordered))?;
+        // this is intentionally N+1 query since this is not a critical performance function and its easy
```
ljedrz, line 807: do we intend to have a `WHERE IN` query later on, just to squeeze out maximum perf? it could be left as a `TODO` or posted as a `help-wanted` GitHub issue

snarkos/main.rs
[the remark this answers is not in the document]
ljedrz: ah, ok - missed it in the diff :+1:

storage/src/storage/sqlite/sync.rs
ljedrz: ah yes, I see it now :+1:; need to get used to this notation

ljedrz (commented):
The SQLite impl looks quite solid to me :+1:; left some comments, the tests need to be adjusted too

niklaslong:
> The SQLite impl looks quite solid to me

Same here, currently testing the branch locally 🧪

PR #1079, opened 2021-08-24
Block Repropagation Improvement

Protryon:
This PR adds a per-peer inbound-populated and outbound-checked cache for block repropagation.

The generally expected flow is:
* Node A receives a block from Peers B, C, D, E, F
* Node A repropagates the block to Peers G, H, I, J, who have not sent Node A the block.

This works in manual testing, needs some unit tests of some form.

### DIFF
--- a/network/src/peers/peer/inbound_handler.rs
+++ b/network/src/peers/peer/inbound_handler.rs
@@ -34,9 +34,8 @@ impl Peer {
         node: &Node,
         network: &mut PeerIOHandle,
         time_received: Option<Instant>,
-        payload: Result<Payload, NetworkError>,
+        payload: Payload,
     ) -> Result<(), NetworkError> {
-        let payload = payload?;
         self.quality.see();
         self.quality.num_messages_received += 1;
         metrics::increment_counter!(inbound::ALL_SUCCESSES);
@@ -70,12 +69,18 @@ impl Peer {
 
                 if node.sync().is_some() {
                     let node = node.clone();
+                    self.block_received_cache.push(&block[..]);
                     task::spawn(async move {
                         // Check if the message hasn't already been processed recently if it's a `Block`.
                         // The node should also reject them while syncing, as it is bound to receive them later.
-                        if node.inbound_cache.lock().await.contains(&block) {
-                            metrics::increment_counter!(misc::DUPLICATE_BLOCKS);
-                            return;
+                        {
+                            let mut inbound_cache = node.inbound_cache.lock().await;
+                            if inbound_cache.contains(&block[..]) {
+                                metrics::increment_counter!(misc::DUPLICATE_BLOCKS);
+                                return;
+                            } else {
+                                inbound_cache.push(&block[..]);
+                            }
                         }
 
                         if node.state() == State::Syncing {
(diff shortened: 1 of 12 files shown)


network/src/peers/peer/inbound_handler.rs
```diff
@@ -76,15 +76,38 @@ impl Peer {
...
+                    metrics::increment_counter!(inbound::BLOCKS);
+
+                    {
+                        let mut inbound_cache = node.inbound_cache.lock().await;
+                        if inbound_cache.contains(&payload) {
+                            metrics::increment_counter!(misc::DUPLICATE_BLOCKS);
+                            return Ok(());
+                        } else {
+                            inbound_cache.push(&payload);
+                        }
```
ljedrz, line 110: the `Block` case was already handled in the branch above, was this code duplicated by accident?

network/src/peers/peer/outbound_handler.rs
```diff
@@ -100,21 +109,45 @@ impl Peer {
...
+            }
             PeerAction::Send(message) => {
                 if matches!(message, Payload::Ping(_)) {
                     self.quality.expecting_pong = true;
                     self.quality.last_ping_sent = Some(Instant::now());
                 }
+
+                self.queued_outbound_message_count.fetch_sub(1, Ordering::SeqCst);
+                metrics::decrement_gauge!(OUTBOUND, 1.0);
+
```
ljedrz, line 143: why was the metric change moved here, before the outbound message has been written to the socket? I think it made more sense for it to be subtracted only once the message is physically out

ljedrz (commented):
left some comments; I agree with the general direction, and would also welcome a test case :+1:

network/src/peers/peer/outbound_handler.rs
Protryon: it's so that cache hits and failures also decrement the outbound metric.

ljedrz: good point wrt `SendBlock`, though to be most correct in terms of the number of items waiting to be written to the socket, we should just add an extra `decrement_gauge` in case of a cache hit and place the existing ones after the writes

Protryon: considering the difference is off by one, i'm inclined to say screw in in favor of cleanliness here. let me know if you disagree.

testing/src/network/sync.rs
```diff
@@ -172,31 +172,26 @@ async fn block_responder_side() {
...
-        });
+    wait_until!(5, node.storage.canon().await.unwrap().block_height == 1);
+    wait_until!(
+        5,
+        matches!(peer2.read_payload().await.unwrap(), Payload::Block(x, Some(1)) if x == block_1)
+    );
+    wait_until!(5, match peer.read_payload().await.unwrap() {
+        Payload::Block(_, _) => unreachable!(),
+        Payload::Ping(1) => true,
+        _ => false,
```
ljedrz, line 194: ok, so if I'm not mistaken, we're checking here that `peer2` is able to obtain the `Block` from the `node` (who got it from `peer`), while `peer` is _not_ getting one from  either `node` or `peer2`? sounds good, though we should separately first check that `peer` gets `Ping(1)` and only _then_ we need to ensure that it doesn't receive a `Block` again; this might require the use of a short-lived `tokio::time::timeout` over `peer.read_payload()`, as `wait_until!` requires some breaking condition

Protryon: `peer2` is not connected to `peer` in this case, which should help the test make sense.

ljedrz: yep, I checked the default test settings and indeed it should not sync peers :+1:

ljedrz: but I forgot that it means the `Ping` from the `Node` won't be delivered, so it should be good as is :ok_hand:

Protryon: it goes beyond that -- the `FakePeer` construction is just a network layer with no logic on top. it can't receive a peer list or connect to anything else.

ljedrz (approved):
LGTM once the new test passes

PR #1080, opened 2021-08-24
Dedicated inbound thread

ljedrz:
This PR adjusts the node's resource handling in order to prioritize the processing of already-read network messages over reading new ones. In scenarios where the node has a lot of connections and they are sending **many** messages (especially big ones, like blocks), the async runtime can just run out of resources it can dedicate to anything else than reading from network sockets.

This PR introduces a thread dedicated to just processing of inbound messages in a blocking fashion; this means that the node's sockets are a lot more likely to become full (instead of the inbound message queue), making the connections that send too many messages likely to drop themselves, reducing the strain on the node and the network (as it limits re-broadcasting from nodes that are potentially "too eager" to send messages).

This PR also re-works the resource management for the miner, using short-lived blocking tasks only for operations that are actually blocking. Some new `clone()`s are introduced, but all the related objects were small or contained `Arc`s, so this should be fine.

The `#[cfg(feature = "test")]` attribute can be used in order to limit the number of changes necessary for tests - otherwise all of them would need to create a `tokio` runtime manually and pass it on to the `Node`, so that the dedicated thread can use `async` code.

Filing this PR as draft, as the earlier https://github.com/AleoHQ/snarkOS/pull/1074 already introduces a `test` feature for `snarkos-network`, and I only realized it's needed late in the implementation. I'll drop the last commit once that PR is in.

### DIFF
--- a/consensus/src/consensus/utility.rs
+++ b/consensus/src/consensus/utility.rs
@@ -16,6 +16,8 @@
 
 use snarkos_storage::PrivateKey;
 
+use tokio::task;
+
 use super::*;
 
 impl Consensus {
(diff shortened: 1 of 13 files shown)


consensus/src/consensus/utility.rs
```diff
@@ -24,97 +26,105 @@ impl Consensus {
...
         new_death_program_ids: Vec<Vec<u8>>,
         recipients: Vec<Address>,
     ) -> Result<TransactionResponse, ConsensusError> {
-        let mut rng = thread_rng();
-        let mut total_value_balance = crate::get_block_reward(block_num);
-
-        for transaction in transactions.iter() {
-            let tx_value_balance = transaction.value_balance;
+        let consensus = self.clone();
+        let (old_records, old_account_private_keys, new_records, memo) = task::spawn_blocking(move || {
```
Protryon, line 36: what's the rationale here? miner should already be in a owned blocking thread, and calling this from `block_on`.

consensus/src/miner.rs
```diff
@@ -191,7 +193,9 @@ impl MineContext {
             );
         }
 
-        let header = self.find_block(&transactions, &self.canon_header, terminator)?;
+        let ctx = self.clone();
```
Protryon, line 196: ditto, why spawn a blocking thread here?

network/src/sync/memory_pool.rs
```diff
@@ -67,13 +67,13 @@ impl Node {
         source: SocketAddr,
         transaction: Vec<u8>,
     ) -> Result<()> {
-        if let Ok(tx) = Testnet1Transaction::read_le(&*transaction) {
-            let inserted = self.expect_sync().consensus.receive_transaction(tx.serialize()?).await;
+        let tx = Testnet1Transaction::read_le(&*transaction)?;
```
Protryon, line 70: this is going to error out when people send us invalid transactions -- is that the intention?

network/src/sync/miner.rs
```diff
@@ -45,20 +43,24 @@ impl MinerInstance {
...
-            self.miner_address.clone(),
-            Arc::clone(&self.node.expect_sync().consensus),
-        )
-        .await?;
-        info!("Miner instantiated; starting to mine blocks");
-
-        let mut mining_failure_count = 0;
-        let mining_failure_threshold = 10;
-
-        Ok(task::spawn_blocking(move || {
```
Protryon, line 61: why do we want to reduce this thread into smaller threads?

Protryon (commented):
what's the rationale for miner changes?
also, what I had more in mind is removing the inbound queue, not running it in a blocking thread -- just handle the individual messages inbound in their own tokio task, spawning blocking for serialization as needed.
we no longer need to use the actor model for inbound I think, since peers, storage, and consensus do.

network/src/sync/memory_pool.rs
ljedrz: yes, and it's acocounted for in the inbound handler and the failure for the peer gets registered; we wouldn't want anyone to be able to spam the node with lots of invalid transactions that don't even parse properly

ljedrz:
missed an RPC test, will correct tomorrow.

network/src/sync/miner.rs
ljedrz: Answering the previous questions on more fine-grained calls to `spawn_blocking` here; we've discussed this once briefly during the community call: blocking tasks should generally not be long-lived, and are expected to be used for operations that finish on their own ([tokio docs](https://docs.rs/tokio/1.10.1/tokio/index.html#cpu-bound-tasks-and-blocking-code)); other than that, a lot of the code underneath `mine_block` is `async`-dense and having `block_on` in some of the underlying operations makes them likely candidates for future issues in tests/refactorings.

TL;DR: while the current implementation does work, it's a bit "hacky" in terms of `tokio`'s design and could cause issues in the future (either "self-inflicted" or when `tokio` decides to enforce some of their documented behavior in a stricter fashion).

consensus/src/consensus/utility.rs
ljedrz: the general answer is provided later on, but more specifically, this is exactly a block containing blocking operations and calls to `thread_rng()`

consensus/src/miner.rs
ljedrz: general answer provided later on, more specifically the same answer as https://github.com/AleoHQ/snarkOS/pull/1080#discussion_r695011085 applies

ljedrz:
@Protryon 
> what I had more in mind is removing the inbound queue, not running it in a blocking thread -- just handle the individual messages inbound in their own tokio task, spawning blocking for serialization as needed.

The issue is that it seems that it's the `recv()` operation that isn't handled quickly enough by the executor itself; we already tried spawning tasks in order to "exit" the function faster, but it didn't provide a result as dramatic as being able to receive the messages in the queue as fast as they arrive. This can't really be worked around with plain task spawning, hence the dedicated thread. Try the branch locally, and you'll see that with these changes inbound queue utilization is close to nothingn even with a **lot** of connections.

Protryon (approved):

niklaslong (approved):

ljedrz:
Merging, as #1079 requires a rebase; note: I have a branch where the inbound queue is removed entirely, but it seems that it requires further adjustment in order to work correctly, as the node is experiencing high outbound queue utilization and some connectivity issues; presumably, the outbound queue isn't processed quickly enough there; I'll do more tests.

PR #1084, opened 2021-08-25
Introduce the block processing time metric

niklaslong:
The PR measures the processing time of an inbound block. The value can likely be fine-tuned in future to differentiate between `SyncBlock` blocks and `Block` blocks but the difference should be negligeable for now.


niklaslong:
Rebased

network/src/sync/blocks.rs
```diff
@@ -75,6 +76,8 @@ impl Node {
             )));
         }
 
+        // Set to `true` if the block was sent in a `Block` messages, `false` if it was sent in a
+        // `SyncBlock` message.
```
ljedrz, line 80: that name isn't great; I think the `is_block_new` variable can have its name changed to `is_non_sync` in order to avoid any confusion

ljedrz:
is this still a draft? looks pretty done to me :slightly_smiling_face:

niklaslong:
Happy to undraft, I'd be cautious about merging because of my previous concerns:
> Leaving this drafted to ensure the impact of Instant::now can be evaluated. In my local tests it seemed to have no impact on the system, though I've also seen big spikes in the inbound queue correlating to slower RTT and block processing times.

ljedrz:
How about a benchmark comparing the "before" and "after" to see the relative impact?

PR #1085, opened 2021-08-25
Remove the inbound queue

ljedrz:
This PR completely removes the inbound queue, aggressively delegating almost all inbound requests to tasks. Initial tests have been very promising (after the _more_ initial ones that were troublesome due to not being aggressive enough with task spawning), but I'm leaving this as a draft while I'm running some more extreme scenarios.

Edit: no issues detected after a good amount of tests.

### DIFF
--- a/metrics/grafana/dashboard.json
+++ b/metrics/grafana/dashboard.json
@@ -21,7 +21,7 @@
   "editable": true,
   "gnetId": null,
   "graphTooltip": 0,
-  "id": 3,
+  "id": 6,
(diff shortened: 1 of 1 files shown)


ljedrz:
Confirmed to be very stable at `--min-peers 300` + syncing.

ljedrz:
No issues detected after running for a few hours; some very short outbound queue spikes, but other than that very good stability with `--min-peers 300` (which is a high number for a node actively syncing blocks).

PR #1093, opened 2021-08-26
Remove the fork length limit

ljedrz:
note: this limit might return in the future, at the current value or a revised one

Cc @damons

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2945,6 +2945,7 @@ dependencies = [
  "blake2",
  "derivative",
  "futures 0.3.17",
+ "hash_hasher",
  "hex",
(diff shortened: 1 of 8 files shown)


ljedrz:
Ok, so it appears that there doesn't need to be a fixed fork length limit. However, there is a great probability that the node will calculate super long fork paths, so I now added a simple cache for block children. I don't think it's a long-term solution, but it provides very good results after the first super-long fork path is calculated and could be used temporarily to facilitate the removal of existing forks.

Also, rebased.

Protryon (approved):

ljedrz:
Rebased.

PR #1099, opened 2021-08-27
Move the crawler to its own binary

niklaslong:
This PR creates a separate binary for crawler nodes (all binaries have been moved to the `bin` directory). This separation will make maintaining the different node types easier. It also removes the sync layer from crawler nodes and the `--is-crawler` command line flag. 

The `snarkos` binary is run by default, other binaries need to be specified via the `--bin` commad line option (e.g. `cargo run --release --bin crawler`. 

More generally, these changes pave the way for peer discovery nodes and sync provider nodes as discussed in #1101.


bin/crawler.rs
```diff
@@ -0,0 +1,114 @@
...
+fn main() -> Result<(), NodeError> {
+    let arguments = ConfigCli::args();
+
+    let mut config: Config = ConfigCli::parse(&arguments)?;
+    config.is_crawler(true);
+    config.check().map_err(|e| NodeError::Message(e.to_string()))?;
+
+    let runtime = runtime::Builder::new_multi_thread()
+        .enable_all()
+        .thread_stack_size(8 * 1024 * 1024)
```
ljedrz, line 108:
```suggestion
```
we don't need that much stack when running the crawler - this value is set that high only for mining purposes

bin/crawler.rs
```diff
@@ -0,0 +1,114 @@
...
+use std::{net::SocketAddr, time::Duration};
+
+use tokio::runtime;
+
+///
+/// Builds a node from configuration parameters.
+///
+/// 1. Creates network server.
+/// 2. Starts rpc server thread.
+/// 3. Starts network server listener.
```
ljedrz, line 38:
```suggestion
```

bin/crawler.rs
```diff
@@ -0,0 +1,114 @@
...
+    // This is done early on, so that the local address can be discovered
+    // before any other object (RPC) needs to use it.
+    let node = Node::new(node_config, None).await?;
+
+    // Initialize metrics framework.
+    node.initialize_metrics().await?;
+
+    // Start listening for incoming connections.
+    node.listen().await?;
+
```
ljedrz, line 72:
```suggestion
```
the crawler doesn't need to listen on its own, no? it should be the one calling the shots wrt making all the connections; this will also make the other nodes not need to special-case its address in any way

bin/crawler.rs
```diff
@@ -0,0 +1,114 @@
...
+    initialize_logger(&config);
+
+    print_welcome(&config);
+
+    let address = format!("{}:{}", config.node.ip, config.node.port);
+    let desired_address = address.parse::<SocketAddr>()?;
+
+    let mut path = config.node.dir.clone();
+    path.push(&config.node.db);
+
```
ljedrz, line 50:
```suggestion
```
if there's no storage, there's no need for this path

bin/crawler.rs
```diff
@@ -0,0 +1,114 @@
...
+    let node_config = NodeConfig::new(
+        desired_address,
+        config.p2p.min_peers,
+        config.p2p.max_peers,
+        config.p2p.bootnodes.clone(),
+        config.node.is_bootnode,
+        config.node.is_crawler,
```
ljedrz, line 57:
```suggestion
        false,
        true,
```

bin/crawler.rs
```diff
@@ -0,0 +1,114 @@
...
+
+    print_welcome(&config);
+
+    let address = format!("{}:{}", config.node.ip, config.node.port);
+    let desired_address = address.parse::<SocketAddr>()?;
+
+    let mut path = config.node.dir.clone();
+    path.push(&config.node.db);
+
+    let node_config = NodeConfig::new(
```
ljedrz, line 51: we should probably introduce some `Builder` pattern for `NodeConfig` for greater flexibility

bin/crawler.rs
```diff
@@ -0,0 +1,114 @@
...
+            config.rpc.username,
+            config.rpc.password,
+        );
+        node.register_task(rpc_handle);
+
+        info!("Listening for RPC requests on port {}", config.rpc.port);
+    }
+
+    // Start the network services.
+    node.start_services().await;
```
ljedrz, line 92: this method will require a tweak or two; for instance, the crawler doesn't need the `NodeState` task to be running

bin/snarkos.rs
```diff
@@ -55,34 +55,6 @@ use snarkvm_utilities::{to_bytes_le, FromBytes, ToBytes};
 use std::{fs, net::SocketAddr, str::FromStr, sync::Arc, time::Duration};
 
 use tokio::runtime;
-use tracing_subscriber::EnvFilter;
-
-fn initialize_logger(config: &Config) {
```
ljedrz, line 60: it would be great if we could move even more common, "`main`-specific" functions like this to some module containing functions usable by any node type

network/src/node.rs
```diff
@@ -336,7 +340,10 @@ impl Node {
         }
 
         // The node can already be at some non-zero height.
-        metrics::gauge!(misc::BLOCK_HEIGHT, self.storage.canon().await?.block_height as f64);
+        if let Some(storage) = &self.storage {
```
ljedrz, line 343: this is fine for now, but I'm guessing that peering-specific nodes that _will_ have storage for the storage of lists of peers will have to be special-cased here too; alternatively, this could already be changed to something like:
```
if self.sync.is_some() {
    <use expect_storage()>
}
```
as storage will always be a given if there's a sync layer, and chain-specific metrics primarily depend on the sync layer being available

network/src/peers/peers.rs
```diff
@@ -286,7 +286,11 @@ impl Node {
         trace!("Broadcasting `Ping` messages");
 
         // Consider peering tests that don't use the sync layer.
-        let current_block_height = self.storage.canon().await?.block_height as u32;
+        let current_block_height = if let Some(storage) = &self.storage {
```
ljedrz, line 289: ditto

ljedrz (commented):
left a few more suggestions, now that I've had a fresh look at it :sunny:

bin/crawler.rs
niklaslong: Indeed, this will require some tweaking but it can (and should) initiate all connections 🤔

bin/crawler.rs
niklaslong: These are all manditory fields at the moment (`NodeConfig` doesn't implement default as that's handled by `Config`) so the builder pattern isn't well suited to this imo?

bin/crawler.rs
niklaslong: I'd keep this here, since we're setting the node type on `Config` in `main` anyway—so long as configuration is coupled at least?

bin/crawler.rs
niklaslong: I've taken another look at this, it'll be easier to remove once we've reduced coupling in the network methinks. This also implies a protocol change, since the `Version` message contains the listening port of the node.

ljedrz: not necessarily - the crawler could still provide the port based on its desired address setting - it would just not be accessible; I think doing this is preferable to having it accept any connections

bin/crawler.rs
ljedrz: ok, let's leave this idea for later

bin/crawler.rs
niklaslong: Made it return the desired addr and it does indeed remove the need for unwrapping it everywhere 👍

ljedrz (approved):
LGTM :+1:

PR #1115, opened 2021-08-31
Non-blocking Merkle tree rebuilds

ljedrz:
The Merkle trees the node is using are most of the time rebuilt in an optimized fashion, reusing as many hashes as possible. However, whenever the number of leaves is a power of 2, the entire tree needs to be rebuilt. This PR spawns a dedicated blocking task whenever such conditions are detected, making it a non-blocking operation.

### DIFF
--- a/consensus/src/ledger/merkle.rs
+++ b/consensus/src/ledger/merkle.rs
@@ -132,4 +132,9 @@ impl<P: MerkleParameters> Ledger for MerkleLedger<P> {
         let calculated_digest = self.commitments.digest();
         self.digest() == calculated_digest
     }
+
+    fn requires_async_task(&self, new_commitments_len: usize, new_serial_numbers_len: usize) -> bool {
+        (self.commitments.len() + new_commitments_len).is_power_of_two()
+            || (self.serial_numbers.len() + new_serial_numbers_len).is_power_of_two()
+    }
 }
--- a/consensus/src/ledger/mod.rs
+++ b/consensus/src/ledger/mod.rs
@@ -37,6 +37,7 @@ use snarkvm_dpc::{
     TransactionScheme,
 };
 
+mod dummy;
 mod merkle;
 pub use merkle::MerkleLedger;
 mod indexed_merkle_tree;
@@ -77,9 +78,17 @@ pub trait Ledger: Send + Sync {
 
     /// checks if a ledgers state is consistent
     fn validate_ledger(&self) -> bool;
+
+    fn requires_async_task(&self, new_commitments_len: usize, new_serial_numbers_len: usize) -> bool;
 }
 
 pub struct DynLedger(pub Box<dyn Ledger>);
+
+impl Default for DynLedger {
+    fn default() -> Self {
+        DynLedger(Box::new(dummy::DummyLedger))
+    }
+}
 
 impl Deref for DynLedger {
     type Target = dyn Ledger;
--- a/consensus/src/ledger/mod.rs
+++ b/consensus/src/ledger/mod.rs
@@ -37,7 +37,7 @@ use snarkvm_dpc::{
     TransactionScheme,
 };
 
-mod dummy;
+pub(crate) mod dummy;
 mod merkle;
 pub use merkle::MerkleLedger;
 mod indexed_merkle_tree;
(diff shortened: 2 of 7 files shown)


Protryon:
I feel we are being too granular with these blocking tasks -- but okay.

consensus/src/ledger/merkle.rs
```diff
@@ -21,13 +21,14 @@ use anyhow::*;
...
 use snarkvm_algorithms::MerkleParameters;
+use tokio::task;
 
 use super::indexed_merkle_tree::IndexedMerkleTree;
 
 pub struct MerkleLedger<P: MerkleParameters> {
     ledger_digests: IndexSet<Digest>,
-    commitments: IndexedMerkleTree<P>,
-    serial_numbers: IndexedMerkleTree<P>,
+    commitments: Option<IndexedMerkleTree<P>>,
```
Protryon, line 30: can we decompose and recompose the merkle ledger rather than taking ownership like this? it's messy.

ljedrz: I wanted to do that, but we can't take ownership of `Ledger`, as it would make it non-object-safe, and remove the possibility of using it as a trait object.

ljedrz: and I think that we can't do this with `MerkleLedger`, as we need the `async` access at a higher layer

consensus/src/ledger/mod.rs
```diff
@@ -77,6 +78,8 @@ pub trait Ledger: Send + Sync {
 
     /// checks if a ledgers state is consistent
     fn validate_ledger(&self) -> bool;
+
+    fn requires_full_rebuilds(&self, new_commitments_len: usize, new_serial_numbers_len: usize) -> bool;
```
Protryon, line 82: can we rename this to something like `require_async_task` to be less context-specific.

consensus/src/ledger/merkle.rs
ljedrz: I will double-check, though :+1:

Protryon: rather than making `extend` async, just keep it sync, have the caller (or a default impl in the trait definition) manage the async end and calling the check for should spawn a blocking task. keep this struct pure to its function -- and we can pass the merkle ledger entirely or by reference without issue.

Protryon: there is also some tricks with swapping a `&mut self` with a default impl and back again if you really need to.

ljedrz: I considered `mem::take`, but the default impl wasn't as trivial as I like; however, it might be negligible; I'll try to make this change simpler

ljedrz:
I think that I just need to make the failing tests `#[tokio::test(flavor = "multi_thread")]` now; checking it locally.

niklaslong (approved):
LGTM 👍

ljedrz:
> I feel we are being too granular with these blocking tasks -- but okay.

@Protryon I don't know if we can do better, as large-scope blocking tasks, e.g. over an entire `Block` import, would still involve a lot of `async` code, which would get delegated to non-blocking tasks and those would still become stuck once a blocking operation is encountered.

Not a huge fan either, but in this case the blocking task is spawned only for each power of 2 of the number of either commitments or serial numbers, meaning this won't happen too often.

consensus/src/ledger/mod.rs
```diff
@@ -77,10 +78,18 @@ pub trait Ledger: Send + Sync {
...
     fn validate_ledger(&self) -> bool;
+
+    fn requires_async_task(&self, new_commitments_len: usize, new_serial_numbers_len: usize) -> bool;
 }
 
 pub struct DynLedger(pub Box<dyn Ledger>);
 
+impl Default for DynLedger {
+    fn default() -> Self {
+        DynLedger(Box::new(dummy::DummyLedger))
```
Protryon, line 89: why do we want a panicky default here?

Protryon (approved):

consensus/src/ledger/mod.rs
ljedrz: that's just so that `mem::take` can work; but fair, I can just use `mem::replace` to avoid accidental use of `DummyLedger` :+1:

ljedrz:
Cc https://github.com/AleoHQ/snarkVM/pull/138 for the measurements of Merkle tree rebuild spikes.

PR #1116, opened 2021-08-31
Non-blocking transaction verification

ljedrz:
Just like Merkle tree rebuilds in https://github.com/AleoHQ/snarkOS/pull/1115, transaction verification was found to be blocking at times, often taking more than 100ms. While spawning a blocking task is not free, it is significantly cheaper than blocking an async worker thread.

An example measurement:
```
10693ms taken by verify_transactions for 263 txs
```

This will be improved by https://github.com/AleoHQ/snarkVM/pull/371, but it's still desirable on its own.

### DIFF
--- a/consensus/src/consensus/inner/commit.rs
+++ b/consensus/src/consensus/inner/commit.rs
@@ -225,7 +225,7 @@ impl ConsensusInner {
         }
 
         // Check that all the transaction proofs verify
-        self.verify_transactions(block.transactions.iter())
+        self.verify_transactions(block.transactions.clone()).await
     }
 
     pub(super) async fn commit_block(&mut self, hash: &Digest, block: &SerialBlock) -> Result<(), ConsensusError> {
@@ -239,7 +239,7 @@ impl ConsensusInner {
         }
 
         let digest = if self.ledger.requires_async_task(commitments.len(), serial_numbers.len()) {
-            let mut ledger = std::mem::replace(&mut self.ledger, DynLedger(Box::new(DummyLedger)));
+            let mut ledger = std::mem::replace(&mut self.ledger, DynLedger(Arc::new(DummyLedger)));
             let (digest, ledger) = tokio::task::spawn_blocking(move || {
                 let digest = ledger.extend(&commitments[..], &serial_numbers[..], &memos[..]);
 
--- a/consensus/src/consensus/inner/transaction.rs
+++ b/consensus/src/consensus/inner/transaction.rs
@@ -16,19 +16,20 @@
 
 use snarkos_storage::VMRecord;
 use snarkvm_dpc::testnet1::instantiated::{Testnet1DPC, Testnet1Transaction};
+use tokio::task;
 
 use crate::DeserializedLedger;
 
 use super::*;
 
 impl ConsensusInner {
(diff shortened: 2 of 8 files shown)


ljedrz:
Making it a draft, as I need to adjust the tests.

consensus/src/consensus/inner/commit.rs
```diff
@@ -224,7 +224,7 @@ impl ConsensusInner {
         }
 
         // Check that all the transaction proofs verify
-        self.verify_transactions(block.transactions.iter())
+        self.verify_transactions(block.transactions.clone()).await
```
niklaslong, line 227: Does this have to be `clone`d?

consensus/src/consensus/inner/transaction.rs
```diff
@@ -71,13 +72,19 @@ impl ConsensusInner {
...
-            deserialized.push(Testnet1Transaction::deserialize(transaction)?);
+            deserialized.push(Testnet1Transaction::deserialize(&transaction)?);
         }
 
-        Ok(self
-            .public
-            .dpc
-            .verify_transactions(&deserialized[..], &self.ledger.deserialize::<Components>()))
+        let consensus = self.public.clone();
+        let ledger = self.ledger.clone();
```
niklaslong, line 79: iirc these are both `Arc`ed? 
(edit: yes they are, spotted below 👍)

consensus/src/consensus/inner/commit.rs
ljedrz: `SerialBlock` is only borrowed in this context, so sadly - yes; however, the cost is nothing compared to blocking happening otherwise

ljedrz:
Update: this PR now also removes all remaining uses of synchronous RPCs in tests, which makes them pass as expected.

consensus/src/consensus/inner/commit.rs
```diff
@@ -239,7 +239,7 @@ impl ConsensusInner {
         }
 
         let digest = if self.ledger.requires_async_task(commitments.len(), serial_numbers.len()) {
-            let mut ledger = std::mem::replace(&mut self.ledger, DynLedger(Box::new(DummyLedger)));
+            let mut ledger = std::mem::replace(&mut self.ledger, DynLedger(Arc::new(DummyLedger)));
```
Protryon, line 242: i don't like changing this from a `Box` to an `Arc`. can we maybe use crossbeam scoped tasks?

ljedrz: perhaps'; I'll look into it :+1:

ljedrz: this might be impossible: https://users.rust-lang.org/t/async-spawn-blocking-but-scoped/38561/3; the alternatives I can see right now are:

- a `Mutex` over/inside `DynLedger`
- a hand-written `Clone` impl for `DynLedger` that checks that there is at most a single strong reference at the moment of cloning or it panics

ljedrz:
Split out the RPC test changes (desirable on their own) and rebased .

Protryon (approved):

ljedrz:
Superseded by https://github.com/AleoHQ/snarkOS/pull/1127.

PR #1120, opened 2021-09-02
Aggro Sync

Protryon:
This PR massively increases network-bound sync performance and significantly reduces any chances of forks not being considered.

The old sync process:
1. Request sync hashes from peers based on block locator hashes
2. Randomly request each novel sync block from a random peer who has it
3. Repeat

The new sync process:
1. Request sync hashes from peers based on block locator hashes
2. For each peer that responds, broadcast back their highest sync hash, request all blocks we don't have at that instant
3. Repeat step # 2 until no one responds for a while

This has the following effects:
* Continuous sync process eliminates cyclical performance observed/expected in old sync process.
* Increase in the number of received duplicate blocks (from effectively zero to scaling linearly by number of peers)
* An effective, naive negotiation for any peers on really distant forks to still be able to send blocks.
* Pulling forks that are ultimately suboptimal, in their entirety.

My current node synced to canon tip from origin in ~18 hours of runtime -- unclean test. Has ~40K forked blocks from the process (not forks, blocks not in canon).

### DIFF
--- a/network/src/inbound/cache.rs
+++ b/network/src/inbound/cache.rs
@@ -14,17 +14,20 @@
 // You should have received a copy of the GNU General Public License
 // along with the snarkOS library. If not, see <https://www.gnu.org/licenses/>.
 
-use snarkvm_dpc::block::BlockHeader;
-
 use circular_queue::CircularQueue;
+use snarkvm_dpc::BlockHeader;
 use twox_hash::xxh3::hash64;
 
 #[derive(Debug, Clone)]
-pub struct Cache<const N: usize> {
+pub struct Cache<const N: usize, const S: usize> {
     queue: CircularQueue<u64>,
 }
 
-impl<const N: usize> Default for Cache<N> {
+const BLOCK_HEADER_SIZE: usize = BlockHeader::size();
+
+pub type BlockCache<const N: usize> = Cache<N, BLOCK_HEADER_SIZE>;
+
+impl<const N: usize, const S: usize> Default for Cache<N, S> {
     fn default() -> Self {
         Self {
             queue: CircularQueue::with_capacity(N),
@@ -32,9 +35,9 @@ impl<const N: usize> Default for Cache<N> {
     }
 }
 
-impl<const N: usize> Cache<N> {
+impl<const N: usize, const S: usize> Cache<N, S> {
     fn hash_block(payload: &[u8]) -> u64 {
-        hash64(&payload[..BlockHeader::size()])
+        hash64(&payload[..S.min(payload.len())])
     }
 
     pub fn contains(&self, payload: &[u8]) -> bool {
--- a/network/src/peers/peer/outbound_handler.rs
+++ b/network/src/peers/peer/outbound_handler.rs
@@ -213,8 +213,8 @@ impl Peer {
                 Ok(PeerResponse::None)
             }
             PeerAction::ExpectingSyncBlocks(amount) => {
-                self.quality.remaining_sync_blocks = amount;
-                self.quality.total_sync_blocks = amount;
+                self.quality.remaining_sync_blocks += amount;
+                self.quality.total_sync_blocks += amount;
                 Ok(PeerResponse::None)
             }
             PeerAction::SoftFail => {
--- a/network/src/peers/peer/outbound_handler.rs
+++ b/network/src/peers/peer/outbound_handler.rs
@@ -208,7 +208,7 @@ impl Peer {
                 if self.quality.remaining_sync_blocks > 0 {
                     self.quality.remaining_sync_blocks -= 1;
                 } else {
-                    warn!("received unexpected or late sync block from {}", self.address);
+                    trace!("received unexpected or late sync block from {}", self.address);
                 }
                 Ok(PeerResponse::None)
             }
--- a/network/src/sync/blocks/aggro.rs
+++ b/network/src/sync/blocks/aggro.rs
@@ -36,7 +36,7 @@ pub struct SyncAggro {
 }
 
 struct BlockLocatorHashes {
-    hashes: IndexSet<BlockHeaderHash>,
+    hashes: Vec<BlockHeaderHash>,
     last_update: Instant,
 }
 
@@ -76,7 +76,7 @@ impl SyncAggro {
         self.base.node.register_block_sync_attempt();
 
         let block_locator_hashes = Arc::new(RwLock::new(BlockLocatorHashes {
-            hashes: SyncBase::block_locator_hashes(&self.base.node).await?.into_iter().collect(),
+            hashes: SyncBase::block_locator_hashes(&self.base.node).await?,
             last_update: Instant::now(),
         }));
 
@@ -86,7 +86,7 @@ impl SyncAggro {
             for peer in &sync_nodes {
                 let mut initial_hashes = self.peer_info.entry(peer.address).or_default().pending_block_hashes.clone();
                 if initial_hashes.is_empty() {
-                    initial_hashes = block_locator_hashes.hashes.iter().cloned().collect();
+                    initial_hashes = block_locator_hashes.hashes.clone();
                 }
                 peer_syncs.push((peer.clone(), initial_hashes));
             }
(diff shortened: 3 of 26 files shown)


ljedrz:
I've started testing this branch and I can immediately say we should probably downgrade some of the exsiting `WARN` logs to `TRACE` now, especially the following 2:
```
WARN received unexpected or late sync block from X
WARN failed receiving block: PreExistingBlock
```

network/src/inbound/cache.rs
```diff
@@ -14,27 +14,30 @@
...
             queue: CircularQueue::with_capacity(N),
         }
     }
 }
 
-impl<const N: usize> Cache<N> {
+impl<const N: usize, const S: usize> Cache<N, S> {
     fn hash_block(payload: &[u8]) -> u64 {
-        hash64(&payload[..BlockHeader::size()])
+        hash64(&payload[..S.min(payload.len())])
```
ljedrz, line 40: is it possible for a `Block` to be smaller than the header :thinking:?

network/src/sync/blocks/batched.rs
```diff
@@ -42,86 +35,20 @@ struct SyncBlock {
...
-            .await?
-            .into_iter()
-            .map(|x| BlockHeaderHash(x.bytes::<32>().unwrap()))
-            .collect::<Vec<_>>();
+        let block_locator_hashes = SyncBase::block_locator_hashes(&self.base.node).await?;
         let mut future_set = vec![];
 
         for peer in sync_nodes.iter() {
-            if let Some(handle) = self.node.peer_book.get_peer_handle(peer.address) {
+            if let Some(handle) = self.base.node.peer_book.get_peer_handle(peer.address) {
```
ljedrz, line 51: this is not strictly related to this PR, but: since we only need the peers' addresses, it is probably preferable to put only those in the `sync_nodes` vector, to reduce the amount of memory and atomic operations

network/src/sync/blocks/aggro.rs
```diff
@@ -0,0 +1,201 @@
...
+                .collect()
+        };
+
+        let hash_requests_sent = self.send_sync_messages(peer_syncs).await?;
+
+        if hash_requests_sent == 0 {
+            return Ok(());
+        }
+
+        let received_hashes = Arc::new(RwLock::new(Cache::<1024, 32>::default()));
```
ljedrz, line 89:
```suggestion
        let received_hashes = Arc::new(RwLock::new(Cache::<1024, 32>::default()));
```
it would be best to refer to `BLOCK_HEADER_SIZE` instead of `32` here

network/src/sync/blocks/aggro.rs
```diff
@@ -0,0 +1,201 @@
...
+
+                            let early_block_states = match node.storage.get_block_states(&hashes_trimmed[..]).await {
+                                Ok(x) => x,
+                                Err(e) => {
+                                    warn!("failed to get block states: {:?}", e);
+                                    return;
+                                }
+                            };
+
+                            let blocks: IndexSet<_> = {
```
ljedrz, line 137: do we need an `IndexSet` (as opposed to a plain `HashSet`) here?

network/src/sync/blocks/base.rs
```diff
@@ -0,0 +1,147 @@
...
+    pub async fn find_sync_nodes(&self) -> Result<Vec<Peer>> {
+        let our_block_height = self.node.storage.canon().await?.block_height;
+        let mut interesting_peers = vec![];
+        for mut node in self.node.peer_book.connected_peers_snapshot().await {
+            let judge_bad = node.judge_bad();
+            if !judge_bad && node.quality.block_height as usize > our_block_height + 1 {
+                interesting_peers.push(node);
+            }
+        }
+        interesting_peers.sort_by(|x, y| y.quality.block_height.cmp(&x.quality.block_height));
```
ljedrz, line 51:
```suggestion
        interesting_peers.sort_by_key(|x| std::cmp::Reverse(x.quality.block_height));
```
`clippy` will suggest this

network/src/sync/blocks/batched.rs
```diff
@@ -133,55 +60,25 @@ impl SyncMaster {
...
+                    SyncInbound::Block(..) => {
+                        warn!("received sync block prematurely");
+                    }
                 }
-            }
-            //todo: fail if peer sends > 1 block hash packet
-            received_block_hashes.len() >= max_message_count
-        })
-        .await;
+                //todo: fail if peer sends > 1 block hash packet
```
ljedrz, line 78: is this `todo` still applicable?

network/src/sync/blocks/aggro.rs
```diff
@@ -0,0 +1,201 @@
...
+use anyhow::*;
+use indexmap::IndexSet;
+use snarkos_storage::Digest;
+use snarkvm_dpc::BlockHeaderHash;
+use tokio::{
+    sync::{mpsc, RwLock},
+    time::Instant,
+};
+
+pub struct SyncAggro {
```
ljedrz, line 29: having seen all the `SyncX` objects I know what their names mean, but it would be nice to have short doc comments on them for future readers

ljedrz (commented):
Left some questions and suggestions; I'll run some more tests (the initial runs were very promising) in the meantime.

network/src/inbound/cache.rs
Protryon: no, but we use digests now -- and truncation is preferable to panicing in general

network/src/sync/blocks/batched.rs
Protryon: yeah, not worth doing since we may need more in the future, and the costs here are trivial

network/src/sync/blocks/aggro.rs
Protryon: no -- but i appreciate the added determinism as a hedge against non-determinism bugs -- and as usual, not perf critical

network/src/sync/blocks/batched.rs
Protryon: it was always just an idea, and remains a valid idea

network/src/sync/blocks/aggro.rs
Protryon: it's block hash size, not header size. don't think we have a constant for that atm, just default digest size.

network/src/sync/blocks/base.rs
Protryon: `clippy` didn't suggest, avoiding for now.

network/src/sync/blocks/aggro.rs
```diff
@@ -0,0 +1,202 @@
...
+                        });
+                    }
+                }
+                false
+            })
+            .await;
+
+        let sync_addresses = sync_nodes.iter().map(|x| x.address).collect::<Vec<_>>();
+
+        self.base.cancel_outstanding_syncs(&sync_addresses[..]).await;
```
niklaslong, line 192: It wasn't immediately obvious to me why this needs to be called at the end of every sync, it would be nice to comment on this here? (It may also not be necessary if state is tracked on the sync layer.)

niklaslong (approved):
This is working really nicely for me locally, left a few comments 🚀

network/src/peers/peer/outbound_handler.rs
```diff
@@ -208,13 +208,13 @@ impl Peer {
...
-                    warn!("received unexpected or late sync block from {}", self.address);
+                    trace!("received unexpected or late sync block from {}", self.address);
                 }
                 Ok(PeerResponse::None)
             }
             PeerAction::ExpectingSyncBlocks(amount) => {
-                self.quality.remaining_sync_blocks = amount;
-                self.quality.total_sync_blocks = amount;
+                self.quality.remaining_sync_blocks += amount;
+                self.quality.total_sync_blocks += amount;
```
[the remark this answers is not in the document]
Protryon: yes, but in a future pr

network/src/sync/blocks/aggro.rs
Protryon: it notifies each peer to reset tracking state. Agreed.

PR #1121, opened 2021-09-02
Feature/metrics catalogue

sadroeck:
## Motivation

This PR introduces a dependency on `metrics-catalogue`, which serves a few purposes:

1. It procedurally generates a `metrics-core` compatible set of trait implementations based on a simple hierarchical struct definition of metrics. This allows the `Stats` struct to be single source of truth, regardless whether they were updated using directly on the struct or via the `metrics` macros
2. It procedurally generates a catalogue of macros, preventing typos & replacing the need for a manually maintained version, such as the one in `metrics::names`.
3. It introduces a Histogram container with fixed windowing, preventing the `drain` behavior as seen in `prometheus-metrics-exporter`. This allows multiple prometheus clients to scrape simultaneously without missing samples.

### Note:

All `counters` currently have a `_total` postfix. For simplicity's sake, I've removed those. Unless there's a significant reason to keep these?

## Test Plan

The changes can/should be tested in multiple ways:

1. HTTP GET request on the metrics endpoint, typically `localhost:9000`
2. Run a local prometheus + grafana & inspect the dashboard. Everything should reported as before

## Related PRs

- https://github.com/AleoHQ/snarkOS/pull/879


ljedrz:
Closing the PR as it's either no longer applicable to testnet2 or would require a rewrite (in which case, feel free to reopen once rebased).

PR #1122, opened 2021-09-02
Upgrade the miner termination log from TRACE to DEBUG

ljedrz:
This doesn't happen too often, so the log can be upgraded for greater visibility.

### DIFF
--- a/consensus/src/miner.rs
+++ b/consensus/src/miner.rs
@@ -144,7 +144,7 @@ impl MineContext {
         let (nonce, proof) = match mined {
             Err(PoswError::SnarkError(SNARKError::Terminated)) => {
                 // technically a race condition, but non-critical
-                trace!("terminated miner due to canon block received");
+                debug!("terminated miner due to canon block received");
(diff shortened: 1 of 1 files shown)


consensus/src/miner.rs
```diff
@@ -144,15 +144,15 @@ impl MineContext {
         let (nonce, proof) = match mined {
             Err(PoswError::SnarkError(SNARKError::Terminated)) => {
                 // technically a race condition, but non-critical
-                trace!("terminated miner due to canon block received");
+                debug!("terminated miner due to canon block received");
```
niklaslong, line 147: Perhaps? 
```suggestion
                debug!("terminated miner due to competing canon block received");
```

niklaslong (approved):

consensus/src/miner.rs
ljedrz: once the new block is recognized as canon, I don't know if it can have any competition per se - an alternative fork might compete with the current canon one, but I don't think it applies to a block of the same height

PR #1124, opened 2021-09-03
Move some MemoryPool methods over to ConsensusInner

ljedrz:
This one was detected as an annoyance when tinkering around https://github.com/AleoHQ/snarkOS/pull/1116; 2 of the `MemoryPool` methods also require access to `Ledger`, so move them to the parent object of both, so that their APIs (and related borrow implications) can be simplified.

The exact behavior of both methods was preserved, notably with `cleanse_memory_pool` still reverting to the original mempool in case of any errors; however, I wonder if instead the errors shouldn't just be skipped or cause the mempool to be fully purged.


niklaslong (approved):

ljedrz:
I added a commit removing the trivial `MemoryPool` tests too.

Protryon (approved):

PR #1126, opened 2021-09-03
Report potential canon tip via the `getnetworkgraph` rpc

niklaslong:
The crawler currently only reports potential fork clusters which don't include the canon chain candidates. This PR adds the crawler's best guess at the canon tip (3+ members and heighest height in the graph) to the values exposed via rpc.

### DIFF
--- a/network/src/topology.rs
+++ b/network/src/topology.rs
@@ -25,6 +25,7 @@ use std::{
 use chrono::{DateTime, Utc};
 use nalgebra::{DMatrix, DVector, SymmetricEigen};
 use parking_lot::{Mutex, RwLock};
+use serde::{Deserialize, Serialize};
 use tokio::sync::mpsc::{self, Receiver, Sender};
 
 // Purges connections that haven't been seen within this time (in hours).
@@ -92,6 +93,13 @@ pub fn nodes_from_connections(connections: &HashSet<Connection>) -> HashSet<Sock
     }
 
     nodes
+}
+
+/// A node cluster (potential forks and tip) in the network, maps height to members.
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+pub struct NodeCluster {
+    pub height: u32,
+    pub members: Vec<SocketAddr>,
 }
 
 /// Message types passed through the `KnownNetwork` channel.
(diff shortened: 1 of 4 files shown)


rpc/documentation/public_endpoints/getnetworkgraph.md
```diff
@@ -13,6 +13,8 @@ None
 | `density`                            | f64        | The density of the known network                                                          |
 | `algebraic_connectivity`             | f64        | The algebraic connectivity, aka the fiedler eigenvalue of the known network               |
 | `degree_centrality_delta`            | f64        | The difference between the largest and the smallest peer count in the network             |
+| `potential_tip`                      | Option     | The crawler's best guess at the canon tip                                                 |
```
ljedrz, line 16:
```suggestion
| `potential_tip`                      | Option     | The crawler's best guess at the canon block height                                        |
```

rpc/src/rpc_impl.rs
```diff
@@ -493,8 +493,10 @@ impl RpcFunctions for RpcImpl {
             })
             .collect();
 
-        let potential_forks = known_network
-            .potential_forks()
+        let (potential_tip, potential_forks) = known_network.potential_forks();
+        let potential_tip = potential_tip.map(|(height, members)| PotentialFork { height, members });
```
ljedrz, line 497: maybe `PotentialFork` could have its name changed to `NodeCluster`, now that it applies to both forks and the possible canon cluster?

network/src/topology.rs
```diff
@@ -218,8 +218,9 @@ impl KnownNetwork {
         self.nodes.read().clone()
     }
 
-    /// Returns a map of the potential forks.
-    pub fn potential_forks(&self) -> HashMap<u32, Vec<SocketAddr>> {
+    #[allow(clippy::type_complexity)]
+    /// Returns the canon tip height and members, and a map of the potential forks.
+    pub fn potential_forks(&self) -> (Option<(u32, Vec<SocketAddr>)>, HashMap<u32, Vec<SocketAddr>>) {
```
ljedrz, line 223: maybe we can reuse the `PotentialFork` (or `NodeCluster`, as suggested later on) here for a simpler return type?

ljedrz (approved):
LGTM with some ergonomics suggestions

network/src/topology.rs
niklaslong: I wanted to avoid a dependency on the RPC types but I can certainly wrap these in a new type?

ljedrz: maybe the RPC type could instead reside here?

rpc/documentation/public_endpoints/getnetworkgraph.md
niklaslong: This doesn't only contain the block height but also the member nodes, I'll make this explicit though 👍

network/src/topology.rs
niklaslong: Fine by me 👍

PR #1127, opened 2021-09-03
Spawn a blocking task for transaction verification

ljedrz:
Supersedes https://github.com/AleoHQ/snarkOS/pull/1116; copying its description over:

Just like Merkle tree rebuilds in #1115, transaction verification was found to be blocking at times, often taking more than 100ms. While spawning a blocking task is not free, it is significantly cheaper than blocking an async worker thread.

An example measurement:

```
10693ms taken by verify_transactions for 263 txs
```

This will be improved by AleoHQ/snarkVM#371, but it's still desirable on its own.

### DIFF
--- a/consensus/src/consensus/inner/agent.rs
+++ b/consensus/src/consensus/inner/agent.rs
@@ -70,10 +70,11 @@ impl ConsensusInner {
 
             match message {
                 ConsensusMessage::ReceiveTransaction(transaction) => {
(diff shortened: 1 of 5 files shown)


ljedrz:
Seems that one RPC test might now need `flavor = multi_thread`; I'll push it as soon as other tests are complete.

niklaslong (approved):

PR #1130, opened 2021-09-06
Use a persistent rayon threadpool for blocking operations

ljedrz:
This PR introduces a persistent `rayon` threadpool that handles all the blocking operations related to inbound `Block`s. This change was confirmed to result in a far more aggressive utilization of CPU cores, resulting in better performance.

Supersedes https://github.com/AleoHQ/snarkOS/pull/1119.


ljedrz:
Needs further analysis; CPU utilization is much better, but responsiveness is lower when there are a lot of connections.

PR #1131, opened 2021-09-07
Add a block processing benchmark

ljedrz:
This PR introduces a benchmark for the import of full blocks. In order to have all the means necessary for the measurements to be accurate, test-only `reset` method was added to both `Ledger` and `Storage` traits, and implemented for `SqliteStorage`.

In addition, there are a few drive-by tweaks to test nodes.


niklaslong (approved):
LGTM 👍

PR #1132, opened 2021-09-07
Sqlite query implementation for longest child path

Protryon:
Improves performance significantly for operations calling longest child path -- not benchmarked.

### DIFF
--- a/storage/src/storage/sqlite/sync.rs
+++ b/storage/src/storage/sqlite/sync.rs
@@ -484,6 +484,55 @@ impl SyncStorage for SqliteStorage {
         })
     }
 
+    fn longest_child_path(&mut self, block_hash: &Digest) -> Result<Vec<Digest>> {
+        /*
+        alternate total tip (slower due to no temp index)
+        total_tip(parent, remaining) AS (
+            SELECT preferred_tip.sub, preferred_tip.length - 1 FROM preferred_tip
+            UNION ALL
+            SELECT children.sub, total_tip.remaining - 1
+            FROM total_tip
+            INNER JOIN children ON children.parent = total_tip.parent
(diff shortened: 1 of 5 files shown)


storage/src/storage/sqlite/sync.rs
```diff
@@ -490,17 +484,53 @@ impl SyncStorage for SqliteStorage {
...
+        /*
+        alternate total tip (slower due to no temp index)
+        total_tip(parent, remaining) AS (
+            SELECT preferred_tip.sub, preferred_tip.length - 1 FROM preferred_tip
+            UNION ALL
+            SELECT children.sub, total_tip.remaining - 1
+            FROM total_tip
+            INNER JOIN children ON children.parent = total_tip.parent
+            WHERE total_tip.remaining > 0
+        )
```
ljedrz, line 497: is this still being considered to be included or can it be removed?

ljedrz (approved):
LGTM :+1:

note: it would be great if we could have some test case for the longest child path; we do have some fork tests, but this could be tested individually as well

niklaslong (approved):

ljedrz (changes requested):
Unfortunately, upon more local tests I ran into the following error:
```
ERROR failed sync process: Invalid column type Null at index: 0, name: hash
```

which breaks the sync process; it needs to be addressed first

niklaslong:
Confirm, seeing the same here—missed it at first.

niklaslong (changes requested):
Confirm, seeing this error locally too 🔍

niklaslong (approved):
Seems resolved 👍

ljedrz (approved):
Going to leave it running for a while more, but I don't seem to be hitting the issue anymore.

PR #1133, opened 2021-09-07
Iterative block tree implementation

Protryon:
This PR removes the recursive implementation of block tree fetching from sqlite in favor of an iterative approach -- not implemented as a SQL query for now.

### DIFF
--- a/storage/src/storage/sync.rs
+++ b/storage/src/storage/sync.rs
@@ -14,7 +14,7 @@
 // You should have received a copy of the GNU General Public License
 // along with the snarkOS library. If not, see <https://www.gnu.org/licenses/>.
 
-use std::collections::VecDeque;
+use std::collections::{HashMap, HashSet, VecDeque};
 
 use anyhow::*;
 use tracing::{debug, trace};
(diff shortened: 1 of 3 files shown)


storage/src/storage/sync.rs
```diff
@@ -156,21 +156,68 @@ pub trait SyncStorage {
 
     /// Gets a tree structure representing all the descendents of [`block_hash`]
     fn get_block_digest_tree(&mut self, block_hash: &Digest) -> Result<DigestTree> {
-        let children = self.get_block_children(block_hash)?;
-        if children.is_empty() {
+        let mut nodes: HashMap<Digest, Vec<Digest>> = HashMap::new();
+        let mut stack = vec![block_hash.clone()];
+        while let Some(hash) = stack.pop() {
+            let children = self.get_block_children(&hash)?;
+            stack.extend_from_slice(&children[..]);
```
ljedrz, line 163: nit: this could be put in the conditional block right below as well

storage/src/storage/sync.rs
```diff
@@ -156,21 +156,68 @@ pub trait SyncStorage {
...
-            let subtree = self.get_block_digest_tree(&child)?;
-            let len = subtree.longest_length() + 1;
-            if len > longest_tree_len {
-                longest_tree_len = len;
+        let mut node_entries: HashSet<Digest> = nodes.keys().cloned().collect();
+
+        let mut trees: HashMap<Digest, DigestTree> = HashMap::new();
+        loop {
+            if nodes.is_empty() {
+                break;
```
ljedrz, line 176: nit: `while !nodes.is_empty() {`

storage/src/storage/sync.rs
```diff
@@ -156,21 +156,68 @@ pub trait SyncStorage {
...
+                trees.insert(
+                    hash.clone(),
+                    DigestTree::Node(hash.clone(), new_children, longest_tree_len),
+                );
+                node_entries.remove(hash);
+                false
+            });
         }
-        Ok(DigestTree::Node(block_hash.clone(), out_children, longest_tree_len))
+        assert_eq!(nodes.len(), 0);
```
ljedrz, line 203:
```suggestion
        assert_eq!(nodes.len(), 0);
```
already ensured by the breaking condition

storage/src/storage/sync.rs
```diff
@@ -156,21 +156,68 @@ pub trait SyncStorage {
...
+        assert_eq!(trees.len(), 1);
```
ljedrz, line 204:
```suggestion
        debug_assert_eq!(trees.len(), 1);
```
unless it's severe enough to warrant a runtime panic?

storage/src/storage/sync.rs
```diff
@@ -156,21 +156,68 @@ pub trait SyncStorage {
...
+        // let mut longest_tree_len = 0usize;
+        // for child in children {
+        //     let subtree = self.get_block_digest_tree(&child)?;
+        //     let len = subtree.longest_length() + 1;
+        //     if len > longest_tree_len {
+        //         longest_tree_len = len;
+        //     }
+        //     out_children.push(subtree);
+        // }
+        // Ok(DigestTree::Node(block_hash.clone(), out_children, longest_tree_len))
```
ljedrz, line 220:
```suggestion
```

storage/src/storage/sync.rs
```diff
@@ -156,21 +156,68 @@ pub trait SyncStorage {
...
+                );
+                node_entries.remove(hash);
+                false
+            });
         }
-        Ok(DigestTree::Node(block_hash.clone(), out_children, longest_tree_len))
+        assert_eq!(nodes.len(), 0);
+        assert_eq!(trees.len(), 1);
+
+        Ok(trees.remove(block_hash).expect("missing block hash tree"))
```
ljedrz, line 206: again a question on panicking here; perhaps we could throw an error instead?

ljedrz (approved):
Left some comments, LGTM otherwise; going to test it a bit more locally, haven't run into issues yet

storage/src/storage/sync.rs
Protryon: it is -- it implies a complete failure of the algorithm to manage a tree, which is an unrecoverable failure.

storage/src/storage/sync.rs
Protryon: again, critical algorithmic failure -- not something that should be recovered from.

Issue #1101, opened 2021-08-27
[Feature] Peer discovery and sync provider nodes


niklaslong:
With the bootnodes coming under increasing pressure and handling mulitple responsibilities including providing peer lists and acting as sync providers to new nodes in the network, it would be desirable to create a lightweight node binary specialised in providing peer lists (perhaps working in concert with a network crawler) and sync provider nodes, designed to share their chainstate. 

The (wip) setup would be as follows:

1. A new peer has a list of addresses of the official peer discovery nodes (we should ditch the name "bootstrapper" to avoid confusion); it could also know a list of sync providers, but they would not be the initial point of "entry" into the network.
2. The peer discovery nodes have regular contact with the crawler(s) and each other; the crawler can let them know what peers no longer appear live, to remove those from active peer list distribution, and recommend peers based on their available capacity and centrality.
3. The peering nodes might want to know which sync provider nodes have the most available capacity and provide those to new peers (perhaps at most one at once?). 
4. If the sync nodes are capped, the peer discovery nodes would not return them.

Tracking:

- [x] merge binaries with sync providers replacing bootnodes: #1101. 
- [x] migrate to using beacons as well as sync providers on testnet.
- [x] reinstate stricter peering rules for sync providers: #1158.

niklaslong:
Implemented by #1221.

PR #1134, opened 2021-09-08
Introduce peer discovery and sync provider nodes

niklaslong:
This PR implements the first steps towards splitting the bootnodes into peer discovery nodes (beacons) and sync provider nodes as described in #1101. It introduces separate binaries for each node type and refactors node configuration.

Although not included here, these changes also pave the way for load balancing the sync providers and leveraging the crawler(s) to improve network cohesion. 

Opening as a draft as this PR introduces architectural changes and seed addresses will need to be updated. 

~~Todo: documentation modifications.~~

### DIFF
--- a/bin/sync_provider.rs
+++ b/bin/sync_provider.rs
@@ -50,7 +50,7 @@ async fn start_server(config: Config) -> anyhow::Result<()> {
     // Construct the node instance. Note this does not start the network services.
     // This is done early on, so that the local address can be discovered
     // before any other object (miner, RPC) needs to use it.
-    let mut node = init_node(&config, Some(storage.clone())).await?;
+    let mut node = init_node(&config, storage.clone()).await?;
 
     // Enable the sync layer.
     node.set_sync(sync);
@@ -63,7 +63,7 @@ async fn start_server(config: Config) -> anyhow::Result<()> {
 
     // Start RPC thread, if the RPC configuration is enabled.
     if config.rpc.json_rpc {
-        init_rpc(&config, node.clone(), Some(storage))?;
+        init_rpc(&config, node.clone(), storage)?;
     }
 
     // Start the network services
--- a/bin/sync_provider.rs
+++ b/bin/sync_provider.rs
@@ -83,7 +83,7 @@ fn main() -> Result<(), NodeError> {
 
     let runtime = runtime::Builder::new_multi_thread()
         .enable_all()
-        .thread_stack_size(8 * 1024 * 1024)
+        .thread_stack_size(4 * 1024 * 1024)
         .build()?;
 
     runtime.block_on(start_server(config))?;
--- a/bin/sync_provider.rs
+++ b/bin/sync_provider.rs
@@ -95,7 +95,7 @@ async fn start_server(config: Config) -> anyhow::Result<()> {
 
     // Start RPC thread, if the RPC configuration is enabled.
     if config.rpc.json_rpc {
-        init_rpc(&config, node.clone(), Some(storage));
+        init_rpc(&config, node.clone(), Some(storage))?;
     }
 
     // Start the network services
--- a/bin/sync_provider.rs
+++ b/bin/sync_provider.rs
@@ -22,6 +22,7 @@ use snarkos::{
     config::{Config, ConfigCli},
     display::{initialize_logger, print_welcome},
     errors::NodeError,
+    init::{init_miner, init_node, init_rpc, init_storage, init_sync},
 };
 use snarkos_consensus::{Consensus, ConsensusParameters, DeserializedLedger, DynLedger, MemoryPool, MerkleLedger};
 use snarkos_network::{config::Config as NodeConfig, MinerInstance, Node, NodeType, Sync};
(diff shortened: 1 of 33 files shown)


bin/sync_provider.rs
```diff
@@ -0,0 +1,92 @@
...
+fn main() -> Result<(), NodeError> {
+    let arguments = ConfigCli::args();
+
+    let mut config: Config = ConfigCli::parse(&arguments)?;
+    config.node.kind = NodeType::SyncProvider;
+    config.check().map_err(|e| NodeError::Message(e.to_string()))?;
+
+    let runtime = runtime::Builder::new_multi_thread()
+        .enable_all()
+        .thread_stack_size(8 * 1024 * 1024)
```
ljedrz, line 86: this much stack is only necessary for mining; that being said, perhaps we still need like 4MiB for proper sync handling

network/src/config.rs
```diff
@@ -17,30 +17,37 @@
...
 use std::{
     net::SocketAddr,
     sync::Arc,
     time::Duration,
     {self},
 };
 
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Serialize, Deserialize)]
+pub enum NodeType {
+    Client, // Sometimes referred to as a "regular" node.
```
ljedrz, line 30: these should probably be doc comments

network/src/config.rs
```diff
@@ -17,30 +17,37 @@
...
+use serde::{Deserialize, Serialize};
 use std::{
     net::SocketAddr,
     sync::Arc,
     time::Duration,
     {self},
 };
 
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Serialize, Deserialize)]
+pub enum NodeType {
```
ljedrz, line 29: been thinking about adding one for a while now :+1:

network/src/config.rs
```diff
@@ -50,56 +57,71 @@ impl Config {
...
-        let mut bootnodes = Vec::with_capacity(bootnodes_addresses.len());
-        for bootnode_address in bootnodes_addresses.iter() {
-            if let Ok(bootnode) = bootnode_address.parse::<SocketAddr>() {
-                bootnodes.push(bootnode);
-            }
-        }
+        // Convert the given seeded nodes into socket addresses.
+        let beacons: Vec<SocketAddr> = beacon_addresses
+            .into_iter()
+            .flat_map(|addr| addr.parse().ok())
```
ljedrz, line 71:
```suggestion
            .filter_map(|addr| addr.parse().ok())
```
it conveys the logic a bit better

network/src/config.rs
```diff
@@ -50,56 +57,71 @@ impl Config {
...
+    pub fn is_client(&self) -> bool {
+        matches!(self.node_type, NodeType::Client)
     }
 
-    /// Returns `true` if this node is a plain node. Otherwise, returns `false`.
+    /// Returns `true` if this node is a peer discovery node. Otherwise, returns `false`.
     #[inline]
-    pub fn is_regular_node(&self) -> bool {
-        !(self.is_bootnode() || self.is_crawler())
+    pub fn is_beacon(&self) -> bool {
```
ljedrz, line 123: I'm wondering if we couldn't deduplicate these methods into a single `is_of_type(t: NodeType)` or something like that

network/src/peers/peers.rs
```diff
@@ -71,7 +71,7 @@ impl Node {
             let number_to_connect = crawling_capacity.saturating_sub(active_peer_count - number_to_disconnect);
 
             (number_to_disconnect, number_to_connect)
-        } else if self.config.is_bootnode() {
+        } else if self.config.is_sync_provider() || self.config.is_beacon() {
             // Bootnodes disconnect down to 80% of their max to leave capacity open for new
```
ljedrz, line 75: nit: you'll probably need an `rg -i` pass for the `bootnode` phrase :microscope:

network/src/peers/peers.rs
```diff
@@ -113,17 +113,17 @@ impl Node {
             }
         }
 
-        // Attempt to connect to a few random bootnodes if the node has no active
-        // connections or if it's a bootnode itself.
-        if self.peer_book.get_active_peer_count() == 0 || self.config.is_bootnode() {
-            let random_bootnodes = self
+        // Attempt to connect to a few random peer provider nodes if the node has no active
```
ljedrz, line 116: they were called "peer discovery nodes" earlier on, I think; it's best to either use a single name or even the `NodeType` in order to avoid confusion

network/src/peers/peers.rs
```diff
@@ -352,11 +351,20 @@ impl Node {
         };
 
         // Limit set size.
-        let peers = peers
+        let mut peers: Vec<SocketAddr> = peers
             .choose_multiple(&mut SmallRng::from_entropy(), crate::SHARED_PEER_COUNT)
```
ljedrz, line 355:
```suggestion
            .choose_multiple(&mut SmallRng::from_entropy(), crate::SHARED_PEER_COUNT - 1)
```

ljedrz: to account for the extra sync provider

network/tests/peers.rs
```diff
@@ -100,7 +100,7 @@ async fn bootnode_peer_propagation() {
             && node_charlie.peer_book.get_active_peer_count() == 2
     };
 
-    // Make sure B and C connect => bootnode propagates peers (without `is_routable` check in this
+    // Make sure B and C connect => becon propagates peers (without `is_routable` check in this
```
ljedrz, line 103:
```suggestion
    // Make sure B and C connect => beacon propagates peers (without `is_routable` check in this
```

rpc/documentation/public_endpoints/getnodeinfo.md
```diff
@@ -8,7 +8,7 @@ None
 
 |     Parameter    |     Type      |                  Description                  |
 |:----------------:|:-------------:|:---------------------------------------------:|
-| `is_bootnode`    | bool          | Flag indicating if the node is a bootnode     |
+| `node_type`      | NodeType      | Flag indicating if the node type              |
```
ljedrz, line 11: only natural to include this change here too :+1:

snarkos/errors/cli.rs
```diff
@@ -28,11 +28,14 @@ pub enum CliError {
...
+    #[error("A crawler node can't also be a miner")]
+    MinerCrawler,
 
-    #[error("The node can't be a bootstrapper and a crawler at the same time")]
-    CrawlerBootstrapper,
+    #[error("The peer discovery node can't also be a miner")]
+    MinerBeacon,
+
+    #[error("The sync provider node can't also be a miner")]
+    MinerSyncProvider,
```
ljedrz, line 38: these miner-related errors could probably be deduplicated into a single `CantMine` one explaining that only client nodes can be miners

ljedrz (approved):
I like this direction :+1:; left some suggestions.

One question: are these types aligned with the related ARC? It could also be linked in the doc comment for `NodeType`.

network/src/peers/peers.rs
niklaslong: Alright, I'll use "beacon" (and clarify that refers to peer discovery nodes in the relevant docs/READMEs).

network/src/config.rs
niklaslong: Might need a macro for this, `matches!(self.node_type, t)` [won't actually work](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=5d8d78bc9f84119df61805a80945633a). I've been refactoring the network crate on another branch, I'll make a note to explore this change there?

niklaslong: Actually, we've got PartialEq 👀

rpc/documentation/public_endpoints/getnodeinfo.md
niklaslong: Nice type there too 😆

network/src/peers/peers.rs
niklaslong: Replace the first address in the list with the sync provider instead, since this is only really applicable to beacons.

niklaslong:
Note: we will need to change the beacon and sync provider addresses before merging.

network/src/node.rs
```diff
@@ -137,7 +137,7 @@ impl Node {
             terminator: Arc::new(AtomicBool::new(false)),
         }));
 
-        if node.config.is_crawler() {
+        if node.config.is_of_type(NodeType::Crawler) {
```
ljedrz, line 140: sorry I haven't noticed this before, but perhaps we can move this method up to `Node`? it would shorten the calls

network/src/peers/peers.rs
```diff
@@ -358,12 +358,12 @@ impl Node {
...
-        // Make sure to include a sync provider in the addresses if this node is a peer discovery
-        // node. In future, sync provider addresses wouldn't be provided if their capacity is
-        // maxed out.
+        // Make sure to include a sync provider in the addresses if this node is a beacon. In
+        // future, sync provider addresses wouldn't be provided if their capacity is maxed out.
         if self.config.is_of_type(NodeType::Beacon) {
             if let Some(random_sync_provider) = self.config.sync_providers().choose(&mut SmallRng::from_entropy()) {
-                peers.push(*random_sync_provider);
+                // Replace to maintain the size of the list.
+                peers[0] = *random_sync_provider;
```
ljedrz, line 366: I'd use `first_mut` instead so we can avoid an (unlikely, but still) panic

network/src/node.rs
niklaslong: Sure!

Protryon:
why are we adding a new binary? Let's just do what we do for bootnodes, miners, and crawlers, and add in a new runtime flag to avoid installation complexity.

Protryon (approved):

niklaslong:
Rebased.

PR #1136, opened 2021-09-09
select! tweaks

ljedrz:
If multiple futures passed to `futures::select` are ready, one is picked at random; this could cause issues (possibly even the stalls we've been seeing), so use `select_biased` or avoid `select` altogether instead, which will ensure that disconnects and timeouts are always triggered when necessary.

Cc https://github.com/AleoHQ/snarkOS/issues/1037, https://github.com/AleoHQ/snarkOS/issues/1066

### DIFF
--- a/network/src/peers/peer/connector.rs
+++ b/network/src/peers/peer/connector.rs
@@ -19,8 +19,7 @@ use std::{
     time::Duration,
 };
 
-use futures::{select, FutureExt};
-use tokio::{net::TcpStream, sync::mpsc};
+use tokio::{net::TcpStream, sync::mpsc, time::timeout};
 
 use snarkos_metrics::{self as metrics, connections::*};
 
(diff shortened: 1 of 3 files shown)


ljedrz:
note: these changes might cause a small discrepancy in queue utilization metrics over a node's runtime, but the fix is important enough to be merged as-is despite that; the metrics can be adjusted later on

Protryon (approved):

Protryon:
The contractual requirements for futures dictate that if `Poll::Ready` is not received from a futures `poll` function, then it can't have mutated internal state. As such, the implementation of `select!`, to my knowledge, will not drop `Poll::Ready` results with meaningful data, but just the order in which `poll` calls are made is not clearly defined -- I would expect that it just loops iteratively until one hits a `Ready`.

Nonetheless, this doesn't look harmful, so approved.

Protryon:
See https://docs.rs/futures-util/0.3.17/src/futures_util/future/select_all.rs.html#55 for detail

niklaslong (approved):

Issue #1066, opened 2021-08-23
[Bug] Syncing seems to freeze up until node is restarted every 50-100k blocks.


Protryon:
It seems that when syncing, every 50-100k blocks, we have a chance to get stuck in a node until the node is restarted.

Unconfirmed, but it seems to me the issue is peer isolation resulting in only being connected to peers on forks since they don't send us sync blocks. It seems plausible that failures accumulate and that the 15 minute expiry timer for failures is overzealous or a bad approach in a congested network (perhaps we should just deprioritize low quality nodes rather than disconnect?)

ljedrz:
Cc https://github.com/AleoHQ/snarkOS/issues/1037

ljedrz:
sounds plausible :+1:

ljedrz:
I haven't encountered this in a while now; marking the issue as `likely fixed`.

ljedrz:
I haven't encountered this issue in a good while now, so it appears to have been fixed.

Issue #1037, opened 2021-08-11
[Bug] The node can stall while syncing blocks


ljedrz:
## 🐛 Bug Report

I'm not sure this is strictly sync-related, but I observed this while syncing blocks with a regular node with ~100 peers:
```
Aug 11 11:23:39.259  INFO Received block from <peer 1> of epoch 1601230268 with hash 44457d6b1953ebce125621b11c9e7f228d0d32f4755ce641fa5a1086b4d9311a (current head 24036)
Aug 11 11:23:43.062  INFO Updating peers
Aug 11 11:23:43.062  INFO Connected to 103 peers
Aug 11 11:23:43.401  WARN unrecoverable failure communicating to inbound peer '<peer 2>': 'Io(Kind(UnexpectedEof))'
Aug 11 11:23:44.496  WARN did not receive block 1/8 'e1be3133178d467a316ee1e4927a0b3fd299da62b62815173b3abad11517034d' by deadline for sync from <peer 3>
Aug 11 11:23:44.496  WARN did not receive block 2/8 '092134fddb052fe33f4a6931646520828e9b2accbf8ea2dc568b9f2d64c53823' by deadline for sync from <peer 4>
Aug 11 11:23:44.496  WARN did not receive block 3/8 '0756e71c32e8ccbcf6ad1f7ca47cf64cbc09ad142eed05a0b2242a5597d139ed' by deadline for sync from <peer 5>
Aug 11 11:23:44.496  WARN did not receive block 4/8 '8ec6953cbdb93d254d5b24152842a46eb5a8543ac5070771f38a8902bac7c728' by deadline for sync from <peer 6>
Aug 11 11:23:44.496  WARN did not receive block 5/8 'e1bed543e4a63369e7167acf927fb5ce504b69d76907456af5036d21a43ff535' by deadline for sync from <peer 7>
Aug 11 11:23:44.496  WARN did not receive block 6/8 '26025434160fbc9863f78869d7331c5c98534470d4ce1737ddb2a4be3d23c9df' by deadline for sync from <peer 8>
Aug 11 11:23:44.496  WARN did not receive block 7/8 '20470d8e5b5f0cdaddfa5901b0fb3b8605289ef71328ac817782ae9b0a9baf2c' by deadline for sync from <peer 9>
Aug 11 11:23:46.325  WARN unrecoverable failure communicating to outbound peer '<peer 10>': 'Io(Kind(UnexpectedEof))'
[...]

PR #1138, opened 2021-09-10
Replace futures::select with tokio::select

ljedrz:
I've been able to reproduce node stalls with peer attrition reliably using my slightly less powerful other PC; the node was barely doing anything (just some internal `tokio` functions were detected by `perf record`) during that time, and surprisingly, it was able to just snap back to life after a long delay, suggesting that it was a glitch related to peer/connection handling and that the removal of a "problem peer" would unblock things.

At that time I started suspecting that this is some `tokio`/`futures` bug, or an issue related to their interop; with `tokio-console` still in its infancy, and with seemingly no related existing GitHub issues for either `tokio` or `futures`, I decided to just try to reduce the possible scope of issues by removing the use of `futures` where `tokio` equivalents existed.

With this PR (on its own), which just replaces the use of `futures::select` with `tokio::select` (and `futures::Future` with `std::future::Future` as a drive-by, though that's just an import cleanup), my node no longer hangs; it still occasionally loses a few peers at once, but it is able to immediately connect to others via regular connection maintenance.

Cc #1037, #1066

### DIFF
--- a/network/src/peers/peer/peer.rs
+++ b/network/src/peers/peer/peer.rs
@@ -16,7 +16,6 @@
 
 use anyhow::*;
 use chrono::Utc;
-use futures::{select_biased, FutureExt};
 use serde::{Deserialize, Serialize};
 use std::{
     net::SocketAddr,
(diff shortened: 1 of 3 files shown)


niklaslong (commented):
LGTM, testing locally as a write this.

Protryon (approved):

niklaslong (approved):

PR #1142, opened 2021-09-13
deal with aggro sync on receiving side

Protryon:
This PR:
* Removes peer failure from returning empty sync

Note, avoiding sending the empty sync on the other end can induce a failure (looks like a timeout) -- so we still send empty syncs.


niklaslong (approved):
LGTM 👍

PR #1147, opened 2021-09-14
Sqlite tree query, fix migration

Protryon:
This PR:
* Makes the block tree fetch a single query for sqlite
* Makes scanning for forks a storage owned operation (only ever called storage anyways), single query
* Fixes migration by not attempting to carry over previous commit state.

### DIFF
--- a/storage/src/storage/sqlite/sync.rs
+++ b/storage/src/storage/sqlite/sync.rs
@@ -599,6 +599,32 @@ impl SyncStorage for SqliteStorage {
         Ok(out)
     }
 
+    fn scan_forks(&mut self, scan_depth: u32) -> Result<Vec<(Digest, Digest)>> {
+        let mut stmt = self.conn.prepare_cached(
+            r"
+            WITH RECURSIVE
+                children(parent, sub, length) AS (
+                    SELECT NULL, (select hash from blocks where canon_height = (select max(canon_height) from blocks)), 0
+                    UNION ALL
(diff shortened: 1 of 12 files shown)


storage/src/storage/sqlite/sync.rs
```diff
@@ -519,6 +523,68 @@ impl SyncStorage for SqliteStorage {
         Ok(out)
     }
 
+    fn get_block_digest_tree(&mut self, block_hash: &Digest) -> Result<DigestTree> {
+        let mut nodes: HashMap<Digest, Vec<Digest>> = HashMap::new();
```
niklaslong, line 527: Futur perf suggestion: leverage `hash_hasher` (see #1047) since this is storing digests?

storage/src/storage/sqlite/sync.rs
```diff
@@ -519,6 +523,68 @@ impl SyncStorage for SqliteStorage {
...
+            .collect::<rusqlite::Result<Vec<(Digest, Digest)>>>()?;
+        for (hash, parent_hash) in out {
+            nodes.entry(parent_hash).or_insert_with(Vec::new).push(hash);
+        }
+
+        // only our base case
+        if nodes.is_empty() {
+            return Ok(DigestTree::Leaf(block_hash.clone()));
+        }
+        let mut node_entries: HashSet<Digest> = nodes.keys().cloned().collect();
```
niklaslong, line 554: ditto

niklaslong (approved):
lgtm, left a few suggestions.

PR #1149, opened 2021-09-14
 Revert "temporarily relax peering rules for sync providers"

niklaslong:
Follow-up to #1134, reinstates the stricter peering rules for sync providers. Only the last commit is new.

This should only be merged once the beacon/sync provider rollout is complete.


niklaslong:
Superseded by #1158.

PR #1150, opened 2021-09-14
Remove BlockHeaderHash

Protryon:
Deprecated in favor of `Digest`

### DIFF
--- a/network/src/sync/blocks/aggro.rs
+++ b/network/src/sync/blocks/aggro.rs
@@ -19,7 +19,6 @@ use std::{sync::Arc, time::Duration};
 use crate::{Cache, Node, Payload, Peer, SyncBase, SyncInbound};
 use anyhow::*;
 use snarkos_storage::Digest;
-use snarkvm_dpc::BlockHeaderHash;
 use tokio::{
     sync::{mpsc, RwLock},
     time::Instant,
@@ -31,7 +30,7 @@ pub struct SyncAggro {
 }
 
 struct BlockLocatorHashes {
-    hashes: Vec<BlockHeaderHash>,
+    hashes: Vec<Digest>,
     last_update: Instant,
 }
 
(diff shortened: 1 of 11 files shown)


network/src/sync/blocks/aggro.rs
```diff
@@ -20,7 +20,6 @@ use crate::{Cache, Node, Payload, Peer, SyncBase, SyncInbound};
 use anyhow::*;
 use indexmap::IndexSet;
```
niklaslong, line 21: I believe this is an unused import, would be good to remove as a fly-by.

network/tests/handshake.rs
```diff
@@ -185,13 +184,13 @@ async fn reject_non_version_messages_before_handshake() {
...
     let mut peer_stream = TcpStream::connect(node.expect_local_addr()).await.unwrap();
-    let block_hash = BlockHeaderHash::new([0u8; 32].to_vec());
+    let block_hash = Digest::from([0u8; 32]);
     write_message_to_stream(Payload::GetBlocks(vec![block_hash]), &mut peer_stream).await;
     assert_node_rejected_message(&node, &mut peer_stream).await;
 
     // GetSync
     let mut peer_stream = TcpStream::connect(node.expect_local_addr()).await.unwrap();
-    let block_hash = BlockHeaderHash::new([0u8; 32].to_vec());
+    let block_hash = Digest::from([0u8; 32]);
```
niklaslong, line 193: Do we want a constant for the block hash length?

niklaslong (approved):

network/src/sync/blocks/aggro.rs
Protryon: done in another PR

network/tests/handshake.rs
Protryon: it's a random test, not worried about it.

PR #1154, opened 2021-09-15
Cleanup metrics

Protryon:
Creates a wrapped channel for more reliable metrics reporting.

### DIFF
--- a/consensus/src/consensus/inner/agent.rs
+++ b/consensus/src/consensus/inner/agent.rs
@@ -14,6 +14,8 @@
 // You should have received a copy of the GNU General Public License
 // along with the snarkOS library. If not, see <https://www.gnu.org/licenses/>.
 
+use snarkos_metrics::wrapped_mpsc;
+
 use crate::consensus::ConsensusMessageWrapped;
 
 use super::*;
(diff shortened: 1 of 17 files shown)


niklaslong:
Running this locally, `inbound_network_messages` is showing negative values. Other queue metrics are flatlining as well.

Protryon:
> Running this locally, `inbound_network_messages` is showing negative values. Other queue metrics are flatlining as well.

I fixed the issue with it going negative.

Flatlining to zero is the expected behavior for a properly functioning message queue? Unless there is a backlog.

niklaslong:
> Flatlining to zero is the expected behavior for a properly functioning message queue?

Indeed, though I never saw it rise despite seeing that whilst running a node on `staging` just prior with the same configuration. I may be wrong but wanted to raise the observation.

niklaslong (approved):

Protryon:
> > Flatlining to zero is the expected behavior for a properly functioning message queue?
> 
> Indeed, though I never saw it rise despite seeing that whilst running a node on `staging` just prior with the same configuration. I may be wrong but wanted to raise the observation.

I was able to see storage climb during an expensive storage action series.

PR #1159, opened 2021-09-16
Fix stackoverflow in fast-forward

Protryon:
`to_vec` was cloning each layer of digest tree, recursively cloning all nodes in the tree. This caused a stack overflow with sufficiently large numbers of blocks.


niklaslong (approved):
lgtm 👍

vinovod:
Fix pulled and checked. Thanks!

PR #1160, opened 2021-09-16
Peer Storage

Protryon:
This PR adds persistent peer storage to snarkOS.

It doesn't change current failure exclusion/temp. banning, but it does give us some persistent metrics to make informed decisions on in a future PR.


network/src/peers/peer_book.rs
```diff
@@ -202,6 +209,19 @@ impl PeerBook {
         .await;
     }
 
+    pub async fn serialize(&self) -> Vec<snarkos_storage::Peer> {
+        self.map_each_peer(move |peer| async move { peer.load().await.map(|x| x.serialize()) })
+            .await
+    }
+
+    pub async fn serialize_disconnected(&self) -> Vec<snarkos_storage::Peer> {
```
ljedrz:
```suggestion
    pub fn serialize_disconnected(&self) -> Vec<snarkos_storage::Peer> {
```

network/src/peers/peers.rs
```diff
@@ -318,7 +327,11 @@ impl Node {
 
         let basic_filter =
             |peer: &Peer| peer.address != remote_address && !self.config.bootnodes().contains(&peer.address);
-        let strict_filter = |peer: &Peer| basic_filter(peer) && peer.is_routable.unwrap_or(false);
+        let strict_filter = |peer: &Peer| {
+            basic_filter(peer)
+                && peer.quality.connection_transient_fail_count == 0
+                && peer.quality.connection_attempt_count > 0
```
ljedrz, line 334: this only means we tried to connect to the peer, but didn't necessarily succeed in doing so; wouldn't a check for `connected_count` - or some new metric like `connection_init_success_count` - be a better fit here?

ljedrz: oh, I see - when a peer gets connected to, the value of `connection_transient_fail_count` is reset; that works too

storage/src/objects/peer.rs
```diff
@@ -0,0 +1,36 @@
...
+pub struct Peer {
+    pub address: SocketAddr,
+    pub block_height: u32,
+    pub first_seen: Option<DateTime<Utc>>,
+    pub last_seen: Option<DateTime<Utc>>,
+    pub last_connected: Option<DateTime<Utc>>,
+    pub blocks_synced_to: u32,
+    pub blocks_synced_from: u32,
+    pub blocks_received_from: u32,
+    pub blocks_sent_to: u32,
```
ljedrz, line 32: are these new items designed to give us the ability to check if the peer is not sending too many requests for blocks or blocks to the node and be grounds for a potential ban? we might also want to add something like a `messages_received` to guard against e.g. a flood of `Ping`s or `GetPeers` messages

ljedrz (approved):
needs a rebase, but LGTM other than that; left some comments

storage/src/objects/peer.rs
Protryon: this is mostly for sync/connection prioritization in some of my future PRs

PR #1170, opened 2021-09-21
Improve block processing time metrics

niklaslong:
Adds some detail to the block processing metrics by also tracking the block verification and commit time.

### DIFF
--- a/consensus/src/consensus/inner/commit.rs
+++ b/consensus/src/consensus/inner/commit.rs
@@ -148,6 +148,8 @@ impl ConsensusInner {
         hash: &Digest,
         block: &SerialBlock,
     ) -> Result<(), ConsensusError> {
+        let now = std::time::Instant::now();
+
         match self.storage.get_block_state(hash).await? {
             BlockStatus::Committed(_) => return Ok(()),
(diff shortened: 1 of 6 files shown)


ljedrz (approved):
LGTM :+1:

ljedrz:
While running this branch and checking the metrics, I thought that perhaps we could improve the name of the `block_processing_time` one now, in order not to confuse the users? Alternatively, we could skip its accounting if the node is currently syncing; I'm just wondering if that metric isn't misleading now :thinking:.

niklaslong:
I'd be happy to remove it, though do we want to keep it as is as a reference to improve the performance of the costly processing? Not tracking it during sync means we won't see many values as most nodes are syncing until they reach the canon tip.

ljedrz:
I'm ok with keeping it, just wondering if there's some better name for it, perhaps `inbound_block_processing_time` (verbose) or perhaps extending the RPC description would help here.

niklaslong:
I've made the clarification in the docs for now. It might soon be worth moving some of the `misc` metrics to a `block` struct in a follow-up, since most of the misc metrics are currently block related (with the exception of the rpc call count).

PR #1179, opened 2021-09-23
Re-structure block metrics

niklaslong:
Follow-up to #1170. Splits the block metrics out from `misc` into the `blocks` mod (and struct, for rpc metrics).

### DIFF
--- a/metrics/src/names.rs
+++ b/metrics/src/names.rs
@@ -67,14 +67,17 @@ pub mod queues {
 }
 
 pub mod misc {
-    pub const BLOCK_HEIGHT: &str = "snarkos_misc_block_height_total";
-    pub const BLOCKS_MINED: &str = "snarkos_misc_blocks_mined_total";
-    pub const DUPLICATE_BLOCKS: &str = "snarkos_misc_duplicate_blocks_total";
-    pub const DUPLICATE_SYNC_BLOCKS: &str = "snarkos_misc_duplicate_sync_blocks_total";
-    pub const ORPHAN_BLOCKS: &str = "snarkos_misc_orphan_blocks_total";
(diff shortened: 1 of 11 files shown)


metrics/src/names.rs
```diff
@@ -67,14 +67,17 @@ pub mod queues {
...
     pub const RPC_REQUESTS: &str = "snarkos_misc_rpc_requests_total";
-    pub const BLOCK_PROCESSING_TIME: &str = "snarkos_misc_block_processing_time";
-    pub const BLOCK_COMMIT_TIME: &str = "snarkos_misc_block_commit_time";
+}
+
+pub mod blocks {
+    pub const HEIGHT: &str = "snarkos_blocks_height_total";
+    pub const MINED: &str = "snarkos_blocks_mined_total";
+    pub const DUPLICATE_BLOCKS: &str = "snarkos_duplicate_blocks_total";
+    pub const DUPLICATE_SYNC_BLOCKS: &str = "snarkos_duplicate_sync_blocks_total";
```
niklaslong: Note: I didn't follow the convention here but I think e.g. `snarkos_blocks_duplicate_sync_blocks_total` didn't add any extra context.

niklaslong: Another option is `snarkos_blocks_duplicate_total` and `snarkos_blocks_duplicate_sync_total`.

ljedrz: I think only the first prefix and the final suffix matter for `prometheus`, though it might be practical to keep a convention; I'd go with:
- `snarkos_blocks_duplicates_total`
- `snarkos_blocks_duplicates_sync_total`

Protryon (approved):

ljedrz (approved):

PR #1182, opened 2021-09-24
A more thorough storage trim

ljedrz:
This PR makes the `--trim-storage` functionality remove transactions that don't belong to canon blocks in addition to removing non-canon blocks. The resulting reduction in database size is significant.

~In order to be able to remove the obsolete transactions, we need to ensure that every tx belonging to a canon block is stored. To do this, the block commit operation is extended with an update to the transaction location table, so that cases where a transaction first delivered in a non-canon block is later followed by a competing block containing the same transaction which becomes canon don't lead to the transaction being removed due to its location not being canon in the end.~

~Filing this as a draft in order to run some more tests and to decide if we want to migrate the current storage (which would require all the existing canon blocks to be partially canonized again) or just wait until testnet2, where no such migration would be necessary.~

Actually, this was much easier than I initially thought, as the transaction locations are connected to transactions and blocks based on their database ids as opposed to hashes, meaning we neither need to perform any migration nor update transaction locations upon commits - this is also part of the reason why this extended trim is so important - all transactions coming from all blocks have their respective locations stored individually, which uses plenty of space.

### DIFF
--- a/storage/src/storage/sqlite/sync.rs
+++ b/storage/src/storage/sqlite/sync.rs
@@ -1195,4 +1195,30 @@ impl SyncStorage for SqliteStorage {
         self.init()?;
         Ok(())
     }
+
+    fn trim(&mut self) -> Result<()> {
+        // Remove transactions belonging to non-canon blocks.
+        self.conn.execute(
+            r"
+            DELETE FROM transactions
+            WHERE id IN (
+                SELECT transactions.id
+                FROM transactions
+                INNER JOIN transaction_blocks on transaction_blocks.transaction_id = transactions.id
+                INNER JOIN blocks on blocks.id = transaction_blocks.block_id
+                WHERE blocks.canon_height IS NULL
+            )
+        ",
+            [],
+        )?;
+
+        // Remove non-canon blocks.
+        self.conn
+            .execute(r"DELETE FROM blocks WHERE blocks.canon_height IS NULL", [])?;
+
+        // Compact the storage file.
+        self.conn.execute("VACUUM", [])?;
+
+        Ok(())
+    }
 }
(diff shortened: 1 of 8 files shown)


niklaslong (approved):

ljedrz:
@Protryon I've tested this PR extensively and it seems to work just fine, but it would be best if you took an extra look if I didn't miss anything wrt the new storage.

storage/src/storage/sqlite/sync.rs
```diff
@@ -1001,4 +1001,30 @@ impl SyncStorage for SqliteStorage {
...
+    fn trim(&mut self) -> Result<()> {
+        // Remove transactions belonging to non-canon blocks.
+        self.conn.execute(
+            r"
+            DELETE FROM transactions
+            WHERE id IN (
+                SELECT transactions.id
+                FROM transactions
+                INNER JOIN transaction_blocks on transaction_blocks.transaction_id = transactions.id
+                INNER JOIN blocks on blocks.id = transaction_blocks.block_id
```
Protryon: we also probably want to delete transactions with no associated blocks (i.e. was trimmed from legacy impl, but not the transaction) -- or if the transaction found its way in there some other way.

can be done by using
```
LEFT JOIN transaction_blocks on transaction_blocks.transaction_id = transactions.id
LEFT JOIN blocks on blocks.id = transaction_blocks.block_id
WHERE blocks IS NULL OR blocks.canon_height IS NULL -- the second check is sufficient, this is just more obvious/readable with no perf cost
```

Protryon (approved):

storage/src/storage/sqlite/sync.rs
ljedrz: adjusted the query to use `LEFT JOIN` :+1:; the suggested `blocks IS NULL` check was faulty (it caused no items to get trimmed), so instead I added one for `transaction_blocks.transaction_id` in order to account for orphan locations.

PR #1187, opened 2021-09-27
Sync tweaks

Protryon:
This PR:
* Splits reconnections to disconnected nodes between the highest claimed block count, and random. This provides us a higher chance to pull a longer chain, but still providing noise and distributing blocks.
* Makes aggro sync rounds last up to 60 seconds, removed interim block locator hash update (was pointless since we just restart after 15, now 60 seconds)
* Allows chain-scanning behavior in aggro sync (was disabled in a regression)
* Adjusts block locator hashes to put proportional hashes first, and makes quadratic hashes look at the the blocks closer to the origin of the chain.

### DIFF
--- a/network/src/peers/peers.rs
+++ b/network/src/peers/peers.rs
@@ -225,7 +225,6 @@ impl Node {
         // Local address must be known by now.
         let own_address = self.expect_local_addr();
 
-        // If this node is not a bootnode, attempt to satisfy the minimum number of peer connections.
         let random_peers: Vec<SocketAddr> = {
             trace!(
                 "Connecting to {} disconnected peers",
--- a/network/src/peers/peers.rs
+++ b/network/src/peers/peers.rs
@@ -244,8 +244,6 @@ impl Node {
                 candidates.sort_unstable_by_key(|peer| peer.quality.last_connected);
             }
 
-            // let addr_iter = candidates.into_iter(); // .iter().map(|peer| peer.address);
-
             if !self.is_of_type(NodeType::Client) {
                 candidates.into_iter().take(count).collect()
             } else {
--- a/network/src/peers/peers.rs
+++ b/network/src/peers/peers.rs
@@ -20,12 +20,13 @@ use std::{
     time::{Duration, Instant},
 };
 
+use itertools::Itertools;
 use rand::{prelude::SliceRandom, rngs::SmallRng, seq::IteratorRandom, SeedableRng};
 use tokio::task;
 
 use snarkos_metrics::{self as metrics, connections::*};
 
-use crate::{message::*, KnownNetworkMessage, NetworkError, Node, NodeType};
+use crate::{message::*, KnownNetworkMessage, NetworkError, Node, NodeType, Peer};
 use anyhow::*;
 
 impl Node {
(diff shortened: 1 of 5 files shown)


network/src/peers/peers.rs
```diff
@@ -224,7 +225,8 @@ impl Node {
         // Local address must be known by now.
         let own_address = self.expect_local_addr();
 
-        let random_peers = {
+        // If this node is not a bootnode, attempt to satisfy the minimum number of peer connections.
```
niklaslong: This comment is slightly misleading as we don't have bootnodes anymore. Might be best to remove it entirely since its use is documented in `update`. 

```suggestion
```

network/src/peers/peers.rs
```diff
@@ -242,14 +244,29 @@ impl Node {
...
             }
 
-            let addr_iter = candidates.iter().map(|peer| peer.address);
-
             if !self.is_of_type(NodeType::Client) {
-                addr_iter.take(count).collect()
+                candidates.into_iter().take(count).collect()
             } else {
-                addr_iter.choose_multiple(&mut SmallRng::from_entropy(), count)
+                let random_count = count / 2;
```
niklaslong, line 250:
```suggestion 
                // Floored if count is odd.
                let random_count = count / 2;
```

network/src/peers/peers.rs
```diff
@@ -242,14 +244,29 @@ impl Node {
...
+                let random_picks: Vec<Peer> = candidates
+                    .iter()
+                    .choose_multiple(&mut SmallRng::from_entropy(), random_count)
+                    .into_iter()
+                    .cloned()
+                    .collect();
+                candidates.sort_unstable_by(|x, y| y.quality.block_height.cmp(&x.quality.block_height));
```
niklaslong, line 257:
```suggestion
                candidates.sort_unstable_by_key(|x| std::cmp::Reverse(x.quality.block_height));
```

storage/src/storage/sync.rs
```diff
@@ -246,40 +246,16 @@ pub trait SyncStorage {
...
 
         // Calculate the average distance between block hashes based on the desired number of locator hashes.
-        let mut proportional_step =
+        let proportional_step =
             (hash_index.min(oldest_fork_threshold as u32) / num_locator_hashes).min(crate::NUM_LOCATOR_HASHES - 1);
 
         // Provide hashes of blocks with indices descending quadratically while the quadratic step distance is
         // lower or close to the proportional step distance.
-        let num_quadratic_steps = (proportional_step as f32).log2() as u32;
+        let num_quadratic_steps = ((hash_index as f32).log2() - (proportional_step as f32).log2()) as u32;
```
niklaslong, line 254: I'd update the comment to document this change. 

(The log subtraction could also be shorter if replaced by a division inside the log, though this may be less clear, just mentioning it's an option.)

niklaslong (commented):
Testing locally at the moment. Looks good, there are some lints to address and I left some suggestions. There is also a broken test according to the CI.

niklaslong:
I'm not able to sync effectively locally. From scratch, I'm receiving a lot of blocks at a higher height but no relevant blocks for the node's chain. Wondering if the locator hashes are causing this.

niklaslong:
Restarted the node and eventually got a good peer to sync with. Seems like it's working fine, though still concerned about the inability to progress I've encountered with both older and fresh dbs.

niklaslong (approved):

PR #1190, opened 2021-10-01
Connection metrics tests 

niklaslong:
Introduces the first few metrics tests (keeping this PR small as it includes other changes). This PR also fixes the cache hits metric which had been forgotten.

Because of how state is tracked globally in the metrics, this PR introduces some functions to clear the state after each test. In addition, these tests will need to be run on a single-threaded executor, otherwise state will leak.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -192,7 +192,10 @@ jobs:
       - run:
           name: Build and run tests
           no_output_timeout: 35m
-          command: cd testing && RUST_MIN_STACK=8388608 cargo test --release -- --nocapture --skip metrics && RUST_MIN_STACK=8388608 cargo test --release metrics -- --test-threads 1
+          command: |
+            cd testing
+            RUST_MIN_STACK=8388608 cargo test --release -- --nocapture --skip metrics
+            cargo test --release metrics -- --test-threads 1
       - clear_environment:
           cache_key: snarkos-testing-cache
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -192,7 +192,7 @@ jobs:
       - run:
           name: Build and run tests
           no_output_timeout: 35m
-          command: cd testing && RUST_MIN_STACK=8388608 cargo test --release -- --nocapture
+          command: cd testing && RUST_MIN_STACK=8388608 cargo test --release -- --nocapture --skip metrics && RUST_MIN_STACK=8388608 cargo test --release metrics -- --test-threads 1
       - clear_environment:
           cache_key: snarkos-testing-cache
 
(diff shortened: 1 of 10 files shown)


.circleci/config.yml
```diff
@@ -192,7 +192,7 @@ jobs:
       - run:
           name: Build and run tests
           no_output_timeout: 35m
-          command: cd testing && RUST_MIN_STACK=8388608 cargo test --release -- --nocapture
+          command: cd testing && RUST_MIN_STACK=8388608 cargo test --release -- --nocapture --skip metrics && RUST_MIN_STACK=8388608 cargo test --release metrics -- --test-threads 1
```
ljedrz: nit: it would be more readable to run 2 separate commands here (still with the same build, though); also, the metrics tests probably don't need a stack this large, but it might still be required due to how the node spawns the `tokio` executor :thinking:

metrics/src/lib.rs
```diff
@@ -99,7 +99,13 @@ pub fn initialize() -> Option<tokio::task::JoinHandle<()>> {
 
 #[cfg(not(feature = "prometheus"))]
 pub fn initialize() -> Option<tokio::task::JoinHandle<()>> {
+    #[cfg(not(feature = "test"))]
     metrics::set_recorder(&crate::stats::NODE_STATS).expect("couldn't initialize the metrics recorder!");
 
+    // Let the errors through in case the recorder has already been set as is likely when running
```
ljedrz, line 105: does this still apply? I thought the single-threaded mode for these tests was there to start with a clean slate every time

testing/src/metrics/mod.rs
```diff
@@ -0,0 +1,135 @@
...
+    let peer = handshaken_peer(node.expect_local_addr()).await;
+
+    // Needed to make sure the values have been updated.
+    wait_until!(5, node.peer_book.get_connected_peer_count() == 1);
+
+    // ...the metrics should reflect this.
+    let metrics = NODE_STATS.snapshot();
+
+    assert_eq!(metrics.connections.all_accepted, 1);
+    assert_eq!(metrics.connections.connected_peers, 1);
```
ljedrz: I'm missing the `all_initiated`, `disconnected` and `connecting` checks here (not as important, but no harm in being extra thorough); the same applies to the post-disconnect checks and the other test

ljedrz (approved):
LGTM with some suggestions :+1:

metrics/src/lib.rs
niklaslong: It does since the recorder is global (and leaks) even when running single threaded. Since we need to initialize the metrics for each test (as we don't want to depend on the running order of the tests), this call will error for all cases except the first in the run.

ljedrz:
note: it seems that one of the RPC tests is flaky (unrelated to this PR).

Issue #1184, opened 2021-09-24
[Bug] Slow RPC response times


evgeny-garanin:
## 🐛 Bug Report

Since latest updates (1.3.16, maybe 1.3.15 too), RPC calls to node sometime take way longer to get response from.
Specifically method "getblockcount".

## Steps to Reproduce

1. Get your node running
2. Run following calls (in .sh script preferably)
curl -s --data-binary '{"jsonrpc": "2.0", "id":"documentation", "method": "getpeerinfo", "params": [] }' -H 'content-type: application/json' http://$1:3030/   | jq '.[].peers?';
echo "NODE INFO:";
curl -s --data-binary '{"jsonrpc": "2.0", "id":"documentation", "method": "getnodeinfo", "params": [] }' -H 'content-type: application/json' http://$1:3030/ | jq '.result?';
printf "CONNECTION COUNT:\t";
curl -s --data-binary '{"jsonrpc": "2.0", "id":"documentation", "method": "getconnectioncount", "params": [] }' -H 'content-type: application/json' http://$1:3030/ | jq '.result?';
printf "BLOCK COUNT:\t\t";
curl -s --data-binary '{"jsonrpc": "2.0", "id":"documentation", "method": "getblockcount", "params": [] }' -H 'content-type: application/json' http://$1:3030/ | jq '.result?';
echo "BLOCKS MINED:";
curl -s --data-binary '{"jsonrpc": "2.0", "id":"documentation", "method": "getnodestats", "params": [] }' -H 'content-type: application/json' http://$1:3030/ | jq '.[].misc?.blocks_mined?';
echo "OVERALL:";
curl -s --data-binary '{"jsonrpc": "2.0", "id":"1", "method": "getnodestats" }' -H 'content-type:application/json' http://$1:3030/ | jq '.result?';
3. Sometimes its instant, sometimes it takes 30+ seconds to respond

## Expected Behavior

Expected to be instant

## Your Environment

-  1.3.16
-  1.24.3, 1.55.0
-  Ubuntu 20.04

zosorock:
I can confirm this :+1:

ljedrz:
It can happen when the node is currently aggressively syncing blocks - this can cause many storage-related requests to get queued, meaning the RPC is waiting its turn.

This behavior is currently expected, and we don't want to prioritize RPCs over network operations, but this can certainly be improved.

Cc https://github.com/AleoHQ/snarkOS/issues/1176, which is a related issue

[...]

PR #1191, opened 2021-10-01
Sqlite write-ahead-log mode

Protryon:
This PR sets sqlite to use WAL mode, to allow reads concurrent with writes and generally better performance.

This PR also adds logic to run `PRAGMA optimize;` every 15 minutes. It apparently benefits greatly from running periodically, according to sqlite docs.

### DIFF
--- a/storage/src/storage/sqlite/mod.rs
+++ b/storage/src/storage/sqlite/mod.rs
@@ -21,7 +21,10 @@ use rusqlite::{
     OpenFlags,
     ToSql,
 };
-use std::path::Path;
+use std::{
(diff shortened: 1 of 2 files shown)


ljedrz (approved):
LGTM :+1: 

Also, it seems that this should fix https://github.com/AleoHQ/snarkOS/issues/1184.

niklaslong (approved):

Protryon:
@ljedrz This is not the full fix for #1184. We need read-only handles in a follow-up PR, but this should improve DB performance.

Issue #1015, opened 2021-08-04
The collection of disconnected peers has a large memory overhead


ljedrz:
We have an `#[ignore]`d (it takes a while to run and is unlikely to detect issues unless heavy refactoring is done) test, `check_node_cleanup`, which checks if heap memory use increases due to many connections being established and severed, and it appears that it currently does:

![massif_leak](https://user-images.githubusercontent.com/3750347/128210528-1bea24f4-9e7b-422b-a852-a55df82637c5.png)

Most of the connection-related memory not being relinquished appears to be allocated by `im::nodes::hamt::Node<A>::insert` and `alloc::sync::Arc<T>::new_uninit` (which is also called by `im::nodes::hamt::Node<A>::insert`).

There also seem to be some "dangling" `Arc`s and/or `Sender`s, as avoiding the aforementioned calls doesn't remove the entire leak yet.

ljedrz:
note: the scale of the leak is not large, and its size appears to be more or less linear in relation to the number of connect+disconnect events.

ljedrz:
After further investigation, it seems that the collection of disconnected peers might just have a pretty severe overhead: even though a single `Peer` object is `208B` in size, the increase in heap memory use caused by all the connect+disconnect events divided by their number is more than `2KiB` (over 10x the size of a `Peer`).

ljedrz:
Removing the `bug` label, as it doesn't appear to be a snarkOS-specific issue, but rather a characteristic of an indirect dependency, [im](https://crates.io/crates/im).

ljedrz:
I double-checked that any number and kind of items stored using `im::HashMap` results in a fairly large memory overhead.

ljedrz:
The improved [node cleanup test](https://github.com/AleoHQ/snarkOS/pull/1022) confirms that we don't actually have any connection-related leaks - it's indeed the aforementioned overhead of the collection of disconnected peers.

ljedrz:
Relinquishing control of this issue, as it's not a priority item and it's a dependency issue.

ljedrz:
Cc https://github.com/AleoHQ/snarkOS/issues/1060

ljedrz:
We will probably want to either restrict the number of cached disconnected peers or keep them in the persistent storage at all times in order to avoid memory growing too much with time, especially for entities like beacons.

ljedrz:
Closing, as it's no longer applicable to testnet2.

PR #1192, opened 2021-10-01
Blocked syncing

Protryon:
This PR changes sync layer to only insert blocks into storage, not validate or attempt to commit them. After each sync round, a fast-forward is run. This eliminates high memory usage in testing, keep queue depth low for storage relatively, and should avoid putting too much pressure on the system. Lower sync pull rate is expected.

### DIFF
--- a/consensus/src/consensus/inner/commit.rs
+++ b/consensus/src/consensus/inner/commit.rs
@@ -158,20 +158,14 @@ impl ConsensusInner {
             BlockStatus::Uncommitted => (),
         }
 
-        let start = Instant::now();
-
         // 1. Verify that the block valid
         if !self.verify_block(block).await? {
             return Err(ConsensusError::InvalidBlock(hash.clone()));
         }
-
-        info!("verify took {} ms", start.elapsed().as_millis());
         let start = Instant::now();
 
         // 2. Insert/canonize block
         self.commit_block(hash, block).await?;
-
-        info!("commit took {} ms", start.elapsed().as_millis());
 
         // 3. Remove transactions from the mempool
         for transaction in block.transactions.iter() {
--- a/consensus/src/consensus/inner/commit.rs
+++ b/consensus/src/consensus/inner/commit.rs
@@ -14,7 +14,6 @@
 // You should have received a copy of the GNU General Public License
 // along with the snarkOS library. If not, see <https://www.gnu.org/licenses/>.
 
-
 use super::*;
 use crate::ledger::dummy::DummyLedger;
 use snarkos_metrics as metrics;
--- a/consensus/src/consensus/inner/commit.rs
+++ b/consensus/src/consensus/inner/commit.rs
@@ -14,7 +14,6 @@
 // You should have received a copy of the GNU General Public License
 // along with the snarkOS library. If not, see <https://www.gnu.org/licenses/>.
 
-use std::time::Instant;
 
 use super::*;
 use crate::ledger::dummy::DummyLedger;
--- a/consensus/src/consensus/inner/commit.rs
+++ b/consensus/src/consensus/inner/commit.rs
@@ -162,7 +162,6 @@ impl ConsensusInner {
         if !self.verify_block(block).await? {
             return Err(ConsensusError::InvalidBlock(hash.clone()));
         }
-        let start = Instant::now();
 
         // 2. Insert/canonize block
         self.commit_block(hash, block).await?;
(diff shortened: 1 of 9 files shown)


consensus/src/consensus/inner/commit.rs
```diff
@@ -14,6 +14,8 @@
 // You should have received a copy of the GNU General Public License
 // along with the snarkOS library. If not, see <https://www.gnu.org/licenses/>.
 
+use std::time::Instant;
```
ljedrz:
```suggestion
```

consensus/src/ledger/merkle.rs
```diff
@@ -134,7 +134,13 @@ impl<P: MerkleParameters> Ledger for MerkleLedger<P> {
...
-            || (self.serial_numbers.len() + new_serial_numbers_len).is_power_of_two()
+        jumps_power_of_two(self.commitments.len(), new_commitments_len)
+            || jumps_power_of_two(self.serial_numbers.len(), new_serial_numbers_len)
     }
 }
+
+fn jumps_power_of_two(start: usize, adding: usize) -> bool {
+    let prior_commitments_depth = (start as f64).log2() as usize;
+    let new_commitments_depth = ((start + adding) as f64).log2() as usize;
+    prior_commitments_depth != new_commitments_depth
```
ljedrz: nit: this function's name is general and it's applied to serial numbers too, so it would be most readable if the variable names didn't reference commitments, but rather have names like `prior_value`, `final_value` etc.

ljedrz:
The `snarkos-testing` CI job fails, and it doesn't seem to be a fluke (a rerun didn't help). The commits could be squashed too.

Other than that, I'm very much in favor of this change :+1:.

ljedrz:
This should fix https://github.com/AleoHQ/snarkOS/issues/1176.

niklaslong:
I'm still seeing some high memory usage but it's definitely lower than it was. Nodes don't need to be restarted as often to alleviate the issue.

Protryon:
@niklaslong I've observed in a node running for ~96 hours that it's been stable at 4.2 GB RSS for 25 min peers.

consensus/src/consensus/inner/commit.rs
Protryon: `Instant` is used here.

niklaslong:
My long running node (started 17 hours ago) is showing 8GB currently, though the min peers is at 50. Perhaps I'm seeing https://github.com/AleoHQ/snarkOS/issues/1015.

Protryon:
> My long running node (started 17 hours ago) is showing 8GB currently, though the min peers is at 50. Perhaps I'm seeing #1015.

Is the node syncing? Mine has been at canon tip for a few days.

niklaslong:
That it is 👍

network/src/sync/blocks/interface.rs
```diff
@@ -143,16 +143,21 @@ impl Node {
...
+                    self.terminator.store(true, Ordering::SeqCst);
+                }
 
-            // This is a non-sync Block, send it to our peers.
-            self.propagate_block(block, height, remote_address);
+                // This is a non-sync Block, send it to our peers.
+                self.propagate_block(block, height, remote_address);
+            }
+        } else {
+            if let Err(e) = self.expect_sync().consensus.shallow_receive_block(block_struct).await {
```
niklaslong: Would be good to address this lint as well.

testing/src/network/sync.rs
```diff
@@ -105,7 +105,7 @@ async fn block_initiator_side() {
 
     // check the blocks have been added to the node's chain
     wait_until!(
-        5,
+        65,
```
niklaslong, line 109: Might be good to comment on the test that it is expected to be long running because of the way sync currently works?

consensus/src/consensus/inner/commit.rs
ljedrz: it is, but via `std::time::Instant::now()`, so either this import is unnecessary (and it's the CI that complained about it), or its use in line 153 needs to be adjusted to avoid the warning

ljedrz (approved):
LGTM, preferably with the import warning removed and the commits squashed

niklaslong (commented):
There are a couple of lints to address but otherwise this is looking good!

ljedrz:
I filed https://github.com/AleoHQ/snarkOS/issues/1197 for an issue that might be introduced with this branch.

PR #1198, opened 2021-10-05
Non-blocking block serialization

ljedrz:
Profiling indicates that block serialization for the purposes of sync provision can block the async executor even for several seconds. The solution is to use a blocking task to serialize all the blocks during that operation.


niklaslong (approved):
LGTM 👍

Protryon (approved):

PR #1200, opened 2021-10-05
Internal state check for block state

### LINKED ISSUE #1197: [Bug] The node can attempt to insert a block that already exists in the db
## 🐛 Bug Report (branch:  `blocked-syncing`)

When my node was syncing blocks, I saw the following in the logs (happened a few times in a row):
```
Oct 05 11:46:06.453 DEBUG failed receiving sync block: UNIQUE constraint failed: blocks.hash

Caused by:
    Error code 2067: A UNIQUE constraint failed
Oct 05 11:46:06.453 DEBUG failed receiving sync block: UNIQUE constraint failed: blocks.hash

Caused by:
    Error code 2067: A UNIQUE constraint failed
```
Suggesting that we are attempting to insert blocks with duplicate hashes somewhere and only the database stops us; this should either be avoided with a check before an insert is attempted, or the log should be more friendly (i.e. the error should be handled more elegantly).

Protryon:
This PR fixes #1197. It also is a candidate fix for similar user reported errors.

Root issue is that storage only guarantees consensus within a specific storage call, not between storage calls. We moved the block state check within insert state to provide a better error message and avoid consistency issues.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2888,6 +2888,7 @@ dependencies = [
  "hash_hasher",
  "hex",
  "indexmap",
+ "metrics",
  "num_enum",
  "parking_lot",
  "rand 0.8.4",
--- a/consensus/src/consensus/inner/agent.rs
+++ b/consensus/src/consensus/inner/agent.rs
@@ -67,14 +67,11 @@ impl ConsensusInner {
                     }
                     Err(e) => {
                         match e {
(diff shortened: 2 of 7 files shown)


ljedrz (approved):
LGTM :+1: 

nit: the `BlockStatus` import needs to be removed from `consensus/src/consensus/mod.rs` now

niklaslong (approved):

PR #1201, opened 2021-10-05
Recommit blocks

Protryon:
This PR:
* Introduces the notion of recommitting, which is asserting that if a block was valid in the past, it's always going to be valid if part of canon. We can avoid revalidating, and if there is a chain of recommittable blocks, then we can skip intermediate hashing by using past-calculated ledger digests.
* Introduces an optimization for recommitting large chains of blocks.

The motivation behind this PR is to resolve an observed phenomenon when syncing at ~400k blocks, where there are 2 valid chains seen based on block 43k competing. This PR should help eliminate that fork by allowing peers to not be stuck recommitting for hours each time they sync a few blocks from one of these forks, and it overtakes the other.

Has been running locally for almost a week, been keeping at canon tip much more reliably.

### DIFF
--- a/consensus/src/consensus/inner/commit.rs
+++ b/consensus/src/consensus/inner/commit.rs
@@ -107,12 +107,12 @@ impl ConsensusInner {
 
                             self.storage.recommit_blockchain(&fork_path.path[0]).await?;
                             let committed_blocks =
-                                self.storage.canon().await?.block_height as u32 - canon_branch_number;
+                                self.storage.canon().await?.block_height - canon_branch_number as usize;
                             if committed_blocks > 0 && self.recommit_taint.is_none() {
                                 self.recommit_taint = Some(canon_branch_number);
                             }
 
-                            for block_hash in &fork_path.path[committed_blocks as usize..] {
+                            for block_hash in &fork_path.path[committed_blocks.min(fork_path.path.len())..] {
                                 if block_hash == hash {
                                     self.verify_and_commit_block(hash, block).await?;
                                 } else {
--- a/consensus/src/consensus/inner/commit.rs
+++ b/consensus/src/consensus/inner/commit.rs
@@ -150,7 +150,7 @@ impl ConsensusInner {
     ) -> Result<(), ConsensusError> {
         let now = std::time::Instant::now();
 
-        match self.storage.get_block_state(hash).await? {
+        match self.recommit_block(hash).await? {
             BlockStatus::Committed(_) => return Ok(()),
             BlockStatus::Unknown => return Err(ConsensusError::InvalidBlock(hash.clone())),
             BlockStatus::Uncommitted => (),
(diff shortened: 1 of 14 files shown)


consensus/src/consensus/inner/commit.rs
```diff
@@ -105,11 +105,18 @@ impl ConsensusInner {
...
-                                if &block_hash == hash {
-                                    self.verify_and_commit_block(hash, block).await?
+                            self.storage.recommit_blockchain(&fork_path.path[0]).await?;
+                            let committed_blocks =
+                                self.storage.canon().await?.block_height as u32 - canon_branch_number;
+                            if committed_blocks > 0 && self.recommit_taint.is_none() {
+                                self.recommit_taint = Some(canon_branch_number);
+                            }
+
+                            for block_hash in &fork_path.path[committed_blocks as usize..] {
```
ljedrz: are we certain `committed_blocks` is always within bounds? otherwise, `.iter().skipcommitted_blocks as usize)` could be used

consensus/src/consensus/inner/mod.rs
```diff
@@ -72,6 +73,8 @@ impl ConsensusInner {
             return Ok(None);
         }
 
+        //todo: partial ledger fix
```
ljedrz: just to check: this is a future TODO, and not something left to do in this PR?

storage/src/storage/sqlite/sync.rs
```diff
@@ -596,48 +651,39 @@ impl SyncStorage for SqliteStorage {
...
+        for (hash, parent_hash, tree_depth) in out.into_iter().rev() {
+            if current_tree_depth.is_none() {
+                current_tree_depth = Some(tree_depth);
+            } else if Some(tree_depth) != current_tree_depth {
+                current_tree_depth = Some(tree_depth);
+
+                past_leaves.clear();
+                std::mem::swap(&mut past_leaves, &mut pending_leaves);
+            }
+            // pending_leaves.entry(parent_hash).and_modify(|inner| inner.push_child(hash)).or_insert_with(|| DigestTree::Leaf(hash));
```
ljedrz:
```suggestion
```

ljedrz:
This should fix https://github.com/AleoHQ/snarkOS/issues/1189.

consensus/src/consensus/inner/mod.rs
Protryon: already dealt with, will remove

consensus/src/consensus/inner/commit.rs
Protryon: added a check, but yes.

ljedrz (approved):
Ok, I think I've tested this change thoroughly enough; I've gone back to lower heights a few times now, getting back to the tip a while later, and while it did take a while (presumably due to it being the first time for my node), it's a definite improvement over the current state.

A tweak or two might be needed to avoid some blocking that I've noticed with higher peer counts, but it's nothing serious and can be done later on.

ljedrz:
One note regarding new nodes: I don't know if it would happen on its own if I waited long enough, but I had to restart mine manually in order to progress above height 128 (at which point the fast forwarding) did the trick.
That being said, it's still an improvement over certain memory exhaustion in a very long sync process.

consensus/src/consensus/inner/mod.rs
```diff
@@ -55,6 +55,7 @@ pub struct ConsensusInner {
     pub ledger: DynLedger,
     pub memory_pool: MemoryPool,
     pub storage: DynStorage,
+    pub recommit_taint: Option<u32>, // height of first recommitted block
```
niklaslong, line 58: How about:
```suggestion
    pub first_recommitted_height Option<u32>,
```

niklaslong:
Similarly to @ljedrz, I've noticed my node going from e.g. 440k to 77k a few times but it seems to be better than it was.

consensus/src/consensus/inner/mod.rs
Protryon: doesn't convey the notion of a tainted state sufficiently, would still require a comment imo.

Protryon:
A node going from 440k to 77k or similar is a current issue, expected, and correct. This PR just makes it not take ~24 hours to catch back up to canon tip.

PR #1204, opened 2021-10-06
Refresh peers even when client node is at capacity

niklaslong:
A node can often get stuck in a cluster of nodes that will block its sync progress. Refreshing a few peers when the node is at capacity aims to alleviate this.

More testing is needed, hence the draft. We may also prefer reconnecting to beacons periodically to refresh the list.

### DIFF
--- a/network/src/peers/peers.rs
+++ b/network/src/peers/peers.rs
@@ -84,7 +84,7 @@ impl Node {
                 0,
             )
         } else {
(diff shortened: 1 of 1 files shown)


ljedrz:
Closing the PR as it's either no longer applicable to testnet2 or would require a rewrite (in which case, feel free to reopen once rebased).

PR #1206, opened 2021-10-06
Fix fork point check

Protryon:
This PR changes the point for which to scan for forks for reversion to the latest received block in a prospective fork, as opposed to the parent hash.

This prevents us from reverting and ending up on the same chain that we are on in fresh-fork scenarios off old canon blocks.

This also indicates some nodes that should be syncing are mining -- which is likely due to the race condition present in node status.

### DIFF
--- a/consensus/src/consensus/inner/commit.rs
+++ b/consensus/src/consensus/inner/commit.rs
@@ -56,10 +56,7 @@ impl ConsensusInner {
                 return Ok(());
             }
             _ => {
-                let fork_path = self
-                    .storage
-                    .get_fork_path(hash, crate::OLDEST_FORK_THRESHOLD)
-                    .await?;
+                let fork_path = self.storage.get_fork_path(hash, crate::OLDEST_FORK_THRESHOLD).await?;
                 match fork_path {
                     ForkDescription::Path(fork_path) => {
                         let new_block_number = fork_path.base_index + fork_path.path.len() as u32;
--- a/consensus/src/consensus/inner/commit.rs
+++ b/consensus/src/consensus/inner/commit.rs
@@ -58,7 +58,7 @@ impl ConsensusInner {
             _ => {
                 let fork_path = self
                     .storage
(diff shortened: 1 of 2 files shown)


ljedrz:
This change breaks the `fork_out_of_order` test, which suggests the node is now less likely to fork to a valid chain too; or is there an issue with the logic of that test?

Protryon:
> This change breaks the `fork_out_of_order` test, which suggests the node is now less likely to fork to a valid chain too; or is there an issue with the logic of that test?

Test was out of date, fixed.

ljedrz (approved):
Looks reasonable to me; so this means that a competing fork will not be able to overtake the chain currently considered to be canon at the same height.

It would be best to also adjust the test's now-outdated comment, which states:
```
// Receive blocks from a sidechain (out of order) that overtakes our current canonical chain.
```

Protryon:
> Looks reasonable to me; so this means that a competing fork will not be able to overtake the chain currently considered to be canon at the same height.
> 
> It would be best to also adjust the test's now-outdated comment, which states:
> 
> ```
> // Receive blocks from a sidechain (out of order) that overtakes our current canonical chain.
> ```

Comment is still accurate

ljedrz:
That comment applies to what happens in the meantime, but doesn't reflect the end result of the test; that being said, the comments within the test could also use some improvement, especially `4.`, which is invalid and has been probably since the beginning. The sections of the test now do the following:

`// 1. Receive irrelevant block.`:

- `block_2_canon` is read
- block height doesn't change, as there's no `block_1_canon`

`// 2. Receive valid sidechain block`:

- `block_1_side` is read
- block height changes, as `block_1_side` applies on top of genesis

`// 3. Receive valid canon block 1 and accept the previous irrelevant block 2`:

- `block_1_canon` is read
- block height changes, as `block_1_side` gets reverted, and `block_1_canon` and `block_2_canon` form a canon chain now

`// 4. Receive valid canon block 1 and accept the previous irrelevant block 2`:

- `block_2_side` is read
- block height doesn't change, as the side chain doesn't become the new canon chain

PR #1208, opened 2021-10-07
Enforce a timeout to speed up peer list collection

### LINKED ISSUE #1207: [Bug] Gathering a peer list for a GetPeers reply can take a long time
## 🐛 Bug Report

I noticed that with slightly higher peer counts (possibly with defaults too, just to a lesser extent), the node can take a long time to collect a list of peers; the internal RTT for that request can go well into seconds (while other requests are handled within a reasonable time frame). Below is an extra time calculation I included just in the body of the function collecting the list of peers:
```
Oct 07 11:24:31.759  WARN collected peers in 30562ms
Oct 07 11:24:31.759  WARN collected peers in 19374ms
Oct 07 11:24:31.759  WARN collected peers in 13831ms
Oct 07 11:24:31.759  WARN collected peers in 28938ms
```

ljedrz:
This PR introduces a deadline to the collection of peer data for the purpose of answering `GetPeers` requests; it will cause the lists of peers to be shorter at times, but it's preferable to not serving them for many seconds.

~It's a draft, as I'm still testing it, since I'm seeing some peer event spikes from time to time, and I'm wondering if it's related.~

Fixes https://github.com/AleoHQ/snarkOS/issues/1207.

### DIFF
--- a/network/src/peers/peer_book.rs
+++ b/network/src/peers/peer_book.rs
@@ -21,12 +21,12 @@ use std::{
         atomic::{AtomicU32, Ordering},
         Arc,
     },
-    time::Instant,
+    time::{Duration, Instant},
 };
 
 use mpmc_map::MpmcMap;
 use snarkos_storage::DynStorage;
-use tokio::net::TcpStream;
+use tokio::{net::TcpStream, time::timeout};
 
 use snarkos_metrics::{self as metrics, connections::*, wrapped_mpsc};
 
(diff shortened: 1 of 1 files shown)


ljedrz:
I was able to confirm that this change doesn't regress the speed at which the queues are processed, and the max `GetPeers` RTT for 50 peers is down from 10s to 10ms (which still isn't ideal, but is a clear improvement; I'm not sure if I should reduce the deadline even further).

network/src/peers/peer_book.rs
```diff
@@ -241,7 +241,16 @@ impl PeerBook {
     }
 
     pub async fn connected_peers_snapshot(&self) -> Vec<Peer> {
-        self.map_each_peer(|peer| async move { peer.load().await }).await
+        let mut futures = Vec::with_capacity(self.connected_peers.len());
+        for (_, peer) in self.connected_peers.inner().iter() {
+            let peer = peer.clone();
+            futures.push(async move { timeout(Duration::from_millis(5), peer.load()).await });
+        }
+        futures::future::join_all(futures)
```
niklaslong, line 249: Is there a difference between this and tokio's `join!` (beyond not needing to allocate a `Vec` for the latter)?

niklaslong (approved):

network/src/peers/peer_book.rs
ljedrz: `tokio` doesn't have a direct equivalent yet, though it's planned for the future; it's probably possible to use it with some extra macro trickery right now, but since it doesn't seem to be causing any issues, I left it as-is.

PR #1212, opened 2021-10-08
Fix some dangling metrics

### LINKED ISSUE #1210: [Bug] Queue-related metrics can "dangle"
## 🐛 Bug Report

The queue-related metrics can accumulate values over time (viewed with both `prometheus` and RPC), which means they aren't always subtracted correctly, presumably when the related channel goes out of scope.

ljedrz:
The `wrapped_mpsc` helper objects, `Sender` and `Receiver`, automatically modify the related queue metrics, and they share a related atomic. `Sender::send` eagerly bumps the metric in order to show that the queue utilization grows, even if it's at its capacity (which is a desirable feature).

However, it's possible for this operation to happen after the `Receiver` side had already been dropped, which means we can have dangling values that never get decremented.

This PR ensures that the `Sender`-side of the `wrapped_mpsc` notices if it had incremented a metric after the `Receiver` had already been dropped, and decrements it.

Fixes https://github.com/AleoHQ/snarkOS/issues/1210.

### DIFF
--- a/metrics/src/wrapped_mpsc.rs
+++ b/metrics/src/wrapped_mpsc.rs
@@ -137,7 +137,9 @@ pub fn channel<T: Send>(metrics_tracker: &'static str, buffer: usize) -> (Sender
 
 impl<T: Send> Drop for Receiver<T> {
     fn drop(&mut self) {
-        let count = self.tracker.swap(0, Ordering::SeqCst) as f64;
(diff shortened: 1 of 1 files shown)


niklaslong (approved):
lgtm 👍 
(haven't tested locally yet)

PR #1217, opened 2021-10-11
Isolate a snarkos-rpc-types crate

ljedrz:
This PR implements the proposal from https://github.com/AleoHQ/snarkOS/issues/438. Its reasoning makes sense, and we might want to expose just the RPC types to external crates.

The one big dependency left is `snarkos-network`, though we probably do want to reuse its objects as much as possible; we might instead want to isolate a `snarkos-network-types` crate for alternative/thinner node implementations in the future.

Picking @howardwu as the reviewer, as it's a design call.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2864,6 +2864,7 @@ dependencies = [
  "snarkos-consensus",
  "snarkos-metrics",
  "snarkos-network",
+ "snarkos-rpc-types",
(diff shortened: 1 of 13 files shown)


ljedrz:
Closing the PR as it's no longer applicable to testnet2. It will probably be filed again against the current branch in the future.

PR #1218, opened 2021-10-11
Add an automatic port forwarding feature, discover node's external address

### LINKED ISSUE #423: [Feature] Use UPnP to open ports
## Motivation

snarkOS requires at least one port to be open and routed to the node PC: 4131. UPnP allows applications to programatically request the router to perform port forwarding.

## Implementation

Perhaps it's possible to use libraries such as https://github.com/sbstp/rust-igd/blob/master/examples/add_port.rs to implement it.

ljedrz:
This PR takes advantage of the gateway's UPnP (if available) in order to enable:
- automatic port forwarding (via `--auto-port-forwarding`; disabled by default)
- the discovery of the node's external IP address

Closes https://github.com/AleoHQ/snarkOS/issues/423.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -106,6 +106,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "attohttpc"
+version = "0.16.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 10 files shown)


niklaslong (approved):
LGTM 👍

Issue #1156, opened 2021-09-16
[Feature] cap data transmission


niklaslong:
This has been discussed previously but never tracked as an issue. SnarkOS currently doesn't cap data transmission, which could be a concern as the node can be easily flooded. I'd also suggest more in depth fuzzing to supplement our current coverage.

vicsn:
Doing a cleanup, closing in favour of more targeted issues in the future <3

PR #1219, opened 2021-10-12
Decouple sync state from peer quality

niklaslong:
This PR endeavours to begin decoupling the sync components from the peering components in the network crate. This is also in preparation of the data transmission caps (#1156). 

This PR also introduces some more documentation and small refactors to improve the safety and readability of the code.


network/src/peers/peer/peer.rs
```diff
@@ -23,16 +23,23 @@ use std::{
...
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct Peer {
+    /// The address of the node's listener socket.
     pub address: SocketAddr,
+    /// The latest broadcast block height of the peer.
+    pub block_height: u32,
+    /// Quantifies the node's connection quality with the peer.
     pub quality: PeerQuality,
+    /// Tracks the node's sync state with the peer.
+    pub sync_state: SyncState,
```
ljedrz, line 39: I'm wondering if we should be serializing this object (and keeping it in the storage), as sync-related numbers are unlikely to be practical once the node disconnects for a while :thinking:... then again, if it's involved in some malicious spamming, those numbers might be immediately suspicious as soon as we load the peer. For now, we can leave it as-is, I guess.

network/src/peers/peer/peer_quality.rs
```diff
@@ -21,49 +21,82 @@ use std::time::Instant;
...
+    }
+
+    pub fn reset(&mut self) {
+        self.set(0);
+    }
+
+    pub fn set(&mut self, amount: u32) {
+        self.remaining_sync_blocks = amount;
+        self.total_sync_blocks = amount;
+    }
```
ljedrz: I don't see `set` used anywhere, so perhaps just `reset` could be left?

network/src/peers/peer/peer.rs
```diff
@@ -175,6 +190,78 @@ impl Peer {
...
+
+        // No need to cover the negative case since we saturate the decrement.
+        if remaining_sync_blocks == 0 {
+            trace!("received unexpected or late sync block from {}", self.address);
+
+            return;
+        }
+
+        // Decrement if > 0.
+        self.sync_state.remaining_sync_blocks = remaining_sync_blocks.saturating_sub(1);
```
ljedrz:
```suggestion
        self.sync_state.remaining_sync_blocks -= 1
```
this is fine, as we are already certain the value is greater than `0`

network/src/peers/peer/peer.rs
```diff
@@ -42,13 +49,21 @@ const FAILURE_THRESHOLD: usize = 5;
 
 impl Peer {
     pub fn new(address: SocketAddr, data: Option<&snarkos_storage::Peer>) -> Self {
+        let mut block_height = Default::default();
```
ljedrz:
```suggestion
        let mut block_height = 0;
```
this should be more self-explanatory for a plain number

network/src/peers/peer/peer.rs
```diff
@@ -57,14 +72,14 @@ impl Peer {
     pub fn serialize(&self) -> snarkos_storage::Peer {
```
ljedrz, line 70: oh, there's a manual serialization impl; in this case maybe we can remove the `Serialize` (and probably `Deserialize`) proc macro impls from the `Peer` object?

network/src/peers/peer/peer.rs
```diff
@@ -175,6 +190,78 @@ impl Peer {
     }
 
     pub(super) fn set_disconnected(&mut self) {
+        self.sync_state.reset();
         self.quality.disconnected();
     }
+
+    pub fn start_rtt_measurement(&mut self) {
```
ljedrz:
```suggestion
    pub fn start_ping_measurement(&mut self) {
```
since we calculate multiple RTTs

network/src/peers/peer/peer.rs
```diff
@@ -175,6 +190,78 @@ impl Peer {
...
+        self.quality.expecting_pong = true;
+        self.quality.last_ping_sent = Some(Instant::now());
+    }
+
+    pub fn stop_rtt_measurement(&mut self) {
```
ljedrz:
```suggestion
    pub fn stop_ping_measurement(&mut self) {
```
same rationale as above

ljedrz (approved):
LGTM with some comments

network/src/peers/peer/peer_quality.rs
niklaslong: Indeed, it used to be common behaviour but no longer is since we accumulate the counts instead of setting them now.

Protryon (approved):

PR #1221, opened 2021-10-12
Beacon and sync provider rollout

niklaslong:
This PR updates the IPs of the beacons and sync provider lists. It also includes a change to make sure sync providers are always loosely connected (just as the beacons are currently and the bootnodes were previously). 

Drafted as the IPs still need to be added to the config.


ljedrz (approved):
LGTM :+1:

PR #1224, opened 2021-10-13
Fix the orphan block metric 

### LINKED ISSUE #1215: [Bug] The orphan block metric is always at 0
## 🐛 Bug Report

This metric's functionality initially overlapped with the one for duplicate blocks (once the block cache was introduced), but it's currently not even incremented anywhere (it presumably got lost during some major refactoring). It either needs to be restored or removed now.

niklaslong:
As per #1215, restores the orphan block metric.


ljedrz (approved):
LGTM :ok_hand:

PR #1231, opened 2021-10-20
Fix sync & trim

Protryon:
This PR:
* Fixes trim behavior to not remove transactions from canon blocks if linked to a non-canon block
* Adds logic to decommit blocks on validate calls when there are invalid blocks in canon (to fix databases from above issue)
* Fixes infinite sync cycle requiring reboots or network timeouts to die.
* Fixes consensus not removing invalid blocks causing sync stalls.


ljedrz (approved):
LGTM 👍

PR #1238, opened 2021-10-25
Calculate the internal mempool RTT even if no response is sent

### LINKED ISSUE #1237: [Bug] The getmemorypool internal RTT metric doesn't change
## 🐛 Bug Report

I'm guessing that it's just not updated in the event that no response is sent (due to the responder's mempool being empty).

ljedrz:
As anticipated, the fix is just an extra update in the event no response is sent - we still want to know how long it takes to obtain the (empty) result.


niklaslong (approved):
LGTM 👍

PR #1241, opened 2021-10-25
Remove the duplicate syncblocks metric

### LINKED ISSUE #1240: The duplicate syncblocks metric is unused
We never increment the `blocks::DUPLICATES_SYNC` metric, we should either remove it if it is not needed or add it back in.

niklaslong:
Removes the duplicate syncblocks metric as it is unused and doesn't provide any added value currently. Re-propagated duplicates are the prime reason for the cache existing in the first place, syncblocks are requested by hash. 

Closes #1240.

### DIFF
--- a/metrics/grafana/dashboard.json
+++ b/metrics/grafana/dashboard.json
@@ -212,14 +212,6 @@
         },
         {
           "exemplar": true,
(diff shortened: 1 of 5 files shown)


ljedrz (approved):
LGTM :+1:

PR #1247, opened 2021-10-28
Speed up indexed maps

ljedrz:
The node does a lot of work handling its Merkle trees; we can speed those operation up by reducing the amount of hashing needed when using already-hashed digests.

In addition, the same can be done to the memory pool.


niklaslong (approved):

ljedrz:
Closing the PR as it's no longer applicable to testnet2.

PR #1253, opened 2021-11-03
Fix missing fmt argument in error message

dtolnay:
## Motivation

Without this, the error message would literally be "invalid length for u32: {}" with curly braces in it instead of the input len.

## Test Plan

`cd storage`
`cargo check`


niklaslong:
Thank you for catching and fixing this!

niklaslong (approved):
Noncritical, let's wait until the testnet 2 changes are in to merge.

ljedrz:
Closing the PR as it's either no longer applicable to testnet2 (I checked both snarkOS and snarkVM).

PR #1285, opened 2021-11-15
Move network objects to their own crate

ljedrz:
This PR moves network objects to their own crate in order to facilitate testing and provide a leaner dependency for any alternative node implementations.

I would later like to do the same for the RPC types. The name of the extra crate can be changed to `snarkos-network-objects` or anything else if you'd prefer.


niklaslong (approved):

ljedrz:
Rebased.

ljedrz:
It would be best if `snarkos_ledger::BlockLocators` could also be moved to its own crate; right now this seems to be the last object that incurs a heavy dependency for any external use.

The end goal here is to just have the objects that are not strictly bound to the node's internals to be exposed to the outside world without all the full node's dependencies; this could even not be use-case-specific, and all such objects could be put under a common crate called e.g. `snarkos-external-types` (which would later also apply to RPC).

ljedrz:
Closing in favor of https://github.com/AleoHQ/snarkOS/pull/1286, which contains this branch.

PR #1286, opened 2021-11-15
Introduce an integration testing framework

ljedrz:
Builds on https://github.com/AleoHQ/snarkOS/pull/1285, including further changes outlined in https://github.com/AleoHQ/snarkOS/pull/1285#issuecomment-968844926.

This PR proposes a basic framework for integration tests (concentrating on the network first), including a thin/synthetic node implementation the full snarkOS node can be tested with/against.

Even though the synthetic node already works and is able to communicate with the full snarkOS node, there are still some wrinkles around the test and node setup that need to be ironed out, so not all of the tests are enabled yet. It would be best to agree on the overall design before proceeding.

### DIFF
--- a/src/network/server.rs
+++ b/src/network/server.rs
@@ -20,6 +20,7 @@ use crate::{
     peers::{Peers, PeersRequest, PeersRouter},
     rpc::initialize_rpc_server,
     Environment,
+    Node,
     NodeType,
 };
 use snarkos_ledger::{storage::rocksdb::RocksDB, LedgerState};
@@ -36,6 +37,7 @@ use tokio::{
 ///
 /// A set of operations to initialize the node server for a specific network.
 ///
+#[derive(Clone)]
 pub(crate) struct Server<N: Network, E: Environment> {
     /// The list of peers for the node.
     peers: Arc<RwLock<Peers<N, E>>>,
@@ -54,29 +56,35 @@ impl<N: Network, E: Environment> Server<N, E> {
     /// Starts the connection listener for peers.
     ///
     #[inline]
-    pub(crate) async fn initialize(
-        node_port: u16,
-        rpc_port: u16,
-        username: String,
-        password: String,
-        miner: Option<Address<N>>,
-    ) -> Result<Self> {
+    pub(crate) async fn initialize(node: &Node, miner: Option<Address<N>>, mut tasks: Tasks<task::JoinHandle<()>>) -> Result<Self> {
+        let node_ip = node.ip.clone();
+        let node_port = node.node.unwrap_or(E::DEFAULT_NODE_PORT);
+        let rpc_ip = node.rpc_ip.clone();
+        let rpc_port = node.rpc.unwrap_or(E::DEFAULT_RPC_PORT);
+
+        #[cfg(not(feature = "test"))]
+        assert!(
+            !(node_port < 4130),
+            "Until configuration files are established, the node port must be at least 4130 or greater"
+        );
+
         // Initialize a new TCP listener at the given IP.
-        let (local_ip, listener) = match TcpListener::bind(&format!("0.0.0.0:{}", node_port)).await {
+        let (local_ip, listener) = match TcpListener::bind(&format!("{}:{}", node_ip, node_port)).await {
             Ok(listener) => (listener.local_addr().expect("Failed to fetch the local IP"), listener),
             Err(error) => panic!("Failed to bind listener: {:?}. Check if another Aleo node is running", error),
         };
 
         // Initialize the ledger storage path.
-        let storage_path = format!(".ledger-{}", (node_port as u16 - 4130) as u8);
-
-        // Initialize the tasks handler.
-        let mut tasks = Tasks::new();
+        #[cfg(not(feature = "test"))]
+        let storage_path = format!(".ledger-{}", (node_port - 4130));
+        #[cfg(feature = "test")]
+        let storage_path = format!(".ledger-{}", node_port);
 
         // Initialize a new instance for managing peers.
         let (peers, peers_router) = Self::initialize_peers(&mut tasks, local_ip);
         // Initialize a new instance for managing the ledger.
-        let (ledger, ledger_router) = Self::initialize_ledger(&mut tasks, &storage_path, &peers_router)?;
+        let path = storage_path.clone();
+        let (ledger, ledger_router) = Self::initialize_ledger(&mut tasks, path, &peers_router).await?;
 
         // Initialize the connection listener for new peers.
         Self::initialize_listener(&mut tasks, local_ip, listener, &peers_router, &ledger_router);
@@ -85,15 +93,20 @@ impl<N: Network, E: Environment> Server<N, E> {
         // Initialize a new instance of the miner.
         Self::initialize_miner(&mut tasks, local_ip, miner, &ledger_router);
 
-        // Initialize a new instance of the RPC server.
-        tasks.append(initialize_rpc_server::<N, E>(
-            format!("0.0.0.0:{}", rpc_port).parse()?,
-            username,
-            password,
-            &peers,
-            LedgerState::open::<RocksDB, _>(&storage_path, true)?,
-            &ledger_router,
-        ));
+        if !node.disable_rpc {
+            // Initialize a new instance of the RPC server.
+            tasks.append(initialize_rpc_server::<N, E>(
+                format!("{}:{}", rpc_ip, rpc_port).parse()?,
+                node.rpc_username.clone(),
+                node.rpc_password.clone(),
+                &peers,
+                LedgerState::open::<RocksDB, _>(&storage_path, true)?,
+                &ledger_router,
+            ));
+        }
+
+        // A small delay to allow all the tasks to get spawned.
+        tokio::time::sleep(std::time::Duration::from_millis(100)).await;
 
         Ok(Self {
             peers,
@@ -138,14 +151,16 @@ impl<N: Network, E: Environment> Server<N, E> {
         // Initialize the peers router process.
         let peers_clone = peers.clone();
         let peers_router_clone = peers_router.clone();
-        tasks.append(task::spawn(async move {
+
+        let task_handle = task::spawn(async move {
             // Asynchronously wait for a peers request.
             // Channel is routing a request to peers.
             while let Some(request) = peers_handler.recv().await {
                 // Hold the peers write lock briefly, to update the state of the peers.
                 peers_clone.write().await.update(request, &peers_router_clone).await;
             }
-        }));
+        });
+        tasks.append(task_handle);
 
         (peers, peers_router)
     }
@@ -155,13 +170,13 @@ impl<N: Network, E: Environment> Server<N, E> {
     ///
     #[inline]
     #[allow(clippy::type_complexity)]
-    fn initialize_ledger(
+    async fn initialize_ledger(
         tasks: &mut Tasks<task::JoinHandle<()>>,
-        storage_path: &str,
+        storage_path: String,
         peers_router: &PeersRouter<N, E>,
     ) -> Result<(Arc<RwLock<Ledger<N, E>>>, LedgerRouter<N, E>)> {
         // Open the ledger from storage.
-        let ledger = Ledger::<N, E>::open::<RocksDB, _>(storage_path)?;
+        let ledger = tokio::task::spawn_blocking(move || Ledger::<N, E>::open::<RocksDB, _>(&storage_path)).await??;
         let ledger = Arc::new(RwLock::new(ledger));
 
         // Initialize an mpsc channel for sending requests to the `Ledger` struct.
@@ -170,13 +185,15 @@ impl<N: Network, E: Environment> Server<N, E> {
         // Initialize the ledger router process.
         let peers_router = peers_router.clone();
         let ledger_clone = ledger.clone();
-        tasks.append(task::spawn(async move {
+
+        let task_handle = task::spawn(async move {
             // Asynchronously wait for a ledger request.
             while let Some(request) = ledger_handler.recv().await {
                 // Hold the ledger write lock briefly, to update the state of the ledger.
                 ledger_clone.write().await.update(request, &peers_router).await;
             }
-        }));
+        });
+        tasks.append(task_handle);
 
         Ok((ledger, ledger_router))
     }
@@ -194,7 +211,8 @@ impl<N: Network, E: Environment> Server<N, E> {
     ) {
         let peers_router = peers_router.clone();
         let ledger_router = ledger_router.clone();
-        tasks.append(task::spawn(async move {
+
+        let task_handle = task::spawn(async move {
             info!("Listening for peers at {}", local_ip);
             loop {
                 // Asynchronously wait for an inbound TcpStream.
@@ -211,7 +229,8 @@ impl<N: Network, E: Environment> Server<N, E> {
                 // Add a small delay to prevent overloading the network from handshakes.
                 tokio::time::sleep(Duration::from_millis(150)).await;
             }
-        }));
+        });
+        tasks.append(task_handle);
     }
 
     ///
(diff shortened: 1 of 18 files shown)


testing/Cargo.toml
```diff
@@ -0,0 +1,50 @@
...
+features = ["macros", "rt-multi-thread", "time"]
+
+[dependencies.structopt]
+version = "0.3"
+
+[dependencies.tracing]
+version = "0.1"
+
+[dependencies.tracing-subscriber]
+version = "0.2"
```
niklaslong: Why not `0.3` here?

testing/src/test_node.rs
```diff
@@ -0,0 +1,310 @@
...
+}
+
+/// snarkOS state required for test purposes.
+#[derive(Clone, Default)]
+struct State {
+    /// The list of known peers; `Pea2Pea` already has its own internal peer handling, but
+    /// snarkOS recognizes nodes by their listening port; this collection facilitates
+    /// the peering experience during tests and aligns it with snarkOS logic.
+    // FIXME: use the `Disconnect` protocol to remove disconnected peers
+    peers: Arc<Mutex<HashSet<SocketAddr>>>,
```
niklaslong: Does this mean pea2pea's internal state is ignored in this case?

testing/src/test_node.rs
```diff
@@ -0,0 +1,310 @@
...
+        // Respond with own challenge request.
+        let own_response = SnarkosMessage::ChallengeResponse(genesis_block_header.clone());
+        trace!(parent: self.node().span(), "sending a challenge response to {}", peer_addr);
+        let msg = own_response.serialize().unwrap();
+        let len = u32::to_le_bytes(msg.len() as u32);
+        conn.writer().write_all(&len).await?;
+        conn.writer().write_all(&msg).await?;
+
+        // Wait for the challenge response to come in.
+        conn.reader().read_exact(&mut buf[..4]).await?;
```
niklaslong: Might be good to have these lengths codefied as shared constants?

testing/src/test_node.rs
```diff
@@ -0,0 +1,310 @@
...
+    }
+}
+
+/// Inbound message processing logic for the test nodes.
+#[async_trait::async_trait]
+impl Reading for TestNode {
+    type Message = SnarkosMessage;
+
+    fn read_message<R: io::Read>(&self, source: SocketAddr, reader: &mut R) -> io::Result<Option<Self::Message>> {
+        let mut buf = [0u8; 8 * 1024];
```
niklaslong, line 260: Do we have a max message size already?

testing/src/test_node.rs
```diff
@@ -0,0 +1,310 @@
...
+            }
+            Err(e) => {
+                error!("a message from {} failed to deserialize: {}", source, e);
+                Err(io::ErrorKind::InvalidData.into())
+            }
+        }
+    }
+
+    async fn process_message(&self, source: SocketAddr, message: Self::Message) -> io::Result<()> {
+        match message {
```
niklaslong, line 282: Suggestion: move the blocks into functions as this set is likely to grow?

testing/tests/common/mod.rs
```diff
@@ -0,0 +1,43 @@
...
+pub fn start_logger() {
+    let filter = match EnvFilter::try_from_default_env() {
+        Ok(filter) => filter.add_directive("mio=off".parse().unwrap()),
+        _ => EnvFilter::default().add_directive("mio=off".parse().unwrap()),
+    };
+    tracing_subscriber::fmt().with_env_filter(filter).with_target(false).init();
+}
+
+#[macro_export]
+macro_rules! wait_until {
```
niklaslong, line 57: _Why hello there old friend, it's good to see you again._ 😏

testing/tests/network/manual_testing.rs
```diff
@@ -0,0 +1,52 @@
...
+/// This test is intended to be run manually in order to monitor the behavior of
+/// a full snarkOS node started independently.
+#[ignore]
+#[tokio::test]
+async fn spawn_inert_node_at_port() {
+    start_logger();
+
+    // FIXME: snarkOS currently expects every node to be on this port, making it difficult
+    // to run local tests with multiple nodes.
+    const PORT: u16 = 4135;
```
niklaslong, line 35: Does snarkOS currently assign a port if it isn't specified?

niklaslong (approved):
LGTM, left some suggestions 👍

ljedrz:
great points, @niklaslong

testing/Cargo.toml
ljedrz: some copy-paste malfunction, it would seem :stuck_out_tongue:; will fix :+1:

testing/src/test_node.rs
ljedrz: no, it isn't - `pea2pea` just doesn't forbid "two-sided" connections, which are not needed here; all the state in this object is on top of the internal networking state of `pea2pea` (which will still forbid duplicate connections, obvious self-connections etc.)

testing/src/test_node.rs
ljedrz: fair; I was just drafting it quickly :+1:

testing/src/test_node.rs
ljedrz: I haven't seen one - this is just an initial placeholder value, will add a `FIXME` :+1:

testing/src/test_node.rs
ljedrz: makes sense - will add a `TODO`

testing/tests/network/manual_testing.rs
ljedrz: `snarkOS` currently uses and expects `4132` (for testnet2); this is just a placeholder that I used for manual tests (which required minor tweaks to the node) until these rules are relaxed (at least for test purposes)

ljedrz:
Addressed review comments.

ljedrz:
Rebased.

ljedrz:
The new tests now pass when run individually; I detected some issue with task accounting, so I temporarily disabled it.

howardwu:
Can we keep block locators in its original place? Alternatively, I can move it to snarkVM so it is out of the way here.

src/node.rs
```diff
@@ -35,12 +41,21 @@ pub struct Node {
     /// Specify the network of this node.
     #[structopt(default_value = "2", short = "n", long = "network")]
     pub network: u16,
+    /// The listener IP of the node.
+    #[structopt(default_value = "0:0:0:0", long = "ip")]
+    pub ip: String,
```
howardwu: Why is this necessary?

src/node.rs
```diff
@@ -35,12 +41,21 @@ pub struct Node {
...
     /// Specify the port for the node server.
-    #[structopt(long = "node")]
-    pub node: Option<u16>,
+    #[structopt(long = "listener-port")]
+    pub listener_port: Option<u16>,
```
howardwu: Please change back to `node`

src/node.rs
```diff
@@ -35,12 +41,21 @@ pub struct Node {
...
+    /// Disable the RPC server.
+    #[structopt(long = "disable-rpc")]
+    pub disable_rpc: bool,
```
howardwu, line 58: This is very messy, let's take another design look.

src/node.rs
```diff
@@ -35,12 +41,21 @@ pub struct Node {
...
+    /// The IP address of the RPC server.
+    #[structopt(default_value = "0:0:0:0", long = "rpc-ip")]
+    pub rpc_ip: String,
```
howardwu: Why is this necessary?

src/node.rs
```diff
@@ -35,12 +41,21 @@ pub struct Node {
...
     /// Specify the port for the RPC server.
-    #[structopt(long = "rpc")]
-    pub rpc: Option<u16>,
+    #[structopt(long = "rpc-port")]
+    pub rpc_port: Option<u16>,
```
howardwu: Please change back to `rpc`

howardwu:
I would prefer `Environment` remain in a primary crate here. It is not "external" to anything here

src/node.rs
ljedrz: Two reasons: 
- users utilizing multiple networks might not want the node to be reachable on all of them
- we don't want test nodes to connect to the testnet, and this allows them to be `127.0.0.1`

src/node.rs
ljedrz: What is messy about it? Not spawning an RPC server in non-RPC tests makes them a lot faster, so it's a desirable feature. Alternatively I could disable that block directly using `#[cfg(not(feature = "test"))]`.

src/node.rs
ljedrz: See https://github.com/AleoHQ/snarkOS/pull/1286#discussion_r750972245.

ljedrz:
> I would prefer Environment remain in a primary crate here. It is not "external" to anything here

I wanted to propose this "split" early on (as it might become more difficult with time), but in general, it's a matter of being able to build outside node (e.g. a dedicated crawler) and RPC implementations faster, without having to build entire snarkOS. The term "external" refers to the fact that these are (except for RPC, which I left for later) objects facing "the external world", i.e. exposed to the network.

I can revert these changes for now, but this is a feature that had been proposed before and which provides clear benefits for the ecosystem, so I would recommend considering it for the future.

ljedrz:
Rebased and addressed most comments.

note: I checked the alternative solution to disabling the RPC server mentioned in https://github.com/AleoHQ/snarkOS/pull/1286#discussion_r750987265, but it was a lot messier - it required multiple imports and `const`s to obtain the extra attribute.

ljedrz:
Rebased and ready for re-review.

src/network/server.rs
```diff
@@ -54,29 +56,35 @@ impl<N: Network, E: Environment> Server<N, E> {
...
             Ok(listener) => (listener.local_addr().expect("Failed to fetch the local IP"), listener),
             Err(error) => panic!("Failed to bind listener: {:?}. Check if another Aleo node is running", error),
         };
 
         // Initialize the ledger storage path.
-        let storage_path = format!(".ledger-{}", (node_port as u16 - 4130) as u8);
-
-        // Initialize the tasks handler.
-        let mut tasks = Tasks::new();
+        #[cfg(not(feature = "test"))]
```
niklaslong, line 78: Might be good to comment on these few lines a little?

src/network/server.rs
```diff
@@ -54,29 +56,35 @@ impl<N: Network, E: Environment> Server<N, E> {
...
     #[inline]
-    pub(crate) async fn initialize(
-        node_port: u16,
-        rpc_port: u16,
-        username: String,
-        password: String,
-        miner: Option<Address<N>>,
-    ) -> Result<Self> {
+    pub(crate) async fn initialize(node: &Node, miner: Option<Address<N>>, mut tasks: Tasks<task::JoinHandle<()>>) -> Result<Self> {
+        let node_ip = node.ip.clone();
```
niklaslong: I'm assuming this is `Copy`?

src/network/server.rs
```diff
@@ -54,29 +56,35 @@ impl<N: Network, E: Environment> Server<N, E> {
...
+        let node_port = node.node.unwrap_or(E::DEFAULT_NODE_PORT);
+        let rpc_ip = node.rpc_ip.clone();
```
niklaslong: ditto ☝️

src/node.rs
niklaslong: Oh, I see this is a `String` (ignore my earlier `Copy` comments), why not use `IpAddr`?

src/node.rs
niklaslong: This should also be an `IpAddr`?

src/node.rs
```diff
@@ -105,25 +119,32 @@ impl Node {
...
-            std::future::pending::<()>().await;
-            Ok(())
         }
+
+        if let Some(peer_ip) = &self.connect {
+            server.connect_to(peer_ip.parse().unwrap()).await?;
+        }
+
+        let server_task = task::spawn(async move {
+            let _server = server;
```
niklaslong, line 142: I'm guessing this is to avoid prematurely dropping the `server`?

testing/src/snarkos_node.rs
```diff
@@ -0,0 +1,73 @@
...
+pub struct SnarkosNode {
+    pub addr: SocketAddr,
+}
+
+impl SnarkosNode {
+    pub async fn default() -> Self {
+        // Procure a free port number for the snarkOS node.
+        // FIXME: due to there being a delay between the port's discovery and its binding by the node, this
+        // method can cause an `AddrInUse` error to occur when multiple tests are run at the same time; only
+        // introspection of a ready node can fully avoid this.
```
niklaslong, line 35: I suppose we can run them with `--test-threads 1` in the meantime?

testing/src/snarkos_node.rs
```diff
@@ -0,0 +1,73 @@
...
+        let free_port = {
+            let temp_socket = TcpListener::bind("127.0.0.1:0").await.unwrap();
+            temp_socket.local_addr().unwrap().port()
+        };
+
+        // Start a snarkOS node with that port.
+        SnarkosNode::with_args(&["--node", &free_port.to_string()]).await
+    }
+
+    pub async fn with_args(extra_args: &[&str]) -> Self {
```
niklaslong, line 47: Some doc here might be good (let's start early 😉), especially as some args are permanent?

testing/tests/network/basic_connectivity.rs
```diff
@@ -0,0 +1,137 @@
...
+};
+
+#[tokio::test(flavor = "multi_thread")]
+async fn test_nodes_can_connect_to_each_other() {
+    // Start 2 test nodes.
+    let test_node0 = TestNode::default().await;
+    let test_node1 = TestNode::default().await;
+
+    // Ensure that the nodes have no active connections.
+    crate::wait_until!(1, test_node0.node().num_connected() == 0 && test_node1.node().num_connected() == 0);
```
niklaslong:
```suggestion
    wait_until!(1, test_node0.node().num_connected() == 0 && test_node1.node().num_connected() == 0);
```
It's imported up top ☝️

testing/tests/network/basic_connectivity.rs
```diff
@@ -0,0 +1,137 @@
...
+async fn handshake_as_initiator_works() {
+    // Start a test node.
+    let test_node = TestNode::default().await;
+
+    // Start a snarkOS node.
+    let test_node_addr = test_node.node().listening_addr().unwrap();
+    SnarkosNode::with_args(&["--node", "0", "--connect", &test_node_addr.to_string()]).await;
+
+    // The snarkOS node should have connected to the test node.
+    crate::wait_until!(5, test_node.node().num_connected() != 0);
```
niklaslong: ditto

testing/tests/network/basic_connectivity.rs
```diff
@@ -0,0 +1,137 @@
...
+    // The test node should be able to connect to the snarkOS node.
+    assert!(test_node.node().connect(snarkos_node.addr).await.is_ok());
+}
+
+#[tokio::test(flavor = "multi_thread")]
+#[ignore = "TODO: currently not possible to connect on demand"]
+async fn node_cant_connect_to_itself() {}
+
+#[tokio::test(flavor = "multi_thread")]
+#[ignore = "TODO: currently not possible to connect on demand"]
```
niklaslong, line 75: Nice 👍

testing/src/snarkos_node.rs
ljedrz: Some of those tests require the executor to be multi-threaded, though; these failures are quite rare.

src/node.rs
ljedrz: I didn't think `structopt` was able to interpret the default value "on the fly" like that, but it appears that it is :+1:.

ljedrz:
Applied review suggestions by @niklaslong.

ljedrz:
Superseded by https://github.com/AleoHQ/snarkOS/pull/1311.

PR #1295, opened 2021-11-16
Mine blocks in the context of a blocking task

ljedrz:
This PR moves the block mining operation into a dedicated blocking task, so that it doesn't block async worker threads. The associated ledger request is still processed asynchronously.

### DIFF
--- a/src/network/ledger.rs
+++ b/src/network/ledger.rs
@@ -248,7 +248,7 @@ impl<N: Network, E: Environment> Ledger<N, E> {
             }
             LedgerRequest::Mine(local_ip, recipient, ledger_router) => {
                 // Process the request to mine the next block.
-                self.mine_next_block(local_ip, recipient, ledger_router);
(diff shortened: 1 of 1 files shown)


niklaslong (approved):
LGTM 👍

howardwu:
Just to confirm, this will continue to spin up a new thread, without causing the ledger to wait on the miner, right?

And for my own learnings, what exactly does this operation do, that differs from the current implementation?

ljedrz:
`tokio` performs all the async operations using its worker threads, and those expect all these operations to be non-blocking. It also has a thread pool dedicated to blocking tasks, which do not affect the execution of async threads even when running blocking operations.

This change still allows the ledger to keep processing further requests (due to still spawning the outer regular `task`), but the mining operation now becomes non-blocking to the async executor due to being executed within a blocking task (which could use an already running blocking thread or spawn a new one).

ljedrz:
Rebased.

howardwu (approved):

PR #1296, opened 2021-11-16
Don't dedicate all the cores to mining

ljedrz:
The node needs to have some resources left to be able to handle network operations.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1775,6 +1775,7 @@ dependencies = [
  "parking_lot",
  "rand",
  "rand_chacha",
+ "rayon",
  "rusty-hook",
  "self_update",
  "serde",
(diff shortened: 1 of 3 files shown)


niklaslong (approved):

howardwu:
Does this slow down all Rayon operations?

ljedrz:
> Does this slow down all Rayon operations?

Yes; `rayon`'s thread pool shouldn't occupy all the cores, though, otherwise the performance of everything running concurrently and not executed within it can suffer. I currently set the number of cores dedicated to the thread pool to be 3/4 of the cores available, but a greater value could probably be used.

howardwu:
This will significantly slow down `Pong` messages, which require checking block headers are well-formed. Naively, it would be 6 seconds, with rayon it is 300-500ms.

I suggest we take a very different approach.

ljedrz:
I'll mark this PR as a draft while I check the implications for `Pong` messages.

ljedrz:
Closing, as this was already introduced elsewhere.

Issue #1290, opened 2021-11-16
[Bug] The read-only database is not catching up to the primary database correctly.


raychu86:
## 🐛 Bug Report

The snarkOS node runs a primary RocksDB database for standard node operations and a read-only `secondary` database for the RPC server. This `secondary` database is uses a `try_catch_up_with_primary` to update it's state. Despite this, the `secondary` seems to occasionally fall behind the primary database and not update properly. 

Initially, we called `try_catch_up_with_primary` every time an RPC endpoint was hit, however we noticed that a high frequency of these `try_catch_up_with_primary` calls jumps the `secondary` database state back and forth between new and old state. This can be seen in the `secondary` database latest sequence number, which was also updating, then reverting back to the original.

To get around the original issue, we have a loop that calls `try_catch_up_with_primary` every X seconds, and only updates the read-only state if the sequence number increases properly. However, this approach still has issues keeping state live/udpated for the `secondary` database.

This is the current iteration of our design - https://github.com/AleoHQ/snarkOS/blob/f9f3210ab8e3b1b7626d133317a1dbfbb5c7715e/ledger/src/state/ledger.rs#L171-L210

## Steps to Reproduce

1. Run a node
2. Query the RPC for block state
3. Look for mismatches in block height or latest block state.

## Expected Behavior

The `secondary` database should catch up to the primary state properly every time the `try_catch_up_with_primary` function is called.

## Your Environment

`testnet2` branch
Latest commit - `8bf81be47aa6be2145abd7a188507848d1dcb4c3`

howardwu:
Closing as we've switched to primary. Pending `rust-rocksdb` new release, we can switch back to secondary.

PR #1298, opened 2021-11-17
Fix secondary database catch-ups

raychu86:
## Motivation

This PR circumvents the issues in #1290. The real issue lies in a bug in RocksDB, which was fixed in `v6.25.0`, however the latest Rust `rocksdb v0.17` crate only supports RocksDB `v6.20.3`. 

The issue of the secondary database state rolling back occurred only when calling `try_catch_up_secondary` when the secondary was already caught up with the primary. To get around this, the PR gives the secondary database access to the primary database's latest block. This allows us to call `try_catch_up_secondary` only when there is actual state to be updated.

### DIFF
--- a/ledger/src/state/ledger.rs
+++ b/ledger/src/state/ledger.rs
@@ -175,7 +175,6 @@ impl<N: Network> LedgerState<N> {
         if ledger.is_read_only() {
             debug!("Loading ledger in read-only mode");
             let mut ledger = ledger.clone();
-            let primary_latest_block = primary_latest_block.clone();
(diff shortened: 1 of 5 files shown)


howardwu:
@raychu86 Are you sure any of the RPC line changes here make a difference?

From what I can tell, it seems just updating the logic in the reader loop in `LedgerState` in the PR would also work?

howardwu:
Closing as we switched to primary for now. Once `rust-rocksdb` is updated, we should be able to switch back without this change.

PR #1305, opened 2021-11-19
Implement initial wallet functionality to retrieve coinbases

### LINKED ISSUE #1258: [Feature] Initial node wallet with coinbase records for miners
## 🚀 Feature

@julesdesmit after thinking through this design further, I actually would suggest the following design:

1. Keep a single pool of records (irrespective of their existence on canon).
2. When fetching records for a specific account, call `ledger.contains_commitment(record.commitment())` on every single record that belongs to the address (i.e. `record.owner() == miner_address`).
3. Return all records that exist in the current canon chain (i.e. anything that returned `true`)

This approach has the advantage that we can support long fork rebases (in case we need to), and the call to `contains_commitment` is a constant/1-step operation in our new DB design. Given we check this for every block, I can say that should be fairly performant (i.e. tens of milliseconds if I had to guess).

jules:
Fixes #1258

### DIFF
--- a/src/network/server.rs
+++ b/src/network/server.rs
@@ -285,8 +285,7 @@ impl<N: Network, E: Environment> Server<N, E> {
         if E::NODE_TYPE == NodeType::Miner {
             if let Some(recipient) = miner {
                 // TODO: add data_path as a constant string when #20 is merged.
-                let wallet =
-                    Wallet::new(recipient.to_string(), ".aleo".to_string(), false).expect("Should be able to open wallet database");
+                let wallet = Wallet::new(&recipient.to_string(), &".aleo", false).expect("Should be able to open wallet database");
                 let ledger_router = ledger_router.clone();
                 tasks.append(task::spawn(async move {
                     loop {
--- a/src/network/server.rs
+++ b/src/network/server.rs
@@ -20,7 +20,8 @@ use crate::{
     peers::{Peers, PeersRequest, PeersRouter},
     rpc::initialize_rpc_server,
     wallet::Wallet,
-    Environment, NodeType,
+    Environment,
+    NodeType,
 };
 use snarkos_ledger::{storage::rocksdb::RocksDB, LedgerState};
 use snarkvm::prelude::*;
--- a/src/network/server.rs
+++ b/src/network/server.rs
@@ -19,8 +19,8 @@ use crate::{
     ledger::{Ledger, LedgerRequest, LedgerRouter},
     peers::{Peers, PeersRequest, PeersRouter},
     rpc::initialize_rpc_server,
-    Environment,
-    NodeType,
+    wallet::Wallet,
+    Environment, NodeType,
 };
 use snarkos_ledger::{storage::rocksdb::RocksDB, LedgerState};
 use snarkvm::prelude::*;
@@ -283,11 +283,13 @@ impl<N: Network, E: Environment> Server<N, E> {
     ) {
         if E::NODE_TYPE == NodeType::Miner {
             if let Some(recipient) = miner {
+                // TODO: add data_path as a constant string when #20 is merged.
+                let wallet = Wallet::new(recipient.to_string(), ".aleo".to_string()).expect("Should be able to open wallet database");
                 let ledger_router = ledger_router.clone();
                 tasks.append(task::spawn(async move {
                     loop {
                         // Start the mining process.
-                        let request = LedgerRequest::Mine(local_ip, recipient, ledger_router.clone());
+                        let request = LedgerRequest::Mine(local_ip, recipient, ledger_router.clone(), wallet.clone());
                         if let Err(error) = ledger_router.send(request).await {
                             error!("Failed to send request to ledger: {}", error);
                         }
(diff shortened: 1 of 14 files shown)


src/wallet/wallet.rs
```diff
@@ -0,0 +1,69 @@
...
+
+    /// Fetch all records from the database.
+    pub fn records(&self) -> Result<Vec<Transaction<N>>> {
+        let mut records = vec![];
+
+        for record in self.db.values() {
+            records.push(record);
+        }
+
+        Ok(records)
```
ljedrz:
```suggestion
        Ok(self.db.values().collect())
```

src/wallet/wallet.rs
```diff
@@ -0,0 +1,69 @@
...
+    address: String,
+    /// A path to a directory containing all coinbase records.
+    data_path: String,
+    /// The RocksDB containing all records for the given address.
+    db: DataMap<N::BlockHash, Transaction<N>>,
+}
+
+impl<N: Network> Wallet<N> {
+    /// Creates a new [`Wallet`].
+    pub fn new(address: String, data_path: String, read_only: bool) -> Result<Self> {
```
ljedrz:
```suggestion
    pub fn new(address: &str, data_path: &str, read_only: bool) -> Result<Self> {
```
this will allow us to drop the earlier calls to `to_string()`

src/wallet/wallet.rs
```diff
@@ -0,0 +1,69 @@
...
+        let path = PathBuf::from(format!("{}/{}/", data_path, address));
+
+        Ok(Self {
+            address: address.clone(),
+            data_path,
+            db: RocksDB::open(path, 0, read_only)?.open_map(&address)?,
```
ljedrz:
```suggestion
            data_path,
            db: RocksDB::open(path, 0, read_only)?.open_map(&address)?,
            address,
```

src/wallet/wallet.rs
```diff
@@ -0,0 +1,63 @@
...
+/// The database will just store coinbase transactions under the block hash that
+/// they belong to.
+#[derive(Clone, Debug)]
+pub struct Wallet<N: Network> {
+    /// The address associated with this wallet.
+    address: String,
+    /// A path to a directory containing all coinbase records.
+    data_path: String,
+    /// The RocksDB containing all records for the given address.
+    db: DataMap<N::BlockHash, Transaction<N>>,
```
ljedrz, line 39: future consideration: if we expect `get_blocks_mined` to be called often, it would be a good idea to cache the values loaded from the database when `Wallet` is created, and to have:
```
canon_blocks_mined: u32,
blocks_mined: u32,
```
in addition to the existing fields, which would need to be updated with `push_record` and the function that marks blocks as canon; the related RPC could then use those 2 fields instead, which would be much faster

src/rpc/documentation/public_endpoints/getblocksmined.md
```diff
@@ -0,0 +1,31 @@
...
+### Arguments
+
+None
+
+### Response
+
+|        Parameter            |  Type  | Description |
+|:---------------------------:|:------:|:-----------:|
+| `canon_blocks_mined`        | number | The amount of canon records corresponding to all addresses found on disk. |
+| `total_blocks_mined`        | number | The amount of all records corresponding to all addresses found on disk. |
```
ljedrz, line 13:
```suggestion
| `canon_blocks_mined`        | number | The number of canon records corresponding to all addresses found on disk. |
| `total_blocks_mined`        | number | The number of all records corresponding to all addresses found on disk. |
```
a grammatical nit

src/rpc/rpc_impl.rs
```diff
@@ -169,6 +170,47 @@ impl<N: Network, E: Environment> RpcFunctions<N> for RpcImpl<N, E> {
         Ok(self.ledger.read().await.get_block_header(block_height)?)
     }
 
+    async fn get_blocks_mined(&self) -> Result<Value, RpcError> {
```
ljedrz, line 173: it would be nice to have a corresponding RPC test, even if it's supposed to return `0`s

src/network/server.rs
```diff
@@ -283,11 +284,13 @@ impl<N: Network, E: Environment> Server<N, E> {
     ) {
         if E::NODE_TYPE == NodeType::Miner {
             if let Some(recipient) = miner {
+                // TODO: add data_path as a constant string when #20 is merged.
+                let wallet = Wallet::new(&recipient.to_string(), &".aleo", false).expect("Should be able to open wallet database");
```
ljedrz, line 288:
```suggestion
                let wallet = Wallet::new(&recipient.to_string(), ".aleo", false).expect("Should be able to open wallet database");
```
also, is there a reason why `recipient` is converted to a `String` here? it would work with the later `format!` without conversion, and the `Wallet` could contain the original type instead

ljedrz (commented):
LGTM in general, left some extra comments; not 100% sure about the panick-y bits (`except`s), but it might make sense for those errors to be grounds for panicking. Haven't had the chance to test it locally yet, will try it out tomorrow.

howardwu:
Closing as snarkOS has implemented this feature

PR #1313, opened 2021-11-23
Fix the snarkos-testing setup, start logging earlier

ljedrz:
This PR removes the workspace dependency on `snarkos-testing` (which shouldn't be there, otherwise it auto-enables the `test` feature), and moves the initialization of the logger to `main`.


howardwu (approved):

howardwu:
LGTM

PR #1318, opened 2021-11-23
Use more fine-grained locks for Ledger and Peers

ljedrz:
The current code uses write locks over the entire `Ledger` and `Peers` objects to process any requests related to them. This effectively causes them to be processed in a sequential fashion, but many (especially the ones related to `Peers`) can be processed concurrently instead.

In addition, this PR enforces `Ledger.block_requests_lock`, which is currently dropped immediately.

### DIFF
--- a/src/network/ledger.rs
+++ b/src/network/ledger.rs
@@ -27,7 +27,6 @@ use snarkvm::dpc::prelude::*;
 
 use anyhow::Result;
 use chrono::Utc;
-use parking_lot::Mutex;
 use rand::thread_rng;
 use std::{
     collections::{HashMap, HashSet},
@@ -40,7 +39,10 @@ use std::{
     },
     time::{Duration, Instant},
 };
-use tokio::{sync::mpsc, task};
+use tokio::{
+    sync::{mpsc, Mutex, RwLock},
+    task,
+};
 
 /// The maximum number of unconfirmed blocks that can be held by the ledger.
 const MAXIMUM_UNCONFIRMED_BLOCKS: u32 = 50;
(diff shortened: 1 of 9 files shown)


ljedrz:
Note: this PR has the potential to greatly improve performance, but it should be tested thoroughly, as everything being processed sequentially could have hidden some issues that would reveal themselves if they were processed concurrently.

ljedrz:
I think I'm seeing an issue (a partial hang) whenever `[Read-Only] Updating ledger state from block` gets triggered; going to investigate it further.

ljedrz:
It appears that this PR is now functional as well; I've been running it without issues for a while now, and I'll keep testing it for a while longer with a miner now.

The PR builds on https://github.com/AleoHQ/snarkOS/pull/1319, which should be merged first.

ljedrz:
Unfortunately, it seems that a fresh node can have difficulty syncing with the ledger-related changes. Back to draft :smiley:.

ljedrz:
Found the issue :sunglasses:.

ljedrz:
Found and fixed a (ledger-lock-related, so the peer-locking PR is unaffected) deadlock in case of a block revert; hopefully this was the last one.

howardwu (approved):

src/network/ledger.rs
```diff
@@ -392,29 +414,27 @@ impl<N: Network, E: Environment> Ledger<N, E> {
...
-                false => trace!("Memory pool already contains unconfirmed block {}", block.height()),
+            // Register the block's height.
+            let block_height = block.height();
+
+            // Add the block to the unconfirmed blocks.
+            if !self.unconfirmed_blocks.write().await.insert(block.previous_block_hash(), block) {
+                trace!("Memory pool already contains unconfirmed block {}", block_height);
+            } else {
+                trace!("Added unconfirmed block {} to memory pool", block_height);
             }
```
howardwu:
```suggestion
            // Retrieve the unconfirmed block height.
            let block_height = block.height();

            // Add the block to the unconfirmed blocks.
            if self.unconfirmed_blocks.write().await.insert(block.previous_block_hash(), block) {
                trace!("Added unconfirmed block {} to memory pool", block_height);
            } else {
                trace!("Memory pool already contains unconfirmed block {}", block_height);
            }
```

howardwu: Minor nit for readability

PR #1319, opened 2021-11-23
Use more fine-grained locks in Peers

ljedrz:
Isolated from https://github.com/AleoHQ/snarkOS/pull/1318 for more convenient reviewing and testing; I seem to be bleeding peers (due to them `spamming unconfirmed blocks (frequency = 5)`) after some time, but this could be caused by some of the network protocol rules that might need to be relaxed (as these changes enable faster communication, which might have been unexpected before).

### DIFF
--- a/src/network/peers.rs
+++ b/src/network/peers.rs
@@ -27,7 +27,7 @@ use std::{
 };
 use tokio::{
     net::TcpStream,
-    sync::{mpsc, oneshot},
+    sync::{mpsc, oneshot, RwLock},
     task,
     time::timeout,
 };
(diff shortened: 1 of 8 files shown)


ljedrz:
note: the second commit from https://github.com/AleoHQ/snarkOS/pull/1318 applies cleanly on top of this PR (which is its first commit)

howardwu:
More of a high-level convention question - when should we be using `Arc<RwLock<Foo>>` as opposed to `RwLock<Foo>`? Want to make sure we maintain consistency in design choice.

src/network/server.rs
```diff
@@ -46,7 +46,7 @@ pub struct Server<N: Network, E: Environment> {
     /// The status of the node.
     status: Status,
     /// The list of peers for the node.
-    peers: Arc<RwLock<Peers<N, E>>>,
+    peers: Arc<Peers<N, E>>,
```
niklaslong: 👍

src/network/peers.rs
```diff
@@ -86,19 +86,19 @@ pub struct Peers<N: Network, E: Environment> {
...
+    connected_peers: RwLock<HashMap<SocketAddr, (u64, OutboundRouter<N, E>)>>,
     /// The set of candidate peer IPs.
-    candidate_peers: HashSet<SocketAddr>,
+    candidate_peers: RwLock<HashSet<SocketAddr>>,
     /// The set of restricted peer IPs.
-    restricted_peers: HashMap<SocketAddr, Instant>,
+    restricted_peers: RwLock<HashMap<SocketAddr, Instant>>,
     /// The map of peers to their first-seen port number, number of attempts, and timestamp of the last inbound connection request.
-    seen_inbound_connections: HashMap<SocketAddr, ((u16, u32), SystemTime)>,
+    seen_inbound_connections: RwLock<HashMap<SocketAddr, ((u16, u32), SystemTime)>>,
```
niklaslong: Note: we might want structs in place of the tuples here (and further up). This is more of a design decision but some unification may not be a bad thing for all these collections.

ljedrz:
>when should we be using `Arc<RwLock<Foo>>` as opposed to `RwLock<Foo>`? Want to make sure we maintain consistency in design choice

Whenever there's an object that requires shared ownership, an `Arc` provides the ability to shallowly clone objects (and pass them on to tasks, threads etc.) that should maintain a consistent state (as opposed to e.g. some `String` that can just be cloned around and it doesn't need to remain consistent) or objects that are just too heavy to clone deeply (even if their state doesn't need to remain consistent everywhere), like large collections.

Since `Arc` use does incur a little bit of runtime overhead, it is also better if multiple objects can be under a single `Arc` (e.g. having `Arc<Peers>` as opposed to plain `Peers` with its children all being `Arc`ed), as long as they can remain immutable or have a lock over them (since `Arc` requires immutability).

src/network/peers.rs
```diff
@@ -506,25 +534,32 @@ impl<N: Network, E: Environment> Peers<N, E> {
...
-            let peer_str = peer_ip.to_string();
-            *peer_ip != &sender && !E::SYNC_NODES.contains(&peer_str.as_str()) && !E::PEER_NODES.contains(&peer_str.as_str())
-        }) {
-            self.send(*peer, message).await;
+        for peer in self
+            .connected_peers()
+            .await
+            .iter()
+            .filter(|peer_ip| {
+                let peer_str = peer_ip.to_string();
```
niklaslong: We should probably store these as `SocketAddr`.

src/network/peers.rs
```diff
@@ -732,18 +767,17 @@ impl<N: Network, E: Environment> Peer<N, E> {
...
+    async fn handler(
         stream: TcpStream,
         local_ip: SocketAddr,
         local_nonce: u64,
         local_status: Status,
         peers_router: &PeersRouter<N, E>,
         ledger_reader: LedgerReader<N>,
         ledger_router: LedgerRouter<N, E>,
-        connected_nonces: &mut T,
+        connected_nonces: Vec<u64>,
```
niklaslong: Do we want a `Nonce` type at some point?

niklaslong (approved):
LGTM, left a few comments for consideration 👍

ljedrz:
@niklaslong agreed with your points, will propose them in a future PR in order to keep this PR self-contained.

ljedrz:
@howardwu this PR can cause RPC to hang after it's been running for a while; I need to find the cause before it can be merged (unless this has already been observed with the main branch)

ljedrz:
I think I was able to find the issue, but I'll keep testing for a bit longer before un-drafting.

ljedrz:
Yep, it appears that the fix did the trick :+1:; I can no longer reproduce the RPC hang.

PR #1320, opened 2021-11-23
Adds filter for duplicate transactions before mining

howardwu:
## Motivation

Adds filter for duplicate transactions before mining

### DIFF
--- a/ledger/src/state/ledger.rs
+++ b/ledger/src/state/ledger.rs
@@ -563,23 +563,19 @@ impl<N: Network> LedgerState<N> {
         let mut transactions: Vec<Transaction<N>> = transactions
             .iter()
             .filter(|transaction| {
-                let mut is_new = true;
-
                 for serial_number in transaction.serial_numbers() {
                     if let Ok(true) = self.contains_serial_number(serial_number) {
-                        is_new = false;
-                        break;
+                        return false;
                     }
                 }
 
                 for commitment in transaction.commitments() {
                     if let Ok(true) = self.contains_commitment(commitment) {
-                        is_new = false;
-                        break;
+                        return false;
                     }
                 }
 
-                is_new
+                true
             })
             .cloned()
             .collect();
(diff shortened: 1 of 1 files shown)


ledger/src/state/ledger.rs
```diff
@@ -558,8 +558,31 @@ impl<N: Network> LedgerState<N> {
         let amount = Block::<N>::block_reward(block_height);
         let coinbase_transaction = Transaction::<N>::new_coinbase(recipient, amount, rng)?;
 
+        // Filter the transactions to ensure they are new, and append the coinbase transaction.
+        // TODO (howardwu): Improve the performance and design of this.
+        let mut transactions: Vec<Transaction<N>> = transactions
+            .iter()
+            .filter(|transaction| {
```
raychu86, line 565: Would a `retain` call be more efficient than a `filter`?

ledger/src/state/ledger.rs
```diff
@@ -558,8 +558,31 @@ impl<N: Network> LedgerState<N> {
...
+                for commitment in transaction.commitments() {
+                    if let Ok(true) = self.contains_commitment(commitment) {
+                        return false;
+                    }
+                }
+
+                true
+            })
+            .cloned()
+            .collect();
```
raychu86, line 581: We should craft the coinbase transaction here and include the transaction fees in the reward.

ledger/src/state/ledger.rs
ljedrz: `transactions` is just a slice, though, so it can't be `retain`ed

raychu86:
Because we just filter out transactions with existing commitments and serial numbers, a transaction could live in the memory pool forever, which seems to open us up to spam attacks. We may want to periodically clean the mempool or do it during block mining.

raychu86 (approved):

ledger/src/state/ledger.rs
raychu86: Will add in a followup PR.

PR #1322, opened 2021-11-23
Add transaction fees to coinbase transaction

raychu86:
## Motivation

This PR adds the transaction fee to the miner reward for each block that is mined.


howardwu:
We should add a test for this if possible

raychu86:
#1547

PR #1326, opened 2021-11-25
Reduce blocking when (de)serializing messages

ljedrz:
This PR introduces deferred non-blocking deserialization for messages that take a long time to parse. This should improve the responsiveness of nodes, especially ones with many active connections.

The object that enables this, `MaybeSerialized`, is intended to be used within `Message`, ~and while it is currently only used for non-blocking deserialization, it could technically easily enable non-blocking serialization as well if need be~ and already applies to the slow bits in both serialization and deserialization.

### DIFF
--- a/src/network/message.rs
+++ b/src/network/message.rs
@@ -25,8 +25,8 @@ use std::{marker::PhantomData, net::SocketAddr};
 use tokio::task;
 use tokio_util::codec::{Decoder, Encoder};
 
-/// This object enables deferred deserialization for objects that take a while to deserialize, in order
-/// to allow message processing to be non-blocking.
+/// This object enables deferred deserialization / ahead-of-time serialization for objects that
+/// take a while to deserialize / serialize, in order to allow these operations to be non-blocking.
 #[derive(Clone, Debug)]
 pub enum MaybeSerialized<T> {
     Deserialized(T),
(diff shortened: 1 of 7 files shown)


src/network/message.rs
```diff
@@ -173,9 +221,9 @@ impl<N: Network, E: Environment> Message<N, E> {
                     _ => return Err(anyhow!("Invalid 'Pong' message: {:?} {:?}", buffer, data)),
                 };
 
-                Self::Pong(is_fork, bincode::deserialize(&data[1..])?)
+                Self::Pong(is_fork, MaybeSerialized::Serialized(data[1..].to_vec()))
```
niklaslong: Ditto.

src/network/message.rs
```diff
@@ -144,14 +192,14 @@ impl<N: Network, E: Environment> Message<N, E> {
         // Deserialize the data field.
         let message = match id {
             0 => Self::BlockRequest(bincode::deserialize(&data[0..4])?, bincode::deserialize(&data[4..8])?),
-            1 => Self::BlockResponse(bincode::deserialize(data)?),
+            1 => Self::BlockResponse(MaybeSerialized::Serialized(data.to_vec())),
             2 => Self::ChallengeRequest(
                 bincode::deserialize(&data[0..4])?,
                 bincode::deserialize(&data[4..6])?,
                 bincode::deserialize(&data[6..14])?,
                 bincode::deserialize(&data[14..18])?,
```
niklaslong, line 216: Would be good to comment on the various indexes used here?

niklaslong (approved):
LGTM 👍

src/network/message.rs
ljedrz: I would like to handle deserialization in a streaming fashion (which would remove the need for indexing altogether), but that's follow-up material.

howardwu (approved):

PR #1327, opened 2021-11-25
Perform clean shutdown upon Ctrl-C

ljedrz:
Note: this PR only provides the means to automatically intercept `Ctrl-C` and perform `Server::shut_down`; further work might be required for that method to perform a fully clean shutdown.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2344,8 +2344,10 @@ dependencies = [
  "memchr",
  "mio",
  "num_cpus",
+ "once_cell",
  "parking_lot",
  "pin-project-lite",
+ "signal-hook-registry",
  "tokio-macros",
  "winapi",
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -110,6 +110,7 @@ features = [
     "macros",
     "net",
     "rt-multi-thread",
+    "signal",
     "sync",
     "time"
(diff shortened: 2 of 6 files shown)


niklaslong (approved):

ljedrz:
Cc @raychu86

raychu86:
Does this ensure all the subtasks within the Peer/Ledger handlers are completed (for example processing a block)? IIRC, the server `shutdown` only flushes the server initialization tasks. 

Practically, I realize we can't really wait for every task to finish; miner tasks should probably be shut down, but tasks that are writing to the database could cause issues if they're shutdown in the process.

ljedrz:
> Does this ensure all the subtasks within the Peer/Ledger handlers are completed

No, this is something that will need to be done in a follow-up. For example, the loop processing the ledger requests should look something like this:
```
loop {
    tokio::select! {
        biased;
        _ = shutdown_receiver.recv() => break,
        message = ledger_handler.recv() => {
            ledger.update(request, **&peers_router).await;
        },
    }
}
```
with `shutdown_receiver` being an `mpsc::Receiver` expecting to receive a message indicating a shutdown. The associated task would then need to be `.await`ed instead of being `abort()`ed, which will require more advanced task accounting.

PR #1331, opened 2021-11-26
Avoid the write()s in Ledger::initialize_peer

ljedrz:
While it would usually preferable to immediately use `write()` when inserting into a `HashMap/Set`, the most probable scenario is that the peer had already been initialized, so a single `read()` will suffice most of the time.


howardwu (approved):

howardwu:
LGTM

PR #1334, opened 2021-11-26
Introduce a BlockRequest object to speed up block request queries

ljedrz:
This PR introduces a `struct BlockRequest` which is compared and hashed only by the block height (as all the related operations check for a block height, with or without the corresponding hash).


niklaslong (approved):

howardwu (approved):
LGTM

PR #1335, opened 2021-11-26
Optimize lock access and reduce the number of queries in Ledger::update_block_requests

ljedrz:
This PR acquires locks just once for the duration of some loops in `Ledger::update_block_requests`, and reduces the number of queries required to add new requests.

### DIFF
--- a/src/network/ledger.rs
+++ b/src/network/ledger.rs
@@ -812,16 +812,35 @@ impl<N: Network, E: Environment> Ledger<N, E> {
                 return;
             }
 
-            // Log each block request to ensure the peer responds with all requested blocks.
-            for block_height in start_block_height..=end_block_height {
-                // If the ledger was reverted, include the expected new block hash for the fork.
-                match ledger_reverted {
-                    true => {
-                        self.add_block_request(peer_ip, block_height, maximum_block_locators.get_block_hash(block_height))
-                            .await
+            // Filter out any pre-existing block requests for the peer.
+            let mut new_block_heights = Vec::new();
+            if let Some(block_requests) = self.block_requests.read().await.get(&peer_ip) {
+                for block_height in start_block_height..=end_block_height {
+                    if !block_requests.contains_key(&block_height.into()) {
+                        new_block_heights.push(block_height);
                     }
-                    false => self.add_block_request(peer_ip, block_height, None).await,
-                };
+                }
+            } else {
+                self.add_failure(peer_ip, format!("Missing block requests for {}", peer_ip)).await;
+            }
+
+            // Log each block request to ensure the peer responds with all requested blocks.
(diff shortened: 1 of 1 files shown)


src/network/ledger.rs
```diff
@@ -812,16 +812,35 @@ impl<N: Network, E: Environment> Ledger<N, E> {
...
+            // Filter out any pre-existing block requests for the peer.
+            let mut new_block_heights = Vec::new();
+            if let Some(block_requests) = self.block_requests.read().await.get(&peer_ip) {
+                for block_height in start_block_height..=end_block_height {
+                    if !block_requests.contains_key(&block_height.into()) {
+                        new_block_heights.push(block_height);
+                    }
+                }
+            } else {
+                self.add_failure(peer_ip, format!("Missing block requests for {}", peer_ip)).await;
```
howardwu, line 825: You might want to add a `return;` here to match the original logic, otherwise it would technically still attempt to acquire a `write` lock in the next step.

howardwu (changes requested):

src/network/ledger.rs
```diff
@@ -812,16 +812,35 @@ impl<N: Network, E: Environment> Ledger<N, E> {
                 return;
             }
 
+            // Filter out any pre-existing block requests for the peer.
+            let mut new_block_heights = Vec::new();
```
niklaslong, line 817: Is it worth pre-allocating the vec here?

src/network/ledger.rs
```diff
@@ -812,16 +812,35 @@ impl<N: Network, E: Environment> Ledger<N, E> {
...
+            if let Some(locked_block_requests) = self.block_requests.write().await.get_mut(&peer_ip) {
+                for block_height in new_block_heights {
+                    // If the ledger was reverted, include the expected new block hash for the fork.
+                    match ledger_reverted {
+                        true => {
+                            self.add_block_request(
+                                peer_ip,
+                                block_height,
+                                maximum_block_locators.get_block_hash(block_height),
+                                locked_block_requests,
```
niklaslong: Perhaps it's worth commenting on the lock holding here or above?

niklaslong (approved):

howardwu (approved):

src/network/ledger.rs
ljedrz: We can't predict how many requests are new; unless the usual case is that all/most of them are guaranteed to be new, but I don't know that :smiley:.

src/network/ledger.rs
ljedrz: Isn't the lock's scope in `if let` sufficient? Or do you mean that it's holding for the duration of the loop?

niklaslong: The latter, especially as the mutable reference is being passed into another function.

PR #1337, opened 2021-11-26
Reduce the scope of the unconfirmed blocks write lock in update_ledger

ljedrz:
Use a vector to register the items to remove, and remove them wholesale afterwards.


howardwu (approved):
LGTM

niklaslong (approved):

PR #1339, opened 2021-11-26
Display tracing logs when running with --display

ljedrz:
This PR is a very simple, quick implementation for redirecting `tracing` logs to `tui`; it can surely be made faster and better-looking, and the log caching logic is not airtight in case of a large volume of pending logs.

It already does work, but I'm filing it as a draft so it can be improved upon.

### DIFF
--- a/src/display/display.rs
+++ b/src/display/display.rs
@@ -16,6 +16,7 @@
 
 use crate::{
     display::{logs::Logs, overview::Overview},
+    initialize_logger,
     network::Server,
     Environment,
 };
@@ -29,8 +30,10 @@ use crossterm::{
 };
 use std::{
     io,
+    thread,
     time::{Duration, Instant},
 };
+use tokio::sync::mpsc;
 use tui::{
     backend::{Backend, CrosstermBackend},
     layout::{Constraint, Direction, Layout},
@@ -68,10 +71,16 @@ pub(crate) struct Display<'a, N: Network, E: Environment> {
     server: Server<N, E>,
     tabs: TabsState<'a>,
     tick_rate: Duration,
+    logs: Logs,
 }
 
 impl<'a, N: Network, E: Environment> Display<'a, N, E> {
-    pub fn start(server: Server<N, E>) -> Result<()> {
+    pub fn start(server: Server<N, E>, verbosity: u8) -> Result<()> {
+        // Initialize the log channel.
+        let (log_sender, log_receiver) = mpsc::channel(1024);
+
+        initialize_logger(verbosity, Some(log_sender));
+
         enable_raw_mode()?;
         let mut stdout = io::stdout();
         execute!(stdout, EnterAlternateScreen, EnableMouseCapture)?;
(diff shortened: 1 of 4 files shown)


src/display/display.rs
```diff
@@ -68,10 +71,16 @@ pub(crate) struct Display<'a, N: Network, E: Environment> {
...
     tabs: TabsState<'a>,
     tick_rate: Duration,
+    logs: Logs,
 }
 
 impl<'a, N: Network, E: Environment> Display<'a, N, E> {
-    pub fn start(server: Server<N, E>) -> Result<()> {
+    pub fn start(server: Server<N, E>, verbosity: u8) -> Result<()> {
+        // Initialize the log channel.
+        let (log_sender, log_receiver) = mpsc::channel(1024);
```
ljedrz, line 80: this number is probably too large, and entries above the log display limit will currently get truncated (they could potentially all be scrolled through, though)

src/display/logs.rs
```diff
@@ -14,28 +14,66 @@
...
 
-        let canvas = Canvas::default()
-            .block(Block::default().borders(Borders::ALL).title("Logs"))
-            .paint(|_ctx| {
-                // ctx.draw(&ball);
-            });
-        f.render_widget(canvas, chunks[0]);
+        let mut new_logs = Vec::new();
+        while let Ok(log) = self.log_receiver.try_recv() {
+            new_logs.push(String::from_utf8(log).unwrap());
```
ljedrz: our own logs do not contain non-UTF-8 characters, hence the `unwrap()`

src/display/logs.rs
```diff
@@ -14,28 +14,66 @@
...
+                }
+            }
+        };
+
+        self.log_cache.extend(new_logs.into_iter().take(self.log_limit));
+
+        let combined_logs = self.log_cache.iter().map(|s| s.as_str()).collect::<String>();
+
+        let combined_logs = Paragraph::new(combined_logs).block(Block::default().borders(Borders::ALL).title("Logs"));
+        f.render_widget(combined_logs, chunks[0]);
```
ljedrz, line 79: currently there's no scrolling, so the new logs are not displayed in a smooth way

src/node.rs
```diff
@@ -155,30 +155,32 @@ impl Node {
...
+        .add_directive("tokio_util=off".parse().unwrap())
+        .add_directive("hyper::proto::h1::conn=off".parse().unwrap())
+        .add_directive("hyper::proto::h1::decode=off".parse().unwrap())
+        .add_directive("hyper::proto::h1::io=off".parse().unwrap())
+        .add_directive("hyper::proto::h1::role=off".parse().unwrap());
+
+    // Initialize tracing.
+    let _ = tracing_subscriber::fmt()
+        .with_env_filter(filter)
+        .with_ansi(log_sender.is_none())
```
ljedrz, line 184: `tui` can surely provide some nice colors for the logs, but this will probably need to be implemented separately, as the `ansi` feature of `tracing` doesn't translate well to `tui`'s `Paragraph`

niklaslong:
I encounter `'main' panicked at 'attempt to subtract with overflow', /Users/nkls/.cargo/registry/src/github.com-1ecc6299db9ec823/tui-0.16.0/src/widgets/canvas/mod.rs:483:36` with `cargo r -- --display`. It seems like a known class of issue with `tui`: https://github.com/fdehau/tui-rs/issues/446.

howardwu (approved):

PR #1355, opened 2021-11-30
Set default RPC address to 127.0.0.1:3032

nomaka:
0.0.0.0:3032 will be vulnerable to a hacker attack


howardwu:
Closing as 0.0.0.0 is the intended functionality. If you are looking to change the RPC address, please use the `--rpc` flag to set snarkOS to a custom address.

PR #1356, opened 2021-11-30
Deferred serialisation/deserialisation of block header in `Ping` messages

niklaslong:
~~Drafted for now, pending some testing.~~
Edit: seems to work well on testnet 2, would be good to benchmark this as a follow-up.

### DIFF
--- a/src/network/peers.rs
+++ b/src/network/peers.rs
@@ -581,6 +581,19 @@ impl<N: Network, E: Environment> Peers<N, E> {
             Some((_, outbound)) => {
                 // Ensure sufficient time has passed before needing to send the message.
                 let is_ready_to_send = match message {
+                    Message::Ping(_, _, _, _, ref mut data) => {
+                        let block_header = if let Data::Object(block_header) = data {
+                            block_header
+                        } else {
+                            panic!("Logic error: the block header shouldn't have been serialized yet.");
+                        };
+
+                        // Perform non-blocking serialisation of the block header.
+                        let serialized_header = bincode::serialize(&block_header).expect("Block header serialization is bugged");
+                        let _ = std::mem::replace(data, Data::Buffer(serialized_header));
+
+                        true
+                    }
                     Message::UnconfirmedBlock(_, _, ref mut data) => {
                         let block = if let Data::Object(block) = data {
                             block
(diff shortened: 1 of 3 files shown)


src/network/peers.rs
```diff
@@ -581,6 +581,19 @@ impl<N: Network, E: Environment> Peers<N, E> {
...
                 let is_ready_to_send = match message {
+                    Message::Ping(_, _, _, _, ref mut data) => {
+                        let block_header = if let Data::Object(block_header) = data {
+                            block_header
+                        } else {
+                            panic!("Logic error: the block header shouldn't have been serialized yet.");
+                        };
+
+                        // Perform non-blocking serialisation of the block header.
+                        let serialized_header = bincode::serialize(&block_header).expect("Block header serialization is bugged");
```
ljedrz, line 594: if I'm not missing anything, this call is blocking and you should use `Data::serialize` here instead

niklaslong: Yep, good catch 👍

niklaslong: I used the same pattern as for the other messages, which I beleive to be correct here? 🤔

ljedrz: I don't have a good way of double-checking this right now, but if my hunch is correct it will be easy to tweak afterwards, along with the other case where this is done in the same fashion; this change does make serialization happen ahead of time, though, so it's a step in the right direction :ok_hand:.

src/network/peers.rs
```diff
@@ -1043,8 +1056,14 @@ impl<N: Network, E: Environment> Peer<N, E> {
                                     peer.node_type = node_type;
                                     // Update the status of the peer.
                                     peer.status.update(status);
-                                    // Update the block header of the peer.
-                                    peer.block_header = block_header;
+
```
ljedrz, line 1061: it can be noted here that this deserialization is deferred and non-blocking, so it's clear why it happens here

src/network/peers.rs
niklaslong: We effectively unwrap the header from the `Data` wrapper, I think all we'd need is not to do that and use the `Data::serialise` method instead?

ljedrz: Yeah, that's what I was thinking, as it's `async`, no?

niklaslong: Should work, I'll file a follow-up to keep the change self-contained 👍

niklaslong: Filed #1357.

niklaslong:
Rebased.

howardwu (approved):

PR #1366, opened 2021-12-01
Update serialization of `getnodestate`

raychu86:
## Motivation

This PR updates the serialization of the `getnodestate` rpc output to reflect a u128 for `latest_cumulative_weight` (instead of using a String).

## Test Plan

A `getnodestate` RPC test has been added.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -95,10 +95,10 @@ version = "1"
 version = "0.27"
 
 [dependencies.serde]
-version = "1.0.130"
+version = "1"
 
 [dependencies.serde_json]
-version = "1.0.72"
+version = "1"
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -95,10 +95,11 @@ version = "1"
 version = "0.27"
 
 [dependencies.serde]
-version = "1"
+version = "1.0.130"
```
ljedrz: these version changes don't do anything (which is confirmed by `Cargo.lock` not changing); in other words, snarkOS already uses these versions

raychu86: Ah good catch. It was mainly just the use of the precision feature that was needed for `serde_json`

howardwu (approved):

PR #1372, opened 2021-12-03
Adds tests for the block request handler

raychu86:
## Motivation

This PR abstracts the handling of the different of peer state and block requests to a helper function `handle_block_requests`.

## Test Plan

Tests have been added for each case.

- **Case 1** - You are ahead of your peer:
    - Do nothing
- **Case 2** - You are behind your peer:
    - **Case 2(a)** - `is_fork` is `None`:
        - Peer is being malicious or thinks you are ahead. Both are issues,
          pick a different peer to sync with.
    - **Case 2(b)** - `is_fork` is `Some(false)`:
        - Request blocks from your latest state
    - **Case 2(c)** - `is_fork` is `Some(true)`:
      - **Case 2(c)(a)** - Common ancestor is within `ALEO_MAXIMUM_FORK_DEPTH`:
        - Roll back to common ancestor, and send block requests to sync.
      - **Case 2(c)(b)** - Common ancestor is NOT within `ALEO_MAXIMUM_FORK_DEPTH`:
        - **Case 2(c)(b)(a)** - You can calculate that you are outside of the `ALEO_MAXIMUM_FORK_DEPTH`:
            - Do not roll back or sync.
            - Remove this peer from peers_state. Call `update_block_requests` again.
        - **Case 2(c)(b)(b)** - You don't know if you are within the `ALEO_MAXIMUM_FORK_DEPTH`:
            - Roll back to most common ancestor and attempt to sync.
            - This means we aren't entirely enforcing `ALEO_MAXIMUM_FORK_DEPTH` precisely.

### DIFF
--- a/src/helpers/block_requests.rs
+++ b/src/helpers/block_requests.rs
@@ -66,7 +66,7 @@ pub async fn handle_block_requests<N: Network, E: Environment>(
         }
 
         // Determine the latest common ancestor.
-        let (latest_common_ancestor, ledger_requires_fork) =
+        let (latest_common_ancestor, ledger_requires_revert) =
             // Case 2(b) - This ledger is not a fork of the peer, it is on the same canon chain.
             if !is_fork {
                 // Continue to sync from the latest block height of this ledger, if the peer is honest.
@@ -114,7 +114,7 @@ pub async fn handle_block_requests<N: Network, E: Environment>(
         let start_block_height = latest_common_ancestor + 1;
         let end_block_height = start_block_height + number_of_block_requests - 1;
 
-        return Some((start_block_height, end_block_height, ledger_requires_fork));
+        return Some((start_block_height, end_block_height, ledger_requires_revert));
     }
 
     None
--- a/src/helpers/block_requests.rs
+++ b/src/helpers/block_requests.rs
@@ -80,11 +80,9 @@ pub async fn handle_block_requests<N: Network, E: Environment>(
                 // Case 2(c)(a) - If the common ancestor is within the fork range of this ledger, proceed to switch to the fork.
                 if latest_block_height.saturating_sub(maximum_common_ancestor) <= E::MAXIMUM_FORK_DEPTH {
                     info!("Discovered a canonical chain from {} with common ancestor {} and cumulative weight {}", peer_ip, maximum_common_ancestor, maximum_cumulative_weight);
-                    // If the latest block is the same as the maximum common ancestor, do not revert.
-                    if latest_block_height != maximum_common_ancestor {
-                        return None;
-                    }
-                    (maximum_common_ancestor, true)
+                    // Revert if the latest block is not the maximum common ancestor.
+                    let requires_revert = latest_block_height != maximum_common_ancestor;
+                    (maximum_common_ancestor, requires_revert)
                 }
                 // Case 2(c)(b) - If the common ancestor is NOT within `MAXIMUM_FORK_DEPTH`.
                 else {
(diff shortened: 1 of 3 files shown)


damons:
Is it time to request a review here?

src/helpers/block_requests.rs
```diff
@@ -0,0 +1,413 @@
...
+    maximal_peer_is_fork: Option<bool>,
+    maximum_block_height: u32,
+    maximum_cumulative_weight: u128,
+    maximum_common_ancestor: u32,
+    first_deviating_locator: Option<&u32>,
+) -> Option<(u32, u32, bool)> {
+    // Case 1 - Ensure the peer has a heavier canonical chain than this ledger.
+    if latest_cumulative_weight >= maximum_cumulative_weight {
+        return None;
+    }
```
ljedrz, line 55: this check is already performed in `update_block_requests`; it should be removed from there so as not to duplicate the logic

src/network/ledger.rs
```diff
@@ -849,6 +832,44 @@ impl<N: Network, E: Environment> Ledger<N, E> {
...
+                    None => first_deviating_locator = Some(block_height),
+                    Some(saved_height) => {
+                        if block_height < saved_height {
+                            first_deviating_locator = Some(block_height);
+                        }
+                    }
+                }
+            }
+        }
+
```
ljedrz, line 872: note: previously this was done as part of `Case 2`

src/network/ledger.rs
```diff
@@ -857,177 +878,116 @@ impl<N: Network, E: Environment> Ledger<N, E> {
...
+            None => {
+                // Case 2(c)(b)(a) - Check if the real common ancestor is NOT within `MAXIMUM_FORK_DEPTH`.
+                // If this peer is outside of the fork range of this ledger, proceed to disconnect from the peer.
+                if let Some(first_deviating_locator) = first_deviating_locator {
+                    if latest_block_height.saturating_sub(*first_deviating_locator) >= E::MAXIMUM_FORK_DEPTH {
+                        debug!(
+                            "Peer {} has exceeded the permitted fork range of the protocol, disconnecting",
+                            peer_ip
+                        );
+                        self.disconnect(peer_ip, "exceeded fork range").await;
```
ljedrz, line 904: there used to be a `return` right below this disconnect, but it seems that now it's only 6 lines later, after 2 scopes are exited

src/helpers/block_requests.rs
```diff
@@ -0,0 +1,413 @@
...
+///     Case 2(c) - `is_fork` is `Some(true)`:
+///             Case 2(c)(a) - Common ancestor is within `MAXIMUM_FORK_DEPTH`:
+///                  - Revert to common ancestor, and send block requests to sync.
+///             Case 2(c)(b) - Common ancestor is NOT within `MAXIMUM_FORK_DEPTH`:
+///                  Case 2(c)(b)(a) - You can calculate that you are outside of the `MAXIMUM_FORK_DEPTH`:
+///                      - Disconnect from peer.
+///                  Case 2(c)(b)(b) - You don't know if you are within the `MAXIMUM_FORK_DEPTH`:
+///                      - Revert to most common ancestor and send block requests to sync.
+///
+pub async fn handle_block_requests<N: Network, E: Environment>(
```
ljedrz, line 42: I don't see any `await`s inside this function, so it doesn't need to be `async`

howardwu:
Closing as this PR has been superseded by #1456, #1457, #1459, #1461, and #1464.

PR #1377, opened 2021-12-04
Update testnet2_ubuntu.sh

unordered-set:
Add missing dependency

## Motivation

(Write your motivation here)

## Test Plan

(Write your test plan here)

## Related PRs

(Link your related PRs here)

### DIFF
--- a/testnet2_ubuntu.sh
+++ b/testnet2_ubuntu.sh
@@ -21,6 +21,7 @@ apt-get install -y \
     make \
     pkg-config \
     tmux \
+    ufw \
     xz-utils
 
 # Install Rust


ljedrz:
I'm not an expert on firewalls, but I'm not sure `ufw` should always be installed, as the firewall might be controlled with alternative software.

unordered-set:
@ljedrz , at https://github.com/AleoHQ/snarkOS/blob/testnet2/testnet2_ubuntu.sh#L47 it uses `ufw`

ljedrz:
True, but my point is that in that case we might instead want to report that it's not installed and that the user should configure a firewall using a tool of their choosing instead of potentially installing an extra one that might not match their setup.

unordered-set:
Ah, we can use a side service to check the external availability. What do you think about it?

raychu86:
We are trying to remove as many assumptions and dependencies as possible in the startup scripts (and will likely refactor them soon). Closing for now.

PR #1393, opened 2021-12-07
Ensure that storage map operations are not interrupted by a shutdown

ljedrz:
This PR is a quick potential fix for https://github.com/AleoHQ/snarkOS/issues/1375. It ensures that some non-atomic storage operations can't be interrupted when the node is shutting down.

The proper solution - which will be a larger undertaking - is recorded as a `FIXME`.


niklaslong (approved):
lgtm 👍

apruden2008 (approved):

PR #1394, opened 2021-12-07
Ignore duplicate blocks, increase desired peer counts

ljedrz:
This PR aims to allow the nodes to handle more connections by ensuring that any duplicate inbound blocks are quickly identified and discarded.

The default numbers for client peers is greatly increased, while the number for miners is increased to a smaller extent.

### DIFF
--- a/src/environment/mod.rs
+++ b/src/environment/mod.rs
@@ -140,8 +140,8 @@ impl<N: Network> Environment for ClientTrial<N> {
         "159.223.124.150:4132", "137.184.192.155:4132", "147.182.213.228:4132", "137.184.202.162:4132", "159.223.118.35:4132",
         "161.35.106.91:4132", "157.245.133.62:4132", "143.198.166.150:4132",
     ];
-    const MINIMUM_NUMBER_OF_PEERS: usize = 11;
-    const MAXIMUM_NUMBER_OF_PEERS: usize = 31;
+    const MINIMUM_NUMBER_OF_PEERS: usize = 41;
+    const MAXIMUM_NUMBER_OF_PEERS: usize = 71;
 }
 
 #[derive(Clone, Debug, Default)]
@@ -156,7 +156,7 @@ impl<N: Network> Environment for MinerTrial<N> {
         "159.223.124.150:4132", "137.184.192.155:4132", "147.182.213.228:4132", "137.184.202.162:4132", "159.223.118.35:4132",
         "161.35.106.91:4132", "157.245.133.62:4132", "143.198.166.150:4132",
     ];
-    const MINIMUM_NUMBER_OF_PEERS: usize = 11;
-    const MAXIMUM_NUMBER_OF_PEERS: usize = 21;
+    const MINIMUM_NUMBER_OF_PEERS: usize = 21;
+    const MAXIMUM_NUMBER_OF_PEERS: usize = 31;
     const COINBASE_IS_PUBLIC: bool = true;
 }
--- a/src/network/peers.rs
+++ b/src/network/peers.rs
@@ -29,10 +29,13 @@ use crate::{
 use snarkvm::dpc::prelude::*;
 
 use anyhow::{anyhow, Result};
+use circular_queue::CircularQueue;
 use futures::SinkExt;
+use once_cell::sync::Lazy;
 use rand::{prelude::IteratorRandom, rngs::OsRng, thread_rng, Rng};
 use std::{
     collections::{HashMap, HashSet},
+    hash::{BuildHasher, Hash, Hasher},
     net::SocketAddr,
     sync::Arc,
     time::{Duration, Instant, SystemTime},
@@ -46,6 +49,7 @@ use tokio::{
 };
 use tokio_stream::StreamExt;
 use tokio_util::codec::Framed;
+use twox_hash::RandomXxHashBuilder64;
 
 /// Shorthand for the parent half of the `Peer` outbound message channel.
 pub(crate) type OutboundRouter<N, E> = mpsc::Sender<Message<N, E>>;
@@ -60,6 +64,9 @@ type PeersHandler<N, E> = mpsc::Receiver<PeersRequest<N, E>>;
 
 /// Shorthand for the parent half of the connection result channel.
 type ConnectionResult = oneshot::Sender<Result<()>>;
+
+/// A cache containing fast hashes of recent blocks.
+static BLOCK_CACHE: Lazy<RwLock<CircularQueue<u64>>> = Lazy::new(|| RwLock::new(CircularQueue::with_capacity(512)));
 
 ///
 /// An enum of requests that the `Peers` struct processes.
@@ -1179,6 +1186,19 @@ impl<N: Network, E: Environment> Peer<N, E> {
                                             warn!("[PeerRestricted] {}", error);
                                         }
                                         break;
+                                    }
+
+                                    // Procure a fast, short hash of the block hash, and check if it's present in the cache.
+                                    let fast_hash = {
+                                        let mut hasher = RandomXxHashBuilder64::default().build_hasher();
+                                        block_hash.hash(&mut hasher);
+                                        hasher.finish()
+                                    };
+                                    if BLOCK_CACHE.read().await.iter().any(|&e| e == fast_hash) {
+                                        trace!("Skipping 'UnconfirmedBlock {}' from {}", block_height, peer_ip);
+                                        continue;
+                                    } else {
+                                        BLOCK_CACHE.write().await.push(fast_hash);
                                     }
 
                                     // Retrieve the last seen timestamp of the received block.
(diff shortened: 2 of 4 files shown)


ljedrz:
During tests I noticed that the node can bleed peers, so some additional tweaking might be needed before their counts can be increased that much.

niklaslong:
Any idea why the node is bleeding peers?

src/network/peers.rs
```diff
@@ -61,6 +65,9 @@ type PeersHandler<N, E> = mpsc::Receiver<PeersRequest<N, E>>;
 /// Shorthand for the parent half of the connection result channel.
 type ConnectionResult = oneshot::Sender<Result<()>>;
 
+/// A cache containing fast hashes of recent blocks.
+static BLOCK_CACHE: Lazy<RwLock<CircularQueue<u64>>> = Lazy::new(|| RwLock::new(CircularQueue::with_capacity(512)));
```
niklaslong, line 69: Just curious, why `512`?

niklaslong (commented):
The cache looks good to me 👍

src/network/peers.rs
ljedrz: This is a somewhat arbitrary pick; it needs to be small enough to be super fast, but also large enough to filter out all recently seen blocks; it might require fine-tuning.

ljedrz:
I also found out that we are currently serializing blocks for each peer separately when we propagate them; it's fixed in the last commit, and with that one in place, my peer count is much more stable.

src/network/peers.rs
```diff
@@ -671,7 +671,13 @@ impl<N: Network, E: Environment> Peers<N, E> {
...
     /// Sends the given message to every connected peer, excluding the sender.
     ///
-    async fn propagate(&self, sender: SocketAddr, message: Message<N, E>) {
+    async fn propagate(&self, sender: SocketAddr, mut message: Message<N, E>) {
+        // Perform ahead-of-time, non-blocking serialization just once for applicable objects.
+        if let Message::UnconfirmedBlock(_, _, ref mut data) = message {
+            let serialized_block = Data::serialize(data.clone()).await.expect("Block serialization is bugged");
+            let _ = std::mem::replace(data, Data::Buffer(serialized_block));
+        }
+
```
ljedrz, line 674: note: I'm not altering the other point where ahead-of-time block serialization is currently performed in order to limit the scope of changes; the only downside of doing it that way is an extra `clone()` per block when propagating.

niklaslong: We could probably `mem::take` but this will require some default impl for the block, which we don't currently have.

apruden2008:
@howardwu I consider this fix could greatly improve node connectivity and enable faster convergence. Please review.

niklaslong (approved):

src/network/peers.rs
```diff
@@ -671,7 +665,13 @@ impl<N: Network, E: Environment> Peers<N, E> {
     ///
     /// Sends the given message to every connected peer, excluding the sender.
     ///
-    async fn propagate(&self, sender: SocketAddr, message: Message<N, E>) {
```
xorinox, line 667: what do you say about this?

`warning: variable does not need to be mutable
   --> src/network/peers.rs:672:51
    |
672 |     async fn propagate(&self, sender: SocketAddr, mut message: Message<N, E>) {
    |                                                   ----^^^^^^^
    |                                                   |
    |                                                   help: remove this `mut`
    |
    = note: `#[warn(unused_mut)]` on by default

warning: `snarkos` (lib) generated 1 warning`

xorinox:
I have patched the code with these changes and I do see errors I have not seen before, could be related, maybe not.  Question is. do you observe these also?


```
2021-12-10T02:29:38.701400Z  WARN [OutboundRouter] Broken pipe (os error 32)
2021-12-10T02:29:38.701424Z  WARN [OutboundRouter] Broken pipe (os error 32)
2021-12-10T02:29:38.701439Z  WARN [OutboundRouter] Broken pipe (os error 32)
2021-12-10T02:29:38.701455Z  WARN [OutboundRouter] Broken pipe (os error 32)
2021-12-10T02:29:38.701367Z  WARN [OutboundRouter] Broken pipe (os error 32)
```

I don't think I have seen these before. But sure this could be coincidence.

src/network/peers.rs
```diff
@@ -595,29 +602,23 @@ impl<N: Network, E: Environment> Peers<N, E> {
...
-                        seen_blocks.insert(block.hash(), SystemTime::now());
+                        seen_blocks.insert(block_hash, SystemTime::now());
                         // Report the unconfirmed block height.
                         if is_ready_to_send {
-                            trace!("Preparing to send 'UnconfirmedBlock {}' to {}", block.height(), peer);
+                            trace!("Preparing to send 'UnconfirmedBlock {}' to {}", block_height, peer);
                         }
 
-                        // Perform non-blocking serialization of the block.
+                        // Perform non-blocking serialization of the block (if it hasn't been serialized yet).
```
howardwu, line 621: Please revert this until a more thorough assessment is made.

The implication is that a malicious block from modified code can admit this change and poison internal `last_seen` filters if it is incorrect.

A simple fix is to check that `block_height == block.height()` and `block_hash == block.hash()` here, however this results in redundant checks. I can take a look at this change more closely later on to ensure we have it enforced at the lowest level possible.

src/environment/mod.rs
```diff
@@ -156,7 +156,7 @@ impl<N: Network> Environment for MinerTrial<N> {
         "159.223.124.150:4132", "137.184.192.155:4132", "147.182.213.228:4132", "137.184.202.162:4132", "159.223.118.35:4132",
         "161.35.106.91:4132", "157.245.133.62:4132", "143.198.166.150:4132",
     ];
-    const MINIMUM_NUMBER_OF_PEERS: usize = 11;
-    const MAXIMUM_NUMBER_OF_PEERS: usize = 21;
+    const MINIMUM_NUMBER_OF_PEERS: usize = 21;
+    const MAXIMUM_NUMBER_OF_PEERS: usize = 31;
```
howardwu, line 160: Has someone verified that these new values do not result in blocking the miner from running?

src/network/peers.rs
```diff
@@ -1181,6 +1188,19 @@ impl<N: Network, E: Environment> Peer<N, E> {
                                         break;
                                     }
 
+                                    // Procure a fast, short hash of the block hash, and check if it's present in the cache.
+                                    let fast_hash = {
+                                        let mut hasher = RandomXxHashBuilder64::default().build_hasher();
```
howardwu, line 1193: Is there anything from `std` lib that can be used with similar performance characteristics?

src/environment/mod.rs
```diff
@@ -140,8 +140,8 @@ impl<N: Network> Environment for ClientTrial<N> {
         "159.223.124.150:4132", "137.184.192.155:4132", "147.182.213.228:4132", "137.184.202.162:4132", "159.223.118.35:4132",
         "161.35.106.91:4132", "157.245.133.62:4132", "143.198.166.150:4132",
     ];
-    const MINIMUM_NUMBER_OF_PEERS: usize = 11;
-    const MAXIMUM_NUMBER_OF_PEERS: usize = 31;
+    const MINIMUM_NUMBER_OF_PEERS: usize = 41;
+    const MAXIMUM_NUMBER_OF_PEERS: usize = 71;
```
howardwu, line 144: Can we increase these to just 25 and 50 if it must go up? My concern is performance overhead

howardwu (changes requested):

src/environment/mod.rs
ljedrz: I'm not running a miner in the testnet, but the increased numbers for the client (which are greater) only led to ~5% avg CPU use; the change in peer counts can be done separately too

src/network/peers.rs
ljedrz: not from the `std`; there's alternatives, but usually not that fast or high-quality; I'm open to using an alternative way of uniquely identifying blocks, though

src/environment/mod.rs
ljedrz: I saw no issues with these (and even larger) values when running a client; the client only handles the network, after all

ljedrz:
I will split this PR into smaller ones to reduce the scope of changes.

ljedrz:
Closing in favor of smaller, more self-contained PRs.

ljedrz:
> I have patched the code with these changes and I do see errors I have not seen before, could be related, maybe not. Question is. do you observe these also?
> 
> ```
> 2021-12-10T02:29:38.701400Z  WARN [OutboundRouter] Broken pipe (os error 32)
> 2021-12-10T02:29:38.701424Z  WARN [OutboundRouter] Broken pipe (os error 32)
> 2021-12-10T02:29:38.701439Z  WARN [OutboundRouter] Broken pipe (os error 32)
> 2021-12-10T02:29:38.701455Z  WARN [OutboundRouter] Broken pipe (os error 32)
> 2021-12-10T02:29:38.701367Z  WARN [OutboundRouter] Broken pipe (os error 32)
> ```
> 
> I don't think I have seen these before. But sure this could be coincidence.

@xorinox I've seen it without these changes too; it's not an issue, just means that a connection was abruptly broken.

PR #1396, opened 2021-12-08
Stricter task accounting

ljedrz:
This PR extends `Peers` with the `Tasks` object, so that all pending connection requests and other peer-related tasks can be aborted upon shutdown.

This should improve the situation with issues like https://github.com/AleoHQ/snarkOS/issues/1380 by making the shutdown process more predictable.

### DIFF
--- a/src/network/peers.rs
+++ b/src/network/peers.rs
@@ -117,7 +117,7 @@ impl<N: Network, E: Environment> Peers<N, E> {
     /// Initializes a new instance of `Peers`.
     ///
     pub(crate) async fn new(
-        tasks: &mut Tasks<JoinHandle<()>>,
+        tasks: Tasks<JoinHandle<()>>,
         local_ip: SocketAddr,
         local_nonce: Option<u64>,
         local_status: &Status,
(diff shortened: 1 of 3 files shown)


ljedrz:
The CI failure is spurious, but I can't rerun it:
```
HTTP request sent, awaiting response... 503 Egress is over the account limit.
2021-12-08 10:10:33 ERROR 503: Egress is over the account limit..
```

niklaslong (approved):
LGTM 👍

niklaslong:
I managed to trigger a rerun, let's see... 👀

apruden2008:
@howardwu please approve this P1 PR

src/network/peers.rs
```diff
@@ -114,14 +114,16 @@ pub struct Peers<N: Network, E: Environment> {
     seen_outbound_connections: RwLock<HashMap<SocketAddr, SystemTime>>,
     /// The map of peers to a map of transaction IDs to their last seen timestamp.
     seen_outbound_transactions: RwLock<HashMap<SocketAddr, HashMap<N::TransactionID, SystemTime>>>,
+    /// The list of tasks spawned by the node.
+    tasks: Tasks<task::JoinHandle<()>>,
```
howardwu: Can someone clarify why a new tasks struct is being introduced here? One on the `server.rs` level is already available and passed through. For `update` functionality, the existing `Tasks` can already be passed through as needed there as well

howardwu (changes requested):

src/network/peers.rs
ljedrz: it's not strictly necessary - it only reduces some shallow cloning, but that shouldn't be too much of an issue; I added a commit removing it from `Peers`

apruden2008:
@ljedrz just to clarify I think you made the changes requested?

ljedrz:
Yes, `Peers` is no longer extended.

ljedrz:
Rebased.

ljedrz:
Rebased.

howardwu (approved):

howardwu:
Merging as the failing case is a flaky test that has already been fixed in v12

PR #1397, opened 2021-12-08
Don't disconnect from sync nodes too soon

ljedrz:
I noticed this while starting a node with high desired peer counts; it would start disconnecting from peers stating it's over the limit, while it was far from it. It appears that it was the condition checking if too many sync nodes are connected - I relaxed it so that it only goes into force when the node has at least the minimum number of peers.

In addition, since it's highly unlikely that this value will change in that context, the number of currently connected peers is now obtained just once, at the beginning, for improved performance

This PR has the potential to improve issues like https://github.com/AleoHQ/snarkOS/issues/1358.

### DIFF
--- a/src/network/peers.rs
+++ b/src/network/peers.rs
@@ -332,12 +332,15 @@ impl<N: Network, E: Environment> Peers<N, E> {
                 }
             }
             PeersRequest::Heartbeat(ledger_reader, ledger_router, prover_router) => {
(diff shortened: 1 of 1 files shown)


niklaslong (approved):

apruden2008:
@howardwu please approve this P1 PR

howardwu (approved):

howardwu:
LGTM

PR #1411, opened 2021-12-10
Perform ahead-of-time serialization just once when propagating blocks

ljedrz:
Split out from https://github.com/AleoHQ/snarkOS/pull/1394; marking as a draft in order to address https://github.com/AleoHQ/snarkOS/pull/1394#discussion_r766332446 first.

The purpose of this PR is to limit the number of times an unconfirmed block is serialized when propagated to `n` peers from `n` to `1`.

### DIFF
--- a/src/network/peers.rs
+++ b/src/network/peers.rs
@@ -585,7 +585,13 @@ impl<N: Network, E: Environment> Peers<N, E> {
     ///
     /// Sends the given message to every connected peer, excluding the sender.
     ///
-    async fn propagate(&self, sender: SocketAddr, message: Message<N, E>) {
+    async fn propagate(&self, sender: SocketAddr, mut message: Message<N, E>) {
+        // Perform ahead-of-time, non-blocking serialization just once for applicable objects.
+        if let Message::UnconfirmedBlock(_, _, ref mut data) = message {
+            let serialized_block = Data::serialize(data.clone()).await.expect("Block serialization is bugged");
+            let _ = std::mem::replace(data, Data::Buffer(serialized_block));
+        }
+
         // Iterate through all peers that are not the sender, sync node, or beacon node.
         for peer in self
             .connected_peers()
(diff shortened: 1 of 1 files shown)


ljedrz:
> Please revert this until a more thorough assessment is made.
> 
> The implication is that a malicious block from modified code can admit this change and poison internal `last_seen` filters if it is incorrect.
> 
> A simple fix is to check that `block_height == block.height()` and `block_hash == block.hash()` here, however this results in redundant checks. I can take a look at this change more closely later on to ensure we have it enforced at the lowest level possible.

@howardwu I checked the call sites, and it seems that the block height and block hash values are already filled in with computed values: https://github.com/AleoHQ/snarkOS/blob/testnet2/src/network/ledger.rs#L320; whenever we're sending an unconfirmed block, those values are either coming from our own node, or are replaced with computed values at ledger level, so they can be trusted.

niklaslong (approved):

ljedrz:
Rebased.

src/network/peers.rs
```diff
@@ -598,29 +598,23 @@ impl<N: Network, E: Environment> Peers<N, E> {
...
                         true
                     }
-                    Message::UnconfirmedBlock(_, _, ref mut data) => {
-                        let block = if let Data::Object(block) = data {
-                            block
-                        } else {
-                            panic!("Logic error: the block shouldn't have been serialized yet.");
-                        };
-
+                    Message::UnconfirmedBlock(block_height, block_hash, ref mut data) => {
```
howardwu: Same comment as in https://github.com/AleoHQ/snarkOS/pull/1394#discussion_r766332446

Please revert this until a more thorough assessment is made.

The implication is that a malicious block from modified code can admit this change and poison internal last_seen filters if it is incorrect.

A simple fix is to check that block_height == block.height() and block_hash == block.hash() here, however this results in redundant checks. I can take a look at this change more closely later on to ensure we have it enforced at the lowest level possible.

howardwu (changes requested):

src/network/peers.rs
niklaslong: It seems like these values are already being checked on receipt: https://github.com/AleoHQ/snarkOS/blob/95a156cda79b4ee411c237089ec3c1a1f9fde514/src/network/peers.rs#L1212

I haven't seen any other relevant flows for these values, so I'm assuming this change would be safe?

ljedrz:
Rebased.

ljedrz:
Rebased.

howardwu (approved):

src/network/peers.rs
howardwu: I still don't believe this is safe, but to keep things moving, I will merge it in. The current assumptions are fine, future assumptions won't necessarily know this flaw.

PR #1413, opened 2021-12-10
fix: ledger not loading the last block when it's just after the batch boundary

### LINKED ISSUE #1375: Node crashed and can't restart with message thread 'main' panicked at 'assertion failed: `(left == right)`
## 🐛 Bug Report

I'm run a miner for 3 days, today I found that it crashed with following message:

thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `AleoLocator { hrp: "al", data: Fp256(5282275773890731157584737813554920628646489886080748107602225735608185483169) }`,
 right: `AleoLocator { hrp: "al", data: Fp256(1884776062570810188821888786119248707234339675300075632591471437638835544945) }`', /root/snarkOS/storage/src/state/ledger.rs:212:9

then miner can not be restarted anylonger with the stored ledger, I must delete the whole ledger and re-sync.

## Expected Behavior

No crashing here.

## My Environment

Dedicated Server: Contabo, 96core, 256G 
 snarkOS Version: 11
Rust Version: 1.56.1
 Computer OS: ubuntu

HarukaMa:
## Motivation

There was a bug when loading the ledger from the disk where the node panic with something like this:

```
2021-12-10T08:15:42.383837Z DEBUG Validating the ledger up to block 501 (100%)
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `AleoLocator { hrp: "al", data: Fp256(2180413841631221483869035365579606912076225239550174444205460207040035170284) }`,
 right: `AleoLocator { hrp: "al", data: Fp256(6625747389475706972641387426803392449301567419990991643240503084430759780944) }`', /Users/MrX/src/snarkOS/storage/src/state/ledger.rs:212:9
```

After observing many logs I thought this might be related to the local ledger length: the ledger having `n * 501 + 1` blocks.

After debugging, it seems it is exactly the case:

Say we have a ledger with 502 blocks (0 - 501). The first batch loads blocks 0-500, and

https://github.com/AleoHQ/snarkOS/blob/dad0f8a6318c8133b4a87217bb9da488bfc50218/storage/src/state/ledger.rs#L191

will set the start block of the next batch to 501. However:

https://github.com/AleoHQ/snarkOS/blob/dad0f8a6318c8133b4a87217bb9da488bfc50218/storage/src/state/ledger.rs#L149

`501 == 501` so the last block will not get loaded. Fortunately,

https://github.com/AleoHQ/snarkOS/blob/dad0f8a6318c8133b4a87217bb9da488bfc50218/storage/src/state/ledger.rs#L210-L212

actually caught this error and panicked.

This PR made the ledger able to load that final block when starting.

## Test Plan

I prepared two local ledgers with 500 and 501 blocks and tested the new code against both ledger to make sure it still works.

## Related PRs

None

### DIFF
--- a/storage/src/state/ledger.rs
+++ b/storage/src/state/ledger.rs
@@ -146,7 +146,7 @@ impl<N: Network> LedgerState<N> {
         // Iterate and append each block hash from genesis to tip to validate ledger state.
         const INCREMENT: u32 = 500;
         let mut start_block_height = 0u32;
-        while start_block_height < latest_block_height {
+        while start_block_height <= latest_block_height {
             // Compute the end block height (inclusive) for this iteration.
             let end_block_height = std::cmp::min(start_block_height.saturating_add(INCREMENT), latest_block_height);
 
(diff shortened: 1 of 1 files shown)


HarukaMa:
Also, although

https://github.com/AleoHQ/snarkOS/blob/dad0f8a6318c8133b4a87217bb9da488bfc50218/storage/src/state/ledger.rs#L147

but the code is loading 501 blocks every pass. Should be fixed too but it's outside of the scope of this PR.

HarukaMa:
Fixed code formatting.

Maybe we can add a pre-commit hook to check it locally?

ljedrz (approved):
Great observation with the off-by-one! I think this could be the final fix to the ledger inconsistency issues. I tested it locally and it works; I'll propose this as the fix in the applicable issue, and let's just double-check that the users who have encountered it can start their node with it in place.

niklaslong (approved):

PR #1414, opened 2021-12-10
Relax the block spam safeguard

ljedrz:
This is a conservative change increasing the number of unconfirmed blocks received from a single peer considered to be spammy from `5` to `10`.

Cc https://github.com/AleoHQ/snarkOS/issues/1403; it might need to be relaxed further, especially if we increase peer counts.


niklaslong (commented):
I'll run this branch for a bit—seems fine at first glance 👍

ljedrz:
This is probably the number 1 cause of my peers being dropped.

niklaslong (approved):
Seems to be running smoothly!

howardwu (approved):

PR #1415, opened 2021-12-10
Enable more blocking threads

ljedrz:
Most of the work that's being done in the blocking threads is offloaded to `rayon` threadpools, so increasing their number will lead to greater average CPU utilization, with `rayon` being able to utilize its threadpools more aggressively.

The default value is `512`, so setting it to the number of available cores is still pretty conservative.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -38,7 +38,7 @@ fn main() -> Result<()> {
         .enable_all()
         .thread_stack_size(8 * 1024 * 1024)
         .worker_threads((num_cpus::get() / 8 * 2).max(1))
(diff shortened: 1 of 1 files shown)


niklaslong (approved):

unordered-set:
Tried it, still does not give a full CPU utilization, and I'm still using `--dev` option :)

ljedrz:
@unordered-set this PR is not intended to result in full CPU utilization, and it's unlikely to cause a difference in `--dev` conditions (as I'm assuming there's few peers involved). You should notice a difference with multiple peers (so that your node receives many messages), regardless of whether your node is mining or not.

howardwu (approved):

PR #1416, opened 2021-12-13
Tweak peer disconnect logic

niklaslong:
When the node is above its max peers limit, it disconnects from randomly selected addresses and adds them to the restricted peers list. There is no reason to restrict these peers as they haven't misbehaved in any way.


ljedrz (approved):
Sounds reasonable :+1:.

niklaslong:
Rebased.

howardwu:
The restriction is intentional, so that peers with misbehaving logic will not intentionally reconnect immediately, perhaps in an attempt to DOS the sync nodes. Note that the restriction is only in place for a few minutes.

PR #1418, opened 2021-12-13
Optimize FromBytes for BlockLocators

ljedrz:
This PR both shortens and optimizes the `FromBytes` impl for `BlockLocators`; it reduces the number of allocations, lookups, and removes checks for non-emptiness that would always return `true`.

### DIFF
--- a/storage/src/helpers/block_locators.rs
+++ b/storage/src/helpers/block_locators.rs
@@ -27,7 +27,7 @@ use snarkvm::{
     },
 };
 
-use rayon::iter::{IntoParallelRefIterator, ParallelIterator};
+use rayon::prelude::*;
 use serde::{de, ser, ser::SerializeStruct, Deserialize, Deserializer, Serialize, Serializer};
 use std::{collections::BTreeMap, ops::Deref};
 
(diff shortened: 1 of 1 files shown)


storage/src/helpers/block_locators.rs
```diff
@@ -87,26 +86,17 @@ impl<N: Network> FromBytes for BlockLocators<N> {
...
-                *header = Some(block_header);
+                block_headers_bytes.push((height, hash, Some(buffer)));
+            } else {
+                block_headers_bytes.push((height, hash, None));
             }
         }
 
+        let block_locators = block_headers_bytes
+            .into_par_iter()
+            .map(|(height, hash, bytes)| (height, (hash, bytes.map(|bytes| BlockHeader::<N>::read_le(&bytes[..]).unwrap()))))
```
niklaslong, line 97: Might be worth adding a note about why the `unwrap` is safe here?

niklaslong (approved):

storage/src/helpers/block_locators.rs
ljedrz: I only preserved the earlier/current behavior here; I don't know if it can be guaranteed.

ljedrz:
Rebased.

howardwu (approved):

PR #1422, opened 2021-12-13
(hot)fixed windows compile

fmu1337:
## Motivation

Help out community, meh

## Test Plan

I'm just got a lot of errors, cause termion isn't support windows OS.

Founded resolve there:
https://github.com/Byron/dua-cli/issues/2#issuecomment-630301477

## Related PRs

(Link your related PRs here)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,6 +23,9 @@ members = ["storage", "testing"]
 default = []
 test = []
 
+[patch.crates-io]
+termion = { git = "https://gitlab.redox-os.org/Jezza/termion.git", branch = "windows-support" }
+
 [dependencies]
 snarkvm = { git = "https://github.com/AleoHQ/snarkVM.git", rev = "471e03b" }
 #snarkvm = { path = "../snarkVM" }


ljedrz:
Thanks for the PR, @fmu1337! We actually shouldn't be using `termion` at all, as the `crossterm` backend is used instead. I tried tinkering with it for a short while, but for some reason it persists in `Cargo.lock` :thinking:.

fmu1337:
> Thanks for the PR, @fmu1337! We actually shouldn't be using `termion` at all, as the `crossterm` backend is used instead. I tried tinkering with it for a short while, but for some reason it persists in `Cargo.lock` 🤔.

This one?
https://github.com/AleoHQ/snarkOS/blob/bc518c3e29faae7323cef516e1e0242cc8ca6c09/Cargo.lock#L2471

Also I found 

.cargo\registry\src\github.com-1ecc6299db9ec823\tui-0.16.0\Cargo.toml 
```
145. [dependencies.termion]
146. version = "1.5"
147. optional = true
```

ljedrz:
> optional = true

Yeah, this is exactly why I'm surprised to see it in the lockfile; it's neither a default feature nor do we enable it manually.

ljedrz:
Upon further digging, I found an issue in `tui` that might be affecting this and filed https://github.com/fdehau/tui-rs/pull/571.

howardwu:
Closing in favor of #1483

PR #1423, opened 2021-12-14
Reduce max future block time to `80s`

niklaslong:
This effectively acts as a damper to block difficulty. 

Note: merging this will result in a hard-fork of the network.


apruden2008:
@niklaslong this will require all nodes to update correct? Nodes that don't update might potentially see their blocks rejected I assume.

scuwan:
> 

i think the nodes are not affected. They don't set the time in the future.

niklaslong:
Rebased.

PR #1427, opened 2021-12-15
Introduces pool functionality to network protocol

### LINKED ISSUE #1492: [Bug][Pool S/W] Failed to initialize a block from given inputs due to Invalid block header
## 🐛 Bug Report

I was testing the pool s/w in feat/mining-pool branch. [mining-pool latest codebase](https://github.com/AleoHQ/snarkOS/commit/eb4fd9133b63323ba90efd1be78f32452f8714cc)
It seems that Prover cannot found the block due to Invalid block header.
I'm not sure it was caused by lack of hash power on prover, but I cannot found to send a PoolResponse to the operator at all.
(FYI, I added some log to tracking block height, difficulty and nonce.)

```
2021-12-22T07:10:18.269644Z TRACE Received 'PoolRequest' from 95.214.55.117:4132
2021-12-22T07:10:18.341725Z DEBUG prover.process_pool_request()
2021-12-22 07:10:18.341817883 UTC --> header.mine_once_unchecked() height 71200, target 660876088681118, nonce hn1qqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqq5dj2mk
2021-12-22 07:10:20.887372256 UTC posw.mine_once_unchecked(), height 71200, block_header.difficulty_target 660876088681118, nonce hn1xsnrwyqmc5l976maevuhlv4ae0tjtyd39q8d5af8kyg4rdptn5zqg7p4mf
2021-12-22 07:10:20.887610660 UTC <-- header.mine_once_unchecked() height 71200, target 660876088681118, nonce hn1xsnrwyqmc5l976maevuhlv4ae0tjtyd39q8d5af8kyg4rdptn5zqg7p4mf
2021-12-22T07:10:20.925679Z DEBUG start new block, share_difficulty 9223372036854775807
2021-12-22T07:10:20.925693Z DEBUG block_header.difficulty_target : 660876088681118
2021-12-22T07:10:20.925695Z DEBUG block_header.nonce : hn1xsnrwyqmc5l976maevuhlv4ae0tjtyd39q8d5af8kyg4rdptn5zqg7p4mf
2021-12-22 07:10:20.925699932 UTC block.new(), Initialize a new block
2021-12-22 07:10:20.926271 UTC block.from() Construct the block. header.difficulty_target 660876088681118, height 71200
2021-12-22 07:10:20.938687287 UTC PoSW difficulty target is not met. Expected 660876088681118, found 730151207245470603
2021-12-22 07:10:20.938721456 UTC Invalid block header
2021-12-22T07:10:20.938821Z TRACE cannot found block Failed to initialize a block from given inputs
```

## Steps to Reproduce

1. Step 1
  pull snarkOS testnet2 branch
  checkout to feat/mining-pool
  build snarkOS
2. Step 2
  run pool operator and 2 prover nodes.
3. Step 3
  check the log

## Expected Behavior

  Prover find unconfirmed block and send a PoolResponse to the operator.

## Your Environment

- snarkOS : [mining-pool latest codebase](https://github.com/AleoHQ/snarkOS/commit/eb4fd9133b63323ba90efd1be78f32452f8714cc) 
- Rust version : rustc 1.57.0 (f1edd0429 2021-11-29) 
- OS : Ubuntu 20.04

jules:
## Motivation

This PR should introduce a rudimentary version of a mining pool, hosted directly on a node, including all changes needed in order for nodes to be able to connect and become pool workers.


xbeastx:
Trying to test this:
```
RUST_BACKTRACE=1 ./snarkos --miner aleo1... --trial --verbosity 2 --miningpool
...
2021-12-17T09:15:44.575148Z DEBUG Validating the ledger up to block 0 (0%)
2021-12-17T09:15:44.732102Z  INFO Ledger successfully loaded at block 0
2021-12-17T09:15:44.864644Z TRACE [Read-Only] Ledger successfully loaded at block 0
2021-12-17T09:15:44.872477Z  INFO Prover successfully initialized
thread 'main' panicked at 'Failed to start the node: IO error: lock hold by current process, acquire time 1639732544 acquiring thread 140411333648640: /root/.aleo/storage/prover-2/LOCK: No locks available', src/main.rs:52:28
stack backtrace:
   0: rust_begin_unwind
             at /rustc/f1edd0429582dd29cccacaf50fd134b05593bd9c/library/std/src/panicking.rs:517:5
   1: core::panicking::panic_fmt
             at /rustc/f1edd0429582dd29cccacaf50fd134b05593bd9c/library/core/src/panicking.rs:100:14
   2: core::result::unwrap_failed
             at /rustc/f1edd0429582dd29cccacaf50fd134b05593bd9c/library/core/src/result.rs:1616:5
   3: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
   4: tokio::park::thread::CachedParkThread::block_on
   5: tokio::runtime::thread_pool::ThreadPool::block_on
   6: tokio::runtime::Runtime::block_on
   7: snarkos::main
```

RocksDB opened twice?

jules:
@xbeastx good catch! I'll fix it up right away.

src/network/mining_pool.rs
```diff
@@ -0,0 +1,324 @@
...
+    current_template: RwLock<Option<BlockTemplate<N>>>,
+    /// Peripheral information on each known miner.
+    /// MinerInfo := (last_submitted, share_difficulty, shares_submitted_since_reset)
+    miner_info: RwLock<HashMap<Address<N>, (i64, u64, u32)>>,
+}
+
+impl<N: Network, E: Environment> MiningPool<N, E> {
+    /// Initializes a new instance of the mining pool.
+    pub async fn open<S: Storage, P: AsRef<Path> + Copy>(
+        tasks: &mut Tasks<JoinHandle<()>>,
```
ljedrz: `Tasks` don't need to be `mut` for any purpose

src/network/mining_pool.rs
```diff
@@ -0,0 +1,324 @@
...
+            ledger_reader,
+            ledger_router,
+            prover_router,
+            current_template: RwLock::new(None),
+            miner_info: RwLock::new(HashMap::new()),
+        });
+
+        if E::NODE_TYPE == NodeType::MiningPool {
+            // Initialize the handler for the mining pool.
+            {
```
ljedrz: there's no need for this extra scope

src/network/mining_pool.rs
```diff
@@ -0,0 +1,324 @@
...
+                    // Check that the block is relevant.
+                    if self.ledger_reader.latest_block_height().saturating_add(1) != block.height() {
+                        warn!("[ProposedBlock] Peer {} sent a stale candidate block.", peer_ip);
+                        return;
+                    }
+
+                    // Check that the block's coinbase transaction owner is the mining pool address.
+                    let records = match block.to_coinbase_transaction() {
+                        Ok(tx) => {
+                            let coinbase_records: Vec<Record<N>> = tx.to_records().collect();
```
ljedrz: can this allocation via `collect()` be avoided? the records are immediately iterated over right afterwards

src/network/mining_pool.rs
```diff
@@ -0,0 +1,324 @@
...
+                            let valid_owner = coinbase_records
+                                .iter()
+                                .map(|r| Some(r.owner()) == self.mining_pool_address)
+                                .fold(false, |a, b| a || b);
```
ljedrz:
```suggestion
                                .any(|r| Some(r.owner()) == self.mining_pool_address)
```

src/network/mining_pool.rs
jules: Well, I do need to use the records later down the line. I can just retrieve the records again down the line but it'll be a bit more bulky on the code side.

src/network/mining_pool.rs
```diff
@@ -0,0 +1,324 @@
...
+                    };
+
+                    let hash_difficulty = sha256d_to_u64(&proof_bytes);
+                    let mut info = self.miner_info.write().await;
+                    let share_difficulty = match info.get(&miner_address) {
+                        Some((_, share_difficulty, _)) => *share_difficulty,
+                        None => {
+                            let share_difficulty = current_template.difficulty_target.saturating_mul(50);
+                            self.miner_info
+                                .write()
```
ljedrz: `miner_info` is already locked for writing at this point

src/network/mining_pool.rs
```diff
@@ -0,0 +1,324 @@
...
+                            return;
+                        }
+                    };
+
+                    let hash_difficulty = sha256d_to_u64(&proof_bytes);
+                    let mut info = self.miner_info.write().await;
+                    let share_difficulty = match info.get(&miner_address) {
+                        Some((_, share_difficulty, _)) => *share_difficulty,
+                        None => {
+                            let share_difficulty = current_template.difficulty_target.saturating_mul(50);
```
ljedrz: what's the reasoning behind `50`?

src/network/mining_pool.rs
```diff
@@ -0,0 +1,324 @@
...
+                    }
+
+                    // Update the score for the miner.
+                    // TODO: add round stuff
+                    if let Err(error) = self.state.add_shares(block.height(), &miner_address, 1) {
+                        warn!("[ProposedBlock] {}", error);
+                    }
+
+                    // Update miner info for this miner.
+                    let mut miner_info = *info.get(&miner_address).expect("miner should have existing info");
```
ljedrz: `miner_info` can easily be mistaken for `info` (`self.miner_info` locked for writing) here

ljedrz: also, why is this not `get_mut`? there should be an `if let Some(x) = info.get_mut(y) { <modify the info> }` here

src/network/mining_pool.rs
jules: Nothing specifically. I felt like it would be low enough for most participants to be able to give one or two shares per block height without overloading the pool too much while I'm working on the round stuff.

src/network/mining_pool.rs
```diff
@@ -0,0 +1,324 @@
...
+                            }
+                        },
+                        None => {
+                            warn!("[ProposedBlock] Peer {} sent a candidate block with a missing proof.", peer_ip);
+                            return;
+                        }
+                    };
+
+                    let hash_difficulty = sha256d_to_u64(&proof_bytes);
+                    let mut info = self.miner_info.write().await;
```
ljedrz: this write lock is held for too long - the related operations should happen right afterwards, and it should be released as soon as it's no longer needed

src/network/mining_pool.rs
```diff
@@ -0,0 +1,324 @@
...
+                // Remove the shares for the given block height.
+                if let Err(error) = self.state.remove_shares(block_height) {
+                    warn!("[BlockHeightClear] {}", error);
+                }
+            }
+            MiningPoolRequest::GetCurrentBlockTemplate(peer_ip, address) => {
+                if let Some(current_template) = &*self.current_template.read().await {
+                    // Ensure this miner exists in the info list first, so we can get their share
+                    // difficulty.
+                    if let None = self.miner_info.read().await.get(&address) {
```
ljedrz:
```suggestion
                    if self.miner_info.read().await.get(&address).is_none() {
```

src/network/mining_pool.rs
```diff
@@ -0,0 +1,324 @@
...
+                        // If not, let's add them.
+                        let share_difficulty = current_template.difficulty_target.saturating_mul(50);
+                        self.miner_info
+                            .write()
+                            .await
+                            .insert(address, (chrono::Utc::now().timestamp(), share_difficulty, 0));
+                    }
```
ljedrz: after having digested this whole block: the [entry API](https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.entry) should be used here instead; the operations performed below should then be done within its block too

src/network/mining_pool.rs
```diff
@@ -0,0 +1,324 @@
...
+                } else {
+                    warn!("[ProposedBlock] No current block template exists");
+                }
+            }
+        }
+    }
+
+    async fn set_block_template(&self, recipient: Address<N>) -> Result<()> {
+        let unconfirmed_transactions = self.memory_pool.read().await.transactions();
+        let mut current_template = self.current_template.write().await;
```
ljedrz: is this write lock needed here already, i.e. should `prepare_block_template` happen with its contents frozen and other potential users not able to access it? otherwise it should only be acquired below it for update purposes

src/network/mining_pool.rs
ljedrz: ah ok, I missed this :+1:

xbeastx:
```2021-12-17T14:13:16.680941Z  INFO Mining pool successfully initialized
thread 'tokio-runtime-worker' panicked at 'A mining pool should have an available Aleo address at all times', .../snarkOS/src/network/mining_pool.rs:147:22
   0: rust_begin_unwind
             at /rustc/f1edd0429582dd29cccacaf50fd134b05593bd9c/library/std/src/panicking.rs:517:5
   1: core::panicking::panic_fmt
             at /rustc/f1edd0429582dd29cccacaf50fd134b05593bd9c/library/core/src/panicking.rs:100:14
   2: core::panicking::panic_display
             at /rustc/f1edd0429582dd29cccacaf50fd134b05593bd9c/library/core/src/panicking.rs:64:5
   3: core::option::expect_failed
             at /rustc/f1edd0429582dd29cccacaf50fd134b05593bd9c/library/core/src/option.rs:1638:5
   4: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
   5: tokio::runtime::task::core::CoreStage<T>::poll
   6: <core::panic::unwind_safe::AssertUnwindSafe<F> as core::ops::function::FnOnce<()>>::call_once
   7: tokio::runtime::task::harness::Harness<T,S>::poll
   8: std::thread::local::LocalKey<T>::with
   9: tokio::runtime::thread_pool::worker::Context::run_task
  10: tokio::runtime::thread_pool::worker::Context::run
  11: tokio::macros::scoped_tls::ScopedKey<T>::set
  12: tokio::runtime::thread_pool::worker::run
  13: tokio::runtime::task::harness::Harness<T,S>::poll
  14: tokio::runtime::blocking::pool::Inner::run
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.```

**Comment by jules** [comment]
@xbeastx did you specify a miner address when starting up a pool? It should've stopped you if not..

**Comment by xbeastx** [comment]
> @xbeastx did you specify a miner address when starting up a pool? It should've stopped you if not..

for sure, like this:
./snarkos --miner aleo1... --trial --verbosity 2 --miningpool

That is because miner_address stored only for Miner node https://github.com/AleoHQ/snarkOS/pull/1427/files#diff-af08c3181737aa5783b96dfd920cd5ef70829f46cd1b697bdb42414c97310e13L132

**Comment by jules** [comment]
@xbeastx thanks for the help! Just pushed a fix.

**Comment by ljedrz** [comment]
I haven't gone over the whole diff yet, but I'll be happy to re-review on Monday.

**Comment by howardwu** [comment]
Note to self: `Prover` and `Worker` can be safely unified as their currently-defined functionality is mutually exclusive.

This will also ensure we don't overutilize resources from multiple thread pools.

**Comment by howardwu** [comment]
Note to self: `Pool` should be renamed to `Operator` and a formal `Pool` struct should be introduced alongside a `Pools` struct (in the likeness of `Peer` to `Peers`).

This functionality is necessary to support delegated proving to multiple pools from a client node perspective.

**Comment by jules** [comment]
@howardwu unless I'm misunderstanding the last message, I don't really understand why there's a need to run multiple pools from the same client; it just seems unnecessary. Also, `Prover` and `Worker` may be unified but just take care that everything is gated correctly - the `Prover` struct will become quite a bit more complex so it's important to take care that all the functionality works as intended.

**Comment by howardwu** [comment]
The purpose here is to support proving in these pools beyond merely PoSW proofs and into program proofs. As mining will be phased out over the next decade, any designs around mining here should be explicitly designed as a subset of proving so that it will sunset without impacting the protocol.

On a more general note, a node should be able to observe the state of multiple pools.

**Comment by howardwu** [comment]
To coordinate, I'm going to attempt the `Prover` / `Worker` unification now, which should help to ensure thread resource utilization remains the same assumptions (which admittedly can still be improved)

File: src/node.rs [hunk 12]
```diff
@@ -91,23 +95,16 @@ impl Node {
...
-                (2, _, _, true, false, false) => self.start_server::<Testnet2, SyncNode<Testnet2>>().await,
-                (2, true, false, false, false, false) => self.start_server::<Testnet2, Miner<Testnet2>>().await,
-                (2, false, false, false, false, false) => self.start_server::<Testnet2, Client<Testnet2>>().await,
-                (2, true, false, false, false, true) => self.start_server::<Testnet2, Operator<Testnet2>>().await,
-                (2, true, false, false, true, false) => self.start_server::<Testnet2, Prover<Testnet2>>().await,
-                (2, true, true, false, false, false) => self.start_server::<Testnet2, MinerTrial<Testnet2>>().await,
-                (2, false, true, false, false, false) => self.start_server::<Testnet2, ClientTrial<Testnet2>>().await,
-                (2, true, true, false, false, true) => self.start_server::<Testnet2, OperatorTrial<Testnet2>>().await,
-                (2, true, true, false, true, false) => self.start_server::<Testnet2, ProverTrial<Testnet2>>().await,
+            None => match (self.network, &self.miner, &self.operator, &self.prover, self.trial, self.sync) {
```
ljedrz, line 97: at this point it might be a good idea to introduce a helper `struct` to hold these init values, or to have a method on `Node` that returns the type of the node based on these values; it's a bit hard to follow a tuple with 6 values in a `match` with many branches

howardwu:
^As of this commit, the operator and prover should now be correctly functioning on share difficulty (without modification to `block.difficulty_target()`). Meaning, blocks should be well-formed now while using a share difficulty target to compute scores.

Note that this has not been tested, so while in theory it is correct, in practice, will need integration testing to verify its behavior.

I also noticed a few corner cases (w.r.t. dishonest provers) that should be addressed.

src/network/message.rs
```diff
@@ -137,9 +146,12 @@ impl<N: Network, E: Environment> Message<N, E> {
             Self::PeerResponse(..) => 6,
             Self::Ping(..) => 7,
             Self::Pong(..) => 8,
-            Self::UnconfirmedBlock(..) => 9,
-            Self::UnconfirmedTransaction(..) => 10,
-            Self::Unused(..) => 11,
+            Self::PoolRegister(..) => 9,
+            Self::PoolRequest(..) => 10,
+            Self::PoolResponse(..) => 11,
+            Self::UnconfirmedBlock(..) => 12,
```
ljedrz: note: this change must occur in tandem with https://github.com/AleoHQ/snarkOS/pull/1466 or a different version update

src/network/operator.rs
```diff
@@ -0,0 +1,335 @@
...
+/// The predefined base share difficulty.
+const BASE_SHARE_DIFFICULTY: u64 = u64::MAX / 2u64.pow(1);
+
+///
+/// An operator for a program on a specific network in the node server.
+///
+#[derive(Debug)]
+pub struct Operator<N: Network, E: Environment> {
+    /// The address of the operator.
+    address: Option<Address<N>>,
```
ljedrz, line 77: can the operator not have an Aleo address? wondering if the `Option` is needed here (possibly for better compatibility with other methods)

jules: It needs one, as the coinbase should be attributed to it.

src/network/operator.rs
```diff
@@ -0,0 +1,335 @@
...
+        // Initialize an mpsc channel for sending requests to the `Operator` struct.
+        let (operator_router, mut operator_handler) = mpsc::channel(1024);
+
+        // Initialize the operator.
+        let operator = Arc::new(Self {
+            address,
+            local_ip,
+            state: Arc::new(OperatorState::open_writer::<S, P>(path)?),
+            block_template: RwLock::new(None),
+            provers: RwLock::new(HashMap::new()),
```
ljedrz: these probably can be initialized with `Default::default()`

src/network/operator.rs
```diff
@@ -0,0 +1,335 @@
...
+            operator_router,
+            memory_pool,
+            peers_router,
+            ledger_reader,
+            ledger_router,
+            prover_router,
+        });
+
+        if E::NODE_TYPE == NodeType::Operator {
```
ljedrz, line 142: (this also applies to line 144): is it possible to call `Operator::open` without the node being an operator? wondering if this predicate is needed

src/network/operator.rs
```diff
@@ -0,0 +1,335 @@
...
+                                .ledger_reader
+                                .get_block_template(recipient, E::COINBASE_IS_PUBLIC, &transactions, &mut thread_rng())
+                                .expect("Should be able to generate a block template");
+
+                            // Acquire the write lock to update the block template.
+                            *operator.block_template.write().await = Some(block_template);
+                        }
+
+                        // Sleep for `5` seconds.
+                        tokio::time::sleep(Duration::from_secs(5)).await;
```
ljedrz: perhaps this could be some new `const` for easier configurability?

src/network/operator.rs
```diff
@@ -0,0 +1,335 @@
...
+    }
+
+    ///
+    /// Performs the given `request` to the operator.
+    /// All requests must go through this `update`, so that a unified view is preserved.
+    ///
+    pub(super) async fn update(&self, request: OperatorRequest<N>) {
+        match request {
+            OperatorRequest::PoolRegister(peer_ip, address) => {
+                if let Some(block_template) = &*self.block_template.read().await {
```
ljedrz:
```suggestion
                if let Some(block_template) = self.block_template.read().await.cloned() {
```
in order not to hold the `read` for the duration of the whole block underneath

src/network/operator.rs
```diff
@@ -0,0 +1,335 @@
...
+                    let share_difficulty = self
+                        .provers
+                        .write()
+                        .await
+                        .entry(address)
+                        .or_insert((Instant::now(), BASE_SHARE_DIFFICULTY, 0))
+                        .1;
+
+                    // Route a `PoolRequest` to the peer.
+                    let message = Message::PoolRequest(share_difficulty, Data::Object(block_template.clone()));
```
ljedrz:
```suggestion
                    let message = Message::PoolRequest(share_difficulty, Data::Object(block_template));
```
if the suggestion above is applied

src/network/operator.rs
```diff
@@ -0,0 +1,335 @@
...
+                    if let Err(error) = self.peers_router.send(PeersRequest::MessageSend(peer_ip, message)).await {
+                        warn!("[PoolRequest] {}", error);
+                    }
+                } else {
+                    warn!("[PoolRegister] No current block template exists");
+                }
+            }
+            OperatorRequest::PoolResponse(peer_ip, block, prover_address) => {
+                if let Some(block_template) = &*self.block_template.read().await {
```
ljedrz:
```suggestion
                if let Some(block_template) = self.block_template.read().await.cloned() {
```
same reasoning as earlier

src/network/operator.rs
```diff
@@ -0,0 +1,335 @@
...
+                    // Check that the block is relevant.
+                    if self.ledger_reader.latest_block_height().saturating_add(1) != block.height() {
+                        warn!("[PoolResponse] Peer {} sent a stale candidate block.", peer_ip);
+                        return;
+                    }
```
ljedrz: note: this check could happen before the owning block

src/network/operator.rs
```diff
@@ -0,0 +1,335 @@
...
+
+                    // Retrieve the coinbase transaction records.
+                    let coinbase_records = match block.to_coinbase_transaction() {
+                        Ok(transaction) => {
+                            // Ensure the owner of the coinbase transaction in the block is the operator address.
+                            let coinbase_records: Vec<Record<N>> =
+                                transaction.to_records().filter(|r| Some(r.owner()) == self.address).collect();
+                            if coinbase_records.len() == 0 {
```
ljedrz:
```suggestion
                            if coinbase_records.is_empty() {
```
even if `clippy` doesn't complain about this right now, it will in the future

src/network/operator.rs
ljedrz: the same applies to at least some of the following blocks, but if all of them only need to happen in case of there being an applicable `Some(block_template)`, it's fine

src/network/operator.rs
```diff
@@ -0,0 +1,335 @@
...
+                            },
+                            None => {
+                                warn!("[PoolResponse] Peer {} sent a candidate block with a missing proof.", peer_ip);
+                                return;
+                            }
+                        };
+
+                        let proof_difficulty = sha256d_to_u64(&proof_bytes);
+                        let share_difficulty = {
+                            let mut provers = self.provers.write().await;
```
ljedrz: is it much more likely that the new prover will be inserted rather than looked up? otherwise I'd do a separate `read` and `write` in order to limit the number of writes

src/network/operator.rs
```diff
@@ -0,0 +1,335 @@
...
+                            match provers.get(&prover_address) {
+                                Some((_, share_difficulty, _)) => *share_difficulty,
+                                None => {
+                                    provers.insert(prover_address, (Instant::now(), BASE_SHARE_DIFFICULTY, 0));
```
ljedrz: general note: if we don't have to be precise to the nanosecond with the timestamp, I'd suggest calling the relatively costly `Instant::now()` before a lock is acquired in order to limit its scope

src/network/operator.rs
```diff
@@ -0,0 +1,335 @@
...
+                    // TODO: add round stuff
+                    // TODO: ensure shares can not be resubmitted
+                    if let Err(error) = self.state.add_shares(block.height(), &prover_address, 1) {
+                        error!("{}", error);
+                    }
+
+                    {
+                        // Update the internal state for this prover.
+                        let mut provers = self.provers.write().await;
+                        let mut prover = *provers.get_mut(&prover_address).expect("prover should have existing info");
```
ljedrz: this should be an `if let Some(ref mut prover) = self.provers.write().await.get_mut(&prover_address) { ... }` block for greater readability; there can still be an `else` branch with a `panic`

src/network/operator.rs
```diff
@@ -0,0 +1,335 @@
...
+                        prover.0 = Instant::now();
+                        prover.2 += 1;
+                        provers.insert(prover_address, prover);
```
ljedrz:
```suggestion
```
we are already modifying the applicable prover in the collection of provers; why insert it again?

src/network/prover.rs
```diff
@@ -102,23 +115,26 @@ impl<N: Network, E: Environment> Prover<N, E> {
     ) -> Result<Arc<Self>> {
         // Initialize an mpsc channel for sending requests to the `Prover` struct.
         let (prover_router, mut prover_handler) = mpsc::channel(1024);
-        // Initialize the prover pool.
-        let pool = ThreadPoolBuilder::new()
+        // Initialize the prover process.
```
ljedrz: nit: I'd call it a threadpool so as not to imply that it's a single process

src/network/prover.rs
```diff
@@ -138,9 +154,194 @@ impl<N: Network, E: Environment> Prover<N, E> {
...
 
+        // Initialize the miner, if the node type is a miner.
+        if E::NODE_TYPE == NodeType::Miner && prover.pool.is_none() {
+            Self::start_miner(tasks, prover.clone(), local_ip).await;
+        }
+
+        // Initialize the prover, if the node type is a prover.
+        if E::NODE_TYPE == NodeType::Prover && prover.pool.is_some() {
+            let prover = prover.clone();
+            task::spawn(async move {
```
ljedrz, line 164: this task needs to be passed to the task handler (`Tasks`) and it would be best if there was an `oneshot` channel that would get notified from within the spawned task that it's been spawned, in order to exit the block only once it's fully ready (like it's already done with other long-running tasks)

src/network/prover.rs
```diff
@@ -138,9 +154,194 @@ impl<N: Network, E: Environment> Prover<N, E> {
...
+                }
+            }
+        }
+    }
+
+    ///
+    /// Sends a `PoolRegister` message to the pool IP address.
+    ///
+    async fn send_pool_register(&self) {
+        if E::NODE_TYPE == NodeType::Prover {
```
ljedrz, line 229: since this method does something only if the node is a prover, should we have a `warn` here in an `else` branch to avoid it being called by non-provers? or do we expect this to be called as a noop by non-provers?

src/network/prover.rs
```diff
@@ -138,9 +154,194 @@ impl<N: Network, E: Environment> Prover<N, E> {
...
+                error!("Missing prover address. Please specify an Aleo address in order to prove");
+            }
+        }
+    }
+
+    ///
+    /// Processes a `PoolRequest` message from a pool operator.
+    ///
+    async fn process_pool_request(&self, operator_ip: SocketAddr, share_difficulty: u64, block_template: BlockTemplate<N>) {
+        if E::NODE_TYPE == NodeType::Prover {
```
ljedrz, line 250: same question as in https://github.com/AleoHQ/snarkOS/pull/1427/files#r772249412

src/network/prover.rs
```diff
@@ -138,9 +154,194 @@ impl<N: Network, E: Environment> Prover<N, E> {
...
+            if let Some(recipient) = self.address {
+                if let Some(pool_ip) = self.pool {
```
ljedrz, line 252: these 2 checks, unlike in `send_pool_register`, don't result in `error!` logs

src/network/prover.rs
```diff
@@ -138,9 +154,194 @@ impl<N: Network, E: Environment> Prover<N, E> {
...
+                    // Refuse work from any pool other than the registered one.
+                    if pool_ip == operator_ip {
+                        // If `terminator` is `false` and the status is not `Peering` or `Mining`
+                        // already, mine the next block.
+                        if !self.terminator.load(Ordering::SeqCst) && !self.status.is_peering() && !self.status.is_mining() {
```
ljedrz, line 257: nit, probably applicable to other spots as well; we might want to get the `status` once rather than obtain the related atomic twice via separate calls to `is_peering` and `is_mining`

src/network/server.rs
```diff
@@ -108,11 +153,20 @@ impl<N: Network, E: Environment> Server<N, E> {
...
             ledger.reader(),
             ledger.router(),
+            operator.router(),
             prover.router(),
         )
         .await;
         // Initialize a new instance of the heartbeat.
-        Self::initialize_heartbeat(&mut tasks, peers.router(), ledger.reader(), ledger.router(), prover.router()).await;
+        Self::initialize_heartbeat(
+            &mut tasks,
```
ljedrz, line 163: nit: `Tasks` never require mutability

Cargo.toml
```diff
@@ -24,11 +24,11 @@ default = []
 test = []
 
 [dependencies]
-snarkvm = { git = "https://github.com/AleoHQ/snarkVM.git", rev = "d3ae7e6" }
-#snarkvm = { path = "../snarkVM" }
+snarkvm = { git = "https://github.com/AleoHQ/snarkVM.git", rev = "d7f6943", features = [ "algorithms" ] }
```
niklaslong: This is still using the version of snarkVM with the adjustable difficulty?

storage/src/state/operator.rs
```diff
@@ -0,0 +1,187 @@
...
+    fn add_shares(&self, block_height: u32, address: &Address<N>, num_shares: u64) -> Result<()> {
+        if let Some(current_shares) = self.shares.get(&block_height)? {
+            let mut new_shares = current_shares.clone();
+
+            // Add the num shares for the address.
+            let address_entry = new_shares.entry(*address).or_insert(0);
+            *address_entry = address_entry.saturating_add(num_shares);
+
+            // Insert the shares for the address.
+            self.shares.insert(&block_height, &new_shares)?;
```
ljedrz: why not just `get_mut(&block_height)` and update that value instead of doing a lookup, cloning the value _and_ doing an `insert` at the end?

storage/src/state/operator.rs
```diff
@@ -0,0 +1,187 @@
...
+            Ok(())
+        } else {
```
ljedrz: this looks like material for `entry(...).or_insert(...)` rather than an `else`

src/network/operator.rs
niklaslong: Might not have come across the reason yet but why the `Option` in that case?

src/network/operator.rs
```diff
@@ -0,0 +1,335 @@
...
+#[derive(Debug)]
+pub struct Operator<N: Network, E: Environment> {
+    /// The address of the operator.
+    address: Option<Address<N>>,
+    /// The local address of this node.
+    local_ip: SocketAddr,
+    /// The state storage of the operator.
+    state: Arc<OperatorState<N>>,
+    /// The current block template that is being mined on by the operator.
+    block_template: RwLock<Option<BlockTemplate<N>>>,
```
niklaslong, line 85: Isn't this also needed at all times?

ljedrz (commented):
Left some comments and suggestions.

src/network/message.rs
jules: The merge target should've been set to `feat/v12`

ljedrz: it is, just double-checking :+1:

storage/src/state/operator.rs
jules: This syntax doesn't apply to the `DataMap` structure, unless I'm misunderstanding the manner in which to apply it here.

storage/src/state/operator.rs
jules: Again, doesn't seem to apply to the rocksdb `DataMap`.

jules:
Addressed some of the comments - I'll leave it to @howardwu to decide direction on the other points.

src/network/operator.rs
```diff
@@ -0,0 +1,336 @@
...
+                // Initialize an update loop for the block template.
+                let operator = operator.clone();
+                let (router, handler) = oneshot::channel();
+                tasks.append(task::spawn(async move {
+                    // Notify the outer function that the task is ready.
+                    let _ = router.send(());
+                    // TODO (julesdesmit): Add logic to the loop to retarget share difficulty.
+                    loop {
+                        // Determine if the current block template is stale.
+                        let is_template_stale = match &*operator.block_template.read().await {
```
niklaslong: Is a template ever stale other than the initial start?

src/network/prover.rs
```diff
@@ -69,8 +76,10 @@ pub enum ProverRequest<N: Network> {
 pub struct Prover<N: Network, E: Environment> {
     /// The state storage of the prover.
     state: Arc<ProverState<N>>,
-    /// The thread pool for the miner.
-    miner: Arc<ThreadPool>,
+    /// The Aleo address of the prover.
+    address: Option<Address<N>>,
```
niklaslong, line 80: Is this optional?

storage/src/state/operator.rs
```diff
@@ -0,0 +1,187 @@
...
+        let operator = Self {
+            shares: SharesState::open(storage)?,
+        };
+
+        info!("Operator successfully initialized");
+        Ok(operator)
+    }
+
+    /// Returns all the shares in storage.
+    pub fn to_shares(&self) -> Vec<(u32, HashMap<Address<N>, u64>)> {
```
niklaslong: Might be nice to introduce a type alias or a struct here?

storage/src/state/operator.rs
```diff
@@ -0,0 +1,187 @@
...
+    pub fn remove_coinbase_record(&self, commitment: &N::Commitment) -> Result<()> {
+        self.shares.remove_record(commitment)
+    }
+}
+
+#[derive(Clone, Debug)]
+#[allow(clippy::type_complexity)]
+struct SharesState<N: Network> {
+    /// The miner shares for each block height.
+    shares: DataMap<u32, HashMap<Address<N>, u64>>,
```
niklaslong: Ditto

storage/src/state/operator.rs
ljedrz: ah, fair, I mixed it up with a `HashMap` :+1:

storage/src/state/operator.rs
ljedrz: nope, it's fine; I was wondering if we could introduce `get_mut` to `DataMap`s, but sadly, we need to deserialize anyway, so there's little benefit to doing that

Cargo.toml
howardwu: It is no longer adjustable

src/network/operator.rs
howardwu: In the case the node isn't an operator, our design still necessitates its instantiation. We just disable all functionality. The overhead of this is minimal (a few hundred bytes in memory, and no CPU time afaik).

In the future, it could be desirable to have an operator live in tandem to a prover on one node, as their roles are mutually exclusive.

src/network/operator.rs
howardwu: The template should go stale every 20 seconds

src/network/operator.rs
howardwu: Yes it is possible

src/network/prover.rs
howardwu: Yes it is possible

crabvk:
Are you sure pools will improve current situation with mining? Because now over 70% of all blocks mined by a very few nodes. Thus, Aleo mining is centralized #1472

HarukaMa:
It looks `PoolResponse` didn't check for the validity of the proof before trying to determine if the block is valid for the network. (I suppose the deserialization method doesn't check for the validity.) It might be possible for a dishonest prover to just send back the block with a proof that only satisfy the share target but is completely irrelevant to the block. The operator doesn't check that and will just add the score for the share and only discover it later when it calls `is_valid()` (or not checking at all as the proof only satisfy the share diff and not the block diff).

src/network/prover.rs
jules: I don't think so, as with the other modules we instantiate everything but basically do nothing in case the incorrect node type is assumed. So we probably shouldn't log anything cause it may just end up causing confusion.

src/node.rs
jules: I've made an issue out of this, and left it out of the scope of this PR for now.

src/network/operator.rs
jules: As we do instantiate the component regardless of node type, so in some cases there will be no address available.

storage/src/state/operator.rs
jules: It won't really work that well cause we need it to be split into two different types for the `DataMap`. I can wrap the hashmap, but I don't really see the need..

jules:
@HarukaMa I will add a check for it, thanks for the suggestion!

HarukaMa:
The proof validation takes some milliseconds so if it's actually needed, there should be some anti-flooding mechanism to avoid DoS to the operator.

HarukaMa:
I've made a fake node to test the mining pool:
https://gist.github.com/HarukaMa/adbe7e9da751a3d4d1e9edeb59bbcf48

(Please change the server before testing. This node is hardcoded with version 11 to test against live testnet2.)

Using this "node", I've found that sometimes the operator won't refresh the block template in time and will send stale works to the prover. You could see logs like:

```
==========================================
Received Ping
Version: 11
Fork Depth: 4096
Node Type: 2
Status: 0
Block Hash: ec85f2383542699315079c66ba5201da9fd6fecea56574ccb6f32d7e1c9fb00a
Block Header: 546734496280ca863632b70b8...
  Previous Ledger Root: 546734496280ca863632b70b8e277c88eabb837976de3117176a2c8c22829111
  Transactions Root: a6765a288c73080b12679f5478e5e752beda43310f5785097c86b57f7515a200
  Metadata: 0d2c0100bbd7c461000000007140d7a47a3b080037da94db010000000000000000000000
    Height: 76813
    Timestamp: 1640290235
    Difficulty Target: 2317197751304305
    Cumulative Weight: 7978932791
  Nonce: af0d927c7bab55a34d42ab7b285c4039185cd21687327d3119ab03bce9548107
  Proof: 030000000000000004000...
==========================================
Received PoolRequest
Share Difficulty: 9223372036854775807
Block Template: c704000000000000aa77...
  Previous Block Hash: aa77c7ad7543d1370265bff3415d5fc739febd8b635eab5f6c9477861065590a
  Block Height: 76811
  Block Timestamp: 1640290203
  Difficulty Target: 2686606088468760
  Cumulative Weight: 7978911160
  Previous Ledger Root: 2f24e1eb47d510b7870d95ee78ceeff07ba20fa78fc71f199f605c1e49f8630a
  Transactions: 01008e20bb6d749540...
==========================================
```

Line 127 is where I try to flood the operator with invalid proofs. If the work is not stale, I can still make the operator check all of the proofs. It seems if the share is invalid, the nonce won't be recorded so the prover can just reuse a bad header.

src/network/operator.rs
```diff
@@ -0,0 +1,402 @@
...
+                    // Check that the block is relevant.
+                    if self.ledger_reader.latest_block_height().saturating_add(1) != block_header.height() {
+                        warn!("[PoolResponse] Peer {} sent a stale candidate block.", peer_ip);
+                        return;
+                    }
+
+                    // Ensure the nonce hasn't been seen before.
+                    if self.known_nonces.read().await.contains(&block_header.nonce()) {
+                        warn!("[PoolResponse] Peer {} sent a duplicate share", peer_ip);
+                        // TODO (julesdesmit): punish?
```
ljedrz, line 298: Just making sure this doesn't need to be addressed yet

src/network/operator.rs
```diff
@@ -0,0 +1,402 @@
...
+                            Some(proof) => proof,
+                            None => {
+                                warn!("[PoolResponse] proof is missing on header");
+                                return;
+                            }
+                        };
+
+                        // NOTE (julesdesmit): Unwraps are here for brevity's sake, and since we do
+                        // it exactly the same in snarkVM, I don't see why we shouldn't use them here.
+                        let inputs = vec![
```
ljedrz: nit: since we only borrow `inputs` later on, maybe this allocation can be avoided?

src/network/operator.rs
jules: I thought about it, but it makes more sense to tackle this from the P2P part of the system by implementing blacklisting logic or something like that. I don't think that logic really belongs here.

src/network/operator.rs
jules: The function currently expects a `&Vec` to be passed to it, so I can't circumvent it for now.

src/network/prover.rs
```diff
@@ -236,6 +207,12 @@ impl<N: Network, E: Environment> Prover<N, E> {
     ///
     pub(super) async fn update(&self, request: ProverRequest<N>) {
         match request {
+            ProverRequest::PoolRequest(operator_ip, share_difficulty, block_template) => {
+                // Process the pool request message.
+                self.process_pool_request(operator_ip, share_difficulty, block_template).await;
+                // Proceed to register the prover to receive the next block template.
+                // self.send_pool_register().await;
```
ljedrz: Making sure this wasn't missed; if not, this could use a `TODO`

src/network/prover.rs
```diff
@@ -250,6 +227,104 @@ impl<N: Network, E: Environment> Prover<N, E> {
...
+                            let result = task::spawn_blocking(move || {
+                                thread_pool.install(move || {
+                                    loop {
+                                        let block_header =
+                                            BlockHeader::mine_once_unchecked(&block_template, &terminator, &mut thread_rng())?;
+
+                                        // Verify the proof against the share difficulty target.
+                                        if let Some(proof) = block_header.proof() {
+                                            // Construct the inputs.
+                                            let inputs = vec![
```
ljedrz: nit: same as before, we can potentially avoid allocating here

src/network/prover.rs
```diff
@@ -272,4 +347,85 @@ impl<N: Network, E: Environment> Prover<N, E> {
...
+                                        if let Err(error) = ledger_router.send(request).await {
+                                            warn!("Failed to broadcast mined block - {}", error);
+                                        }
+                                    }
+                                    Ok(Err(error)) | Err(error) => trace!("{}", error),
+                                }
+                            }));
+                        }
+                        // Sleep for 2 seconds.
+                        tokio::time::sleep(std::time::Duration::from_secs(2)).await;
```
ljedrz: nit: could probably be `const`ified for greater visibility; could be done in a follow-up

src/network/prover.rs
jules: It can be deleted - it caused an infinite loop of messages before and I tentatively commented it out to see what would happen.

src/network/server.rs
```diff
@@ -90,15 +100,50 @@ impl<N: Network, E: Environment> Server<N, E> {
...
+            local_ip,
+            prover.memory_pool(),
+            peers.router(),
+            ledger.reader(),
+            ledger.router(),
+            prover.router(),
+        )
+        .await?;
+
+        // TODO (howardwu): This is a hack for the prover.
```
ljedrz, line 127: is the hack that possibly an extra connection is attempted instead of checking for an existing connection first?

jules: @howardwu

src/node.rs
```diff
@@ -115,6 +122,19 @@ impl Node {
         }
     }
 
+    /// Returns the storage path of the operator.
+    pub(crate) fn operator_storage_path(&self, _local_ip: SocketAddr) -> PathBuf {
```
ljedrz, line 126: future TODO: it would probably be best to have methods like these implemented in the context of some trait that applies to different types of nodes

src/network/operator.rs
ljedrz: it can be done later; in general there's no reason to use `&Vec<T>` instead of `&[T]`, so it could be done in a pass targeting such APIs

ljedrz (approved):
Re-reviewed.

howardwu:
@julesdesmit I recommend bumping the message version to `13` in this PR, and updating the snarkVM `rev`.

howardwu:
@julesdesmit I think the concept of `operator.round` might be vulnerable to a reorg attack. One could in theory connect to a pool operator, intentionally mine a fork that's heavier, cause it to rebase, and "de-sync" the concept of `round` from `block_height`. The risk is one could change the expected payout for past participants for the wrong round, based on current logic afaik. I'm going to proactively switch `round` to `block_height`, however, please feel free to chime in and happy to revert if I'm missing something conceptually.

howardwu:
@ all - I think the operator might be susceptible to a reward withholding attack as well. We currently choose to store the coinbase record that the prover reports back with (and check just that the owner is the operator), however the value balance isn't validated. This means, a prover could change the coinbase record to still have the original operator address, but with a value balance of 0.

I think there are two ways to resolve this:
1. We recompute the expected reward based on the transactions in the block, and the expected coinbase reward.
2. We could store the originally generated coinbase record when the operator computes the block template.

While 1) is more precise (as a prover may choose to restructure which transactions to include in a block, thereby changing the expected reward), 2) is a far simpler solution to implement (and it is reliable). I will defer to 2) for now. Please feel free to chime in if others have differing opinions.

howardwu:
I'm adding in a V12-compliance check that blocks about 100,000 contain non-hiding proofs.

howardwu:
@julesdesmit I think based on the new logic, we no longer need the method `Block::from_unchecked`. I'm going to preliminarily remove it, happy to add it back in if there was something I missed in my review.

howardwu:
I've added additional block header checks to ensure they match the given inputs from the block template.

howardwu:
I'm upgrading the share tracker key from a block height to a tuple `(block height, coinbase record commitment)` to account for forks.

PR #1431, opened 2021-12-15
Speed up storage validation for block headers

ljedrz:
This PR not only reduces the overall storage validation time to a **third** of the current value, but it also checks for the existence of all the block headers, as opposed to just a half of them.


niklaslong (approved):

ljedrz:
In addition, in order to increase validation performance further, we should consider either increasing the value of `INCREMENT` or making it dependent on the overall number of blocks.

PR #1433, opened 2021-12-15
fix #1432

### LINKED ISSUE #1432: Break the outer loop in Ledger::add_block
```
for requests in self.block_requests.read().await.values() {
    for request in requests.keys() {
        // If the unconfirmed block conflicts with a requested block on a fork, skip.
        if request.block_height == unconfirmed_block_height {
            if let Some(requested_block_hash) = request.block_hash {
                if unconfirmed_block.hash() != requested_block_hash {
                    is_block_on_fork = true;
                    break;
                }
            }
        }
    }
}
```
After the variable `is_block_on_fork` is set to `true`, the execution of the first layer `for` loop is meaningless

kaola526:
After the variable "is_block_on_fork" is set to true, exit the outer loop

### DIFF
--- a/src/network/ledger.rs
+++ b/src/network/ledger.rs
@@ -561,6 +561,9 @@ impl<N: Network, E: Environment> Ledger<N, E> {
                         }
                     }
                 }
+                if is_block_on_fork {
+                    break;
+                }
             }
 
             // If the unconfirmed block is not on a fork, attempt to add it as the next block.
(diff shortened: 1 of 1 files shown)


ljedrz:
Makes sense; however, in order to make this more idiomatic, we should add a label to the outer loop and use it instead of duplicating the check. Example: https://doc.rust-lang.org/rust-by-example/flow_control/loop/nested.html.

ljedrz (approved):
LGTM :ok_hand:

PR #1436, opened 2021-12-15
Speed up CircularMap

ljedrz:
This PR aims to improve the performance of `CircularMap` by using a `HashMap` in addition to the underlying `CircularQueue`, so that lookups are no longer `O(n)`.

What is more, this PR fixes the currently invalid behavior where the underlying circular queue's order is reversed every time a call to `CircularMap::remove` is performed, by using `CircularQueue::asc_iter` instead of `::iter`.

### DIFF
--- a/src/helpers/circular_map.rs
+++ b/src/helpers/circular_map.rs
@@ -16,24 +16,24 @@
 
 use circular_queue::CircularQueue;
 
+use std::{collections::HashMap, hash::Hash};
+
 ///
 /// A helper struct to maintain a bounded number of elements in a map.
 ///
-/// Note that this implementation can likely be improved significantly.
-/// Due to the fact that the number of elements stored in the map is small,
-/// the performance overhead should not be significant.
-///
 #[derive(Clone, Debug)]
-pub struct CircularMap<K: Clone + PartialEq, V: Clone, const N: u32> {
-    queue: CircularQueue<(K, V)>,
+pub struct CircularMap<K: Clone + PartialEq + Eq + Hash, V: Clone, const N: u32> {
+    map: HashMap<K, V>,
+    queue: CircularQueue<Option<K>>,
 }
 
-impl<K: Clone + PartialEq, V: Clone, const N: u32> CircularMap<K, V, N> {
+impl<K: Clone + PartialEq + Eq + Hash, V: Clone, const N: u32> CircularMap<K, V, N> {
     ///
     /// Initializes a new instance of a circular map, of pre-defined size.
     ///
     pub fn new() -> Self {
         Self {
+            map: HashMap::with_capacity(N as usize),
             queue: CircularQueue::with_capacity(N as usize),
         }
     }
@@ -42,38 +42,28 @@ impl<K: Clone + PartialEq, V: Clone, const N: u32> CircularMap<K, V, N> {
     /// Returns `true` if the circular map is empty.
     ///
     pub fn is_empty(&self) -> bool {
-        self.queue.is_empty()
+        self.map.is_empty()
     }
 
     ///
     /// Returns the number of key-value pairs in the circular map.
     ///
     pub fn len(&self) -> usize {
-        self.queue.len()
-    }
-
-    ///
-    /// Returns the maximum capacity of the circular map.
-    ///
-    pub fn capacity(&self) -> usize {
-        self.queue.capacity()
+        self.map.len()
     }
 
     ///
     /// Returns `true` if the given key exists in the circular map.
     ///
     pub fn contains_key(&self, key: &K) -> bool {
-        self.queue.iter().any(|(k, _)| k == key)
+        self.map.contains_key(key)
     }
 
     ///
     /// Returns the value for the given key from the map, if it exists.
     ///
     pub fn get(&self, key: &K) -> Option<&V> {
-        match self.queue.iter().find(|(k, _)| k == key) {
-            Some((_, value)) => Some(value),
-            None => None,
-        }
+        self.map.get(key)
     }
 
     ///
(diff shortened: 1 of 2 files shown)


ljedrz:
note: the performance of `CircularMap::remove` remains weak; I'll need to test it a bit more, but I think the solution here will be to just _not_ call it at all and instead rely on the map being overwritten with newer objects.

ljedrz:
Just added a commit with `remove` unused for test purposes.

ljedrz:
Rebased, removed `CircularMap::{capacity, remove}`, and added `CircularMap::clear`.

niklaslong (approved):
LGTM 👍

src/network/ledger.rs
```diff
@@ -622,20 +620,17 @@ impl<N: Network, E: Environment> Ledger<N, E> {
...
                 // Set the terminator bit to `true` to ensure the miner resets state.
                 self.terminator.store(true, Ordering::SeqCst);
 
-                // Lock unconfirmed_blocks for further processing.
-                let mut unconfirmed_blocks = self.unconfirmed_blocks.write().await;
-
-                // Ensure the removed blocks are not in the unconfirmed blocks.
-                for removed_block in removed_blocks {
-                    unconfirmed_blocks.remove(&removed_block.previous_block_hash());
-                }
```
howardwu: Please keep this logic. While we have minimized the "stuck on wrong fork due to fast-forward with incorrect next block" case (which this logic mitigates), it is better to keep it as it is critical for syncing.

src/network/ledger.rs
```diff
@@ -565,8 +565,8 @@ impl<N: Network, E: Environment> Ledger<N, E> {
 
             // If the unconfirmed block is not on a fork, attempt to add it as the next block.
             match is_block_on_fork {
-                // Filter out the undesirable unconfirmed blocks, if it exists.
-                true => self.unconfirmed_blocks.write().await.remove(&unconfirmed_previous_block_hash),
```
howardwu: Please keep this logic. See below for explanation

src/helpers/circular_map.rs
```diff
@@ -82,28 +76,27 @@ impl<K: Clone + PartialEq, V: Clone, const N: u32> CircularMap<K, V, N> {
...
         } else {
             false
         }
     }
 
     ///
-    /// Removes the key-value pair for the given key from the circular map.
+    /// Removes all the entries from the circular map.
     ///
-    pub fn remove(&mut self, key: &K) {
```
howardwu, line 95: Please keep an implementation of this.

howardwu (changes requested):

ljedrz:
In that case I'll look into ways of speeding up the `remove` operation (perhaps it can be done "wholesale"), as it's quite slow.

src/helpers/circular_map.rs
```diff
@@ -82,7 +72,11 @@ impl<K: Clone + PartialEq, V: Clone, const N: u32> CircularMap<K, V, N> {
     ///
     pub fn insert(&mut self, key: K, value: V) -> bool {
         if !self.contains_key(&key) {
-            self.queue.push((key, value));
+            if let Some(Some(popped)) = self.queue.push(Some(key.clone())) {
+                self.map.remove(&popped);
+            }
+            self.map.insert(key, value);
```
ljedrz, line 78: performance note: since we are using this collection only for unconfirmed blocks (where the key is a block hash and the value is the corresponding block), this method could be made faster by doing the following:
```
if self.map.insert(key, value).is_none() {
    if let Some(Some(popped)) = self.queue.push(Some(key.clone())) {
        self.map.remove(&popped);
    }
    // insertion already happened
```
which avoids a lookup; overwriting the pre-existing key-value pair would change nothing, as a specific block hash can only correspond to a single block.

src/helpers/circular_map.rs
```diff
@@ -93,17 +87,26 @@ impl<K: Clone + PartialEq, V: Clone, const N: u32> CircularMap<K, V, N> {
...
     pub fn remove(&mut self, key: &K) {
-        let mut queue = CircularQueue::with_capacity(N as usize);
-        for (k, v) in self.queue.iter() {
-            if key != k {
-                queue.push((k.clone(), v.clone()));
+        if self.map.remove(key).is_some() {
+            for k in self.queue.asc_iter_mut() {
+                if k.as_ref() == Some(key) {
+                    *k = None;
+                    return;
```
ljedrz, line 94: this is the reason why the circular queue now contains `Option<K>`: this allows us to remove the desired keys without having to rewrite the entire queue with every removal

src/helpers/circular_map.rs
```diff
@@ -93,17 +87,26 @@ impl<K: Clone + PartialEq, V: Clone, const N: u32> CircularMap<K, V, N> {
...
+                }
             }
         }
-        self.queue = queue;
+    }
+
+    ///
+    /// Removes all the entries from the circular map.
+    ///
+    pub fn clear(&mut self) {
```
ljedrz, line 103: using this method instead of replacing the whole map with `Default::default` reuses the existing allocations instead of re-allocating

ljedrz:
@howardwu `test_block_requests_case_2a` appears to be flaky - I can reproduce the CI failure in `testnet2` as well from time to time.

src/helpers/circular_map.rs
ljedrz: I could also use the entry API here; it's not a blocker, though

howardwu:
Discussed offline.

The fix is a simple one and I will get that out shortly.

src/helpers/circular_map.rs
ljedrz: FWIW, the entry API solution would be:
```
pub fn insert(&mut self, key: K, value: V) -> bool {
    match self.map.entry(key.clone()) {
        Entry::Occupied(_) => return false,
        Entry::Vacant(entry) => {
            entry.insert(value);
        }
    }

    if let Some(Some(popped)) = self.queue.push(Some(key)) {
        self.map.remove(&popped);
    }

    true
}
```
however, I would prefer to introduce some benchmarks before potentially changing it.

howardwu (approved):

PR #1437, opened 2021-12-16
fix bug: stack overflow

scuwan:
## Motivation

fix bug.
 just increase the stack size can solve the problem， i've been running tests for over 48 hours.
```
2021-12-14T22:35:22.430245Z TRACE Received 'UnconfirmedBlock' from 65.108.78.154:4132
2021-12-14T22:35:22.430342Z TRACE Skipping 'UnconfirmedBlock 42339' from 65.108.78.154:4132
2021-12-14T22:35:22.430356Z TRACE Received 'BlockRequest' from 65.108.98.24:4132

thread '<unknown>' has overflowed its stack
fatal runtime error: stack overflow
./run.sh: line 9: 61464 Aborted                 (core dumped) ./snarkos --miner aleo1a0xqxfcsnnktd2c9knemw0qhq6jwkpwdvthz7cxjwk6z6npu8g9qtsh698 --verbosity 2 --trial --node 0.0.0.0:14539 --rpc 0.0.0.0:13539
```
## Test Plan

(Write your test plan here)

## Related PRs

(Link your related PRs here)

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -36,14 +36,14 @@ fn main() -> Result<()> {
     // Initialize the runtime configuration.
     let runtime = runtime::Builder::new_multi_thread()
         .enable_all()
-        .thread_stack_size(8 * 1024 * 1024)
+        .thread_stack_size(64 * 1024 * 1024)
         .worker_threads((num_cpus::get() / 8 * 2).max(1))
         .max_blocking_threads(num_cpus::get())
         .build()?;
 
     // Initialize the parallelization parameters.
     rayon::ThreadPoolBuilder::new()
-        .stack_size(8 * 1024 * 1024)
+        .stack_size(64 * 1024 * 1024)
         .num_threads((num_cpus::get() / 8 * 5).max(1))
         .build_global()
         .unwrap();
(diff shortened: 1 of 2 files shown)


scuwan:
i think the problem is in the runtime stack，but just adjust the stack of runtime and rayon global pool

ljedrz:
We shouldn't keep increasing the already high stack sizes; can you provide more information on your setup and the conditions under which the stack overflow occurred?

ljedrz (changes requested):
We neither want to bump stack sizes to 64MB nor change `MAXIMUM_FUTURE_BLOCK_TIME` ad-hoc from 120s to 2s.

scuwan:
> We shouldn't keep increasing the already high stack sizes; can you provide more information on your setup and the conditions under which the stack overflow occurred?

i set RUST_BACKTRACE=full, but no more information. but，increase the runtime stack size，it disappeared. maybe because I added peers. but i still think there are potential risks.

ljedrz:
> maybe because I added peers

how many peers have you added? were there any other tweaks? in general, even though bumping the stack size "solves" this, we should avoid doing that, as it just covers up any underlying issues.

PR #1443, opened 2021-12-16
strict future block time

scuwan:
## Motivation

Attackers can take advantage of the loose time limit to attack the network. e.g
1. miner a mined a block， weight(24000) ，block time diff（2s），ignore the propagation time
2. attacker set the block timestamp_a(time.now()) + MAXIMUM_FUTURE_BLOCK_TIME，weight(~4000)，win the block with block time diff 122s（2s）
3. the next block weight will be 8w （assume block time is "timestamp_a + 10s"）
4. the sencond block weight will be 160w (assume block time is "timestamp_a + 10 + 100s")
5. the third block weight will be 3200w
6. then the network will freeze

I have observed similar things happening on the network.
Miners should be responsible for their local time（ntp）
## Test Plan

(Write your test plan here)

## Related PRs

(Link your related PRs here)


ljedrz:
Closing as a duplicate of https://github.com/AleoHQ/snarkOS/pull/1423; feel free to discuss the time limit there.

PR #1445, opened 2021-12-16
Reduce LedgerState locking, merge 2 collections

ljedrz:
I first wanted to just reduce the scope of some read locks, but in the end it occurred to me that 2 related collections could (and should, as they are always used in tandem) be collapsed into a single one.


niklaslong (approved):
Seems reasonable 👍

howardwu (approved):

PR #1447, opened 2021-12-16
Speed up network message serialization & propagation

ljedrz:
This is achieved by:
- using `Bytes` instead of `Vec<u8>` inside `Data` (shallow copying for message propagation)
- writing directly into a `W: Write` instead of using intermediate vectors
- not using an intermediate vector to calculate the length of the outbound message

I've done some testing and it seems to work fine, but I welcome thorough reviews, as it's pretty low-level stuff.

### DIFF
--- a/src/network/message.rs
+++ b/src/network/message.rs
@@ -210,7 +210,7 @@ impl<N: Network, E: Environment> Message<N, E> {
     /// Serializes the given message into bytes.
     #[inline]
     pub fn serialize(&self) -> Result<Vec<u8>> {
-        Ok([self.id().to_le_bytes().to_vec(), self.data()?].concat())
+        Ok([&self.id().to_le_bytes()[..], &self.data()?].concat())
     }
 
     /// Deserializes the given buffer into a message.
--- a/src/network/message.rs
+++ b/src/network/message.rs
@@ -21,10 +21,10 @@ use crate::{
 use snarkos_storage::BlockLocators;
 use snarkvm::{dpc::posw::PoSWProof, prelude::*};
 
-use ::bytes::{Buf, Bytes, BytesMut};
+use ::bytes::{Buf, BufMut, Bytes, BytesMut};
 use anyhow::{anyhow, Result};
 use serde::{de::DeserializeOwned, Serialize};
-use std::{marker::PhantomData, net::SocketAddr};
+use std::{io::Write, marker::PhantomData, net::SocketAddr};
 use tokio::task;
 use tokio_util::codec::{Decoder, Encoder};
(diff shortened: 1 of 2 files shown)


src/network/message.rs
```diff
@@ -250,28 +258,18 @@ impl<N: Network, E: Environment> Encoder<Message<N, E>> for Message<N, E> {
...
+        dst.extend_from_slice(&0u32.to_le_bytes());
 
-        // Ensure the message does not exceed the maximum length limit.
-        if buffer.len() > E::MAXIMUM_MESSAGE_SIZE {
-            return Err(std::io::Error::new(
-                std::io::ErrorKind::InvalidData,
-                format!("Frame of length {} is too large.", buffer.len()),
-            )
-            .into());
-        }
```
ljedrz, line 295: this check is not needed for outbound messages - it is only required for the inbound ones (as those are the only ones that can be maliciously crafted to be too long)

niklaslong (commented):
The code looks good to me, will run and test 👍

niklaslong (approved):

src/network/message.rs
```diff
@@ -250,28 +258,18 @@ impl<N: Network, E: Environment> Encoder<Message<N, E>> for Message<N, E> {
     type Error = anyhow::Error;
 
     fn encode(&mut self, message: Message<N, E>, dst: &mut BytesMut) -> Result<(), Self::Error> {
-        // Serialize the message into a buffer.
-        let buffer = message.serialize()?;
+        // Prepare the room for the length of the payload.
+        dst.extend_from_slice(&0u32.to_le_bytes());
```
howardwu, line 294: Can you help me understand what this line is doing?

ljedrz: In order to avoid an intermediate allocation of a vector for the serialized message (whose length could then be known at this point, as it's done right now), we prepare 4B of room in advance, so that we can calculate that length _after_ it has been written to `dst` (after those 4B) and then overwrite those first 4B with the length (line 271).

It preserves the current product of serialization while avoiding allocations.

ljedrz: That line is the same as `dst.extend_from_slice(&[0, 0, 0, 0]);`

ljedrz:
Rebased.

ljedrz:
Rebased. @howardwu this PR removes several allocations for every message sent and removes deep clones when propagating messages, so I'd consider it likely to noticeably impact performance (positively), especially with many active connections.

ljedrz:
Rebased.

howardwu (approved):

PR #1448, opened 2021-12-16
Remove an allocation from Decoder::decode impl for Message

ljedrz:
This removes an intermediate vector allocation whenever a network message is received.


niklaslong (approved):
👍

howardwu (approved):

PR #1454, opened 2021-12-17
Revert "Speed up storage validation for block headers"

howardwu:
Reverts AleoHQ/snarkOS#1431

This PR is the right idea, but requires some touch ups to adhere our desired design goals. Notably:
- the `get_block_header_count` method should have **private** visibility in the `BlockState` struct
- we should remove exposing the `get_block_header_count` as it could be confused for "latest_block_height"
    - Remark: while in most cases, the two should be equivalent (off-by-one), corrupted state would imply a mismatch.
- **Most importantly**, this PR makes validation **less restrictive**/"**less safe**" as we no longer synthesize the block header and implicitly verify that the block header SNARK proof is valid

### DIFF
--- a/storage/src/state/ledger.rs
+++ b/storage/src/state/ledger.rs
@@ -28,7 +28,7 @@ use rand::{CryptoRng, Rng};
 use rayon::prelude::*;
 use serde::{Deserialize, Serialize};
 use std::{
-    collections::{BTreeMap, HashSet},
+    collections::BTreeMap,
     path::Path,
     sync::{
         atomic::{AtomicBool, AtomicU32, Ordering},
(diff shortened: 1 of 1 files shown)


ljedrz:
Perhaps we could make this check on-demand instead of it being mandatory whenever a node is started (in which case it could check all the headers)? It takes quite some time already, and that amount will grow with the number of blocks.

As for security, my reasoning was that all the headers stored in the database had initially either been produced by our own node (in which case we can trust the proof) or deserialized from another node (so the proof had already been verified).

howardwu:
@ljedrz would you be open to reviewing the `BlockHeader` struct implementation in snarkVM `testnet2` branch?

I think you will find a number of low-hanging performance improvements within 30 minutes or 1 hour of review.

You can also run `cd dpc && cargo bench --bench block`, which will show you the performance for serializing and deserializing `Block`, `BlockHeader`, `Transactions`, `Transaction`, `Transition`, etc. with `ToBytes/FromBytes`, `bincode`, and `serde_json`

I think improving `BlockHeader` performance could remedy some of the discussion here.

howardwu:
Unrelated but related: you'll notice `FromBytes` is always double the speed of `bincode::deserialize` which may be something we want to switch to in snarkOS for the `Message` enum

For `BlockResponse`, `Ping`, `UnconfirmedBlock`, etc, these would be significantly faster

apruden2008:
Thanks @howardwu for review. @ljedrz did we make any similar changes or does this motivate work in other areas? 

I can also create an issue for the optimization you describe here.

ljedrz (approved):

ljedrz:
@apruden2008 if you mean any changes where deserialization is omitted, then no; I did it here only due to storage entries already having undergone validation earlier on (as previously described [here](https://github.com/AleoHQ/snarkOS/pull/1454#issuecomment-996635662)). For those reasons I consider this extra check to be a bit of an overkill, but I don't mind trying to improve performance elsewhere before considering this route, so I'm ok with reverting this.

PR #1461, opened 2021-12-17
Refactor Ledger::update_block_requests (2)

ljedrz:
Builds on https://github.com/AleoHQ/snarkOS/pull/1459, only the last commit is new. Can probably be made a bit tidier, but in this shape it is already nearly completely compatible with [#1372](https://github.com/AleoHQ/snarkOS/pull/1372) (the tests will come in a later PR). As with the earlier PRs, there should be no functional difference from the current logic.

This is the 4th PR in the Ledger::update_block_requests refactoring series (#1456).


howardwu:
This is very nice, thank you @ljedrz . I will proceed to unify this with the minor changes I made.

howardwu (approved):

PR #1462, opened 2021-12-18
feat: choose sync target peer randomly

dtynn:
## Motivation
choose one single sync target from a batch of peers with the same maximum weight and the same height

btw, the Cargo.lock is outdated somehow, as the #1455 is already merged


dtynn:
seems that sync target selection has been rafactored.
close the pr for now

PR #1463, opened 2021-12-18
Minor updates and cleanups

howardwu:
## Motivation

- Increases ledger validation to increments of 1000 blocks
- Small tweaks in `latest_block_hashes_and_headers`
- Reenable precommit hook (that was my bad!)

### DIFF
--- a/storage/src/state/ledger.rs
+++ b/storage/src/state/ledger.rs
@@ -464,15 +464,15 @@ impl<N: Network> LedgerState<N> {
         // Determine the number of latest block headers to include as block locators (linear).
         let num_block_headers = std::cmp::min(MAXIMUM_LINEAR_BLOCK_LOCATORS, block_locator_height);
 
+        // Acquire the read lock for the latest block hashes and block headers.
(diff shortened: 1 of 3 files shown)


storage/src/state/ledger.rs
```diff
@@ -464,23 +464,23 @@ impl<N: Network> LedgerState<N> {
         // Determine the number of latest block headers to include as block locators (linear).
         let num_block_headers = std::cmp::min(MAXIMUM_LINEAR_BLOCK_LOCATORS, block_locator_height);
 
+        // Acquire the read lock for the latest block hashes and block headers.
+        let latest_block_hashes_and_headers = self.latest_block_hashes_and_headers.read();
```
ljedrz: what's the motivation for acquiring this read lock for an extended scope? I think it was I who reduced it before; does this object need to be locked from writing until the end of this function's body?

howardwu: Ah, I did it mainly to reduce the number of collect() calls, however you're right this may not be desirable.

howardwu: Reverting it now

PR #1464, opened 2021-12-18
Updates block request handler cases and tests

howardwu:
## Motivation

- Updates block request handler cases and tests
- Adds `Case` enum and tracks return outcomes for testing
- Formalizes `Case 0` and `Case 2(c)(c)`

### DIFF
--- a/src/helpers/block_requests.rs
+++ b/src/helpers/block_requests.rs
@@ -98,6 +98,19 @@ pub fn find_common_ancestor<N: Network>(canon: &LedgerState<N>, block_locators: 
     Ok((maximum_common_ancestor, first_deviating_locator))
 }
 
+/// A case annotation enum for the block request handler.
+#[derive(Debug, PartialEq, Eq)]
+pub(crate) enum Case {
+    Zero,
+    One,
+    TwoA,
+    TwoB,
+    TwoCA,
+    TwoCBA,
+    TwoCBB,
+    TwoCC,
+}
+
 /// The successful outcome of a block request handler.
 #[derive(Debug, PartialEq, Eq)]
 pub(crate) struct BlockRequestHandlerProceed {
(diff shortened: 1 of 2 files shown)


src/helpers/block_requests.rs
```diff
@@ -98,102 +98,124 @@ pub fn find_common_ancestor<N: Network>(canon: &LedgerState<N>, block_locators:
     Ok((maximum_common_ancestor, first_deviating_locator))
 }
 
+/// A case annotation enum for the block request handler.
```
ljedrz, line 101: suggestion: perhaps this enum could be the place to document all the cases (as doc comments for each variant), as opposed to `handle_block_requests`?

howardwu: Moving it over now

howardwu: Done

src/helpers/block_requests.rs
```diff
@@ -98,102 +98,133 @@ pub fn find_common_ancestor<N: Network>(canon: &LedgerState<N>, block_locators:
...
+#[derive(Debug, PartialEq, Eq)]
+pub(crate) enum Case {
+    /// Case 0 - The common ancestor is greater than the latest block height:
+    ///     - Abort. This is an internal error that needs to be remedied promptly.
+    Zero,
```
ljedrz, line 106: note: this case - while pretty trivial - doesn't have a corresponding test

src/helpers/block_requests.rs
```diff
@@ -98,102 +98,133 @@ pub fn find_common_ancestor<N: Network>(canon: &LedgerState<N>, block_locators:
...
+/// based on the cases as described in the `Case` enum.
 ///
-pub fn handle_block_requests<E: Environment, N: Network>(
+pub(crate) fn handle_block_requests<N: Network, E: Environment>(
     latest_block_height: u32,
     latest_cumulative_weight: u128,
-    peer_ip: SocketAddr,
-    is_fork: bool,
+    maximal_peer: SocketAddr,
+    maximal_peer_is_on_fork: Option<bool>,
```
ljedrz, line 166: what does `Some(false)` mean or how does it differ from `None`? if I recall correctly, this used to be (or was proposed to be in https://github.com/AleoHQ/snarkOS/pull/1372) an option only because if the `maximal_peer` was `None`, this was `None` as well; the existence of a `maximal_peer` (which is guaranteed at this stage) should indicate that `maximal_peer_is_on_fork` can only be `true` or `false` if the current logic is to be preserved

howardwu: Today, we only consider the case when `is_on_fork` is `Some(_)`, however the `undecided/unknown` case (i.e. `None` case) is important because it delineates clearly you are ahead in the honest case, and can detect malicious behavior if you are behind and it is `None` for the dishonest case.

src/helpers/block_requests.rs
howardwu: Adding a test case for 0 now

ljedrz (approved):
There's some issue with one of the tests, but once that one is cleared, this looks good to me; I particularly like that all the possible outcomes are synthesized into a dedicated `enum`.

src/helpers/block_requests.rs
howardwu: Done

howardwu:
> There's some issue with one of the tests, but once that one is cleared, this looks good to me; I particularly like that all the possible outcomes are synthesized into a dedicated `enum`.

The issue with that one test is a test-specific typo, there was a 1 in 1000 chance of flakiness that has now been fixed.

PR #1476, opened 2021-12-20
Introduce a MapId object containing the DataMap prefix

ljedrz:
This PR facilitates further improvements to the `rocksdb` prefix setup. An immediate benefit is that it'll also be harder to make a typo when picking a context/prefix and to have their full list in a single spot.

### DIFF
--- a/storage/src/state/ledger.rs
+++ b/storage/src/state/ledger.rs
@@ -16,7 +16,7 @@
 
 use crate::{
     helpers::BlockLocators,
-    storage::{DataMap, Map, Storage},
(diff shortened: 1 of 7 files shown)


ljedrz:
The CI failure is unrelated.

ljedrz:
Cc https://github.com/AleoHQ/snarkOS/pull/1417.

niklaslong (approved):

niklaslong:
Looks like CI is unhappy with one of the tests. It's the same flaky test as was reported earlier, I think.

PR #1477, opened 2021-12-20
Don't accept connections when at max peers

ljedrz:
This should improve the performance of sync nodes and any other nodes that are the target of many inbound connections.


howardwu:
Merging, again the flaky test case has already been fixed in an incoming PR.

LGTM

howardwu (approved):

PR #1480, opened 2021-12-20
Feat: Prometheus exporter

niklaslong:
This PR adds a prometheus exporter and wires it up to the node behind the `prometheus` feature flag. No metrics are currently implemented to avoid conflicts with ongoing work.


niklaslong:
Rebased.

niklaslong:
Rebased.

src/network/server.rs
```diff
@@ -404,4 +408,11 @@ impl<N: Network, E: Environment> Server<N, E> {
         // Wait until the heartbeat task is ready.
         let _ = handler.await;
     }
+
+    #[cfg(feature = "prometheus")]
+    fn initialize_metrics(tasks: &mut Tasks<task::JoinHandle<()>>) {
+        tasks.append(snarkos_metrics::initialize().expect("couldn't initialise the metrics"));
+
+        // TODO: set the block height here, as it could be non-zero.
```
ljedrz:
```suggestion
```

metrics/src/names.rs
```diff
@@ -0,0 +1,57 @@
...
+    pub const TIMEOUTS_INIT: &str = "snarkos_handshakes_timeouts_init_total";
+    pub const TIMEOUTS_RESP: &str = "snarkos_handshakes_timeouts_resp_total";
+}
+
+pub mod blocks {
+    pub const HEIGHT: &str = "snarkos_blocks_height_total";
+    pub const MINED: &str = "snarkos_blocks_mined_total";
+    pub const DUPLICATES: &str = "snarkos_blocks_duplicates_total";
+    pub const ORPHANS: &str = "snarkos_blocks_orphan_total";
+}
```
ljedrz: we can probably remove these for now, since we might want to have alternative metrics this time around; the first ones I'd be interested in would be internal RTTs, for instance

ljedrz (commented):
Left some suggestions :ok_hand:

metrics/src/names.rs
niklaslong: Sure, meant to keep them as a reminder of what we had but the `testnet1` branch still exists, so that's fine 👍

src/network/server.rs
niklaslong: Well, we do actually want to set the block height there no?

ljedrz: The same principle applies as with my suggestion to remove the (currently) unused `const`s - we don't use it yet. Other than that, you're correct.

ljedrz (approved):
LGTM :+1:

niklaslong:
Rebased.

niklaslong:
Rebased.

howardwu (approved):

howardwu (approved):

PR #1485, opened 2021-12-21
Don't allow to mine in future at all

xbeastx:
## Motivation

There is a group of guys who own a lot of power (aleo1gqa7nht8lv93ksgmfy8hnz028r776evnv5kllkcsx2ycek5ryq9suprafw) and they mine in advance and keeping blocks in private. After some time they send them to network and everyone is accepting this because of this condition:
```rust
        // Ensure the next block timestamp is within the declared time limit.
        let now = chrono::Utc::now().timestamp();
        if block.timestamp() > (now + MAXIMUM_FUTURE_BLOCK_TIME) {
            return Err(anyhow!("The given block timestamp exceeds the time limit"));
        }
```

To prevent this as soon as possible we can restrict future time as much as possible. I propose less then 20 seconds to not allow this guys to mine block on same high level of difficulty every time. Or they will always increase difficulty (that is not possible) or they will stay fair to mine in correct time (now)

### DIFF
--- a/storage/src/state/ledger.rs
+++ b/storage/src/state/ledger.rs
@@ -48,8 +48,8 @@ pub const MAXIMUM_BLOCK_LOCATORS: u32 = MAXIMUM_LINEAR_BLOCK_LOCATORS.saturating
 /// TODO (howardwu): Reconcile this with the equivalent in `Environment`.
 const MAXIMUM_FORK_DEPTH: u32 = 4096;
 
-/// The maximum future block time - 2 minutes.
-const MAXIMUM_FUTURE_BLOCK_TIME: i64 = 120;
+/// The maximum future block time - 19 seconds.
+const MAXIMUM_FUTURE_BLOCK_TIME: i64 = 19;
 
 ///
 /// A helper struct containing transaction metadata.


xorinox:
The real issue here is, that the network is not coordinated on what "now" means.  On my miners "now" is sys clock, that is synchronized with internet time. On any other miner it might be sys clock, that is not synchronized and be in the past or future depending on how much it deviates.  Depending on how much it deviates, some miner will allow blocks with less or more than the intended MAXIMUM_FUTURE_BLOCK_TIME, because the block time was a sys clock on some miner somewhere in the network that was not synchronized with internet time. I think that the challenge request, could include a mechanism to conclude on a true "now" and only then allow peering.

xbeastx:
> and be in the past or future depending on how much it deviates

I understand. But what I want say it's not our (or majority miners) problem... it's problem of people which run miner out of clock. It can be specified in requirements section for example... it's like you have bad internet connection and will slow thing down and not working correctly. It's your problem I mean.

howardwu:
This change does not do what you think it does.

Modifying to 19 seconds in the future only prevents withholding future blocks up to 19 seconds. It does not prevent a timestamp from being 19 seconds in the future and still be admitted.

PR #1487, opened 2021-12-21
Fix sync node hangs

ljedrz:
I'm still going to test it for a while, but it seems to fix the recent sync node hangs; the reduction in maximum peers is not related to the hang, but we shouldn't need that many (with the number of sync nodes we're running) and the sync nodes will be more responsive with a lower maximum.

Cc https://github.com/AleoHQ/snarkOS/issues/1358, https://github.com/AleoHQ/snarkOS/issues/1378


niklaslong:
I can confirm this seems to fix the issue 👍

niklaslong (approved):
Preemptively approved.

howardwu (approved):

howardwu:
LGTM

I may kill the experiment (i.e. delete that commented out code) if this works. Thanks for the help.

PR #1495, opened 2021-12-22
Feat/rpc large

amousa11:
## Motivation

When attempting to send a large transaction with the sendtransaction rpc endpoint, hyper was failing to read the entire body of the request. To fix this, I changed the request body deserialization to use hyper::body::to_bytes.

## Test Plan

Manually tested each RPC endpoint in postman(success and failure cases) to ensure there were no regressions.
Added a unit test with the large transaction in question.


howardwu (approved):

src/rpc/rpc.rs
```diff
@@ -1324,6 +1378,51 @@ mod tests {
...
+            let response = client.post(&url).json(&request).send().await?.text().await?;
+
+            Ok(serde_json::Value::from_str(&response)?)
+        }
+
+        // Send a new request that calls the `sendtransaction` endpoint to the RPC server.
+        let result = send_transaction(caller()).await;
+        assert!(result.is_ok());
+
+        let response = send_transaction(caller()).await.expect("Test RPC failed to process request");
```
howardwu: @amousa11 is the intent to `send_transaction` twice?

PR #1504, opened 2021-12-27
feat: add a subcommand to revert blocks to a specific height

xander988:
## Motivation

(Write your motivation here)

## Test Plan

(Write your test plan here)

## Related PRs

(Link your related PRs here)

### DIFF
--- a/src/node.rs
+++ b/src/node.rs
@@ -473,9 +473,6 @@ impl MinerRevert {
         let ledger_storage_path = node.ledger_storage_path(ip);
         let ledger = snarkos_storage::LedgerState::<Testnet2>::open_writer::<RocksDB, _>(ledger_storage_path).unwrap();
         let block_height = ledger.latest_block_height().saturating_sub(self.revert);
-        if block_height <= 0u32 {
-            return Err(anyhow!("revert height must before latest_block_height"));
-        }
         let block = ledger
             .revert_to_block_height(block_height)
             .expect("Failed to remove the last block");
--- a/src/node.rs
+++ b/src/node.rs
@@ -389,7 +389,6 @@ impl MinerSubcommand {
         match self.commands {
             MinerCommands::Stats(command) => command.parse(),
             MinerCommands::Revert(command) => command.parse(),
-
         }
     }
 }
@@ -480,10 +479,8 @@ impl MinerRevert {
         let block = ledger
             .revert_to_block_height(block_height)
             .expect("Failed to remove the last block");
-        return Ok(format!(
-            "Revert len={} success",
-            block.len(),
-        ));
+        return Ok(format!("Revert len={} success", block.len(),));
+
     }
 }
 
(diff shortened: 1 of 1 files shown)


damons:
Would you like to request review?  Are you ready?

xander988:
> Would you like to request review? Are you ready?

xander988:
> Would you like to request review? Are you ready?

Yes, sure. I'm ready. What should i do

xander988:
> Would you like to request review? Are you ready?

Yes, sure. I'm ready. What should i do

src/node.rs
```diff
@@ -453,6 +456,30 @@ impl MinerStats {
     }
 }
 
+#[derive(StructOpt, Debug)]
+pub struct MinerRevert {
+    #[structopt()]
```
ljedrz, line 461: does this attribute do anything?

src/node.rs
```diff
@@ -453,6 +456,30 @@ impl MinerStats {
...
+    revert: u32,
+}
+
+impl MinerRevert {
+    pub fn parse(self) -> Result<String> {
+        // Initialize the node.
+        let node = Node::from_iter(&["snarkos", "--norpc", "--verbosity", "0"]);
+
+        let ip = "0.0.0.0:1000".parse().unwrap();
```
ljedrz, line 470: since there is no reason to connect, the IP should be localhost; as for the port, it should be the snarkOS default port number (or it should be configurable), otherwise the revert won't work

src/node.rs
```diff
@@ -453,6 +456,30 @@ impl MinerStats {
...
+
+        // Initialize the ledger storage.
+        let ledger_storage_path = node.ledger_storage_path(ip);
+        let ledger = snarkos_storage::LedgerState::<Testnet2>::open_writer::<RocksDB, _>(ledger_storage_path).unwrap();
+        let block_height = ledger.latest_block_height().saturating_sub(self.revert);
+        let block = ledger
+            .revert_to_block_height(block_height)
+            .expect("Failed to remove the last block");
+        return Ok(format!("Revert len={} success", block.len()));
```
ljedrz:
```suggestion
        Ok(format!("Revert len={} success", block.len()))
```

ljedrz (commented):
I'm not sure if this should be officially supported (as this is a workaround and shouldn't be needed), but other than that, this functionality is pretty straightforward and should work just fine.

src/node.rs
xander988: I will extend the struct to support specfying port to get ledge info.

src/node.rs
xander988: Yep, I agree with you. The port will have a default value, or get through the command. what's your idea? If you agree with me, I will push the changes. By the way, as I followed the above code style, the changes will break the legacy style.

raychu86:
We aren't looking to add this feature to our official releases at the moment. We can revisit this if it becomes necessary.

PR #1505, opened 2021-12-27
[WIP] Feat/dtynn/jsonrpc2

dtynn:
Hi @howardwu 
I would like to get feedback from the team about if you will accept a pr like this, while it is still in the early stage.

## Motivation
1. avoid handwritten rpc implementaions
2. rpc client constructor

## Test Plan
WIP

## Related PRs
no


ljedrz:
The issue with `jsonrpc` crates is that they are effectively deprecated at this point, with `jsonrpsee` being the in-progress successor; at one point in time the handwritten implementation was introduced to remove all the outdated dependencies including the then-incompatible `tokio` below 1.0, and I think it would be best to switch to `jsonrpsee` once it's ready.

PR #1513, opened 2021-12-29
Added Cargo-related files to .gitignore

evgeny-garanin:
## Motivation

(Write your motivation here)

## Test Plan

(Write your test plan here)

## Related PRs

(Link your related PRs here)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 .vscode
 **/*.idea/
 **/target
+**/debug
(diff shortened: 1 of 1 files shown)


howardwu:
@evgeny-garanin We publish the `Cargo.lock` because some Linux builds and embedded OS users require it in order to compile. AFAIK this is due to a limitation in the Rust compiler on those OS targets.

PR #1528, opened 2022-01-06
Revert "Updates `PoolResponse` from block header to PoSW proof"

howardwu:
Reverts AleoHQ/snarkOS#1527

Reverting for a precautionary measure with Cargo.


howardwu:
This PR has been unreverted on `testnet2`.

The issue was caused by a bug in how Cargo indexes branches for a git dependency. It has been resolved manually on our end.

PR #1529, opened 2022-01-06
Refactors `BlockRequest` and adds tests for it

howardwu:
## Motivation

Refactors `BlockRequest` and adds tests for it

## Test Plan

Adds the following tests:
- `test_block_request`
- `test_block_request_eq`

### DIFF
--- a/src/helpers/block_request.rs
+++ b/src/helpers/block_request.rs
@@ -1,0 +1,124 @@
+// Copyright (C) 2019-2021 Aleo Systems Inc.
+// This file is part of the snarkOS library.
+
+// The snarkOS library is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
+// The snarkOS library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with the snarkOS library. If not, see <https://www.gnu.org/licenses/>.
+
+use snarkvm::dpc::prelude::*;
+
+use std::hash::{Hash, Hasher};
+
+///
+/// A request for a block with the specified height and an optional hash.
+///
+#[derive(Clone, Debug)]
+pub struct BlockRequest<N: Network> {
+    block_height: u32,
+    block_hash: Option<N::BlockHash>,
+}
+
+impl<N: Network> BlockRequest<N> {
+    /// Returns the block height stored in the request.
+    pub fn block_height(&self) -> u32 {
+        self.block_height
+    }
+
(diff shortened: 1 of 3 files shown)


src/helpers/block_request.rs
```diff
@@ -0,0 +1,124 @@
...
+        }
+    }
+
+    #[test]
+    fn test_block_request_eq() {
+        let rng = &mut thread_rng();
+
+        let a = BlockRequest::<Testnet2>::from(1000);
+        let b = BlockRequest::<Testnet2>::from((1000, None));
+        assert_eq!(a, b);
```
niklaslong: Why no iteration in this case?

src/helpers/block_request.rs
```diff
@@ -0,0 +1,124 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use snarkvm::{dpc::testnet2::Testnet2, prelude::UniformRand};
+
+    use rand::{thread_rng, Rng};
+
+    const ITERATIONS: usize = 1_000_000;
```
niklaslong: I don't see any harm in iterating (beyond making the tests slower) but is there any reason you suspect different values would change the behaviour in this case?

ljedrz: For this many iterations I'd rather use `thread_rng` only to seed a different, faster source of randomness that would be used in the loop; it's something I'd like to do for other tests too, including snarkVM ones.

howardwu: We can do that in one go to keep it consistent in design.

The rationale for multiple iterations is more to ensure the Default::default() block hash Eq case is accounted for.

I put 1 million iterations arbitrarily, it could be 1000 as well.

src/helpers/block_request.rs
howardwu: Good point, will add

src/helpers/block_request.rs
```diff
@@ -0,0 +1,124 @@
...
+
+    #[test]
+    fn test_block_request() {
+        let rng = &mut thread_rng();
+
+        for _ in 0..ITERATIONS {
```
ljedrz: I'm not sure if running these checks for many random heights and hashes is more likely to detect issues than just testing 2 cases (any height + `None` and any height + `Some(any hash)`) for equality and conversion; after all, the actual values of the hash and height are not converted further or used in any operations.

ljedrz (commented):
Moving the object to its own module is a good idea, but I'm not convinced that running many iterations for random values is more likely to detect issues, as commented [here](https://github.com/AleoHQ/snarkOS/pull/1529/files#r780128641).

src/helpers/block_request.rs
howardwu: Added it in

src/helpers/block_request.rs
howardwu: I dropped the iteration count to 5.

src/helpers/block_request.rs
howardwu: I dropped the iteration count to 5.

raychu86 (approved):

Issue #1450, opened 2021-12-16
[Bug] Ledger validation is affected by the tokio threadpool


ljedrz:
@zosorock reported that setting the number of worker threads to 1% of the number of logical cores available considerably slows ledger validation down; this could indicate a resource use logic bug.

ljedrz:
@zosorock I tested it locally, and wasn't able to reproduce: changing the number of async worker threads resulted in no change to the amount of time needed to validate the ledger (which shouldn't be affected); can you double-check if your changes were related to the [following code](https://github.com/AleoHQ/snarkOS/blob/testnet2/src/main.rs#L37-L42)?
```
let runtime = runtime::Builder::new_multi_thread()
        .enable_all()
        .thread_stack_size(8 * 1024 * 1024)
        .worker_threads((num_cpus::get() / 8 * 2).max(1)) // <- this is the value I was changing
        .max_blocking_threads(num_cpus::get())
        .build()?;
```

ljedrz:
Closing, as I haven't been able to confirm it and there hasn't been any further activity. Please feel free to reopen if you encounter this again.

PR #1530, opened 2022-01-06
Unifies the thread pool for the node

howardwu:
## Motivation

Unifies the thread pool for the node.

As more node types appear, unifying the thread pool will ensure we don't overutilize resources, and makes it easier to track and debug potential utilization issues.

This PR should be merged after #1529.

### DIFF
--- a/src/environment/mod.rs
+++ b/src/environment/mod.rs
@@ -18,6 +18,7 @@ use crate::helpers::{NodeType, Status, Tasks};
 use snarkvm::dpc::Network;
 
 use once_cell::sync::OnceCell;
+use rayon::{ThreadPool, ThreadPoolBuilder};
 use std::{
     collections::HashSet,
     fmt::Debug,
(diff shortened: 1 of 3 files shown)


howardwu:
Relevant to the discussion if #1450 as well

niklaslong (approved):
LGTM 👍

src/environment/mod.rs
```diff
@@ -104,6 +105,18 @@ pub trait Environment: 'static + Clone + Debug + Default + Send + Sync {
...
         TERMINATOR.get_or_init(|| Arc::new(AtomicBool::new(false)))
     }
+
+    /// Returns a thread pool for the node to perform intensive operations.
+    fn thread_pool() -> &'static Arc<ThreadPool> {
+        static POOL: OnceCell<Arc<ThreadPool>> = OnceCell::new();
+        POOL.get_or_init(|| {
+            Arc::new(ThreadPoolBuilder::new()
+                .stack_size(8 * 1024 * 1024)
+                .num_threads((num_cpus::get() * 7 / 8).max(2))
```
niklaslong, line 115: Note that this will result in a lot of threads dedicated to networking on systems with a large number of cores.

ljedrz: This threadpool is only applicable to mining/proving-related operations; the network bits are handled by the `tokio` threadpool  (minus some (de)serialization that is handled via a global `rayon` threadpool); it's easy to get them mixed up, though, and this PR simplifies things a bit; it might be a good idea to adjust this threadpool's name so that it's more clear that it's a dedicated `rayon` threadpool (as opposed to the `tokio` or global `rayon` one).

ljedrz (approved):
LGTM, but I'd tweak the name so it's more clear which threadpool this is at the callsites.

PR #1532, opened 2022-01-07
Move the block locator length check, deduplicate some logic in update_peer

ljedrz:
I was looking at `Pong` message handling, hoping to find potential performance improvements, and while bigger potential wins require more work, I've found two possible immediate improvements (the 2 commits in this PR).

### DIFF
--- a/storage/src/helpers/block_locators.rs
+++ b/storage/src/helpers/block_locators.rs
@@ -14,6 +14,7 @@
 // You should have received a copy of the GNU General Public License
 // along with the snarkOS library. If not, see <https://www.gnu.org/licenses/>.
 
+use crate::MAXIMUM_BLOCK_LOCATORS;
 use snarkvm::{
     dpc::{BlockHeader, Network},
     utilities::{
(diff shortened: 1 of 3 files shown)


niklaslong (approved):
Nice 👍

storage/src/helpers/block_locators.rs
```diff
@@ -76,6 +77,12 @@ impl<N: Network> FromBytes for BlockLocators<N> {
     fn read_le<R: Read>(mut reader: R) -> IoResult<Self> {
         let num_locators: u32 = FromBytes::read_le(&mut reader)?;
 
+        // Check that the number of block_locators is less than the total MAXIMUM_BLOCK_LOCATORS.
+        if num_locators > MAXIMUM_BLOCK_LOCATORS {
+            error!("The list of block locators is too long");
+            return Err(ErrorKind::Other.into());
```
ljedrz, line 83: My only concern here is that the peer that sends too many block locators will get insta-kicked (i.e. without going through the `add_failure` mechanism) by the node, but since this is technically protocol violation and this deserialization is costly, it might be preferable to do that.

raychu86 (approved):

PR #1533, opened 2022-01-07
Feat: initial crawler implementation

niklaslong:
This PR introduces a network-based crawler implementation. There are two new crates:
- `crawler`: which, self-explanatorily, contains the crawler impl
- `synthetic_node`: which contains the bare-bones `pea2pea`-based node which powers both the peers used in some tests and the crawler. 

Currently, the data is stored in memory and isn't exportable yet; how we implement this is still tbd.

This is a draft and will probably still change but any feedback at this early stage is welcome.


crawler/Cargo.toml
```diff
@@ -0,0 +1,60 @@
...
+    "decentralized",
+    "zero-knowledge"
+]
+categories = [ "cryptography", "operating-systems" ]
+license = "GPL-3.0"
+edition = "2018"
+
+[dependencies.snarkos]
+path = ".."
+features = ["test"]
```
ljedrz: why are we using this feature here? presumably only so that there is no storage path clash with the proper node?

crawler/Cargo.toml
```diff
@@ -0,0 +1,60 @@
+[package]
+name = "snarkos-crawler"
+version = "2.0.0"
+authors = [ "The Aleo Team <hello@aleo.org>" ]
+description = "A decentralized operating system"
```
ljedrz: the description needs adjustment

crawler/Cargo.toml
```diff
@@ -0,0 +1,60 @@
...
+    "cryptography",
+    "blockchain",
+    "decentralized",
+    "zero-knowledge"
+]
+categories = [ "cryptography", "operating-systems" ]
+license = "GPL-3.0"
+edition = "2018"
+
+[dependencies.snarkos]
```
ljedrz, line 19: note: the crawler would be more lightweight and faster to compile if the network components of snarkOS were moved to a crate of their own, as proposed in the past (it would be outside of this PR's scope, though)

crawler/src/known_network.rs
```diff
@@ -0,0 +1,256 @@
...
+    net::SocketAddr,
+};
+use time::OffsetDateTime;
+
+use crate::connection::{nodes_from_connections, Connection};
+
+// Purges connections that haven't been seen within this time (in hours).
+const STALE_CONNECTION_CUTOFF_TIME_HRS: i64 = 4;
+
+const CRAWL_INTERVAL_MINS: i64 = 10;
```
ljedrz, line 30: needs a comment to clarify whether it's an interval for an entire pass or for revisiting known connections

crawler/src/main.rs
```diff
@@ -0,0 +1,299 @@
...
+
+    crawler.run_periodic_tasks();
+
+    tokio::spawn(async move {
+        loop {
+            tokio::time::sleep(Duration::from_secs(5)).await;
+            println!("Nodes: {}", crawler.known_network.nodes().len());
+            println!("Connections: {}", crawler.known_network.connections().len());
+        }
+    });
```
ljedrz: this can probably be moved into `run_periodic_tasks`

crawler/src/main.rs
```diff
@@ -0,0 +1,299 @@
...
+    fn deref(&self) -> &Self::Target {
+        &self.synth_node
+    }
+}
+
+impl Crawler {
+    /// Creates a default crawler node with the most basic network protocols enabled.
+    pub async fn default() -> Self {
+        let config = Config {
+            listener_ip: Some(IpAddr::V4(Ipv4Addr::LOCALHOST)),
```
ljedrz: this needs to be configurable

crawler/src/main.rs
```diff
@@ -0,0 +1,299 @@
...
+            desired_listening_port: Some(4132),
```
ljedrz: ditto

crawler/src/main.rs
```diff
@@ -0,0 +1,299 @@
...
+        self.update_peers();
+    }
+}
+
+/// Inbound message processing logic for the test nodes.
+#[async_trait::async_trait]
+impl Reading for Crawler {
+    type Message = ClientMessage;
+
+    fn read_message<R: io::Read>(&self, source: SocketAddr, reader: &mut R) -> io::Result<Option<Self::Message>> {
```
ljedrz, line 189: in order to achieve the best performance (and to reduce the required size of the read buffer), we should probably identify the type of the client message by its id and discard (i.e. skip the bytes of) the unwanted messages (like blocks) altogether, rather than fully deserializing them; this is likely already partially achieved due to the small buffer size :slightly_smiling_face:

crawler/src/main.rs
```diff
@@ -0,0 +1,299 @@
...
+        // FIXME: use the maximum message size allowed by the protocol or (better) use streaming deserialization.
+        let mut buf = [0u8; 8 * 1024];
```
ljedrz, line 191: once https://github.com/AleoHQ/snarkOS/pull/1533/files#r780981233 is done, this buffer can be as large as the biggest message that needs to be fully deserialized by the crawler (which is much smaller than the protocol limit)

crawler/src/main.rs
```diff
@@ -0,0 +1,299 @@
...
+            Err(e) => {
+                error!("a message from {} failed to deserialize: {}", source, e);
+                Err(io::ErrorKind::InvalidData.into())
+            }
+        }
+    }
+
+    async fn process_message(&self, source: SocketAddr, message: Self::Message) -> io::Result<()> {
+        // We may need to handle other messages as well.
+        match message {
```
ljedrz, line 218: this logic is missing the recent pool-related variants

crawler/src/main.rs
```diff
@@ -0,0 +1,299 @@
...
+            _ => return Err(io::ErrorKind::InvalidData.into()), // Peer is not following the protocol.
+        }
+
+        Ok(())
+    }
+}
+
+impl Crawler {
+    async fn process_peer_request(&self, source: SocketAddr) -> io::Result<()> {
+        // Don't share peers for the time being.
```
ljedrz: needs a `TODO` for easier tracking

crawler/src/main.rs
```diff
@@ -0,0 +1,299 @@
...
+        self.known_network.update_height(source, block_height);
+        self.known_network.received_height(source);
+
+        let genesis = Testnet2::genesis_block();
+        let msg = ClientMessage::Pong(
+            None,
+            Data::Object(BlockLocators::<Testnet2>::from(
+                vec![(genesis.height(), (genesis.hash(), None))].into_iter().collect(),
+            )),
+        );
```
ljedrz: we could perform this op just once for better perf and just send the ready bytes whenever they're needed

synthetic_node/Cargo.toml
```diff
@@ -0,0 +1,47 @@
+[package]
+name = "snarkos-synthetic-node"
+version = "2.0.0"
+authors = [ "The Aleo Team <hello@aleo.org>" ]
+description = "A decentralized operating system"
```
ljedrz: needs adjustment

ljedrz (commented):
LGTM in general, left some comments :ok_hand:

crawler/Cargo.toml
niklaslong: Agreed!

crawler/Cargo.toml
niklaslong: This is actually left over from the synthetic node refactor, it runs fine without 👍

ljedrz: cool; as long as there is no persistent storage associated with the crawler, there will be no clash

crawler/src/main.rs
niklaslong: We may just want to delete this—there's no reason for the crawler to share peers (unless we hope that this could increase the connectivity of the network) 🤔

crawler/src/main.rs
niklaslong: Same in the test node, will add 👍

niklaslong:
Rebased.

PR #1538, opened 2022-01-10
Validate locators searching by their height

ljedrz:
`get_block_height` is super slow compared to `get_block_hash`; using the latter to validate the block locators reduces the duration of the associated loop from hundreds of ms to **zero** ms.


niklaslong (approved):
Seems fine to me 👍

HarukaMa (changes requested):
If the ledger has forked, the incoming block hash for the forked block won't match the one on the ledger and it'll be reported as an error, effectively stopping the node from recovering from the fork situation.

ljedrz:
It's weird that the tests haven't caught this; I'll see if I can salvage some of this perf improvement.

ljedrz:
Superseded by https://github.com/AleoHQ/snarkOS/pull/1548.

PR #1542, opened 2022-01-11
Use atomic batch writes in the storage

ljedrz:
This PR utilizes rockdb's atomic batch operations in order to address any remaining database inconsistency issues. This makes the use of locks during node shutdown obsolete. As a nice side effect, this should also make most database-related operations faster.

In order to limit the scope of API changes and to preserve the ergonomics of `DataMap` objects, batch operations are implemented in the existing `Map` trait:
- `Map::{insert, remove}` are extended with an optional batch id argument
- `Map::{prepare_batch, execute_batch, discard_batch}` are added

The following steps are required in order to execute a bundle of write operations atomically:
1. obtain a batch id using `Map::prepare_batch`
2. perform a number of `Map::{insert, remove}` operations using the id from point 1
3. call `Map::execute_batch` with the id from point 1 (or `Map::discard_batch` to cancel it)

notes:
- `Map::{prepare_batch, execute_batch, discard_batch}`, not unlike `Map::refresh`, can be performed on any implementer of `Map`; in other words, `execute_batch` can be called on a different object than `prepare_batch` just fine
- while it might make sense to **force** batching in more methods (i.e. by calling `prepare_batch` and later `execute_batch` within them even if they are called without a batch id) than just in `try_fixing_inconsistent_state`, it is the only method that was called both on its own and as part of another batch; all the others, like `BlockState::add_block`, are only called in the context of a "broader" batch, and they are not public, so I didn't consider it necessary
- `is_read_only` is removed from `DataMap`s, as it's a property of the actual storage shared by them all; instead, they now all contain a pointer to the whole `RocksDB` object

### DIFF
--- a/src/network/ledger.rs
+++ b/src/network/ledger.rs
@@ -163,7 +163,7 @@ impl<N: Network, E: Environment> Ledger<N, E> {
         self.ledger_router.clone()
     }
 
-    pub(super) async fn shut_down(&self) -> (Arc<Mutex<()>>, Arc<Mutex<()>>) {
+    pub(super) async fn shut_down(&self) {
         debug!("Ledger is shutting down...");
 
         // Set the terminator bit to `true` to ensure it stops mining.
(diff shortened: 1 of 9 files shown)


niklaslong (approved):
LGTM as per our discussions—will make a second pass with fresh eyes in the morning.

howardwu:
This is great and much needed for snarkOS.

Before we proceed to review and merge this PR, if you feel it is stable and ready, can we first do a burn in test of this approach with a set of nodes to ensure it continues to behave reliably (and that shut downs do not fail/stall)?

ljedrz:
@howardwu that would be great; I've been testing it locally, but thorough burn-ins with more nodes are recommended for any storage-related changes 👍.

ljedrz:
Rebased.

jules (approved):

ljedrz:
Rebased.

ljedrz:
Superseded by https://github.com/AleoHQ/snarkOS/pull/1639.

PR #1544, opened 2022-01-11
Update snarkVM rev

raychu86:
## Motivation

This PR is a routine update to the latest `snarkVM` rev. In particular, this `snarkVM` version exposes a `cuda` feature that introduces the initial WIP GPU implementation for MSMs used in mining operations.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -70,9 +70,6 @@ version = "0.4"
 version = "0.14"
 features = [ "http1", "runtime", "server", "tcp" ]
 
-[dependencies.indexmap]
-version = "1.8.0"
-
 [dependencies.json-rpc-types]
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -70,6 +70,9 @@ version = "0.4"
 version = "0.14"
 features = [ "http1", "runtime", "server", "tcp" ]
 
+[dependencies.indexmap]
```
ljedrz: I don't see `IndexMap` actually being used, other than it being added to the TOML; am I missing something?

raychu86: Ah good catch! Removed.

ljedrz (approved):
LGTM

PR #1545, opened 2022-01-12
Add CI for `storage` and `testing` modules

raychu86:
## Motivation

This PR adds CI for the `storage` and `testing` modules. 

Additionally, it fixes a `get_block_locators` off-by-one issue by skipping the genesis block in the initial collection of block locator headers.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -52,7 +52,7 @@ jobs:
     docker:
       - image: cimg/rust:1.56.1
     resource_class: 2xlarge
-    parallelism: 20
+    parallelism: 10
     steps:
       - checkout
       - setup_environment:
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -70,6 +70,51 @@ jobs:
       - clear_environment:
           cache_key: snarkos-stable-cache
 
+  storage:
+    docker:
+      - image: cimg/rust:1.56.1
```
ljedrz: note: the current stable version is `1.57.0`, we might want to update that one as well

storage/src/state/ledger.rs
```diff
@@ -476,6 +476,7 @@ impl<N: Network> LedgerState<N> {
             .latest_block_hashes_and_headers
             .read()
             .asc_iter()
+            .skip_while(|(_, header)| header.height() == 0) // Skip the genesis block.
```
ljedrz:
```suggestion
            .filter(|(_, header)| header.height() != 0) // Skip the genesis block.
```
otherwise this will not skip the genesis block unless it's the very first one is the iterator, which is not certain

storage/src/state/ledger.rs
```diff
@@ -840,7 +841,10 @@ impl<N: Network> LedgerState<N> {
         self.latest_block_hashes_and_headers
             .write()
             .push((block.hash(), block.header().clone()));
+        println!("got to here");
```
ljedrz:
```suggestion
```

storage/src/state/ledger.rs
```diff
@@ -840,7 +841,10 @@ impl<N: Network> LedgerState<N> {
...
         *self.latest_block_locators.write() = self.get_block_locators(block.height())?;
+        println!("got to here after");
+
```
ljedrz:
```suggestion
```

ljedrz (commented):
Left a few suggestions; since we're at it, we might also want to restore the network integration tests: https://github.com/AleoHQ/snarkOS/pull/1338.

howardwu (approved):

PR #1546, opened 2022-01-12
Shorten storage prefixes

ljedrz:
This PR shortens all the storage prefixes in order to speed up lookups, pave way for internal rocksdb optimizations, and to reduce the storage size. This change is paired with automatic one-off migration, which is quite snappy. In order to achieve this, the following changes are introduced:
- `Storage::{import, export}` methods are changed to more tweakable and future-proof file-oriented implementations
- `Storage::migrate` is introduced and run automatically as part of `Storage::open`

For reference, below is a comparison between the current prefixes and this proposal:
```
current prefixes:

        | length prefix | N::NETWORK_ID | length prefix |      map id       |   key    |   value  |
|-------|---------------|---------------|---------------|-------------------|----------|----------|
| size  |  u32 (4B) LE  |  u16 (2B) LE  |  u32 (4B) LE  | <varies: 6 to 18> | <varies> | <varies> |
| bytes |  2, 0, 0, 0   |  2, 0 (now)   |    <varies>   |      <varies>     | <varies> | <varies> |

prefix length: 16B - 28B
```

```
this proposal:

        | N::NETWORK_ID |   map id    |   key    |   value  |
|-------|---------------|-------------|----------|----------|
| size  |  u16 (2B) LE  | u16 (2B) LE | <varies> | <varies> |
| bytes |  2, 0 (now)   |   <varies>  | <varies> | <varies> |

prefix length: 4B
```

As of today, this reduces the storage size by ~7%. Further optimizations facilitated by these changes will follow in future PRs.

### DIFF
--- a/storage/src/storage/rocksdb/tests.rs
+++ b/storage/src/storage/rocksdb/tests.rs
@@ -20,6 +20,13 @@ fn temp_dir() -> std::path::PathBuf {
     tempfile::tempdir().expect("Failed to open temporary directory").into_path()
 }
 
+fn temp_file() -> std::path::PathBuf {
+    tempfile::NamedTempFile::new()
+        .expect("Failed to open temporary file")
+        .path()
+        .to_owned()
+}
+
 #[test]
 fn test_open() {
     let _storage = RocksDB::open(temp_dir(), 0, false).expect("Failed to open storage");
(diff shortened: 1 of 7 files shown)


storage/src/storage/rocksdb/tests.rs
```diff
@@ -103,11 +110,35 @@ fn test_reopen() {
...
+        storage.export(&file).expect("Failed to export storage");
+    }
+
+    let storage = RocksDB::open(temp_dir(), 0, false).expect("Failed to open storage");
+    storage.import(&file).expect("Failed to import storage");
+
+    let map = storage.open_map::<u32, String>(MapId::Test).expect("Failed to open data map");
+
+    for i in 0..100 {
+        assert_eq!(map.get(&i).expect("Failed to insert"), Some(i.to_string()));
```
niklaslong, line 142:
```suggestion
        assert_eq!(map.get(&i).expect("Failed to get record"), Some(i.to_string()));
```

niklaslong (approved):
LGTM 👍

ljedrz:
Since I'm currently working on some storage benchmarks, I decided to check if this change yields a noticeable improvement, and it does result in a moderate performance boost, especially for smaller keys:
```
lookup_by_height        time:   [636.35 ns 637.43 ns 638.66 ns]                              
                        change: [-10.239% -9.9988% -9.7846%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 2 outliers among 100 measurements (2.00%)
  2 (2.00%) high mild

lookup_by_hash          time:   [838.34 ns 839.21 ns 840.10 ns]                            
                        change: [-3.0862% -2.8064% -2.5321%] (p = 0.00 < 0.05)
                        Performance has improved.
```
note: this comparison was made on top of https://github.com/AleoHQ/snarkOS/pull/1548 in order for `lookup_by_hash` to measure actual lookup performance (as opposed to being dominated by proof validation)

ljedrz:
Rebased.

jules (approved):

ljedrz:
In addition, I already included the changes from https://github.com/AleoHQ/snarkOS/pull/1417 here.

howardwu:
Lets save this one for `testnet3` branch (next week), and then we can remove the `migration` method in the PR

ljedrz:
Ok; I'll revert that commit and keep it in history in case we ever do need the means for fine-grained migration. Also, I propose to keep the altered way of importing/exporting the data, as it is simpler, more suitable for binary data (as opposed to JSON), much faster (as it omits deserialization), has tests and allows partial imports/exports.

ljedrz:
Superseded by https://github.com/AleoHQ/snarkOS/pull/1639.

PR #1548, opened 2022-01-12
Greatly speed up block height lookups and contains_key calls

ljedrz:
Supersedes https://github.com/AleoHQ/snarkOS/pull/1538.

This reduces the time needed to validate inbound block locators from hundreds of ms to **zero** ms. It will also boost any other operation which calls `get_block_height`.

### DIFF
--- a/storage/src/state/ledger.rs
+++ b/storage/src/state/ledger.rs
@@ -18,7 +18,7 @@ use crate::{
     helpers::BlockLocators,
     storage::{DataMap, Map, MapId, Storage},
 };
-use snarkvm::dpc::prelude::*;
+use snarkvm::{dpc::prelude::*, prelude::FromBytes};
 
 use anyhow::{anyhow, Result};
 use circular_queue::CircularQueue;
@@ -29,6 +29,7 @@ use rayon::prelude::*;
 use serde::{Deserialize, Serialize};
 use std::{
     collections::{BTreeMap, HashSet},
+    io::{self, Read},
     path::Path,
     sync::{
         atomic::{AtomicBool, AtomicU32, Ordering},
@@ -1303,8 +1304,23 @@ impl<N: Network> BlockState<N> {
 
     /// Returns the block height for the given block hash.
     fn get_block_height(&self, block_hash: &N::BlockHash) -> Result<u32> {
-        match self.block_headers.get(block_hash)? {
-            Some(block_header) => Ok(block_header.height()),
+        // Performance note: this method is using `Storage::get_raw` in order to avoid costly
+        // (and duplicate) proof verification, which happens automatically upon full deserialization.
+        match self.block_headers.get_raw(block_hash)? {
+            Some(raw_block_header) => {
+                // Ensure that any changes to the header's format don't go unnoticed.
+                debug_assert_eq!(N::NETWORK_ID, 2);
+                // Wrap the raw bytes in a reader.
+                let mut block_header_reader = io::Cursor::new(raw_block_header);
+                // Deserialize the bytes before the heights, skipping their values.
+                let _: N::LedgerRoot = FromBytes::read_le(&mut block_header_reader)?;
+                let _: N::TransactionsRoot = FromBytes::read_le(&mut block_header_reader)?;
+                // Obtain the block height.
+                let mut raw_block_height = [0u8; 4];
+                block_header_reader.read_exact(&mut raw_block_height)?;
+                let block_height = u32::from_le_bytes(raw_block_height);
+                Ok(block_height)
+            }
             None => return Err(anyhow!("Block {} missing from block headers map", block_hash)),
         }
     }
--- a/storage/src/storage/rocksdb/map.rs
+++ b/storage/src/storage/rocksdb/map.rs
@@ -93,6 +93,23 @@ impl<'a, K: Serialize + DeserializeOwned, V: Serialize + DeserializeOwned> Map<'
     }
 
     ///
+    /// Returns the raw value for the given key from the map, if it exists.
+    ///
+    fn get_raw<Q>(&self, key: &Q) -> Result<Option<Vec<u8>>>
+    where
+        K: Borrow<Q>,
+        Q: Serialize + ?Sized,
+    {
+        let mut key_buf = self.context.clone();
+        key_buf.reserve(bincode::serialized_size(&key)? as usize);
+        bincode::serialize_into(&mut key_buf, &key)?;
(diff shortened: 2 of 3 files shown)


storage/src/state/ledger.rs
```diff
@@ -1285,8 +1286,14 @@ impl<N: Network> BlockState<N> {
 
     /// Returns the block height for the given block hash.
     fn get_block_height(&self, block_hash: &N::BlockHash) -> Result<u32> {
-        match self.block_headers.get(block_hash)? {
-            Some(block_header) => Ok(block_header.height()),
+        // Performance note: this method is using `Storage::get_raw` in order to avoid costly
+        // (and duplicate) proof verification, which happens automatically upon full deserialization.
+        match self.block_headers.get_raw(block_hash)? {
+            Some(raw_block_header) => {
+                let raw_block_height: [u8; 4] = raw_block_header[64..][..4].try_into().unwrap(); // safe - we stored the headers ourself
```
raychu86: My worry here is that if we change the serialization of block headers for mainnet or subsequent testnets (will likely add additional fields), this will be left unchecked.

ljedrz: Hmm, how about if I deserialize `HeaderMetadata`? I'll see how this can be guarded against better :+1: .

ljedrz: Actually, there's access to `Network` there, so we can assert this is the expected `NETWORK_ID`.

raychu86: I think the NETWORK_ID flag should suffice for now. Once we have additional network ids here, we can look into adding a match case for it.

niklaslong (approved):
LGTM 👍

raychu86 (dismissed):

ljedrz:
Rebased and added 2 extra small commits:
- [perf: use Storage::get_raw in Storage::contains_key](https://github.com/AleoHQ/snarkOS/pull/1548/commits/6bd745d2c8f2e4f97a3ce1433a2154c112eb0240) (we don't need to deserialize the value in those calls)
- [refactor: rewrite Storage::get using Storage::get_raw](https://github.com/AleoHQ/snarkOS/pull/1548/commits/58f5cf4a96edf25276c418e41d721b7b969ab83e) (code deduplication)

ljedrz:
I'm currently working on some storage benchmarks and they confirm the gains from these changes (and the result below is just the `get_block_height`-specific improvement):
```
lookup_by_hash          time:   [858.88 ns 861.34 ns 864.27 ns]                            
                        change: [-99.998% -99.998% -99.998%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high mild
```

ljedrz:
And below are the effects of just [perf: use Storage::get_raw in Storage::contains_key](https://github.com/AleoHQ/snarkOS/pull/1548/commits/6bd745d2c8f2e4f97a3ce1433a2154c112eb0240) on the `LedgerState::contains_X` methods:
```
blocks_lookup_by_height time:   [479.89 ns 480.82 ns 481.87 ns]                                    
                        change: [-30.139% -29.995% -29.856%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 2 outliers among 100 measurements (2.00%)
  1 (1.00%) low mild
  1 (1.00%) high severe

txs_lookup_by_id        time:   [667.34 ns 668.88 ns 670.68 ns]                              
                        change: [-50.577% -50.430% -50.282%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 4 outliers among 100 measurements (4.00%)
  4 (4.00%) high mild

txs_lookup_by_commitment                                                                             
                        time:   [812.51 ns 813.86 ns 815.13 ns]
                        change: [-14.584% -14.387% -14.202%] (p = 0.00 < 0.05)
                        Performance has improved.

txs_lookup_by_serial_number                                                                             
                        time:   [789.79 ns 791.33 ns 792.81 ns]
                        change: [-20.740% -20.561% -20.391%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) low mild
```

storage/src/state/ledger.rs
raychu86: Currently rethinking the approach of using `get_raw`. 

I don't believe dealing with bytes directly is the approach we want to move forward with. It introduces potential issues down the line and also doesn't sit well with the abstractions we currently have.

raychu86:
An alternative approach without using `get_raw` would be to use the `block_heights` DataMap with some sort of iterator. Of course this will be much slower (O(N)), but will keep our base storage module simple.

ljedrz:
`DataMap`s already use iterators internally; there's no going around this one, I'm afraid, if we want storage to be fast, and having plain lookups taking hundreds of milliseconds is unacceptable in terms of the negative influence on async operations.

storage/src/state/ledger.rs
ljedrz: We are always dealing with raw bytes under the hood in our storage; as long as deserialization is computationally heavy, we need to find ways to avoid it in places where it's just not needed.

ljedrz:
@raychu86 I adjusted the first commit so that the partial deserialization doesn't involve any slicing or fixed lengths - instead, the values before the height get deserialized and dropped; this should still be cheap enough. Also, rebased.

howardwu:
What exactly is the use case that motivates this change?

If we want to call directly into bytes, we might as well write snarkOS in C++.

ljedrz:
The motivation is that `get_block_height` lookups take many milliseconds, which is thousands of times longer than the other storage lookups are (hundreds of nanoseconds), which currently causes them to be blocking operations in async contexts, especially when validating block locators. This was reported by @HarukaMa, who indicated that an associated operator-related operation was noticably slow.

`get_raw` is intended to be used only in 3 scenarios:
- in the implementation of `contains_key` (because then the values don't need to be deserialized at all, and this method allows that)
- in the implementation of `get` (to deduplicate code, as it already does the same thing internally)
- in specific, very limited cases like `get_block_height`, where the alternative is a blocking operation, which can and will disrupt all the other async operations of the node, especially with many peers

ljedrz:
Rebased.

ljedrz:
Closing in favor of https://github.com/AleoHQ/snarkOS/pull/1573 to get much of these performance improvments; if we could extend the `BlockHeader` with `LedgerRoot` and `TransactionsRoot`, we would almost get the full performance win of this PR by doing cleaner partial deserialization.

PR #1549, opened 2022-01-12
Expose `cuda` feature flag

raychu86:
## Motivation

This PR simply updates the snarkVM rev and exposes the `cuda` feature.

This CUDA implementation is experimental and currently a work in progress. If the node is started up with the `cuda` feature flag and the machine does not have a supported GPU, then mining will not work properly.


ljedrz (approved):
Don't have CUDA, so can't test mining, but LGTM :+1:

howardwu (approved):

PR #1553, opened 2022-01-13
Perf: use `parking_lot::RwLock` where possible 

niklaslong:
These collections never held the lock over an `await` point making it possible to switch to more performant `parking_lot` variants. 

I'm currently testing this change, hence the draft.


ljedrz (commented):
I agree with the general direction, but I'd wait with changes like these until we have internal RTT metrics, so we are confident that there are no pathological cases where we would be holding the any of these locks for too long (I'm mainly thinking of handshakes and disconnects here).

niklaslong:
Rebased.

aleo-svc-exp:

niklaslong:
Closing this PR as it is no longer applicable.

PR #1556, opened 2022-01-15
Revert "Move the block locator length check, deduplicate some logic in update_peer"

howardwu:
Reverts AleoHQ/snarkOS#1532

This check, while desirable, does not account for all entry points to create a BlockLocator struct.


howardwu:
Superseded by #1558

PR #1566, opened 2022-01-17
Only enable snappy feature of rocksdb

HarukaMa:
## Motivation

I was developing something on top of snarkOS and when compiling the project the following error occurred:

```
warning: Linking globals named 'bz_internal_error': symbol multiply defined!

error: failed to load bc of "librocksdb_sys-ca644c3a1310c35b.librocksdb_sys.cb77860e-cgu.0.rcgu.o":

```

I can only find one reference [here](https://github.com/rust-rocksdb/rust-rocksdb/issues/557). By only enabling `snappy` I'm able to compile the project.

According to https://github.com/rust-rocksdb/rust-rocksdb/blob/522b4967389881749405b5872f46d0b8d0625b81/src/db_options.rs#L993, snappy should be the compression type in use so disabling other features should be fine.

## Test Plan

Tried to load existing databases by running the node for a while and they loaded fine.

## Related PRs

(Link your related PRs here)


storage/Cargo.toml
```diff
@@ -50,6 +50,8 @@ version = "1"
 [dependencies.rocksdb]
 version = "0.17"
 optional = true
+default-features = false
+features = ["snappy"]
```
howardwu:
```suggestion
default-features = false  # See PR #1566 (https://github.com/AleoHQ/snarkOS/pull/1566)
features = ["snappy"]     #
```

howardwu (approved):
Seems reasonable

PR #1575, opened 2022-01-18
Allow the node to auto-forward its listnening port using UPnP

### LINKED ISSUE #423: [Feature] Use UPnP to open ports

ljedrz:
The testnet2 equivalent of https://github.com/AleoHQ/snarkOS/pull/1218; the only difference is that the external address is not registered, but it can be extended with that functionality if need be.

Supersedes https://github.com/AleoHQ/snarkOS/pull/1218.
Closes https://github.com/AleoHQ/snarkOS/issues/423.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -151,6 +151,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "attohttpc"
+version = "0.16.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "fdb8867f378f33f78a811a8eb9bf108ad99430d7aad43315dd9319c827ef6247"
+dependencies = [
+ "http",
+ "log",
+ "url",
+ "wildmatch",
+]
+
+[[package]]
 name = "atty"
 version = "0.2.14"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


ljedrz:
The CI error is unrelated:
```
running 1 test
test helpers::block_requests::tests::test_block_requests_case_2cbb ... FAILED

failures:

---- helpers::block_requests::tests::test_block_requests_case_2cbb stdout ----
thread 'helpers::block_requests::tests::test_block_requests_case_2cbb' panicked at 'assertion failed: `(left == right)`
  left: `AbortAndDisconnect(TwoCBA, "exceeded fork range")`,
 right: `Proceed(TwoCBB, BlockRequestHandlerProceed { start_block_height: 904, end_block_height: 1153, ledger_is_on_fork: true })`', src/helpers/block_requests.rs:572:13
```

niklaslong (dismissed):
LGTM 👍

ljedrz:
Rebased.

howardwu:
I think this is worth including. Question: is there a way to harmonize this feature with the design, so that we adhere to an implicit design for using upnp (as opposed to having the user explicitly set a upnp flag)?

For example, is there a concept of using this first, then defaulting to basic tcp/ip if it does not hold?

ljedrz:
There's a reason why I proposed this functionality to be opt-in rather than being automatic; if for whatever reason someone is changing their listening port on a regular basic, or if they are using internal nodes that are communicating with the network only via a single proxy exposed to it, or if there are any other reasons why someone wouldn't want their port to be opened automatically (e.g. a complex network configuration), it would potentially break their setups and/or eventually leave them with many open ports that they might not be aware about.

That being said, this functionality only requires the user to run with the `--upnp` flag once, after which they can forget about it even being there.

ljedrz:
Rebased against testnet3.

aleo-svc-exp:

PR #1576, opened 2022-01-18
Use an alias for the current network

ljedrz:
It's something that has been on my mind for a while now - in order for future default network changes to be simpler, we can have a global type alias for the current `N::Network`.

Note: it's currently set separately in both `snarkos` and `snarkos-storage`, but the one in `snarkos` could technically be based on the `snarkos-storage` one. In any case, it's still 2 places to adjust as opposed to dozens.

### DIFF
--- a/src/environment/mod.rs
+++ b/src/environment/mod.rs
@@ -27,6 +27,8 @@ use std::{
     sync::{atomic::AtomicBool, Arc},
 };
 
+pub type CurrentNetwork = snarkvm::dpc::testnet2::Testnet2;
+
 #[rustfmt::skip]
 pub trait Environment: 'static + Clone + Debug + Default + Send + Sync {
     type Network: Network;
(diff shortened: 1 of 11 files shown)


niklaslong (approved):

ljedrz:
@howardwu let me know if this change is desirable, and I'll rebase it (as there's some conflict).

jules (approved):

howardwu:
Yes, let's go for it

ljedrz:
@howardwu now that we're considering adding a helper `snarkos-utils` crate (in https://github.com/AleoHQ/snarkOS/pull/1615), perhaps it would be best that that one is introduced first, after which we could put the type alias there, so that it's defined in a single spot only.

ljedrz:
Superseded by https://github.com/AleoHQ/snarkOS/pull/1657.

PR #1577, opened 2022-01-18
Make message deserialisation generic over `Read`

niklaslong:
This refactors the message deserialisation code to be generic over `Read`. Note that this removes the possible per-connection panics that could occur on the buffer slicing. 

Drafted as testing is in progress.

### DIFF
--- a/src/network/message.rs
+++ b/src/network/message.rs
@@ -24,7 +24,11 @@ use snarkvm::{dpc::posw::PoSWProof, prelude::*};
 use ::bytes::{Buf, BufMut, Bytes, BytesMut};
 use anyhow::{anyhow, Result};
 use serde::{de::DeserializeOwned, Serialize};
-use std::{io::Write, marker::PhantomData, net::SocketAddr};
+use std::{
+    io::{Cursor, Write},
+    marker::PhantomData,
+    net::SocketAddr,
+};
 use tokio::task;
 use tokio_util::codec::{Decoder, Encoder};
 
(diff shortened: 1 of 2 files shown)


ljedrz (approved):
Nice, it's much safer now. LGTM :+1:

src/network/message.rs
```diff
@@ -223,61 +227,77 @@ impl<N: Network, E: Environment> Message<N, E> {
 
     /// Deserializes the given buffer into a message.
     #[inline]
-    pub fn deserialize(buffer: &[u8]) -> Result<Self> {
-        // Ensure the buffer contains at least the length of an ID.
-        if buffer.len() < 2 {
-            return Err(anyhow!("Invalid message buffer"));
-        }
+    pub fn deserialize<R: Read>(reader: &mut R) -> Result<Self> {
```
howardwu: Should we be clarifying that this `Read` must be "cursored", perhaps requiring the `BufRead` trait?

howardwu: There is a risk that future implementers are unaware and use methods/logic that reconsumes existing bytes. This particular case has cropped up snarkVM serialization before.

niklaslong: `BufRead` is an optimisation (e.g. faster reading line by line) that isn't applicable in this specific case. That being said, I've extended the trait bound to include `Seek`, which provides the cursoring guarantee.

jules (approved):

PR #1578, opened 2022-01-18
Startup scripts - updates, new ones for mining pool

wcannon-aleo:
## Motivation

The run-client and run-miner scripts have been updated for the updated functionality

- enable use of environment variables (e.g. miner address)
- only restart the snarkos executable when the code has updated and must be recompiled

Two new scripts have been added to make running a mining pool easier.

- run-operator makes it obvious to run a mining pool operator
- run-prover makes it simple to add a miner to the mining pool

## Test Plan

The scripts were tested multiple times on multiple servers.  

## Related PRs

None


run-prover.sh
```diff
@@ -0,0 +1,66 @@
...
+  OPERATOR_IP_ADDRESS=$REPLY
+fi
+
+if [ "${OPERATOR_IP_ADDRESS}" == "" ]
+then
+  echo "IP Address of Operator server is required to run a prover"
+  exit 1
+fi
+
+COMMAND="cargo run --release -- --prover ${MINER_ADDRESS} --pool ${OPERATOR_IP_ADDRESS}:4132 --trial --verbosity 2"
```
zosorock: I would not assume it is running on port 4132 so make OPERATOR_ADDRESS == IP+port.

run-prover.sh
```diff
@@ -0,0 +1,66 @@
...
+  STATUS=$(git pull)
+
+  if [ "${STATUS}" != "Already up to date." ]; then
+    echo "Updated code found, rebuilding and relaunching miner"
+    cargo clean
+    kill -INT $!; sleep 2; $COMMAND &
+  fi
+
+  sleep 1800;
+done
```
zosorock: I would add a newline for the sake of consistency

run-prover.sh
wcannon-aleo: What do you think of this instead?  Allow for an optional ENV VAR, and if not found then set port = 4132.

run-prover.sh
wcannon-aleo: done!

run-prover.sh
wcannon-aleo: Just added this functionality to make the port changeable via env vars

zosorock (approved):
lgtm 🚢

howardwu:
LGTM overall. I have one clarifying question:

In the `run-client` and `run-miner` scripts, the logic was updated to no longer terminate the existing node every 30 minutes, but still run the update checker for Github changes. Does this properly terminate if updates are made? Has it been manually checked?

wcannon-aleo:
> 
@howardwu The short answer is Yes.  The backgrounded $COMMAND process is terminated if an update is needed.  This was tested successfully (several times).  All testing was done on an ubuntu 20.04 server.  Would you like some additional testing or other checks?

howardwu:
LGTM, thanks @wcannon-aleo !

PR #1589, opened 2022-01-24
Fix lints from the updated nightly

ljedrz:
It seems that the nightly image was updated, causing the pending PRs to fail the `clippy` job. This PR takes care of the current clippy lints.

### DIFF
--- a/src/network/ledger.rs
+++ b/src/network/ledger.rs
@@ -236,14 +236,17 @@ impl<N: Network, E: Environment> Ledger<N, E> {
                 // Update the block requests.
                 self.update_block_requests().await;
 
+                let block_requests = self.number_of_block_requests().await;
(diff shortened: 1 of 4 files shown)


ljedrz:
Oof, seems that my nightly was just 2 days shy of one new lint; checking it out now.

ljedrz:
I'm not 100% sure the new `format_args` vs. `await` error is legitimate, but I haven't found any related bug reports, and working around it is trivial and doesn't have downsides.

niklaslong (approved):
LGTM 👍

jules (approved):

raychu86 (approved):

ljedrz:
This appears to be a bug after all: https://github.com/rust-lang/rust/issues/93274. That being said, the workaround doesn't hurt.

howardwu (approved):

PR #1601, opened 2022-02-01
Add a case for ledger validation with different increments, default increment is now `10,000`

ljedrz:
This PR introduces an `#[ingore]`d test case that allows us to load the ledger with different validation increment values; we can use it to double-check any future changes to logic applicable to ledger validation. Ran it a few times locally with no issues.

This slightly alters the signature of `LedgerState::open_writer`.


storage/tests/ledger_validation.rs
```diff
@@ -0,0 +1,47 @@
...
+
+const NUM_BLOCKS: usize = 1_000;
+const NUM_CHECKS: usize = 100;
+
+#[test]
+#[ignore = "takes a while to run (most of which is deserialization)"]
+fn test_ledger_validation() {
+    let mut test_blocks = fs::read(format!("benches/blocks_{}", NUM_BLOCKS)).expect(&format!("Missing the test blocks file"));
+    let blocks: Vec<Block<Testnet2>> = bincode::deserialize(&mut test_blocks).expect("Failed to deserialize a block dump");
+    assert_eq!(blocks.len(), NUM_BLOCKS - 1);
```
niklaslong, line 35: Maybe just comment on the `len` not including genesis here?

niklaslong (approved):
LGTM 👍

storage/tests/ledger_validation.rs
ljedrz: Fair enough, as this test is also missing the comment on how the test blocks were created. I'll include that soon :tm:.

storage/tests/ledger_validation.rs
```diff
@@ -0,0 +1,55 @@
...
+    // Read the test blocks; note: they don't include the genesis block, as it's always available when creating a ledger.
+    // note: the `blocks_100` and `blocks_1000` files were generated on a testnet2 storage using `LedgerState::dump_blocks`.
+    let mut test_blocks = fs::read(format!("benches/blocks_{}", NUM_BLOCKS)).expect(&format!("Missing the test blocks file"));
+    let blocks: Vec<Block<Testnet2>> = bincode::deserialize(&mut test_blocks).expect("Failed to deserialize a block dump");
+    assert_eq!(blocks.len(), NUM_BLOCKS - 1);
+
+    // Prepare a test ledger and an iterator of blocks to insert.
+    let temp_dir = tempfile::tempdir().expect("Failed to open temporary directory").into_path();
+    {
+        let ledger = LedgerState::open_writer::<RocksDB, _>(&temp_dir, NUM_BLOCKS as u32).expect("Failed to initialize ledger");
```
ljedrz:
```suggestion
        let ledger = LedgerState::open_writer::<RocksDB, _>(&temp_dir, 1).expect("Failed to initialize ledger");
```
just noticed this

storage/src/state/ledger.rs
```diff
@@ -94,12 +94,14 @@ pub struct LedgerState<N: Network> {
...
     /// Opens a new writable instance of `LedgerState` from the given storage path.
+    /// The `validation_increment` parameter determines the number of blocks to be
+    /// handled during the incremental validation process.
     /// For a read-only instance of `LedgerState`, use `LedgerState::open_reader`.
     ///
     /// A writable instance of `LedgerState` possesses full functionality, whereas
     /// a read-only instance of `LedgerState` may only call immutable methods.
     ///
-    pub fn open_writer<S: Storage, P: AsRef<Path>>(path: P) -> Result<Self> {
+    pub fn open_writer<S: Storage, P: AsRef<Path>>(path: P, validation_increment: u32) -> Result<Self> {
```
howardwu: This interface is not intuitive.

One option is to make it a parameter in Environment, otherwise it should be removed.

ljedrz: There needs to be a way of parametrizing it, otherwise it can't be tested; in order for the interface to be more intuitive, there can be another function, perhaps a `#[doc(hidden)]` one, that gets called by `open_writer`; I'll propose it in the next commit.

howardwu: Thank you for the change, it looks great

PR #1602, opened 2022-02-01
Use jsonrpsee as the RPC server impl

### LINKED ISSUE #754: Change the JSON-RPC crate when the successor is available
The currently used crate, [jsonrpc-http-server](https://crates.io/crates/jsonrpc-http-server) is going to be superseded by [jsonrpsee-http-server](https://crates.io/crates/jsonrpsee-http-server). We need to change our implementation once the successor becomes stable, otherwise we'll be carrying a lot of old dependencies.

This issue is nothing new, it's just a reminder for the future.

ljedrz:
While I still need to do some research around applying constraints outlined in https://github.com/AleoHQ/snarkOS/issues/1597, I can already provide some insight into how the `jsonrpsee`-powered implementation is going to look like. The decision to finally make the switch was motivated by the crate no longer warning about its instability and the required functionalities becoming too advanced for a hand-written implementation (which was always only intended to be temporary in the first place).

This PR closes https://github.com/AleoHQ/snarkOS/issues/754 and likely https://github.com/AleoHQ/snarkOS/issues/1597 as well, though that could also be applied as a follow-up.


src/rpc/tests/mod.rs
```diff
@@ -0,0 +1,726 @@
...
+    let ledger_state = new_ledger_state::<Testnet2, RocksDB, PathBuf>(Some(directory.clone()));
+    assert_eq!(0, ledger_state.latest_block_height());
+
+    // Initialize a new account.
+    let account = Account::<Testnet2>::new(&mut thread_rng());
+    let address = account.address();
+
+    // Mine the next block.
+    let (block_1, _) = ledger_state
+        .mine_next_block(address, true, &[], &Default::default(), &mut thread_rng())
```
ljedrz: TODO: using the pre-existing ledger from storage benchmarks could be faster

src/rpc/tests/mod.rs
```diff
@@ -0,0 +1,726 @@
...
```
ljedrz: TODO: using the pre-existing ledger from storage benchmarks could be faster

src/rpc/tests/mod.rs
```diff
@@ -0,0 +1,726 @@
...
+    let ledger_state = new_ledger_state::<Testnet2, RocksDB, PathBuf>(Some(directory.clone()));
+    assert_eq!(0, ledger_state.latest_block_height());
+
+    // Initialize a new account.
+    let account = Account::<Testnet2>::new(&mut rng);
+    let address = account.address();
+
+    // Mine the next block.
+    let (block_1, _) = ledger_state
+        .mine_next_block(address, true, &[], &terminator, &mut rng)
```
ljedrz: TODO: using the pre-existing ledger from storage benchmarks could be faster

src/rpc/tests/mod.rs
```diff
@@ -0,0 +1,726 @@
...
+            .first()
+            .unwrap()
+            .ciphertexts()
+            .any(|expected| response == *expected)
+    );
+}
+
+#[tokio::test]
+async fn test_get_ledger_proof() {
+    let mut rng = ChaChaRng::seed_from_u64(123456789);
```
ljedrz: TODO: don't use a hard-coded seed

src/rpc/tests/mod.rs
ljedrz: sadly, a custom account is used here

niklasad1:
Hey, you shouldn't enable `serde-json/arbitrary-precision` when using `jsonrpsee`, see https://github.com/paritytech/jsonrpsee/issues/480.

You will likely run into weird errors for instance when the request id is a number.

src/rpc/context.rs
```diff
@@ -0,0 +1,361 @@
...
+    (server_addr, task)
+}
+
+fn create_rpc_module<N: Network, E: Environment>(rpc_context: RpcContext<N, E>) -> Result<RpcModule<RpcContext<N, E>>, JsonrpseeError> {
+    let mut module = RpcModule::new(rpc_context);
+
+    // Public methods.
+
+    module.register_async_method("latestblock", |_rpc_params, rpc_context| async move {
+        rpc_context.latest_block().map_err(|e| JsonrpseeError::Request(e.to_string())).await
```
niklasad1: the intention is that you should use https://docs.rs/jsonrpsee-core/latest/jsonrpsee_core/error/enum.Error.html#method.to_call_error here if you don't care about the specific JSON-RPC error code and optional data.

ljedrz:
> Hey, you shouldn't enable `serde-json/arbitrary-precision` when using `jsonrpsee`, see [paritytech/jsonrpsee#480](https://github.com/paritytech/jsonrpsee/issues/480).
> 
> You will likely run into weird errors for instance when the request id is a number.

I don't even know why we need that one, everything seems to compile just fine without it; @howardwu?

ljedrz:
Ok, I now know where that feature might be used:
```
thread 'rpc::tests::test_get_block_template' panicked at 'called `Result::unwrap()` on an `Err` value: Error("u128 is not supported", line: 0, column: 0)', src/rpc/rpc_impl.rs:165:12
```

ljedrz:
Ok, I think the PR is ready for reviews; for now:
- the temporary JSON-RPC server is replaced with `jsonrpsee`
- a basic, quite conservative limit of 10 concurrent requests is introduced, with each request just counting as 1
- the `serde-json/arbitrary-precision` feature is an open point; to remove it, we could just not serve `cumulative_weight` or send it as a hex-encoded `String` (I implemented the latter for now)

src/rpc/context.rs
```diff
@@ -0,0 +1,309 @@
...
+    })?;
+
+    module.register_async_method("sendtransaction", |rpc_params, rpc_context| async move {
+        let string = std::mem::take(&mut rpc_params.parse::<[String; 1]>()?[0]);
+        rpc_context.send_transaction(string).map_err(JsonrpseeError::to_call_error).await
+    })?;
+
+    // Private methods.
+
+    // "createtransaction" => {
```
niklaslong, line 256: Kept for reference?

ljedrz: they were already commented out - presumably temporarily - so I preserved them so it's later easier to bring them back

Cargo.toml
```diff
@@ -68,15 +68,9 @@ version = "2.0"
...
-[dependencies.hyper]
-version = "0.14"
-features = [ "http1", "runtime", "server", "tcp" ]
-
-[dependencies.json-rpc-types]
-version = "1.0.2"
-
-[dependencies.jsonrpc-core]
-version = "18"
+[dependencies.jsonrpsee]
```
niklaslong: Do we want to feature gate the rpc module?

ljedrz: probably not during the testnet, but in the future, likely; I could add it as a default feature

src/rpc/context.rs
```diff
@@ -0,0 +1,309 @@
...
+/// RPC failures do not affect the rest of the node.
+pub async fn initialize_rpc_server<N: Network, E: Environment>(
+    rpc_server_addr: SocketAddr,
+    rpc_server_context: RpcContext<N, E>,
+) -> (SocketAddr, tokio::task::JoinHandle<()>) {
+    let access_control = AccessControlBuilder::default().build(); // TODO(ljedrz): adjust to only accept the desired hosts?
+
+    let server = HttpServerBuilder::new()
+        .set_access_control(access_control)
+        .register_resource(ALL_CONCURRENT_REQUESTS, ALL_CONCURRENT_REQUESTS_LIMIT, 1) // Limit the number of requests handled at a time.
```
niklaslong: Suggestion: comment on the fact that `1` here is the default?

ljedrz: `1` is actually not the default here - the default would be `0`; I can add an extra comment though

ljedrz: unless you meant that it's effectively the default value for every call, which it is; yeah, a comment will help :smile:

src/rpc/tests/mod.rs
```diff
@@ -0,0 +1,717 @@
...
+    E::tasks().append(rpc_server_handle);
+
+    rpc_server_addr
+}
+
+#[tokio::test]
+async fn test_handle_rpc() {
+    // Initialize a new RPC server and create an associated client.
+    let rpc_server_addr = new_rpc_server::<Testnet2, Client<Testnet2>, RocksDB>(None).await;
+    let rpc_client = HttpClientBuilder::default()
```
niklaslong: Might be worth abstracting like the server. Unless you think we'll need direct access to the build options here in future?

ljedrz:
Some jobs seem to be failing for unrelated reasons:
```
W: Failed to fetch http://archive.ubuntu.com/ubuntu/dists/focal/InRelease  Could not connect to archive.ubuntu.com:80 (91.189.88.152), connection timed out Could not connect to archive.ubuntu.com:80 (91.189.88.142), connection timed out
```

niklaslong (commented):
Left some comments, looks good other than that `u128` panic in the tests.

ljedrz:
> Left some comments, looks good other than that `u128` panic in the tests.

I already proposed the hex-encoding solution in https://github.com/AleoHQ/snarkOS/pull/1602/commits/fb1ec165dcd5380c9296024a29172007de1fabbf.

src/rpc/tests/mod.rs
ljedrz: we're mostly interested in the server capabilities, and the client is only used in tests; if you mean that we should have a shorter call to create it, I'm all for it :+1:

niklaslong: Yes, just noticed the repetition. I don't mind code being less DRY in tests but couldn't see any upsides to it here.

ljedrz: Nah, you're 100% correct, I was just in a copy-paste frenzy and forgot that it could have been compressed.

niklaslong:
>I already proposed the hex-encoding solution in fb1ec16

My bad, didn't update my local branch after your recent push 👍

dvdplm:
Happy to see this happen, we (Parity) are fans of Aleo and would love to see you as users. Let us know if we can do anything to help. :)

niklaslong (approved):

howardwu (approved):

howardwu:
LGTM.

Due to the number of systems depending on the current format for `cumulative_weight` (as a `u128` rather than a `hex-string`), I've included `arbitrary-precision` to prevent breakages. Note that existing systems already pass in the `id` as a string for RPC, so the numeric `id` shouldn't be an issue. (Our documentation also reflects a string `id` in its examples)

We should investigate a real solution to displaying a `u128` without breakage of the `id` field, as the `cumulative_weight` is likely the **most** important value returned from `getnodestate` (it defines how heavy the canonical chain is, which is read by the explorer, miners, and mining pool operators)

PR #1618, opened 2022-02-08
Return shares for prover regardless of canon block status

jules:


storage/src/state/operator.rs
```diff
@@ -129,10 +129,7 @@ impl<N: Network> SharesState<N> {
                 if !shares.contains_key(prover) {
                     None
                 } else {
-                    match ledger.contains_commitment(&coinbase_record.commitment()) {
-                        Ok(true) => shares.get(prover).copied(),
-                        Ok(false) | Err(_) => None,
-                    }
+                    shares.get(prover).copied()
```
ljedrz: this doubles the lookup, no? instead, we should just do `.filter_map(|((_, coinbase_record), shares)| shares.get(prover).copied())`

jules: completely right - just updated it

ljedrz (approved):
LGTM 👍

niklaslong (approved):

howardwu (approved):

howardwu:
Great catch @julesdesmit

PR #1622, opened 2022-02-09
Ensure the operator is connected to the prover before sending work

jules:
This avoids an infinite loop in which a worker continually
asks for work but cant get it back from the pool as it
is not properly connected.

### DIFF
--- a/src/network/operator.rs
+++ b/src/network/operator.rs
@@ -241,6 +241,26 @@ impl<N: Network, E: Environment> Operator<N, E> {
         match request {
             OperatorRequest::PoolRegister(peer_ip, address) => {
                 if let Some(block_template) = self.block_template.read().await.clone() {
+                    // Ensure that we're connected to the prover.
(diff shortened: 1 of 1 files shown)


ljedrz:
Other than the fact that I agree that we should instead have something like https://github.com/AleoHQ/snarkOS/issues/1554 in place, this workaround should be adjusted as follows:
1. the first few early returns in `Peers::update` should send `Err(<some meaningful error>)` to the `Sender` in `connection_result`, so we can actually know if we weren't already connected or if there's some other connection attempt error
2. instead of ignoring the result of the corresponding `Receiver` (`let _ = handler.await;`), we should check whether we succeeded at connecting, were already connected or if there was an error and potentially act differently if ultimately we're not connected after all

ljedrz:
The CI jobs should succeed once https://github.com/AleoHQ/snarkOS/pull/1625 is merged and this PR is rebased on top of it.

aleo-svc-exp:

PR #1624, opened 2022-02-09
Provide peers with the reason for disconnecting from them

### LINKED ISSUE #1384: [Feature] Provide better information on why a disconnect happens
## 🚀 Feature

Failure to connect to a sync node is often reported as an issue; we should either provide better disconnect details or verify that the node's ports are open and accessible during startup in order for this to be more clear to the users.

ljedrz:
This PR extends `Message::Disconnect` with a `DisconnectReason`, which allows peers to know why they were disconnected from. This change is backwards-compatible, returning `DisconnectReason::NoReasonGiven` for older messages.

This will aid all the issues where the reason behind frequent disconnects is unknown.

Closes https://github.com/AleoHQ/snarkOS/issues/1384.


ljedrz:
The CI jobs should succeed once https://github.com/AleoHQ/snarkOS/pull/1625 is merged and this PR is rebased on top of it.

niklaslong (approved):
Great, this will be most useful! 👍

howardwu (approved):

PR #1630, opened 2022-02-12
Fix handshake disconnect

ljedrz:
https://github.com/AleoHQ/snarkOS/pull/1624 introduced a bug I missed during tests: the counter-connect attempt sends a `Disconnect::YourPortIsClosed` even if the connection succeeds. This PR fixes this, and improves related disconnect logs.
 note: only the https://github.com/AleoHQ/snarkOS/commit/2e3ffbfa939e28d567b55b41465af20fbaf81cc7 commit is relevant, the other is an oddly large `cargo fmt` change.

### DIFF
--- a/src/network/peer.rs
+++ b/src/network/peer.rs
@@ -35,7 +35,7 @@ use crate::{
 };
 use snarkvm::dpc::prelude::*;
 
-use anyhow::{anyhow, Result};
+use anyhow::{anyhow, bail, Result};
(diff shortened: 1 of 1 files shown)


ljedrz:
I'll merge this myself as a hotfix, as right now we can have peers whose listening port is open rejected during the handshake. It doesn't happen every time, but it can cause some annoying connectivity issues.

PR #1639, opened 2022-02-16
Pending PR bundle

ljedrz:
This bundle consists of the following PRs applied in the given order:
- https://github.com/AleoHQ/snarkOS/pull/1542
- https://github.com/AleoHQ/snarkOS/pull/1546
- https://github.com/AleoHQ/snarkOS/pull/1615
- https://github.com/AleoHQ/snarkOS/pull/1603

As requested, it does not have a migration routine, and thus is incompatible with the current storage. However, it can still be used with a fresh storage to sync with the network, and I will be running tests on it today.


ljedrz:
In addition, this PR now also changes the RocksDB compression algorithm from `Snappy` (which is the crate's default for compatibility reasons) to `LZ4`, which has a higher compression ratio, and is faster at both compression and decompression ([source](https://github.com/facebook/zstd#benchmarks)).

note: `zstd`, which is also available in RocksDB, has a higher compression ratio than `LZ4`, but lower (de)compression speed.

ljedrz:
I've encountered no issues during tests :ok_hand:.

ljedrz:
Update: https://github.com/AleoHQ/snarkOS/pull/1639/commits/d364d710062b7d8b822517f967b271a2ea9a7508 now enables the use of the canon reader, now that `rocksdb` uses a version above `6.25` (`6.28.2`).

howardwu:
I have not reviewed these PRs, but merging them in and will make subsequent PRs as needed.

PR #1641, opened 2022-02-17
Faster storage tests and benchmarks

ljedrz:
The storage benchmarks and the `test_ledger_validation` test are currently very slow to execute due to the low deserialization speed of the 1k test blocks. This PR takes advantage of the `Storage::import` functionality paired with a storage dump at 1k blocks that allow the aforementioned tests and benchmarks to run much faster, making `test_ledger_validation` viable for running automatically, as opposed to being `#[ignore]`d (as is the case right now).


ljedrz:
Rebased.

storage/benches/opening.rs
```diff
@@ -14,35 +14,33 @@
...
+        .import("benches/storage_1k_blocks")
+        .expect("Couldn't import the test ledger");
+    // Drop the ledger, as we will be benchmarking re-opening it.
+    drop(ledger);
 
     c.bench_function("Ledger::open_writer", |b| {
         b.iter(|| {
             let _ledger: LedgerState<Testnet2> =
-                LedgerState::open_writer_with_increment::<RocksDB, _>(&temp_dir, NUM_BLOCKS as u32).expect("Failed to initialize ledger");
+                LedgerState::open_writer_with_increment::<RocksDB, _>(&temp_dir, 1_000).expect("Failed to initialize ledger");
```
niklaslong: We probably still want this to be a constant, even if it is scoped (num blocks)?

storage/benches/insertion.rs
```diff
@@ -14,37 +14,47 @@
...
+        LedgerState::open_writer_with_increment::<RocksDB, _>(&temp_dir1, 1).expect("Failed to initialize ledger");
+    // Import a dump of a ledger containing 1k blocks.
+    ledger1
+        .storage()
+        .import("benches/storage_1k_blocks")
+        .expect("Couldn't import the test ledger");
+    // Reopen the ledger so that it applies the storage changes to its in-memory components.
+    drop(ledger1);
+    let ledger1: LedgerState<Testnet2> =
+        LedgerState::open_writer_with_increment::<RocksDB, _>(&temp_dir1, 1_000).expect("Failed to initialize ledger");
```
niklaslong: ditto

storage/benches/opening.rs
ljedrz: I considered it, but it's the only place where it's used, and the only reason for altering it before was the option to reduce that number for performance, which is no longer needed.

niklaslong (approved):
LGTM, left a few comments 👍

ljedrz:
Rebased; also, for the reference, below is a comparison between the current setup time of the `lookups` benchmark and the improvement this PR provides:
```
before: 243.67s
after:  788.48ms
```

ljedrz:
After some additional benchmarking, the extra commit https://github.com/AleoHQ/snarkOS/pull/1641/commits/5de93fcfe1d4d62d228e3a23e108573df761a27c was found to improve the performance of storage operations.

howardwu (approved):

PR #1651, opened 2022-02-23
Feat: basic connectivity metrics

niklaslong:
This PR adds some basic connectivity metrics. I've also added some tests that currently cover initialisation and connect/disconnect. These need to be run sequentially as the recorder used for tests stores state globally. Each test is responsible for clearing its recorder state.

Drafted as a Grafana dashboard still needs to be added and CI likely needs some adjusting.

### DIFF
--- a/.integration/tests/metrics/mod.rs
+++ b/.integration/tests/metrics/mod.rs
@@ -16,8 +16,8 @@
 
 use crate::wait_until;
 
+use snarkos_integration::{ClientNode, TestNode};
 use snarkos_metrics as metrics;
-use snarkos_testing::{ClientNode, TestNode};
 
 use pea2pea::Pea2Pea;
 
--- a/.integration/tests/metrics/mod.rs
+++ b/.integration/tests/metrics/mod.rs
@@ -23,35 +23,14 @@ use pea2pea::Pea2Pea;
 
 #[tokio::test]
 async fn metrics_initialization() {
-    // Start a test node.
-    let _test_node = TestNode::default().await;
-
     // Initialise the metrics, we need to call this manually in tests.
-    let snapshotter = metrics::initialize();
-
-    // Start a snarkOS node.
-    let _client_node = ClientNode::default().await;
+    let metrics = metrics::TestMetrics::new();
 
     // Verify the metrics have been properly initialised, expect the block height to be set.
-    assert_eq!(
-        metrics::get_metric(&snapshotter, metrics::blocks::HEIGHT),
-        metrics::MetricVal::Gauge(0.0)
-    );
-    assert_eq!(
-        metrics::get_metric(&snapshotter, metrics::peers::CONNECTED),
-        metrics::MetricVal::Gauge(0.0)
-    );
-    assert_eq!(
-        metrics::get_metric(&snapshotter, metrics::peers::CANDIDATE),
-        metrics::MetricVal::Gauge(0.0)
-    );
-    assert_eq!(
-        metrics::get_metric(&snapshotter, metrics::peers::RESTRICTED),
-        metrics::MetricVal::Gauge(0.0)
-    );
-
-    // Clear the recorder to avoid the global state bleeding into other tests.
-    metrics::clear_recorder();
+    assert_eq!(metrics.get_val_for(metrics::blocks::HEIGHT), metrics::MetricVal::Gauge(0.0));
+    assert_eq!(metrics.get_val_for(metrics::peers::CONNECTED), metrics::MetricVal::Gauge(0.0));
+    assert_eq!(metrics.get_val_for(metrics::peers::CANDIDATE), metrics::MetricVal::Gauge(0.0));
+    assert_eq!(metrics.get_val_for(metrics::peers::RESTRICTED), metrics::MetricVal::Gauge(0.0));
 }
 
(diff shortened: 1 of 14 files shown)


niklaslong:
Rebased.

.integration/tests/metrics/mod.rs
```diff
@@ -0,0 +1,111 @@
...
+        metrics::get_metric(&snapshotter, metrics::peers::CANDIDATE),
+        metrics::MetricVal::Gauge(0.0)
+    );
+    assert_eq!(
+        metrics::get_metric(&snapshotter, metrics::peers::RESTRICTED),
+        metrics::MetricVal::Gauge(0.0)
+    );
+
+    // Clear the recorder to avoid the global state bleeding into other tests.
+    metrics::clear_recorder();
```
ljedrz: if all the metrics tests begin with setting up a `Snapshotter` and end with a call to `metrics::clear_recorder()`, perhaps we could make our own wrapper `struct` for the former that would call the latter in a custom `Drop` impl?

.integration/tests/metrics/mod.rs
```diff
@@ -0,0 +1,111 @@
...
+
+use snarkos_integration::{ClientNode, TestNode};
+use snarkos_metrics as metrics;
+
+use pea2pea::Pea2Pea;
+
+#[tokio::test]
+async fn metrics_initialization() {
+    // Start a test node.
+    let _test_node = TestNode::default().await;
```
ljedrz: is this needed in order for the test to work?

.integration/tests/metrics/mod.rs
```diff
@@ -0,0 +1,111 @@
...
+
+    // Initialise the metrics, we need to call this manually in tests.
+    let snapshotter = metrics::initialize();
+
+    // Start a snarkOS node.
+    let _client_node = ClientNode::default().await;
```
ljedrz: ditto

network/src/ledger.rs
```diff
@@ -510,12 +513,17 @@ impl<N: Network, E: Environment> Ledger<N, E> {
                 // Attempt to add the unconfirmed block as the next block in the canonical chain.
                 false => match self.canon.add_next_block(&unconfirmed_block) {
                     Ok(()) => {
+                        // Allocation necessary if prometheus is enabled.
```
ljedrz:
```suggestion
```
since there is no functional difference either way, no need for the comment; also, I think it was meant to say "Assignment" instead

network/src/ledger.rs
```diff
@@ -510,12 +513,17 @@ impl<N: Network, E: Environment> Ledger<N, E> {
...
+                        let latest_block_height = self.canon.latest_block_height();
                         info!(
                             "Ledger successfully advanced to block {} ({})",
-                            self.canon.latest_block_height(),
+                            latest_block_height,
                             self.canon.latest_block_hash()
                         );
 
+                        #[cfg(any(feature = "test", feature = "prometheus"))]
+                        metrics::gauge!(metrics::blocks::HEIGHT, latest_block_height as f64);
```
ljedrz, line 524: this point isn't bad, but I wonder if the end of the happy path in `LedgerState::add_next_block` (as soon as `LedgerState::latest_block` is updated) is not a more future-proof place to update this metric :thinking:

ljedrz: hmm, on second thought, I can see how `snarkos-storage` is not something that needs to be concerned with metrics; let's leave it as is, then :+1:

network/src/ledger.rs
```diff
@@ -562,7 +570,12 @@ impl<N: Network, E: Environment> Ledger<N, E> {
 
         match self.canon.revert_to_block_height(block_height) {
             Ok(removed_blocks) => {
-                info!("Ledger successfully reverted to block {}", self.canon.latest_block_height());
+                // Allocation necessary if prometheus is enabled.
```
ljedrz:
```suggestion
```
(same reasoning as before)

network/src/peers.rs
```diff
@@ -560,6 +600,12 @@ impl<N: Network, E: Environment> Peers<N, E> {
                 if let Err(error) = outbound.send(message).await {
                     trace!("Outbound channel failed: {}", error);
                     self.connected_peers.write().await.remove(&peer);
+
+                    #[cfg(any(feature = "test", feature = "prometheus"))]
+                    {
+                        let number_of_connected_peers = self.number_of_connected_peers().await;
+                        metrics::gauge!(metrics::peers::CONNECTED, number_of_connected_peers as f64);
+                    }
```
ljedrz, line 608: this metric would ideally only be updated at `PeersRequest::{PeerConnected, PeerDisconnected}`; would it be possible to instead remove `self.connected_peers.write().await.remove(&peer);`, the selected section, and instead to send a `PeerDisconnected` message to the peer handler from here?

ljedrz (commented):
LGTM in general, left a few comments :+1:

network/src/peers.rs
niklaslong: I agree, though I think this implies changes to logic outside of metrics, happy to submit a follow up to address this though 👍

ljedrz (approved):

howardwu (approved):

PR #1652, opened 2022-02-24
More comprehensive resource management

ljedrz:
This PR is a follow-up to https://github.com/AleoHQ/snarkOS/pull/1615; it extends the current resource management setup with the possibility of deregistering individual resources based on an identifier that needs to be procured in advance. Why in advance? Two reasons:

1. While task accounting could be done in the following fashion:
```
pub fn register_task<T>(&self, future: T)
where
    T: Future + Send + 'static,
    T::Output: Send + 'static,
{
    let resource_id = self.procure_id();
    let resources = self.clone();

    let task_handle = tokio::task::spawn(async move {
        let _ = future.await;
        resources.deregister(resource_id);
    });
    let resource = Resource::Task(task_handle);

    self.register(resource, Some(resource_id));
}
```
this would only allow their automatic "self-destruction", but not ad-hoc/on-demand individual termination, e.g. if it needs to be triggered elsewhere. In other words, this design provides greater utility.

2. This is not feasible for threads, because those can't be aborted from "the outside" (hence the need for an `abort_sender` when they are registered, which is then incorporated inside the body of their inner function. Working around this is technically possible, but would be quite ugly and restrictive.

I also decided to be explicit and always require an optional resource id argument instead of introducing a method like `Resources::register_anonymous_task` that would internally call `::register_task` and pass `None` as the resource id; this was done in order to always be aware what the implications of not providing an id are, and all callsites have appropriate comments.

Another design consideration is the `ResourceId`; in order to maintain the current lockless structure, the resource ids are only ever increasing, so it's possible for a resource to get an id going into thousands (and beyond) with only a handful of resources actually registered. It is possible to reuse expired ids, but it would increase the complexity of implementation and likely have a slight performance impact. Since `usize` is typically 64 bits in size these days, we are VERY unlikely to ever run out of available ids, even if the node is running non-stop for years.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2693,6 +2693,7 @@ dependencies = [
  "snarkvm",
  "tokio",
  "tracing",
+ "tracing-subscriber",
 ]
 
 [[package]]
--- a/environment/Cargo.toml
+++ b/environment/Cargo.toml
@@ -33,7 +33,7 @@ version = "0.8.0"
 
 [dependencies.tokio]
 version = "1"
-features = ["sync", "rt"]
+features = ["sync", "rt", "time"]
 
 [dependencies.tracing]
 version = "0.1"
--- a/environment/Cargo.toml
+++ b/environment/Cargo.toml
@@ -37,3 +37,10 @@ features = ["sync", "rt"]
 
 [dependencies.tracing]
 version = "0.1"
+
+[dev-dependencies.tracing-subscriber]
+version = "0.3"
+
+[dev-dependencies.tokio]
+version = "1"
+features = ["macros", "rt-multi-thread", "time"]
--- a/environment/src/helpers/resources.rs
+++ b/environment/src/helpers/resources.rs
@@ -14,7 +14,29 @@
 // You should have received a copy of the GNU General Public License
 // along with the snarkOS library. If not, see <https://www.gnu.org/licenses/>.
 
+use std::{
+    collections::HashMap,
+    sync::{
+        atomic::{AtomicUsize, Ordering},
+        Arc,
+    },
+};
 use tokio::sync::{mpsc, oneshot};
+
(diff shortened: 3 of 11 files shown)


ljedrz:
Note: when I missed one change related to this PR, I also saw and fixed some `clippy` lints; they are in https://github.com/AleoHQ/snarkOS/pull/1652/commits/df059447875a0c73c7646c4366197a41fd10d7fe.

ljedrz:
Note: when I missed one change related to this PR, I also saw and fixed some `clippy` lints; they are in https://github.com/AleoHQ/snarkOS/pull/1652/commits/df059447875a0c73c7646c4366197a41fd10d7fe.

ljedrz:
Rebased.

niklaslong (approved):

howardwu:
Can we have it so we don't define a `let task` variable, instead just `E::register_task({the task})`?

ljedrz:
We can, though note that then we'll have some callsites like:
```
E::resources().register_task(tokio::task::spawn(async move {
    ...
}), None); // this is potentially many lines later
```
And it might be unclear what the `None` refers to.

ljedrz:
Rebased and re-formatted in the last commit. @howardwu your call if it's preferable, feel free to keep or drop the final commit.

howardwu:
How about changing the syntax of `register_task` to `register_task(Option<id>, task)` to address the visibility/readability concern?

ljedrz:
Sounds like a good idea 👍.

ljedrz:
Rebased and changed the argument position.

ljedrz:
I noticed that sometimes the aborted tasks might not have enough time to free their resources, and two extra places where tasks needed to be deregistered - fixed.

howardwu (approved):

ljedrz:
Rebased.

PR #1654, opened 2022-02-26
Update Rust to 1.59 in CI, update edition to 2021

howardwu:
## Motivation
Update Rust to 1.59 in CI, update edition to 2021


ljedrz (approved):
LGTM; note: edition 2021 only requires Rust 1.56+, but I'm all for a higher requirement for the sake of future-proofing; also, we can later add:
```
rust-version = "1.59"
```
to the `[package]` sections of the crates to enforce this new requirement.

PR #1662, opened 2022-03-03
Feat: network based crawler implementation

niklaslong:
Supersedes #1533. 

This PR introduces a network-based crawler implementation. There are two new crates:
- `.crawler`: which, self-explanatorily, contains the crawler impl.
- `.synthetic_node`: which contains the bare-bones pea2pea-based node which powers both the peers used in some tests and the crawler.

Currently, the data is stored in memory and isn't exportable yet; how we implement this is still tbd.

This is a fairly primitive implementation and still needs to be tested but the main structures are there.


.crawler/src/constants.rs
```diff
@@ -0,0 +1,25 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with the snarkOS library. If not, see <https://www.gnu.org/licenses/>.
+
+use snarkos_environment::{Client, CurrentNetwork, Environment};
+
+pub const DESIRED_CONNECTIONS: usize = <Client<CurrentNetwork>>::MINIMUM_NUMBER_OF_PEERS * 3;
```
ljedrz: Are we using this `const` anywhere? Also, I wonder if we need to control the number of connections (except for some maximum), as the crawler doesn't have to maintain any at any given time to function properly - it just needs to initiate them briefly from time to time to collect new network data.

.crawler/src/constants.rs
```diff
@@ -0,0 +1,25 @@
...
+pub const LOG_INTERVAL_SECS: u64 = 10;
+pub const MAXIMUM_NUMBER_OF_PEERS: usize = 10000;
```
ljedrz: We can probably handle that many at a time, but perhaps we should decrease it (to maybe 1k?) so as not to run into open file descriptor issues.

.crawler/src/crawler.rs
```diff
@@ -0,0 +1,274 @@
...
+    Node as Pea2PeaNode,
+    Pea2Pea,
+};
+use std::{convert::TryInto, io, net::SocketAddr, ops::Deref, sync::Arc, time::Duration};
+use structopt::StructOpt;
+use tokio::task;
+use tracing::*;
+
+// CLI
+#[derive(Debug, StructOpt)]
```
ljedrz, line 38: I wonder if we need to use `structopt` in the crawler if the only parameter is the listening address; I think just `clap` (a dependency of `structopt`) would suffice. In any case, it's not a big deal, just a thought for later.

.crawler/src/crawler.rs
```diff
@@ -0,0 +1,274 @@
...
+    pub fn send_pings(&self) {
+        let node = self.clone();
+        task::spawn(async move {
+            let genesis = CurrentNetwork::genesis_block();
+            let ping_msg = ClientMessage::Ping(
+                MESSAGE_VERSION,
+                MAXIMUM_FORK_DEPTH,
+                node.node_type(),
+                node.state(),
+                genesis.hash(),
```
ljedrz, line 102: not an immediate concern, but we might want to track the network's top height and use that one here instead so as not to suggest that there's a node out there that can't sync properly :smiley:.

.crawler/src/crawler.rs
```diff
@@ -0,0 +1,274 @@
...
+        task::spawn(async move {
+            loop {
+                // Disconnect from peers we have just crawled.
+                for addr in node.known_network.addrs_to_disconnect() {
+                    node.node().disconnect(addr).await;
+                }
+
+                // Connect to peers we haven't crawled in a while.
+                for addr in node.known_network.addrs_to_connect() {
+                    if !node.node().is_connected(addr) {
```
ljedrz: this check is not needed (especially since we're ignoring errors), as `pea2pea` already does this internally; if we do notice related `WARN`s, we should instead make sure the collections of addresses to connect to is maintained correctly

.crawler/src/crawler.rs
```diff
@@ -0,0 +1,274 @@
...
+                MAXIMUM_FORK_DEPTH,
+                node.node_type(),
+                node.state(),
+                genesis.hash(),
+                Data::Object(genesis.header().clone()),
+            );
+
+            loop {
+                if node.node().num_connected() != 0 {
+                    info!(parent: node.node().span(), "sending out Pings");
```
ljedrz: nit: we have some logs starting with a capital letter and some with a lowercase one; we should be consistent, and probably maintain the convention from snarkOS (uppercase)

.crawler/src/crawler.rs
```diff
@@ -0,0 +1,274 @@
...
+        let mut buf = [0u8; 8 * 1024];
+
+        reader.read_exact(&mut buf[..MESSAGE_LENGTH_PREFIX_SIZE])?;
+        let len = u32::from_le_bytes(buf[..MESSAGE_LENGTH_PREFIX_SIZE].try_into().unwrap()) as usize;
+
+        if reader.read_exact(&mut buf[..len]).is_err() {
+            return Ok(None);
+        }
+
+        match ClientMessage::deserialize(&mut io::Cursor::new(&buf[..len])) {
```
ljedrz, line 197: TODO: only partially deserialize the client messages at first; we need to first read the message id, and immediately reject message types we're not interested in, especially blocks, txs, pongs and anything else that has deferred deserialization in the true node.

.crawler/src/crawler.rs
```diff
@@ -0,0 +1,274 @@
...
+            ClientMessage::BlockRequest(_start_block_height, _end_block_height) => {}
+            ClientMessage::BlockResponse(_block) => {}
+            ClientMessage::Disconnect(reason) => {
+                debug!("Peer {} disconnected for the following reason: {:?}", source, reason);
+            }
+            ClientMessage::PeerRequest => self.process_peer_request(source).await?,
+            ClientMessage::PeerResponse(peer_ips) => self.process_peer_response(source, peer_ips).await?,
+            ClientMessage::Ping(version, _fork_depth, _peer_type, _peer_state, _block_hash, block_header) => {
+                // Deserialise the block header.
+                let block_header = block_header.deserialize().await.unwrap();
```
ljedrz, line 221: TODO: we should probably manually deserialize the header, as we only need the height, and we need to be able to quickly handle any number of such messages

.crawler/src/crawler.rs
```diff
@@ -0,0 +1,274 @@
...
+        }
+
+        Ok(())
+    }
+}
+
+// Helper methods.
+impl Crawler {
+    async fn process_peer_request(&self, source: SocketAddr) -> io::Result<()> {
+        let peers = vec![];
```
ljedrz: we should probably share some peers with the nodes that request them; they can be chosen at random for now

.crawler/src/crawler.rs
```diff
@@ -0,0 +1,274 @@
...
+
+        Ok(())
+    }
+
+    async fn process_ping(&self, source: SocketAddr, version: u32, block_height: u32) -> io::Result<()> {
+        // Ensure the message protocol version is not outdated.
+        if version < MESSAGE_VERSION {
+            warn!(parent: self.node().span(), "dropping {} due to outdated version ({})", source, version);
+            return Err(io::ErrorKind::InvalidData.into());
+        }
```
ljedrz: TODO: we should probably maintain a detailed list of non-compliant peers so we can report their numbers and reasons for non-compliance with the protocol

.crawler/src/crawler.rs
ljedrz: on second thought, that would require us to send proper block locators in `Pong`s :thinking:; let's leave it be, then, at least for now

.crawler/src/crawler.rs
```diff
@@ -0,0 +1,274 @@
...
+
+        // Update the known network nodes and update the crawl state.
+        self.known_network.update_height(source, block_height);
+        self.known_network.received_height(source);
+
+        let genesis = CurrentNetwork::genesis_block();
+        let msg = ClientMessage::Pong(
+            None,
+            Data::Object(
+                BlockLocators::<CurrentNetwork>::from(vec![(genesis.height(), (genesis.hash(), None))].into_iter().collect()).unwrap(),
```
ljedrz, line 313: TODO: we'll be sending this bad boy out very often, so we might as well create this object just once and copy it over whenever needed.

.crawler/src/crawler.rs
```diff
@@ -0,0 +1,274 @@
...
+            ),
+        );
+
+        info!(parent: self.node().span(), "sending a Pong to {}", source);
```
ljedrz: `INFO` for individual messages is likely a bit too verbose; I'd leave that level for the desired metrics, and keep everything else at `DEBUG`

.crawler/src/known_network.rs
```diff
@@ -0,0 +1,257 @@
...
+};
+use time::OffsetDateTime;
+
+use crate::connection::{nodes_from_connections, Connection};
+
+// Purges connections that haven't been seen within this time (in hours).
+const STALE_CONNECTION_CUTOFF_TIME_HRS: i64 = 4;
+
+// The interval for revisiting connections.
+const CRAWL_INTERVAL_MINS: i64 = 10;
```
ljedrz: we might want to put these `const`s in `constants.rs`, so there's only one spot where everything crawler-related can be tweaked easily

.crawler/src/known_network.rs
```diff
@@ -0,0 +1,257 @@
...
+
+/// Node information collected while crawling.
+#[derive(Debug, Default, Clone)]
+pub struct NodeMeta {
+    last_height: u32,
+    // Set on disconnect.
+    last_crawled: Option<OffsetDateTime>,
+
```
ljedrz, line 37: I know that we'll also have this info available in connections, but we might also want to keep the node's address here, we might need it in the future for more detailed analysis

ljedrz (commented):
Looking good so far :ok_hand:! Left some comments, all the TODOs can be considered follow-up material.

.crawler/src/constants.rs
```diff
@@ -0,0 +1,25 @@
...
+pub const MESSAGE_VERSION: u32 = <Client<CurrentNetwork>>::MESSAGE_VERSION;
+pub const PING_INTERVAL_SECS: u64 = 5;
+pub const PEER_INTERVAL_SECS: u64 = 3;
+pub const SYNC_NODES: &'static [&'static str] = <Client<CurrentNetwork>>::SYNC_NODES;
```
ljedrz:
```suggestion
pub const SYNC_NODES: &'static [&'static str] = <ClientTrial<CurrentNetwork>>::SYNC_NODES;
```
otherwise we'd be using local addresses; feel free to apply this change directly in your further changes, as you'll need to adjust imports too

.crawler/src/crawler.rs
```diff
@@ -0,0 +1,274 @@
...
+}
+
+/// Inbound message processing logic for the crawler nodes.
+#[async_trait::async_trait]
+impl Reading for Crawler {
+    type Message = ClientMessage;
+
+    fn read_message<R: io::Read>(&self, source: SocketAddr, reader: &mut R) -> io::Result<Option<Self::Message>> {
+        // FIXME: use the maximum message size allowed by the protocol or (better) use streaming deserialization.
+        let mut buf = [0u8; 8 * 1024];
```
ljedrz:
```suggestion
        let mut buf = [0u8; 64 * 1024];
```

.crawler/src/crawler.rs
```diff
@@ -0,0 +1,274 @@
...
+
+        reader.read_exact(&mut buf[..MESSAGE_LENGTH_PREFIX_SIZE])?;
+        let len = u32::from_le_bytes(buf[..MESSAGE_LENGTH_PREFIX_SIZE].try_into().unwrap()) as usize;
+
```
ljedrz, line 184: needs a `len <= buf.len()` check

.crawler/src/crawler.rs
ljedrz: note: this won't need to be as large as soon as we first decode the message id and skip the bytes that belong to messages we don't need

.synthetic_node/src/lib.rs
```diff
@@ -0,0 +1,254 @@
...
+        locked_peers.retain(|peer| peer.connected_addr != disconnecting_addr);
+        assert_eq!(locked_peers.len(), initial_len - 1)
+    }
+}
+
+/// Enables tracing for all synth node instances (usually scoped by test).
+pub fn enable_tracing() {
+    use tracing_subscriber::{fmt, EnvFilter};
+
+    fmt().with_test_writer().with_env_filter(EnvFilter::from_default_env()).init();
```
ljedrz: we might want to silence `pea2pea` logs when the PR is ready, otherwise the logs might get too verbose

ljedrz:
```
2022-03-03T10:06:57.136507Z  INFO snarkos_crawler::crawler: Known addresses: 196
2022-03-03T10:06:57.136550Z  INFO snarkos_crawler::crawler: Known connections: 3085
```
Seems to be working (once I adjusted the sync node addrs), but it might need some further tweaking to improve crawling performace - pretty sure it could be much faster.

.crawler/src/crawler.rs
niklaslong: Will mark this as a TODO 👍

niklaslong:
I've updated this PR with the critical comments and marked TODOs for the rest. There's still room for improvement but this may well be a good point to consider a first merge, given that it's a blocker for other changes.

.crawler/src/main.rs
```diff
@@ -0,0 +1,44 @@
...
+#[tokio::main]
+async fn main() {
+    // Read configuration options.
+    let opts = Opts::from_args();
+
+    // Configure and start crawler.
+    let crawler = Crawler::new(opts).await;
+
+    // Enable tracing for the crawler.
+    // snarkos_synthetic_node::enable_tracing();
```
ljedrz: I wouldn't disable the logs completely (especially since we don't have an RPC to get the network details yet), but rather filter them out in the config inside `enable_tracing`, so as not to display `TRACE` and `DEBUG` logs from `pea2pea` (too verbose for the crawler's needs).

.crawler/src/crawler.rs
```diff
@@ -0,0 +1,287 @@
...
+    type Message = ClientMessage;
+
+    fn read_message<R: io::Read>(&self, source: SocketAddr, reader: &mut R) -> io::Result<Option<Self::Message>> {
+        // FIXME: use the maximum message size allowed by the protocol or (better) use streaming deserialization.
+        let mut buf = [0u8; 64 * 1024];
+
+        reader.read_exact(&mut buf[..MESSAGE_LENGTH_PREFIX_SIZE])?;
+        let len = u32::from_le_bytes(buf[..MESSAGE_LENGTH_PREFIX_SIZE].try_into().unwrap()) as usize;
+
+        if len > buf.len() || reader.read_exact(&mut buf[..len]).is_err() {
```
ljedrz: if `len > buf.len()`, we should return `Err(io::ErrorKind::InvalidData.into())`, as this is breaking the protocol; we currently don't allocate the maximum possible size (as we will want to ignore most large messages), but anything that's above our buffer length is a no-no

ljedrz:
I applied some fixes based on a round of tests I did against the current testnet; the TODOs remain follow-up material.

howardwu (approved):

howardwu:
Note there is a lot of hardcoded handshake logic here. It will likely be broken as I update peer designs.

PR #1663, opened 2022-03-03
Squash a few perf TODOs

ljedrz:
Takes care of 3 perf-related TODOs.

### DIFF
--- a/storage/src/state/ledger.rs
+++ b/storage/src/state/ledger.rs
@@ -956,17 +956,8 @@ impl<N: Network> LedgerState<N> {
 
         // Compute the transactions inclusion proof.
         let transactions_inclusion_proof = {
-            // TODO (howardwu): Optimize this operation.
-            let index = transactions
-                .transaction_ids()
-                .enumerate()
-                .filter_map(|(index, id)| match id == transaction_id {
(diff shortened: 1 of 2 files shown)


storage/src/state/ledger.rs
```diff
@@ -956,17 +956,8 @@ impl<N: Network> LedgerState<N> {
...
-                .transaction_ids()
-                .enumerate()
-                .filter_map(|(index, id)| match id == transaction_id {
-                    true => Some(index),
-                    false => None,
-                })
-                .collect::<Vec<_>>();
-            assert_eq!(1, index.len()); // TODO (howardwu): Clean this up with a proper error handler.
-            transactions.to_transactions_inclusion_proof(index[0], transaction_id)?
+            let index = transactions.transaction_ids().position(|id| id == transaction_id).unwrap();
```
ljedrz, line 959: note on the `unwrap()` - it's no less safe than `index[0]` that's currently used, and if I'm not mistaken, line 941 guarantees that we have already received the block with the transaction in question, so it's safe

niklaslong (approved):

howardwu (approved):

PR #1668, opened 2022-03-07
Improve the crawler, add a related test

ljedrz:
This PR is a collection of cleanups and performance improvements for the crawler plus a test that confirms that it works as intended. I'm not adding that test to the list of CI jobs due to https://github.com/AleoHQ/snarkOS/issues/1667, but it does pass :ok_hand:.

### DIFF
--- a/.crawler/tests/peering.rs
+++ b/.crawler/tests/peering.rs
@@ -1,0 +1,76 @@
+// Copyright (C) 2019-2022 Aleo Systems Inc.
+// This file is part of the snarkOS library.
+
+// The snarkOS library is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
+// The snarkOS library is distributed in the hope that it will be useful,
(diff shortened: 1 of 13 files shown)


.crawler/tests/peering.rs
```diff
@@ -0,0 +1,76 @@
...
+    let crawler = Crawler::new(opts).await;
+
+    // "Seed" the crawler with the address of the first node.
+    crawler.known_network.add_node(test_node_addrs[0]);
+
+    // Initialize the crawler.
+    crawler.run_periodic_tasks();
+
+    let now = Instant::now();
+    loop {
```
niklaslong: Could we use `wait_until!` here?

ljedrz: This is _exactly_ `wait_until` :smile:; hmm I could try moving it around in `snarkos-integration` (so it can be imported from there) or find some other, better place for it; I'll do that tomorrow :ok_hand:.

niklaslong (approved):
LGTM 👍

ljedrz:
Rebased and took care of the review comment.

PR #1672, opened 2022-03-09
Further improvements to the network crawler

ljedrz:
Builds on https://github.com/AleoHQ/snarkOS/pull/1668, only the last 2 commits are new (squashed due to rebases becoming tricky; [relevant diff](https://github.com/AleoHQ/snarkOS/pull/1672/files/fd2709527cfbe8a5471f557a9b8c05ee58edfc47..6fcc8a7065aeee0fa29305afab6d5d841c3256e6)).

This PR documents the network-based crawler, makes it _much_ faster (the same improvement extends to the underlying synthetic node used in the network integration tests), and tweaks its default configuration.

### DIFF
--- a/.crawler/src/known_network.rs
+++ b/.crawler/src/known_network.rs
@@ -15,8 +15,10 @@
 // along with the snarkOS library. If not, see <https://www.gnu.org/licenses/>.
 
 use parking_lot::RwLock;
+use snarkos_environment::helpers::{NodeType, State};
 use std::{
     collections::{HashMap, HashSet},
+    fmt,
     net::SocketAddr,
 };
 use time::OffsetDateTime;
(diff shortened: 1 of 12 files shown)


.crawler/src/known_network.rs
```diff
@@ -26,37 +28,90 @@ use crate::{
...
 
     fn needs_refreshing(&self) -> bool {
-        if let Some(last_crawled) = self.last_crawled {
-            (OffsetDateTime::now_utc() - last_crawled).whole_minutes() > CRAWL_INTERVAL_MINS
+        if let Some(timestamp) = self.timestamp {
+            let crawl_interval = if self.state.is_some() {
+                CRAWL_INTERVAL_MINS
+            } else {
+                // Delay further connection attempts to nodes that are hard to connect to.
+                self.connection_failures as i64
```
niklaslong, line 111: Why is the number of connection failures being used as the crawl interval in this case? Is it just to have some non-zero value to use as an interval if a peer is hard to connect to?

niklaslong (approved):
👍

.crawler/src/known_network.rs
ljedrz: if we can't connect to some node, it's just a simple way to retry more quickly than in the success case (as we want to quickly reconnect to the temporarily unavailable nodes), but with longer and longer delays in case it keeps failing

ljedrz:
Rebased.

PR #1674, opened 2022-03-10
Finishing touches for the network crawler

ljedrz:
This PR is a set of small tweaks to the crawler, its docs and the related test. In addition, it now logs the average handshake time.

### DIFF
--- a/.crawler/src/known_network.rs
+++ b/.crawler/src/known_network.rs
@@ -77,7 +77,7 @@ pub struct NodeMeta {
     listening_addr: SocketAddr,
     // The details of the node's state.
     pub state: Option<NodeState>,
-    // The last connection success/failure timestamp.
+    // The last interaction timestamp.
     timestamp: Option<OffsetDateTime>,
     // The number of lists of peers received from the node.
     received_peer_sets: u8,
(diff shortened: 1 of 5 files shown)


.crawler/src/known_network.rs
```diff
@@ -66,42 +68,36 @@ impl fmt::Debug for NetworkSummary {
...
-            listening_addr,
-            state: None,
-            timestamp: None,
-            received_peer_sets: 0,
-            connection_failures: 0,
-        }
-    }
-
+    // Resets the node's values which determine whether the crawler should stay connected to it.
+    // note: it should be called when a node is disconnected from after it's been crawled successfully
```
niklaslong, line 93:
```suggestion
    // note: it should be called when a peer is disconnected from after it's been crawled successfully
```

niklaslong: Actually we use `node` throughout here, ignore this 👍

.crawler/src/known_network.rs
```diff
@@ -307,6 +315,8 @@ impl KnownNetwork {
         let mut states = HashMap::with_capacity(nodes.len());
         let mut types = HashMap::with_capacity(nodes.len());
         let mut heights = HashMap::with_capacity(nodes.len());
+
+        let mut handshake_times = Vec::with_capacity(nodes.len());
```
niklaslong, line 307: Should we be using a bounded circular queue here instead?

ljedrz: no need, as this collection is created ad-hoc for the currently known nodes only

niklaslong (approved):

PR #1682, opened 2022-03-11
Add feature-gated persistent storage to the network crawler

ljedrz:
This PR adds optional persistent storage to the network crawler. The new dedicated features are `postgres` and `postgres-tls`.

Cc @zosorock


niklaslong (approved):
LGTM 🚀

PR #1687, opened 2022-03-14
Include more network metrics in the crawler, store connections

ljedrz:
This PR introduces some extra node-level and network-level metrics, refactors some of the code, and also stores the found connections in case persistent storage is used.

In addition, a `clippy` pass was done on the crawler code.

Cc @zosorock

### DIFF
--- a/.crawler/src/metrics.rs
+++ b/.crawler/src/metrics.rs
@@ -1,0 +1,416 @@
+// Copyright (C) 2019-2022 Aleo Systems Inc.
+// This file is part of the snarkOS library.
+
+// The snarkOS library is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
+// The snarkOS library is distributed in the hope that it will be useful,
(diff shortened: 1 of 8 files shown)


.crawler/src/metrics.rs
```diff
@@ -0,0 +1,416 @@
...
+            // Returns all metrics set to `0`.
+            return None;
+        }
+
+        let node_count = nodes.len();
+        let connection_count = connections.len();
+        let density = calculate_density(node_count as f64, connection_count as f64);
+
+        // Create an index of nodes to introduce some notion of order the rows and columns all matrices will follow.
+        let index: BTreeMap<SocketAddr, usize> = nodes.keys().enumerate().map(|(i, addr)| (*addr, i)).collect();
```
niklaslong, line 159: We'll want this index to be sorted—order preservation is very helpful for debugging and correlating data over multiple sets of these measurements.

ljedrz: You mean sorted as in a more specified order? As this list already has a predictable order (i.e. is somewhat sorted), which depends on `addr`.

niklaslong: Actually, the ordering of `SocketAddr` seems sufficient here 👍

niklaslong (approved):

PR #1690, opened 2022-03-16
Allow multiple addresses when using --connect

ljedrz:
This PR allows multiple comma-separated addresses to be connected to when using `--connect`; in addition, supplying malformed addresses no longer causes a panic.

Cc @zosorock

### DIFF
--- a/snarkos/node.rs
+++ b/snarkos/node.rs
@@ -196,9 +196,17 @@ impl Node {
             let _display = Display::<N, E>::start(server.clone(), self.verbosity)?;
         };
 
-        // Connect to a peer if one was given as an argument.
-        if let Some(peer_ip) = &self.connect {
-            let _ = server.connect_to(peer_ip.parse().unwrap()).await;
+        // Connect to peer(s) if given as an argument.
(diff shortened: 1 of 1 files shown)


snarkos/node.rs
```diff
@@ -196,9 +196,16 @@ impl Node {
...
-        // Connect to a peer if one was given as an argument.
-        if let Some(peer_ip) = &self.connect {
-            let _ = server.connect_to(peer_ip.parse().unwrap()).await;
+        // Connect to peer(s) if given as an argument.
+        if let Some(peer_ips) = &self.connect {
+            for peer_ip in peer_ips.split(',') {
+                let addr: SocketAddr = if let Ok(addr) = peer_ip.parse() {
+                    addr
+                } else {
+                    continue;
```
niklaslong, line 206: We might want to notify the user that they've supplied malformed addresses?

ljedrz: yeah, we can :ok_hand:

niklaslong (approved):

howardwu (approved):

PR #1692, opened 2022-03-16
Tweaks to the node's cleanups

ljedrz:
This PR is the result of further work on the node cleanup tests; it enables the connection cleanup tests by default, tweaks the other cleanup tests and resource-related logs, and introduces the following changes found while working on them:
- [store the peer's current height instead of last header](https://github.com/AleoHQ/snarkOS/commit/ba5f9d582fc5545d241086784f4b993ab48f29d5) (reduces per-peer memory use)
- ~[remove the default dev sync node address](https://github.com/AleoHQ/snarkOS/commit/ae9a308da6cb64599b5ba0c65caabf0037624690) (it can interfere with network integration tests)~
- [don't attempt to ask for peers without any peers](https://github.com/AleoHQ/snarkOS/commit/ec6de4018f2aaa1862192b4b28d6411c01102fee) (a small performance improvement)

### DIFF
--- a/network/src/peers.rs
+++ b/network/src/peers.rs
@@ -432,10 +432,12 @@ impl<N: Network, E: Environment> Peers<N, E> {
                 // Skip if the number of connected peers is above the minimum threshold.
                 match number_of_connected_peers < E::MINIMUM_NUMBER_OF_PEERS {
                     true => {
-                        trace!("Sending request for more peer connections");
-                        // Request more peers if the number of connected peers is below the threshold.
-                        for peer_ip in self.connected_peers().await.iter().choose_multiple(&mut OsRng::default(), 3) {
-                            self.send(*peer_ip, Message::PeerRequest).await;
+                        if number_of_connected_peers > 0 {
+                            trace!("Sending requests for more peer connections");
+                            // Request more peers if the number of connected peers is below the threshold.
+                            for peer_ip in self.connected_peers().await.iter().choose_multiple(&mut OsRng::default(), 3) {
+                                self.send(*peer_ip, Message::PeerRequest).await;
+                            }
                         }
                     }
                     false => return,
--- a/.integration/src/client_node.rs
+++ b/.integration/src/client_node.rs
@@ -64,16 +64,16 @@ impl ClientNode {
 
         ClientNode { server }
     }
+
+    pub async fn shut_down(&self) {
+        self.server.shut_down().await;
+    }
 }
 
 // Remove the storage artifacts after each test.
 impl Drop for ClientNode {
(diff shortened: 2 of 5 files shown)


niklaslong (approved):

environment/src/lib.rs
```diff
@@ -55,7 +55,7 @@ pub trait Environment: 'static + Clone + Debug + Default + Send + Sync {
     /// The list of beacon nodes to bootstrap the node server with.
     const BEACON_NODES: &'static [&'static str] = &[];
     /// The list of sync nodes to bootstrap the node server with.
-    const SYNC_NODES: &'static [&'static str] = &["127.0.0.1:4135"];
```
howardwu: Please revert this, it breaks development guide instructions.

network/src/peers.rs
```diff
@@ -427,9 +427,9 @@ impl<N: Network, E: Environment> Peers<N, E> {
                 }
 
                 // Skip if the number of connected peers is above the minimum threshold.
-                match number_of_connected_peers < E::MINIMUM_NUMBER_OF_PEERS {
+                match number_of_connected_peers > 0 && number_of_connected_peers < E::MINIMUM_NUMBER_OF_PEERS {
```
howardwu: Is there a reason to only request more peers if we already are connected to peers?

environment/src/lib.rs
ljedrz: happy to revert it for now, but we should probably avoid this interfering with tests; I can propose an adjustment that doesn't affect the dev mode separately :+1:

network/src/peers.rs
ljedrz: if we aren't connected to anyone, we won't be able to send any peer requests; this is just a log cleanup (plus a minor performance improvement due to not calling `connected_peers`), as we currently report that we are `Sending requests for more peers connections` even if no such request is actually sent (since `self.connected_peers()` returns an empty collection).

ljedrz:
Rebased and addressed review comments.

environment/src/lib.rs
ljedrz: done in https://github.com/AleoHQ/snarkOS/pull/1708

network/src/peers.rs
howardwu: Shouldn't that say `self.candidate_peers()`? Otherwise this logic doesn't seem to work.

ljedrz: ah, you meant the early return; I adjusted the logic so that the change only applies to the `true` branch :+1:

howardwu (approved):

PR #1693, opened 2022-03-16
Use a proper codec with Framed

ljedrz:
The current setup is a bit of a hack, and this PR fixes that. No (en/de)coding changes are introduced - just a helper object.

I initially wanted to use [LengthDelimitedCodec](https://docs.rs/tokio-util/latest/tokio_util/codec/length_delimited/struct.LengthDelimitedCodec.html) (which we basically partially implement ourselves), but that one only works with `Bytes`, which would currently require an extra allocation for each network message; I'll revisit this option later on.

### DIFF
--- a/environment/src/network/message.rs
+++ b/environment/src/network/message.rs
@@ -21,15 +21,11 @@ use crate::{
 use snarkvm::{dpc::posw::PoSWProof, prelude::*};
 
 use ::bytes::{Buf, BufMut, Bytes, BytesMut};
-use anyhow::{anyhow, Result};
+use anyhow::{bail, Result};
 use serde::{de::DeserializeOwned, Deserialize, Serialize};
-use std::{
-    io::{Cursor, Seek, Write},
-    marker::PhantomData,
-    net::SocketAddr,
-};
+use std::{io::Write, marker::PhantomData, net::SocketAddr};
 use tokio::task;
-use tokio_util::codec::{Decoder, Encoder};
+use tokio_util::codec::{Decoder, Encoder, LengthDelimitedCodec};
 
 /// This object enables deferred deserialization / ahead-of-time serialization for objects that
 /// take a while to deserialize / serialize, in order to allow these operations to be non-blocking.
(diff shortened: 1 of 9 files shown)


niklaslong (approved):
LGTM 👍

ljedrz:
Actually, having tinkered with `codec` a bit more, I don't think there would be any performance penalty; however, due to there being a bit of a backlog of PRs, I'd prefer to iterate on this change once they are merged.

howardwu:
Sorry, why do we need a dedicated codec struct? This seems redundant

ljedrz:
@howardwu this was an incremental step towards a later implementation that would take advantage of `tokio-util`'s [`LengthDelimitedCodec`](https://docs.rs/tokio-util/latest/tokio_util/codec/length_delimited/struct.LengthDelimitedCodec.html) (which can handle the logic we currently implement manually). I was initially worried about potentially introducing extra allocations, but in the end I was able to actually reduce their number (by no longer having to allocate `Data` vectors. I updated the PR to contain the full target implementation.

niklaslong (approved):

ljedrz:
Rebased and applied https://github.com/AleoHQ/snarkOS/pull/1705.

ljedrz:
I've been fuzz-testing (using `cargo-fuzz`) the newly introduced `MessageCodec` for a while now with no issues, happy to add the fuzz-test to the repo if desired.

environment/src/network/message.rs
```diff
@@ -255,152 +251,149 @@ impl<N: Network, E: Environment> Message<N, E> {
...
-    pub fn deserialize<R: Read + Seek>(reader: &mut R) -> Result<Self> {
-        // Read the message ID.
-        let id: u16 = bincode::deserialize_from(&mut *reader)?;
-
-        // Helper function to read all the remaining bytes from a reader.
-        let read_to_end = |reader: &mut R| -> Result<Bytes> {
-            let mut data = vec![];
-            reader.read_to_end(&mut data)?;
+    pub fn deserialize(mut bytes: BytesMut) -> Result<Self> {
+        // Make sure there is an id available.
```
howardwu:
```suggestion
        // Ensure there is at least a message ID in the buffer.
```

environment/src/network/message.rs
```diff
@@ -255,152 +251,149 @@ impl<N: Network, E: Environment> Message<N, E> {
...
-        };
+        // Read the message ID.
+        let id: u16 = bytes.get_u16_le();
 
         // Deserialize the data field.
         let message = match id {
-            0 => Self::BlockRequest(bincode::deserialize_from(&mut *reader)?, bincode::deserialize_from(&mut *reader)?),
-            1 => Self::BlockResponse(Data::Buffer(read_to_end(&mut *reader)?)),
+            0 => {
+                let mut reader = bytes.reader();
```
howardwu, line 266: Is there no way to use `bytes.reader()` directly?

environment/src/network/message.rs
```diff
@@ -255,152 +251,149 @@ impl<N: Network, E: Environment> Message<N, E> {
...
-        if source.len() < 4 + length {
-            // The full message has not yet arrived.
-            //
-            // We reserve more space in the buffer. This is not strictly
-            // necessary, but is a good idea performance-wise.
-            source.reserve(4 + length - source.len());
-
-            // We inform `Framed` that we need more bytes to form the next frame.
+        // Decode a frame containing bytes belonging to a message.
+        let bytes = if let Some(bytes) = self.codec.decode(source)? {
```
howardwu: nit: Can you turn this into a `match` statement?

environment/src/network/message.rs
ljedrz: `BytesMut::reader` takes `self` by value and it's used twice in this block, so it's necessary to introduce a binding here.

ljedrz:
Applied review suggestions and rebased.

howardwu (approved):

PR #1694, opened 2022-03-16
Restructure crawler args, introduce a new storage table

ljedrz:
This PR consists of the following changes related to the network crawler:
- the postgres-specific arguments were moved to a dedicated `struct`
- the `nodes` table was renamed to `node_states` and its last 3 columns were moved to a new `node_centrality` table
- `node_states` now enforces uniqueness on the `(ip, port, timestamp)` triplet and has a related index; the records are now no longer duplicated

I initially considered only updating the node centrality metrics (without adding the new table - see https://github.com/AleoHQ/snarkOS/pull/1694/commits/82dd6f9d44ee17643945d934ced0b1bd43f3b310), but that would erase the older values, which would remove the related history for nodes that can't be connected to.

Cc @zosorock

### DIFF
--- a/.crawler/src/storage.rs
+++ b/.crawler/src/storage.rs
@@ -16,8 +16,9 @@
 
 use std::collections::HashSet;
 #[cfg(feature = "postgres-tls")]
-use std::fs;
+use std::{fs, path::PathBuf};
 
+use clap::Parser;
 #[cfg(feature = "postgres-tls")]
 use native_tls::{Certificate, TlsConnector};
 #[cfg(feature = "postgres-tls")]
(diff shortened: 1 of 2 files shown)


.crawler/src/storage.rs
```diff
@@ -74,15 +101,23 @@ pub async fn initialize_storage(opts: &Opts) -> Result<Client, anyhow::Error> {
...
             port                INTEGER NOT NULL,
-            timestamp           TIMESTAMP WITH TIME ZONE,
-            type                SMALLINT,
-            version             INTEGER,
-            state               SMALLINT,
-            height              INTEGER,
-            handshake_ms        INTEGER,
+            timestamp           TIMESTAMP WITH TIME ZONE NOT NULL,
             connection_count    SMALLINT,
             prestige_score      REAL,
```
niklaslong, line 128: Just a note for later, we should either change the corresponding value name in the struct or comment on the fact that they are equivalent.

niklaslong (approved):

PR #1696, opened 2022-03-19
Network refactor: routeless

niklaslong:
The purpose of this PR is to explore the possibility of shifting from actors to shared memory within the snarkOS network implementation. This will, in time, allow for the removal of routing logic that isn't necessary with the new model as every structure now has access to the `NetworkState`. 

This is a work in progress. I'd consider the main structural changes to be in place, though there is still quite a bit of refactoring, optimising and cleanup to be done. I would not consider the APIs stable at this point as I plan on breaking up the newly introduced methods further.

### DIFF
--- a/network/Cargo.toml
+++ b/network/Cargo.toml
@@ -32,6 +32,9 @@ version = "0.2"
 version = "0.3.21"
 features = [ "thread-pool" ]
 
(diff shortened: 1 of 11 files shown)


niklaslong:
Closing for the time being as this was purely exploratory.

PR #1697, opened 2022-03-20
Simplify the handshake impl in the synth node

ljedrz:
Builds on https://github.com/AleoHQ/snarkOS/pull/1693; only the last 2 commits are new, of which only https://github.com/AleoHQ/snarkOS/commit/b310471f6306f4aef10143a39614f65ea454aef1 is of interest.

This PR makes the handshake impl of the synthetic node much more similar to the one in the true node.

In addition, `pea2pea` is updated to `0.35`.

### DIFF
--- a/.crawler/Cargo.toml
+++ b/.crawler/Cargo.toml
@@ -31,6 +31,9 @@ version = "0.1"
 [dependencies.bincode]
 version = "1"
 
+[dependencies.bytes]
+version = "1"
+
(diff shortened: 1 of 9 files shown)


niklaslong (approved):
🎉

ljedrz:
Due to the conflicts in this PR and the one it builds on, I'll close this one, rebase the original and then append these changes to it.

PR #1699, opened 2022-03-22
Restore missing CI integration tests

ljedrz:
It seems that after https://github.com/AleoHQ/snarkOS/pull/1688 they are compiled, but not run.

I think I found the missing config bit, going to mark the PR as a draft until I've confirmed it.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -58,7 +58,7 @@ commands:
             TEST_NAMES=$(circleci tests split test_names.txt)
             for i in $(echo $TEST_NAMES | sed "s/ / /g")
             do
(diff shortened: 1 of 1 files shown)


ljedrz:
Ok, so this restores the integration test runs to the previous state; ~however, it seems that some of them have not been running before either; I'll open a related issue~. Edit: actually, they are all there, but under separate tabs due to this run being parallelized :ok_hand:.

niklaslong (approved):

ljedrz:
Merging this one, as it is a trivial fix, the results are clear, and we don't want to miss any `.integration` errors with further PRs.

PR #1703, opened 2022-03-24
Add an automatic node self-update feature

ljedrz:
This PR aims to replace the existing auto-update scripts by performing periodic checks on a specified branch of snarkOS, downloading any updates, rebuilding the node and restarting it with the same arguments as the ones it was started with. It also uses `git` and expects the host to be able to build snarkOS.

It is feature-gated and currently not enabled by default (but it probably should be once we decide to use it and update the docs) and - not unlike the scripts - only works on Unix-like systems. Btw, we might want to recommend that users use WSL by default due to our `clang` build requirement regardless.

Open points:
- [ ] when should it become a default feature
- [ ] what to do with the existing scripts
- [ ] should the auto-update check interval be changed (I set it to 15 minutes)


niklaslong (dismissed):

ljedrz:
the `clippy` job failure is unrelated: https://github.com/briansmith/ring/issues/1469

ljedrz:
Rebased.

ljedrz:
Rebased.

aleo-svc-exp:

PR #1712, opened 2022-03-31
Add fuzz tests for the network message decoder

ljedrz:
This PR adds a new workspace, `fuzz`, initialized and set up as per the `cargo-fuzz` [docs](https://rust-fuzz.github.io/book/cargo-fuzz/tutorial.html). It doesn't interfere with the snarkOS workspace.

I've been running it locally for a good amount of time and detected no issues so far.


fuzz/Cargo.toml
```diff
@@ -0,0 +1,28 @@
+[package]
+name = "snarkos-fuzz"
+version = "0.0.0"
```
niklaslong, line 3: Do we want to match the snarkOS version here?

fuzz/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+authors = ["Automatically generated"]
+publish = false
+edition = "2018"
```
niklaslong: Why not `2021`?

niklaslong (approved):
Left a few comments but LGTM 👍

fuzz/Cargo.toml
ljedrz: this was auto-generated, I don't think it's intended to ever be published; if anything, I think we should rather add some README which would be updated with subsequent runs, stamped with the specific SHA of the repo (as the version is not updated as often)

fuzz/Cargo.toml
ljedrz: it was auto-generated; I guess I can bump it to 2021

fuzz/Cargo.toml
niklaslong: Perfect—so long as we keep track of what code corresponds to which run 👍

ljedrz:
Added `Cargo.lock` to `.gitignore`, as it's not needed.

ljedrz:
Added a second fuzz target that specifically targets deserialization (as opposed to the network message decoder).

ljedrz:
Rebased.

howardwu (approved):

PR #1719, opened 2022-04-04
Feat: round trip time metrics for request/response cycles

niklaslong:
This PR introduces some internal RTT metrics for certain request/response cycles. It also updates the grafana dashboard to display these. Unfortunately there is currently no perfectly clean way of introducing them with feature flags so the occasional `Option` is necessary.

The `Pong` message isn't technically a cycle but as it does some operations based on the block locators, I thought it would be interesting to have a sense of how expensive that is.

### DIFF
--- a/metrics/src/names.rs
+++ b/metrics/src/names.rs
@@ -15,7 +15,7 @@
 // along with the snarkOS library. If not, see <https://www.gnu.org/licenses/>.
 
 pub const GAUGE_NAMES: [&str; 4] = [blocks::HEIGHT, peers::CONNECTED, peers::CANDIDATE, peers::RESTRICTED];
-pub const HISTOGRAM_NAMES: [&str; 2] = [internal_rtt::PING, internal_rtt::PONG];
+pub const HISTOGRAM_NAMES: [&str; 3] = [internal_rtt::PING, internal_rtt::PONG, internal_rtt::BLOCK_REQUEST];
 
 pub mod blocks {
     pub const HEIGHT: &str = "snarkos_blocks_height_total";
@@ -30,4 +30,5 @@ pub mod peers {
 pub mod internal_rtt {
     pub const PING: &str = "snarkos_internal_rtt_ping";
     pub const PONG: &str = "snarkos_internal_rtt_pong";
+    pub const BLOCK_REQUEST: &str = "snarkos_internal_rtt_block_request";
 }
--- a/metrics/src/names.rs
+++ b/metrics/src/names.rs
@@ -14,8 +14,8 @@
 // You should have received a copy of the GNU General Public License
 // along with the snarkOS library. If not, see <https://www.gnu.org/licenses/>.
 
-pub const GAUGE_NAMES: [&str; 4] = [blocks::HEIGHT, peers::CONNECTED, peers::CANDIDATE, peers::RESTRICTED];
-pub const HISTOGRAM_NAMES: [&str; 4] = [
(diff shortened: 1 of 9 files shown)


metrics/src/names.rs
```diff
@@ -15,6 +15,12 @@
 // along with the snarkOS library. If not, see <https://www.gnu.org/licenses/>.
 
 pub const GAUGE_NAMES: [&str; 4] = [blocks::HEIGHT, peers::CONNECTED, peers::CANDIDATE, peers::RESTRICTED];
+pub const HISTOGRAM_NAMES: [&str; 4] = [
```
ljedrz: nit: I think we could just name those `GAUGES`, `HISTOGRAMS` etc.

niklaslong:
The Grafana dashboard can be seen here: https://github.com/AleoHQ/snarkOS/pull/1720#issuecomment-1088637071

ljedrz (approved):
LGTM :+1:

niklaslong:
Rebased.

network/src/peers.rs
```diff
@@ -84,7 +84,7 @@ pub enum PeersRequest<N: Network, E: Environment> {
     /// PeerRestricted := (peer_ip)
     PeerRestricted(SocketAddr),
     /// SendPeerResponse := (peer_ip)
```
howardwu: nit: Documentation needs updating

network/src/peer.rs
```diff
@@ -589,16 +614,28 @@ impl<N: Network, E: Environment> Peer<N, E> {
...
                                         warn!("[Pong] {}", error);
                                     }
                                 },
                                 Message::Pong(is_fork, block_locators) => {
+                                    // Unfortunately can't be feature-flagged because of the enum
+                                    // it's passed around in.
+                                    let _rtt_start_instant: Option<Instant> = None;
+
+                                    #[cfg(any(feature = "test", feature = "prometheus"))]
+                                    let _rtt_start_instant = Some(rtt_start);
```
howardwu, line 633: nit: Can you wrap this in a scope, so it is easily comment-out-able
```rs
let _foo = {
    #[cfg(not)]
    bar
    #[cfg]
    baz
};
```

network/src/peer.rs
```diff
@@ -526,12 +544,19 @@ impl<N: Network, E: Environment> Peer<N, E> {
                                     break;
                                 },
                                 Message::PeerRequest => {
+                                    // Unfortunately can't be feature-flagged because of the enum
+                                    // it's passed around in.
+                                    let _rtt_start_instant: Option<Instant> = None;
+
+                                    #[cfg(any(feature = "test", feature = "prometheus"))]
+                                    let _rtt_start_instant = Some(rtt_start);
+
```
howardwu, line 553: nit: ditto on scoping this so it is containerized

howardwu:
More broadly as a question, is there really no better way to integrate metrics into the structs other than to just inject commands into the codebase? I would be surprised if this was the best way to design a solution for metrics

network/src/peer.rs
niklaslong: I can add this in, though the scoping unfortunately won't make it comment-out-able because `_rtt_start_instant` is passed around in the message enum (hence the variable shadowing when the feature flag is set).

niklaslong: The easiest way to comment out the functionality is just by not running with `--features prometheus`.

niklaslong:
Rebased.

howardwu (approved):

PR #1725, opened 2022-04-06
Feature: get all the ciphertexts from the LedgerState

IAvecilla:
## Motivation

Get all the ciphertexts directly from `LedgerState` to avoid iterating over all the transactions and commitments to get them.

## Test Plan

There is a new test created in `storage/src/state/tests.rs`
- `test_get_all_ciphertexts` - This test asserts that this function gets all the ciphertexts in the ledger without a specific order.

### DIFF
--- a/storage/src/state/ledger.rs
+++ b/storage/src/state/ledger.rs
@@ -467,7 +467,7 @@ impl<N: Network> LedgerState<N> {
     }
 
     // Returns all the ciphertexts in ledger.
-    pub fn get_ciphertexts(&self) -> Result<impl Iterator<Item = N::RecordCiphertext> + '_> {
+    pub fn get_ciphertexts(&self) -> impl Iterator<Item = Result<N::RecordCiphertext>> + '_ {
         self.blocks.get_ciphertexts()
     }
 
@@ -1298,7 +1298,7 @@ impl<N: Network> BlockState<N> {
     }
 
     // Returns all the record ciphertexts.
-    fn get_ciphertexts(&self) -> Result<impl Iterator<Item = N::RecordCiphertext> + '_> {
+    fn get_ciphertexts(&self) -> impl Iterator<Item = Result<N::RecordCiphertext>> + '_ {
         self.transactions.get_ciphertexts()
     }
 
@@ -1578,11 +1578,8 @@ impl<N: Network> TransactionState<N> {
     }
 
     // Returns all the record ciphertexts.
-    fn get_ciphertexts(&self) -> Result<impl Iterator<Item = N::RecordCiphertext> + '_> {
-        Ok(self
-            .commitments
-            .keys()
-            .map(move |commitment| self.get_ciphertext(&commitment).unwrap()))
+    fn get_ciphertexts(&self) -> impl Iterator<Item = Result<N::RecordCiphertext>> + '_ {
+        self.commitments.keys().map(move |commitment| self.get_ciphertext(&commitment))
     }
 
     /// Returns the transition for a given transition ID.
--- a/storage/src/state/ledger.rs
+++ b/storage/src/state/ledger.rs
@@ -466,7 +466,7 @@ impl<N: Network> LedgerState<N> {
         self.blocks.get_previous_ledger_root(block_height)
     }
 
-    // Returns the record ciphertexts for a given block.
+    // Returns all the ciphertexts in ledger.
     pub fn get_ciphertexts(&self) -> Result<Vec<N::RecordCiphertext>> {
         self.blocks.get_ciphertexts()
     }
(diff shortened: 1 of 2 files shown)


storage/src/state/ledger.rs
```diff
@@ -1567,6 +1577,14 @@ impl<N: Network> TransactionState<N> {
         Err(anyhow!("Commitment {} is missing in storage", commitment))
     }
 
+    // Returns all the record ciphertexts.
+    fn get_ciphertexts(&self) -> Result<impl Iterator<Item = N::RecordCiphertext> + '_> {
+        Ok(self
+            .commitments
+            .keys()
+            .map(move |commitment| self.get_ciphertext(&commitment).unwrap()))
```
ljedrz: why does this function return a `Result` if we're calling `unwrap` on every item? we should probably return an `impl Iterator<Item = Result<N::RecordCiphertext>> + '_` instead

storage/src/state/tests.rs
```diff
@@ -32,6 +32,14 @@ fn create_new_ledger<N: Network, S: Storage>() -> LedgerState<N> {
     LedgerState::open_writer_with_increment::<S, _>(temp_dir(), 1).expect("Failed to initialize ledger")
 }
 
+fn to_hash_set_test<N: Network>(ciphertexts: Vec<N::RecordCiphertext>) -> HashSet<N::RecordCiphertext> {
+    let mut set = HashSet::new();
+    for ciphertext in ciphertexts {
+        set.insert(ciphertext);
+    }
+    set
+}
```
ljedrz: I don't think this helper is needed, this is just `ciphertexts.into_iter().collect::<HashSet<_>>()`

storage/src/state/tests.rs
```diff
@@ -346,3 +354,34 @@ fn test_transaction_fees() {
...
+    ledger.add_next_block(&block).expect("Failed to add next block to ledger");
+
+    let expected_ciphertexts_set = to_hash_set_test::<Testnet2>(
+        ledger
+            .get_blocks(0, ledger.latest_block_height())
+            .unwrap()
+            .iter()
+            .flat_map(|block| block.commitments())
+            .map(|commitment| ledger.get_ciphertext(commitment).unwrap())
+            .collect::<Vec<_>>(),
```
ljedrz:
```suggestion
            .collect::<HashSet<_>>(),
```
(which renders the call to `to_hash_set_test` redundant)

storage/src/state/tests.rs
```diff
@@ -346,3 +354,34 @@ fn test_transaction_fees() {
...
+    );
+    let ciphertexts_set = to_hash_set_test::<Testnet2>(ledger.get_ciphertexts().unwrap().collect());
```
ljedrz:
```suggestion
    let ciphertexts_set = ledger.get_ciphertexts().unwrap().collect();
```

storage/src/state/tests.rs
```diff
@@ -346,3 +354,34 @@ fn test_transaction_fees() {
...
+    // Initialize a new ledger.
+    let ledger = create_new_ledger::<Testnet2, RocksDB>();
+
+    // Initialize a new account.
+    let account = Account::<Testnet2>::new(&mut thread_rng());
+    let address = account.address();
+
+    // Mine the next block.
+    let (block, _) = ledger
+        .mine_next_block(address, true, &[], &terminator, rng)
```
ljedrz: instead of mining, we should load a single pre-existing block for a much faster testing experience, like in the `test_get_blocks` test in `snarkos-rpc`.

ljedrz: note to self: we should make sure we're not mining in _any_ test unless absolutely necessary.

ljedrz:
Left some comments; also, these changes should be squashed into 1-2 commits.

storage/src/state/ledger.rs
IAvecilla: I agree with this. The change is already done on the last commit.

storage/src/state/tests.rs
IAvecilla: You're right. I deleted this helper function and I changed the collect type on the last commit.

storage/src/state/tests.rs
IAvecilla: Great! I was mining a block to ensure that the function was getting all the actual ciphertexts but I think the test works well with only the genesis block. I deleted that part of the test on the last commit.

ljedrz: BTW, I already proposed the aforementioned change in https://github.com/AleoHQ/snarkOS/pull/1708/commits/c45463aa98f8c6ed64962647c3183f36f172893c.

howardwu (approved):

PR #1726, opened 2022-04-08
Disable some resource logs

ljedrz:
They are a bit too verbose, and we don't need to actively inspect them. The logs in question are only commented out, in order to restore them more easily if need be.

Cc @zosorock

### DIFF
--- a/environment/src/helpers/resources.rs
+++ b/environment/src/helpers/resources.rs
@@ -103,11 +103,6 @@ impl Resources {
             while let Some(request) = receiver.recv().await {
                 match request {
                     ResourceRequest::Register(resource, id) => {
-                        let _resource_name = match resource {
-                            Resource::Task(..) => "task",
-                            Resource::Thread(..) => "thread",
-                        };
-
                         if resources.insert(id, resource).is_some() {
                             error!("A resource with id {} already exists!", id);
                         }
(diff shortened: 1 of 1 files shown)


environment/src/helpers/resources.rs
```diff
@@ -103,15 +103,15 @@ impl Resources {
...
+                        let _resource_name = match resource {
                             Resource::Task(..) => "task",
                             Resource::Thread(..) => "thread",
                         };
 
                         if resources.insert(id, resource).is_some() {
                             error!("A resource with id {} already exists!", id);
                         } else {
-                            trace!("Registered a {} as resource {}", resource_name, id);
+                            // trace!("Registered a {} as resource {}", _resource_name, id);
```
howardwu: If these aren't necessary can we remove the commented out traces?

ljedrz: Sure thing.

environment/src/helpers/resources.rs
```diff
@@ -103,15 +103,13 @@ impl Resources {
             while let Some(request) = receiver.recv().await {
                 match request {
                     ResourceRequest::Register(resource, id) => {
-                        let resource_name = match resource {
+                        let _resource_name = match resource {
                             Resource::Task(..) => "task",
                             Resource::Thread(..) => "thread",
                         };
```
howardwu:
```suggestion
```

howardwu (approved):

PR #1727, opened 2022-04-08
Introduce an object with combined state

ljedrz:
This PR is purely a refactoring and despite the rather large diff, it introduces no changes to the current functionalities, other than the order of some operations when the node is started.

There are 2 main changes:
1. [rename `Status` to `RawStatus` and `State` to `Status`](https://github.com/AleoHQ/snarkOS/commit/0851bd0c10456a35234ad92e6d4d7b8ceda194db): both of these objects refer to the same thing, so their names should match as well; also, this allows the new combined state object to be named `State`
2. [introduce a `State` object](https://github.com/AleoHQ/snarkOS/commit/0ee6be33ec34c154b56fb5c112610950b5a9d70e): the current node structure involves a few actor-like objects that communicate with one another; this is fine, but the state of the actors overlaps, can't really be fully isolated, and adds plenty of extra complexity; this change proposes a combined `State` object shared between the main actors, allowing the communication between them to be much simpler, while not changing the way their channels are utilized or the scopes of their functionalities; this setup reduces code complexity and will enable further improvements to the overall structure

### DIFF
--- a/network/src/ledger.rs
+++ b/network/src/ledger.rs
@@ -22,7 +22,7 @@ use crate::{
     ProverRouter,
 };
 use snarkos_environment::{
-    helpers::{block_locators::*, NodeType, State},
+    helpers::{block_locators::*, NodeType, Status},
     network::{Data, DisconnectReason, Message},
     Environment,
 };
@@ -71,12 +71,12 @@ pub enum LedgerRequest<N: Network> {
     /// Heartbeat := (prover_router)
     Heartbeat(ProverRouter<N>),
     /// Pong := (peer_ip, node_type, status, is_fork, block_locators)
-    Pong(SocketAddr, NodeType, State, Option<bool>, BlockLocators<N>, Option<Instant>),
+    Pong(SocketAddr, NodeType, Status, Option<bool>, BlockLocators<N>, Option<Instant>),
     /// UnconfirmedBlock := (peer_ip, block, prover_router)
     UnconfirmedBlock(SocketAddr, Block<N>, ProverRouter<N>),
 }
 
-pub type PeersState<N> = HashMap<SocketAddr, Option<(NodeType, State, Option<bool>, u32, BlockLocators<N>)>>;
+pub type PeersState<N> = HashMap<SocketAddr, Option<(NodeType, Status, Option<bool>, u32, BlockLocators<N>)>>;
 
 ///
 /// A ledger for a specific network on the node server.
@@ -350,7 +350,7 @@ impl<N: Network, E: Environment> Ledger<N, E> {
                     };
 
                     // If the peer is not a sync node and is syncing, and the peer is ahead, proceed to disconnect.
-                    if *node_type != NodeType::Sync && *status == State::Syncing && cumulative_weight > latest_cumulative_weight {
+                    if *node_type != NodeType::Sync && *status == Status::Syncing && cumulative_weight > latest_cumulative_weight {
                         // Append the peer to the list of disconnects.
                         peer_ips_to_disconnect.push(*peer_ip);
                     }
@@ -419,7 +419,7 @@ impl<N: Network, E: Environment> Ledger<N, E> {
         let mut status = E::status().get();
 
         // If the node is shutting down, skip the update.
-        if status == State::ShuttingDown {
+        if status == Status::ShuttingDown {
             trace!("Ledger is shutting down");
             // Set the terminator bit to `true` to ensure it stops mining.
             E::terminator().store(true, Ordering::SeqCst);
@@ -427,14 +427,14 @@ impl<N: Network, E: Environment> Ledger<N, E> {
         }
         // If there is an insufficient number of connected peers, set the status to `Peering`.
         else if self.peers_state.read().await.len() < E::MINIMUM_NUMBER_OF_PEERS {
-            status = State::Peering;
+            status = Status::Peering;
         }
         // If the ledger is out of date, set the status to `Syncing`.
         else {
             // Update the status to `Ready` or `Mining`.
             status = match status {
-                State::Mining => State::Mining,
-                _ => State::Ready,
+                Status::Mining => Status::Mining,
+                _ => Status::Ready,
             };
 
             // Retrieve the latest cumulative weight of this node.
@@ -450,7 +450,7 @@ impl<N: Network, E: Environment> Ledger<N, E> {
                     // If the cumulative weight is greater than MAXIMUM_LINEAR_BLOCK_LOCATORS, set the status to `Syncing`.
                     if cumulative_weight.saturating_sub(latest_cumulative_weight) > MAXIMUM_LINEAR_BLOCK_LOCATORS as u128 {
                         // Set the status to `Syncing`.
-                        status = State::Syncing;
+                        status = Status::Syncing;
                         break;
                     }
                 }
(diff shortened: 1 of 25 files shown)


niklaslong (commented):
I'll rebase this onto the branch with the RTT metrics and run it for a bit—should give us a perf sanity check.

niklaslong (approved):
The RTT metrics would suggest the performance is unchanged—I'll leave the branch running for a while longer.

I think this is a desirable first step towards consolidating state. The need for sync primitives should dissapear once the channels are consolidated. 👍

network/src/peer.rs
```diff
@@ -264,11 +251,7 @@ impl<N: Network, E: Environment> Peer<N, E> {
                         trace!("Sending '{}-B' to {}", message.name(), peer_ip);
                         outbound_socket.send(message).await?;
 
-                        // Initialize a status variable.
-                        let status = Status::new();
-                        status.update(peer_status);
```
howardwu, line 272: Why was this removed during the process of a handshake?

network/src/peer.rs
```diff
@@ -582,35 +560,35 @@ impl<N: Network, E: Environment> Peer<N, E> {
                                     // Update the node type of the peer.
                                     peer.node_type = node_type;
                                     // Update the status of the peer.
-                                    peer.status.update(status);
```
howardwu, line 610: If we can directly assign `status`, do we even need the `update()` method? Can/should it be removed?

network/src/prover.rs
```diff
@@ -61,80 +60,55 @@ pub enum ProverRequest<N: Network> {
...
 /// A prover for a specific network on the node server.
 ///
-#[derive(Debug)]
 pub struct Prover<N: Network, E: Environment> {
     /// The state storage of the prover.
-    state: Arc<ProverState<N>>,
+    prover_state: Arc<ProverState<N>>,
     /// The Aleo address of the prover.
-    address: Option<Address<N>>,
+    pub address: Option<Address<N>>,
```
howardwu: revert: Why is this `pub` visibility? Can we move it into `State` if it needs to be accessed?

Also, why is `State` `pub` as well? Can we have `E: Environment` provide it?

network/src/prover.rs
```diff
@@ -318,13 +328,15 @@ impl<N: Network, E: Environment> Prover<N, E> {
     ///
     /// Initialize the miner, if the node type is a miner.
     ///
-    async fn start_miner(prover: Arc<Self>, local_ip: SocketAddr) {
+    pub async fn start_miner(&self) {
```
howardwu: revert: Why is this `pub`? These methods are meant to be internal, otherwise there is "crossing of wires" which can lead to unintentional node halts

network/src/prover.rs
```diff
@@ -203,13 +207,13 @@ impl<N: Network, E: Environment> Prover<N, E> {
     ///
     /// Sends a `PoolRegister` message to the pool IP address.
     ///
-    async fn send_pool_register(&self) {
+    pub async fn send_pool_register(&self) {
```
howardwu: revert: Why is this `pub`? These methods are meant to be internal, otherwise there is "crossing of wires" which can lead to unintentional node halts

network/src/ledger.rs
```diff
@@ -195,13 +169,13 @@ impl<N: Network, E: Environment> Ledger<N, E> {
     /// Performs the given `request` to the ledger.
     /// All requests must go through this `update`, so that a unified view is preserved.
     ///
-    pub(super) async fn update(&self, request: LedgerRequest<N>) {
+    pub async fn update(&self, request: LedgerRequest<N>) {
```
howardwu: revert: Why is `super` removed? The goal is not ensure this is not callable outside of this scope, out of order executions lead to node halts

network/src/operator.rs
```diff
@@ -204,42 +161,40 @@ impl<N: Network, E: Environment> Operator<N, E> {
...
-        self.state.get_provers()
+        self.operator_state.get_provers()
     }
 
     ///
     /// Performs the given `request` to the operator.
     /// All requests must go through this `update`, so that a unified view is preserved.
     ///
-    pub(super) async fn update(&self, request: OperatorRequest<N>) {
+    pub async fn update(&self, request: OperatorRequest<N>) {
```
howardwu: Revert: why is `super` removed? The goal is not ensure this is not callable outside of this scope, out of order executions lead to node halts

network/src/state.rs
```diff
@@ -0,0 +1,173 @@
...
+        Self {
+            local_ip,
+            peers: Default::default(),
+            ledger: Default::default(),
+            operator: Default::default(),
+            prover: Default::default(),
+        }
+    }
+
+    pub async fn initialize_peers(self: &Arc<Self>, peers: Peers<N, E>, mut peers_handler: PeersHandler<N, E>) {
```
howardwu, line 58: remove: Why are these `initialize` methods `pub`?

howardwu:
Overall, this structure is the way to go. Need to ensure the visibility does not leak, I've had some bad experiences with node halts when these methods are misused during future modifications

network/src/peer.rs
ljedrz: I noticed this by accident - it's just redundant, as it creates an intermediate `status` binding that takes the value of `peer_status`, while we can just use `peer_status`

network/src/peer.rs
ljedrz: with `Peer`, we have the option of mutating this field during synchronous access (so we can just assign to it directly), but for the immutable `Node` shared across multiple threads we still need to use `Status::update` that uses atomics

ljedrz:
Rebased and applied some of the review suggestions related to APIs (some of the `pub`s seem to have been left over from an intermediate step of the refactoring).

network/src/prover.rs
ljedrz: `state` doesn't need to be `pub` and it was removed; `address` is `pub` due to the `Operator` depending on it and there not being a method for it (as is the case with e.g. `MemoryPool`); I added a commit moving it to `State` like you suggested

network/src/state.rs
ljedrz: it's because they are called by the `Server`, and can also be called in tests that don't require the functionalities of the `Server` (i.e. just the state); it's very unlikely that those would be called by accident or more than once

howardwu:
Have we been able to burn-in test this PR?

jhaineault1:
TEST TEST TEST 
     Snarkos on $devnet-ue2c-arm-a1e568c376e78b13956d66e24f4b4c84 has been recompiled and docker containers have been rebuilt. 
     IP: $18.223.50.176 
     LOGS: $https://app.datadoghq.com/logs/livetail?query=pr%3A1727%20Report&cols=host%2Ccontainer_name

jhaineault1:
Snarkos on devnet-ue2c-arm-a1e568c376e78b13956d66e24f4b4c84 has been recompiled and docker containers have been rebuilt. 
     IP: 18.223.50.176 
     LOGS: https://app.datadoghq.com/logs/livetail?query=pr%3A1727%20Report&cols=host%2Ccontainer_name

jhaineault1:
Snarkos on devnet-ue2c-arm-a1e568c376e78b13956d66e24f4b4c84 has been recompiled and docker containers have been rebuilt. 
     LOGS: https://app.datadoghq.com/logs/livetail?query=pr%3A1727%20Report&cols=host%2Ccontainer_name

jhaineault1:
Snarkos on devnet-ue2c-arm-a1e568c376e78b13956d66e24f4b4c84 has been recompiled and docker containers have been rebuilt. 
     LOGS: https://app.datadoghq.com/logs/livetail?query=pr%3A1727%20Report&cols=host%2Ccontainer_name

zosorock:
> Have we been able to burn-in test this PR?

Yes, it has been running on an airtight devnet since last Thursday.

howardwu (approved):

PR #1731, opened 2022-04-12
Safer storage traits

ljedrz:
The ledger currently accesses storage via a single abstraction (`LedgerState`) with a `read_only` flag that determines what operations can/should be run using it. However, this applies only to the top layer and one could still perform writes in the read-only instance using the more low-level objects or in the top layer if the flag is not checked or ignored; in fact, storage is currently written to in `LedgerState::open_reader`.

This PR introduces new traits and marker types that ensure that neither the lowest-level objects related to storage, nor any abstraction built on top of them, can perform any writes if it's created in read-only mode. This is checked at compile time.

The most notable changes are:
- traits `StorageAccess` and `StorageReadWrite` are added
- marker types `ReadOnly` and `ReadWrite` are added
- the `Storage` trait now has an associated `Access` type
- the `Map` trait is split into `MapRead` and `MapWrite`
- `LedgerState::open_reader` no longer performs any storage writes (only uses `MapRead::refresh`)
- the storage-related objects have an additional `T: StorageAccess` bound (some with the default set to `ReadWrite`)
- `LedgerState` no longer has (or needs) the `read_only` member

### DIFF
--- a/storage/src/state/ledger.rs
+++ b/storage/src/state/ledger.rs
@@ -16,7 +16,7 @@
 
 #[cfg(any(test, feature = "test"))]
 use crate::storage::rocksdb::RocksDB;
-use crate::storage::{DataMap, Map, MapId, Storage};
+use crate::storage::{DataMap, MapId, MapRead, MapReadWrite, Storage, StorageAccess, StorageReadWrite};
 use snarkos_environment::helpers::{block_locators::*, Resource};
 use snarkvm::dpc::prelude::*;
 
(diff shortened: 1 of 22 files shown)


niklaslong (commented):
I've rebased this onto the recent rtt metrics PR branch, will run for a while.

storage/src/state/ledger.rs
```diff
@@ -265,22 +115,6 @@ impl<N: Network> LedgerState<N> {
...
-
-            ledger
-                .ledger_roots
-                .insert(&genesis.previous_ledger_root(), &genesis.height(), Some(batch))?;
-            ledger.blocks.add_block(genesis, Some(batch))?;
-
-            // Execute the pending storage batch.
-            ledger.ledger_roots.execute_batch(batch)?;
-        }
-
```
ljedrz, line 283: I think this is the only functional change in this PR - since there is `Storage::refresh` that we keep using to catch up to the primary instance of storage (with the reader being the secondary instance), we don't need to insert the genesis block if it's not found; this was removed due to the operations involved being direct writes, and no longer being available to the reader (in addition to not being needed)

howardwu (approved):

PR #1753, opened 2022-05-03
Feature: iterator for get_blocks function

IAvecilla:
## Motivation

Optimize the `get_blocks` function to be faster and avoid using `collect` because of future memory issues with a great number of blocks. I did some benchmarking with Criterion, trying different scenarios and implementations of this function and these were the results.

|                           | 1 Block         | 143 Blocks        | 203 Blocks        | 400 Blocks       |
|:---------------|:------------ | --------------- |:--------------- |:--------------- |
| std-iterator       |  620 ms       | 19.5 s                 | 28 s                  | 54 s                    |
| std-iterator + par_bridge (std + rayon) | 760 ms (+22%) | 21 s (+5%)   | 30.6 s (+10%) | 58.5 s(+9%) |
| Parallel-iterator (rayon)        | 244 ms (-60%) | 5.6 s (-70%) | 8 s (-71%)   | 15 s (-72%)   |

These results are approximations of the real values.
The first implementation returns an Iterator from the std library.
The second returns an iterator from the std library and uses the `par_bridge` function to transform it into a parallel one whenever it's possible (this `par_bridge` function doesn't guarantee to keep the same order of the original iterator as it says in the [documentation](https://docs.rs/rayon/1.0.3/rayon/iter/trait.ParallelBridge.html)).
The third implementation returns a `ParallelIterator` from rayon and uses that directly.
I had the best results using the third implementation, so I went with that.

I encountered a little problem in the `src/network/peer.rs` file, there was a `for` loop and some `async/await` code and I used a `collect` right now to make it work, that is a work in progress and I want to find a way to avoid that and use the `ParallelIterator` without the `collect` and the `for` loop.

Here are the reports and graphics made by Criterion for you to check the results in more detail 
[report.zip](https://github.com/IAvecilla/snarkOS/files/8612334/report.zip)
The report has three directories:
- Get Blocks ParallelIterator Benches
- Get Blocks std-iter + par_bridge Benches
- Get Blocks std-iter Benches

Each one has a directory for each case of benchmarking (1 block, 143 blocks, 203 blocks and 403 blocks).
The report folder inside each case has all the graphics of interest.

## Test Plan

There is a new test `storage/src/tests.rs` named `test_get_blocks_iterator` that checks that this function works properly with a couple of blocks.

### DIFF
--- a/src/network/peer.rs
+++ b/src/network/peer.rs
@@ -480,16 +480,8 @@ impl<N: Network, E: Environment> Peer<N, E> {
                                         continue;
                                     }
                                     // Retrieve the requested blocks.
-                                    let blocks: Vec<Block<N>> = match ledger_reader.get_blocks(start_block_height, end_block_height) {
-                                        Ok(blocks) => match blocks.collect() {
-                                            Ok(blocks) => blocks,
-                                            Err(error) => {
-                                                if let Err(error) = ledger_router.send(LedgerRequest::Failure(peer_ip, format!("{}", error))).await {
-                                                    warn!("[Failure] {}", error);
-                                                }
-                                                continue;
-                                            }
-                                        },
+                                    let blocks: Vec<Block<N>> = match ledger_reader.get_blocks(start_block_height, end_block_height).and_then(|blocks| blocks.collect()) {
+                                        Ok(blocks) => blocks,
                                         Err(error) => {
                                             // Route a `Failure` to the ledger.
                                             if let Err(error) = ledger_router.send(LedgerRequest::Failure(peer_ip, format!("{}", error))).await {
--- a/src/network/peer.rs
+++ b/src/network/peer.rs
@@ -481,7 +481,15 @@ impl<N: Network, E: Environment> Peer<N, E> {
                                     }
                                     // Retrieve the requested blocks.
                                     let blocks: Vec<Block<N>> = match ledger_reader.get_blocks(start_block_height, end_block_height) {
-                                        Ok(blocks) => blocks.filter_map(|block_result| block_result.ok()).collect(),
+                                        Ok(blocks) => match blocks.collect() {
+                                            Ok(blocks) => blocks,
+                                            Err(error) => {
+                                                if let Err(error) = ledger_router.send(LedgerRequest::Failure(peer_ip, format!("{}", error))).await {
+                                                    warn!("[Failure] {}", error);
+                                                }
+                                                continue;
+                                            }
+                                        },
                                         Err(error) => {
                                             // Route a `Failure` to the ledger.
                                             if let Err(error) = ledger_router.send(LedgerRequest::Failure(peer_ip, format!("{}", error))).await {
--- a/src/network/peer.rs
+++ b/src/network/peer.rs
@@ -480,7 +480,7 @@ impl<N: Network, E: Environment> Peer<N, E> {
                                     }
                                     // Retrieve the requested blocks.
                                     let blocks = match ledger_reader.get_blocks(start_block_height, end_block_height) {
-                                        Ok(blocks) => blocks,
+                                        Ok(blocks) => blocks.filter_map(|block_result| block_result.ok()),
                                         Err(error) => {
                                             // Route a `Failure` to the ledger.
                                             if let Err(error) = ledger_router.send(LedgerRequest::Failure(peer_ip, format!("{}", error))).await {
--- a/src/network/peer.rs
+++ b/src/network/peer.rs
@@ -37,6 +37,7 @@ use snarkvm::dpc::prelude::*;
 
 use anyhow::{anyhow, bail, Result};
 use futures::SinkExt;
+use rayon::iter::ParallelIterator;
 use std::{
     collections::HashMap,
     net::SocketAddr,
@@ -479,8 +480,8 @@ impl<N: Network, E: Environment> Peer<N, E> {
                                         continue;
                                     }
                                     // Retrieve the requested blocks.
-                                    let blocks = match ledger_reader.get_blocks(start_block_height, end_block_height) {
-                                        Ok(blocks) => blocks.filter_map(|block_result| block_result.ok()),
+                                    let blocks: Vec<Block<N>> = match ledger_reader.get_blocks(start_block_height, end_block_height) {
+                                        Ok(blocks) => blocks.filter_map(|block_result| block_result.ok()).collect(),
                                         Err(error) => {
                                             // Route a `Failure` to the ledger.
                                             if let Err(error) = ledger_router.send(LedgerRequest::Failure(peer_ip, format!("{}", error))).await {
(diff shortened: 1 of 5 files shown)


ljedrz (changes requested):
Looks reasonable to me; however, we should preserve the current behavior of returning an error for the whole operation instead of just filtering errors out via `filter_map`.

IAvecilla:
> Looks reasonable to me; however, we should preserve the current behavior of returning an error for the whole operation instead of just filtering errors out via `filter_map`.

Great! I changed the `filter_maps` for handling errors correctly. I think that part in `peer.rs` could be better. Let me know if you have any ideas for that, I'll work to think of a better solution.

ljedrz:
I just noticed this PR was filed against the `testnet2` branch, while `testnet3` is the current one; I changed it in the PR, but you'll need to rebase it too (removing the merge commits etc.).

IAvecilla:
> I just noticed this PR was filed against the `testnet2` branch, while `testnet3` is the current one; I changed it in the PR, but you'll need to rebase it too (removing the merge commits etc.).

Hi @ljedrz the other day Howard told us that the `get_blocks` function should be implemented in both testnet2 and 3.
We think that the best way to do this is not to rebase this PR to testnet3 and maintain it on testnet2. Later we’ll open a new PR to testnet3. Do you think that is a good idea?

ljedrz:
I see that https://github.com/AleoHQ/snarkOS/pull/1725 was merged into `testnet2` indeed; I usually prefer to merge things into the current branch (`testnet3` here) first and to backport them into legacy branches afterwards, but since the first changes are already in `testnet2`, I'll switch this PR to point back at it and I'd just like to ask you to check out my suggestions and later squash the changes into a single commit.

src/network/peer.rs
```diff
@@ -479,8 +480,16 @@ impl<N: Network, E: Environment> Peer<N, E> {
                                         continue;
                                     }
                                     // Retrieve the requested blocks.
-                                    let blocks = match ledger_reader.get_blocks(start_block_height, end_block_height) {
-                                        Ok(blocks) => blocks,
+                                    let blocks: Vec<Block<N>> = match ledger_reader.get_blocks(start_block_height, end_block_height) {
```
ljedrz:
```suggestion
                                    let blocks: Vec<Block<N>> = match ledger_reader.get_blocks(start_block_height, end_block_height).and_then(|blocks| blocks.collect()) {
```

ljedrz: this will simplify the structure of the match statement below

src/rpc/rpc_impl.rs
```diff
@@ -75,7 +76,13 @@ impl<N: Network, E: Environment> RpcFunctions<N> for RpcContext<N, E> {
...
     async fn get_blocks(&self, start_block_height: u32, end_block_height: u32) -> Result<Vec<Block<N>>, RpcError> {
         let safe_start_height = max(start_block_height, end_block_height.saturating_sub(E::MAXIMUM_BLOCK_REQUEST - 1));
-        Ok(self.ledger.get_blocks(safe_start_height, end_block_height)?)
+        let blocks: Result<Vec<Block<N>>, _> = self
+            .ledger
+            .get_blocks(safe_start_height, end_block_height)?
+            .map(|block_result| block_result)
+            .collect();
+
+        Ok(blocks?)
```
ljedrz:
```suggestion
        blocks
```
I think this simpler way should be possible here instead?

ljedrz: or even better, to not even have a `blocks` binding at all (if the omission of `?` is possible)

storage/src/state/ledger.rs
```diff
@@ -877,12 +881,12 @@ impl<N: Network> LedgerState<N> {
...
         let start_block_height = latest_block_height.saturating_sub(number_of_blocks);
-        let blocks: BTreeMap<u32, Block<N>> = self
+        let blocks: Result<BTreeMap<u32, Block<N>>> = self
             .get_blocks(start_block_height, latest_block_height)?
-            .iter()
-            .map(|block| (block.height(), block.clone()))
+            .map(|block| block.map(|block| (block.height(), block)))
             .collect();
 
+        let blocks = blocks?;
```
ljedrz:
```suggestion
```
you can just use `?` in line `887`

storage/src/state/tests.rs
```diff
@@ -347,6 +348,32 @@ fn test_transaction_fees() {
...
+
+    // Load a test block into the ledger.
+    ledger_state.add_next_block(&blocks[0]).expect("Failed to add a test block");
+    let blocks_result: Vec<_> = ledger_state
+        .get_blocks(0, ledger_state.latest_block_height() + 1)
+        .unwrap()
+        .filter_map(|block_result| block_result.ok())
+        .collect();
+
+    drop(ledger_state);
```
ljedrz:
```suggestion
```
no need

src/network/peer.rs
IAvecilla: This is great! much cleaner now. The change was made in this commit [f4eccff](https://github.com/AleoHQ/snarkOS/pull/1753/commits/f4eccff4fbc5864b1ce3aeb3d519f1edf79bad88)

src/rpc/rpc_impl.rs
IAvecilla: I agree that wrapping `blocks` in an `Ok` and using `?` is not so good. I can't omit the `?` because of the error type of the result that is a `RpcError` but maybe the new change is better. Let me know what you think or if you think of another solution. The change was made in [f4eccff](https://github.com/AleoHQ/snarkOS/pull/1753/commits/f4eccff4fbc5864b1ce3aeb3d519f1edf79bad88)

storage/src/state/ledger.rs
IAvecilla: Great! I had to add some extra type annotations to get `collect` working but I think is better now. The change was made in this commit [f4eccff](https://github.com/AleoHQ/snarkOS/pull/1753/commits/f4eccff4fbc5864b1ce3aeb3d519f1edf79bad88)

storage/src/state/tests.rs
IAvecilla: Done! The change was made in this commit [f4eccff](https://github.com/AleoHQ/snarkOS/pull/1753/commits/f4eccff4fbc5864b1ce3aeb3d519f1edf79bad88)

src/rpc/rpc_impl.rs
```diff
@@ -75,7 +76,11 @@ impl<N: Network, E: Environment> RpcFunctions<N> for RpcContext<N, E> {
     /// Returns up to `MAXIMUM_BLOCK_REQUEST` blocks from the given `start_block_height` to `end_block_height` (inclusive).
     async fn get_blocks(&self, start_block_height: u32, end_block_height: u32) -> Result<Vec<Block<N>>, RpcError> {
         let safe_start_height = max(start_block_height, end_block_height.saturating_sub(E::MAXIMUM_BLOCK_REQUEST - 1));
-        Ok(self.ledger.get_blocks(safe_start_height, end_block_height)?)
+        Ok(self
+            .ledger
+            .get_blocks(safe_start_height, end_block_height)?
+            .map(|block_result| block_result)
```
ljedrz:
```suggestion
```
shouldn't this call be safe to remove?

ljedrz (approved):
LGTM with a final nit

src/rpc/rpc_impl.rs
IAvecilla: Yes, You're right. I already deleted that with another type annotation that wasn't necessary. The change was made in this commit [e3567ed](https://github.com/AleoHQ/snarkOS/pull/1753/commits/e3567edba73f49b758d9402cf8c1edcdcf2cf2d3)

howardwu (approved):

PR #1759, opened 2022-05-06
Feature: get all the ciphertexts from the LedgerState

IAvecilla:
## Motivation

This PR is related to #1725. It has some improvements in performance and now it returns all ciphertexts in the correct order.
The motivation is to get all the ciphertexts directly from LedgerState to avoid iterating over all the transactions and commitments to get them.

## Test Plan

There is a new test created in `storage/src/state/tests.rs`
- `test_get_all_ciphertexts` - This test asserts that this function gets all the ciphertexts in the ledger without a specific order.

PR #1760, opened 2022-05-06
Feature: get ciphertexts improvement

IAvecilla:
## Motivation

Improve performance in `get_ciphertexts` method of `LedgerState`. This improvement was added in another PR to Testnet 3 #1759.

## Test Plan

There are no new tests but the `test_get_all_ciphertexts` has been modified because the use of HashSet is no longer needed.

### DIFF
--- a/storage/src/state/ledger.rs
+++ b/storage/src/state/ledger.rs
@@ -471,7 +471,7 @@ impl<N: Network> LedgerState<N> {
     }
 
     // Returns all the ciphertexts in ledger.
-    pub fn get_ciphertexts(&self) -> impl Iterator<Item = Result<N::RecordCiphertext>> + '_ {
+    pub fn get_ciphertexts(&self) -> impl Iterator<Item = N::RecordCiphertext> + '_ {
         self.blocks.get_ciphertexts()
     }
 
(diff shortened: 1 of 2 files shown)


ljedrz (approved):
Looks fine to me, but I'm already getting these changes mixed up between `testnet2` and `testnet3`; @howardwu if we even intend to still use the former, we should probably rather only merge into `testnet3` and later backport if they are needed in `testnet2`, otherwise we're likely to get things mixed up

storage/src/state/ledger.rs
```diff
@@ -1578,8 +1578,10 @@ impl<N: Network> TransactionState<N> {
     }
 
     // Returns all the record ciphertexts.
-    fn get_ciphertexts(&self) -> impl Iterator<Item = Result<N::RecordCiphertext>> + '_ {
-        self.commitments.keys().map(move |commitment| self.get_ciphertext(&commitment))
+    fn get_ciphertexts(&self) -> impl Iterator<Item = N::RecordCiphertext> + '_ {
+        self.transitions
+            .values()
+            .flat_map(|(_, _, transition)| transition.ciphertexts().cloned().collect::<Vec<N::RecordCiphertext>>())
```
ljedrz, line 1584: is the `collect` needed here?

howardwu:
@IAvecilla we'll need to resolve the merge conflict here

storage/src/state/ledger.rs
IAvecilla: I think so, the `cloned` part gives me the ownership of the ciphertexts in that transition and the `collect` allocates the iterator over those ciphertexts.

IAvecilla:
> @IAvecilla we'll need to resolve the merge conflict here

Done! 👍

aleo-svc-exp:

IAvecilla:
These changes apply only for `testnet2` because `testnet3` has a big refactor on this logic, it will be better to do the necessary changes on the testnet3 if necessary

PR #1768, opened 2022-05-12
Add a ciphertext_lookup_by_commitment benchmark

ljedrz:
This should help the ciphertext-lookup-related work.

Cc @IAvecilla

### DIFF
--- a/storage/benches/lookups.rs
+++ b/storage/benches/lookups.rs
@@ -97,17 +97,25 @@ fn lookups(c: &mut Criterion) {
         })
     });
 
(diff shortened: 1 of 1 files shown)


niklaslong (approved):
LGTM 👍

ljedrz:
Note: this new benchmark stands out in terms of numbers (a few `ms` instead of `<1us` on my PC), because it performs deserialization (since there is no `LedgerState::contains_ciphertext` method and I don't know if it's desirable).

howardwu (approved):

PR #1776, opened 2022-05-20
Fix new clippy lints

niklaslong:
Fixes the new Rust `1.63` lints.

### DIFF
--- a/snarkos/display/aleo.rs
+++ b/snarkos/display/aleo.rs
@@ -18,6 +18,8 @@ use snarkvm::dpc::prelude::*;
 
 use colored::*;
 
+use std::fmt::Write;
+
 pub fn welcome_message() -> String {
     let mut output = String::new();
     output += &r#"
@@ -85,7 +87,8 @@ pub fn notification_message<N: Network>(miner: Option<Address<N>>) -> String {
     .bold();
 
     if let Some(miner) = miner {
-        output += &format!(
+        let _ = write!(
+            output,
             "
      Your Aleo miner address is {}
 
--- a/snarkos/updater.rs
+++ b/snarkos/updater.rs
@@ -16,6 +16,7 @@
 
 use colored::Colorize;
 use self_update::{backends::github, version::bump_is_greater, Status};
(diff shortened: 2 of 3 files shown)


ljedrz (approved):
LGTM :+1:

snarkos/updater.rs
```diff
@@ -34,7 +35,7 @@ impl Updater {
 
         let mut output = "List of available versions\n".to_string();
         for release in releases {
-            output += &format!("  * {}\n", release.version);
+            let _ = writeln!(output, "  * {}", release.version);
```
howardwu, line 38: Does this need the `let _ = ` part?

snarkos/display/aleo.rs
```diff
@@ -85,7 +87,8 @@ pub fn notification_message<N: Network>(miner: Option<Address<N>>) -> String {
     .bold();
 
     if let Some(miner) = miner {
-        output += &format!(
+        let _ = write!(
```
howardwu, line 90: Does this need the `let _ = ` part?

snarkos/updater.rs
ljedrz: ~there could be a `?` at the end instead, but~ since this macro is writing to memory (`output`), there is no need to account for any errors

snarkos/display/aleo.rs
ljedrz: same reasoning as in my other comment; not having `let _ =` would cause an `unused Result that must be used` warning, and there is no possibility of an error (unless it's an OOM, in which case there are bigger issues), so the `Result` doesn't need to be handled

snarkos/updater.rs
niklaslong: And the error itself isn't castable to the specific type returned from the function, at least not currently (so `?` won't work without some additional changes).

ljedrz: ah cool, I missed the fact that the error was incompatible :+1:

howardwu (approved):

PR #1797, opened 2022-06-14
Update deps

ljedrz:
Updates `clap`, `jsonrpsee`, and applies `cargo update` afterwards.

### DIFF
--- a/.crawler/Cargo.toml
+++ b/.crawler/Cargo.toml
@@ -49,7 +49,7 @@ optional = true
 version = "0.12"
 
 [dependencies.pea2pea]
-version = "0.37"
+version = "0.38"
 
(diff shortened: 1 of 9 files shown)


niklaslong (approved):

howardwu (approved):

howardwu:
@ljedrz does the `short` still work with these changes? I know clap is undergoing some major feature changes, but want to make sure these commands have not changed.

ljedrz:
`short` without a right-hand value uses the first character (of the variant's name) and `long` without one uses the full name, i.e.
```
#[clap(long = "addr", short = 'a')]
addr: SocketAddr,
```
is equivalent to:
```
#[clap(long, short)]
addr: SocketAddr,
```

PR #1871, opened 2022-08-19
Fix CI failures

ljedrz:
This PR comments out some test features that are currently unused and a few old tests that need to be fixed first.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -68,35 +68,35 @@ commands:
 
 jobs:
 
-  environment:
-    docker:
-      - image: cimg/rust:1.59
-    resource_class: xlarge
-    parallelism: 1
-    steps:
-      - checkout
-      - setup_environment:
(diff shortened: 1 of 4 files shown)


ljedrz:
While I'm not exactly sure why the CI reports a dependency issue in the `environment` and `snarkos` jobs (as I can't reproduce it locally), we currently don't have any tests there, so we can disable them for the time being.

raychu86:
Can you help fix them instead of leaving them commented out?

For example, the rocksdb tests can be updated to address the breaking changes.

ljedrz:
This was an initial proposal to quickly unblock other work that kept getting bogus failures, it might still be a good idea until the failures are fixed; as for rocksdb, this is already in progress as https://github.com/AleoHQ/snarkOS/pull/1878.

ljedrz:
Closing, as it's no longer applicable.

PR #1872, opened 2022-08-19
Rename RPC args to HTTP, use those values for the warp server

### LINKED ISSUE #1875: [Bug] The node's HTTP server (the replacement for the RPC one) panics when attempting to bind
It happens immediately on the node's start-up, it doesn't affect the other functionalities.

There's already a fixing PR ready, reporting it here in case the users encounter it before it's merged.

ljedrz:
This fixes the initial panic if port 80 is in use (which is extremely likely), adjusts the settings to the current server setup and allows the user to specify the address or disable it.

Closes https://github.com/AleoHQ/snarkOS/issues/1875.

### DIFF
--- a/snarkos/cli.rs
+++ b/snarkos/cli.rs
@@ -57,7 +57,7 @@ pub struct CLI {
     pub connect: Option<String>,
 
     /// Specify the IP address and port for the HTTP server.
-    #[clap(parse(try_from_str), default_value = "127.0.0.1:3033")]
+    #[clap(parse(try_from_str), default_value = "0.0.0.0:3033")]
     pub rest: SocketAddr,
     /// If the flag is set, the node will not initialize the HTTP server.
     #[clap(long)]
--- a/snarkos/cli.rs
+++ b/snarkos/cli.rs
@@ -56,18 +56,12 @@ pub struct CLI {
     #[clap(long = "connect")]
     pub connect: Option<String>,
 
-    /// Specify the IP address and port for the RPC server.
-    #[clap(parse(try_from_str), default_value = "0.0.0.0:3033", long = "rpc")]
-    pub rpc: SocketAddr,
-    /// Specify the username for the RPC server.
-    #[clap(default_value = "root", long = "username")]
-    pub rpc_username: String,
-    /// Specify the password for the RPC server.
-    #[clap(default_value = "pass", long = "password")]
-    pub rpc_password: String,
-    /// If the flag is set, the node will not initialize the RPC server.
+    /// Specify the IP address and port for the HTTP server.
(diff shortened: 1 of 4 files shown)


niklaslong (approved):

snarkos/cli.rs
```diff
@@ -48,18 +48,12 @@ pub struct CLI {
...
-    /// Specify the username for the RPC server.
-    #[clap(default_value = "root", long = "username")]
-    pub rpc_username: String,
-    /// Specify the password for the RPC server.
-    #[clap(default_value = "pass", long = "password")]
-    pub rpc_password: String,
-    /// If the flag is set, the node will not initialize the RPC server.
+    /// Specify the IP address and port for the HTTP server.
+    #[clap(parse(try_from_str), default_value = "127.0.0.1:3033", long = "http")]
+    pub http: SocketAddr,
```
raychu86: Can we rename `http` and `nohttp` to `rest` and `norest`?

snarkos/cli.rs
```diff
@@ -48,18 +48,12 @@ pub struct CLI {
...
-    pub rpc: SocketAddr,
-    /// Specify the username for the RPC server.
-    #[clap(default_value = "root", long = "username")]
-    pub rpc_username: String,
-    /// Specify the password for the RPC server.
-    #[clap(default_value = "pass", long = "password")]
-    pub rpc_password: String,
-    /// If the flag is set, the node will not initialize the RPC server.
+    /// Specify the IP address and port for the HTTP server.
+    #[clap(parse(try_from_str), default_value = "127.0.0.1:3033", long = "http")]
```
raychu86:
```suggestion
    #[clap(parse(try_from_str), default_value = "0.0.0.0:3033", long = "http")]
```

We tried `127.0.0.1` and it does not work

ljedrz: It doesn't work locally? I just tried
```
curl -X GET http://localhost:3033/testnet3/latest/hash
```
and it worked fine for me. IMO it's the safer default (it can still be set to `0.0.0.0` on startup), but it's your call; I'll add a commit

ljedrz:
Rebased and applied comments.

damons:
Binding to all IP addresses by default will generally cause confusion.  Does this patch also attempt to bind to port 80 by default?  This requires running as root, and IMO, that's a security problem.

ETA on this?  This issue affects everyone in the community on first-run experience.

ljedrz:
> Binding to all IP addresses by default will generally cause confusion.

I'd also prefer it to be set to localhost instead, it should work fine as long as the requests are local.

> Does this patch also attempt to bind to port 80 by default?

This patch doesn't change that value, but I would also recommend it to be set to something different.

ljedrz:
Closing, as it's no longer applicable.

PR #1876, opened 2022-08-19
Use blocking tasks when calling Ledger::add_next_block

ljedrz:
This operation is not fast enough to be called in async contexts without a blocking task.


ljedrz:
Superseded by https://github.com/AleoHQ/snarkOS/pull/1880.

PR #1878, opened 2022-08-19
Test suite for storage module

ilitteri:
## Motivation

We want to test the storage proper functionality.

## Test Plan

Runnig `cd snarkos/store && cargo test` should run the following test suite:

```
test_a_value_cannot_be_removed_twice
test_a_value_that_was_not_inserted_cannot_be_removed
test_a_key_is_contained_after_a_value_was_inserted_with_it
test_a_key_is_not_contained_if_no_value_was_inserted_with_it
test_open_map_opens_a_datamap_for_a_given_dataid
test_can_iter_on_keys_after_inserting
test_trying_to_get_a_value_associated_to_a_non_existent_key_returns_none
test_can_iter_on_pairs_after_inserting
test_can_iter_on_values_after_inserting
test_open
test_reopen
test_a_value_that_was_inserted_can_be_retrieved_with_its_associated_key
test_a_value_that_was_inserted_can_be_removed_with_its_associated_key
```


snarkos/store/rocksdb/tests.rs
```diff
@@ -14,185 +14,170 @@
...
+    assert!(expected_result.is_ok());
+}
 
 #[test]
-fn test_insert_and_contains_key() {
-    let storage = RocksDB::open(temp_dir(), 0).expect("Failed to open storage");
-    let mut map: TestMap = storage.open_map(DataID::Test).expect("Failed to open data map");
+#[serial]
+fn test_a_key_is_not_contained_if_no_value_was_inserted_with_it() {
+    let test_map = setup_test_map(3);
```
ljedrz, line 65: with the current setup, any test using a map with ID `3` will remove the storage for the (non-test) `testnet3`

snarkos/store/mod.rs
```diff
@@ -68,6 +68,8 @@ pub enum DataID {
     OutputRecordMap,
     OutputRecordNonceMap,
     OutputExternalRecordMap,
+    // Test
```
facundoolano: can you elaborate here?

Cargo.toml
```diff
@@ -127,6 +127,9 @@ optional = true
 [dev-dependencies.rusty-hook]
 version = "0.11.2"
 
+[dev-dependencies.serial_test]
+version = "0.9.0"
```
ljedrz, line 131: instead of running the tests one-by-one, which is a workaround, we shouldn't be using `&'static` for storage in the first place; this will require snarkVM-side changes first

snarkos/store/mod.rs
```diff
@@ -68,6 +68,8 @@ pub enum DataID {
...
+    Test,
```
ljedrz: this variant should be the very last one in order no to perturb the IDs that follow it (we want the tests to be as close to regular runs as possible)

snarkos/store/rocksdb/tests.rs
```diff
@@ -14,185 +14,170 @@
...
-fn test_insert_and_values() {
-    let storage = RocksDB::open(temp_dir(), 0).expect("Failed to open storage");
-    let mut map = storage.open_map::<u32, String>(DataID::Test).expect("Failed to open data map");
-    map.insert(123456789, "123456789".to_string()).expect("Failed to insert");
-
-    let mut values = map.values();
-    assert_eq!(Some("123456789".to_string()), values.next().map(|v| v.to_string()));
-    assert_eq!(None, values.next());
+#[serial]
+#[ignore = "Removing a key that does not exist does not result in an error"]
```
ljedrz, line 107: why ignore this test? it should be able to run just fine, and it's not slow or anything

snarkos/store/rocksdb/tests.rs
```diff
@@ -14,185 +14,170 @@
...
-    {
-        let storage = RocksDB::open(directory, 0).expect("Failed to open storage");
-        let map = storage.open_map::<u32, String>(DataID::Test).expect("Failed to open data map");
-        assert_eq!(
-            Some("123456789".to_string()),
-            map.get(&123456789).expect("Failed to get").map(|v| v.to_string())
-        );
-    }
+#[serial]
+#[ignore = "Removing a key that does not exist does not result in an error"]
```
ljedrz, line 118: ditto

snarkos/store/rocksdb/tests.rs
```diff
@@ -14,185 +14,170 @@
...
 
-    // Sample `ITERATION` random field elements to store.
-    for i in 0..ITERATIONS {
-        let value = Scalar::<CurrentNetwork>::rand(rng);
-        map.insert(i, value).expect("Failed to insert");
+#[test]
+#[serial]
+fn test_reopen() {
+    {
+        let test_map = setup_test_map(13);
```
ljedrz, line 172: we should probably be picking the ID at random in all the tests if we want to try as many as possible out in different conditions

ilitteri: Do you mean the `network_id` or the `DataID`? If you mean the frist one I think its ok, and if you mean the `DataID` what do you think about iterate through all the enum values? (if this ends to be too slow, we'll go with the random)

snarkos/store/rocksdb/tests.rs
ilitteri: We ignored this test because we saw that it doesn't do what we thought it should do. Now that you mention it, it wasn't ok to ignore them.

ilitteri: Is it ok for us to assume that this API should have this behaviour?

snarkos/store/rocksdb/tests.rs
ilitteri: You're right!

snarkos/store/rocksdb/tests.rs
ljedrz: The data id; the network id should probably always be the current one (though right now it would interfere with the non-test db). As for iterating, I'd suggest having a `static` ID that starts with a random value and is bumped with each test and wraps around, but it likely wouldn't work due to the use of `serial`; just doing all IDs in every test is redundant, it's good enough that a few get tested with every run.

snarkos/store/rocksdb/tests.rs
ilitteri: What do you think about receiving a `dev` parameter in the `open` method and replace the `None` later used in the `aleo_ledger_dir` call?

snarkos/store/rocksdb/tests.rs
ljedrz: Either way might make sense; I guess it's up to Howard

snarkos/store/rocksdb/tests.rs
ljedrz: it's an option, though IMO the best way would be to alter `aleo_std::aleo_ledger_dir` to use something like the `tempfile` crate in tests (note: the `dev` param in `aleo_ledger_dir` indicates dev mode, which is *not* the same as a test), which would provide automatic directory deletion on `Drop`

ljedrz (changes requested):
while I'm happy to have some storage tests even with the `serial` workaround, the [directory issue](https://github.com/AleoHQ/snarkOS/pull/1878/files#r950547637) must be resolved before this PR can be merged

snarkos/store/rocksdb/tests.rs
ilitteri: I agree with your opinion. I can work on something like that.

snarkos/store/mod.rs
```diff
@@ -85,4 +85,6 @@ pub enum DataID {
     KeyValueIDMap,
     KeyMap,
     ValueMap,
+    // A value for testing purposes
+    Test,
```
raychu86, line 89: Generally don't like sprinkling feature flags everywhere, but we should only expose the `Test` enum in the testing context. 

Adding `#[cfg(test)]` might be sufficient

snarkos/store/rocksdb/tests.rs
```diff
@@ -14,185 +14,170 @@
...
+use serial_test::serial;
+use snarkvm::compiler::{Map, MapRead};
 
-// use snarkvm::compiler::{Map, MapRead};
+use std::fs;
 
-pub(crate) fn temp_dir() -> std::path::PathBuf {
-    tempfile::tempdir().expect("Failed to open temporary directory").into_path()
+fn remove_test_dir(network_id: u16) {
+    let _ = fs::remove_dir_all(format!("~/.aleo/storage/.ledger-{}", network_id));
```
raychu86, line 27: This will likely interfere with the existing ledger (which i believe lives at `.ledger-3`). We used the temp_dir previously to avoid the confusion.

On a related note, we are likely going to need to modify the `.ledger-{}` storage paths to add support for custom testnets, to solve the same collision issues.

PR #1879, opened 2022-08-19
Test Coverage

ilitteri:
## Motivation

We want to add code coverage to this codebase using [Tarpaulin](https://github.com/xd009642/tarpaulin) and [Codecov](https://about.codecov.io/)

### DIFF
--- a/.github/workflows/codecov.yml
+++ b/.github/workflows/codecov.yml
@@ -1,0 +1,23 @@
+name: Codecov
+
+on:
+  push:
+    branches: [ testnet3 ]
+  pull_request:
+    branches: [ '*' ]
+
+jobs:
+  check:
+    name: snarkos
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+
+      - name: coverage
+        run: docker run --security-opt seccomp=unconfined -v "${PWD}:/volume" xd009642/tarpaulin /bin/bash -c "apt-get update; apt-get install -y libclang-dev; cargo tarpaulin --workspace -t 1800"
+
+      - name: Upload coverage to codecov.io
+        uses: codecov/codecov-action@v2
+        with:
+          fail_ci_if_error:     true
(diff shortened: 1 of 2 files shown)


facundoolano (approved):
LGTM although I think it would be preferable to put this in CircleCI so we have all CI stuff in a single place

ljedrz:
I think the best place for CI test coverage would be snarkVM, as many functions in snarkOS are not usable on their own and require more comprehensive integration tests; that being said, it might be practical for the storage operations, or any other self-contained functionalities that would be added in the future.

ljedrz (commented):
I think the best place for CI test coverage would be snarkVM (and likely only for some of its components), as many functions in snarkOS are not usable on their own and require more comprehensive integration tests; that being said, it might be practical for the storage operations, or any other self-contained functionalities that would be added in the future.

facundoolano:
> I think the best place for CI test coverage would be snarkVM (and likely only for some of its components), as many functions in snarkOS are not usable on their own and require more comprehensive integration tests; that being said, it might be practical for the storage operations, or any other self-contained functionalities that would be added in the future.

yeah I think if it's not too much effort to set it up, the best is to do it by default on all repos, but we don't necessarily need to attempt to increase the coverage if the tests to do it would be awkward to implement due to dependencies. 

But now that you mention it @ljedrz can you point to some examples of functions that would be hard to test because they can be run on their own? (this will help us to familiarize with the repo, and also could point to potential improvements --e.g. if they are hard to test maybe they shouldn't live in this repo?)

ljedrz:
> can you point to some examples of functions that would be hard to test because they can **not** be run on their own?

for example any function/method that involves a long-running task, e.g. sending pings or handling inbound network messages

> if they are hard to test maybe they shouldn't live in this repo?

this repo contains a binary (i.e. it's not a library), so I wouldn't say that functions/methods like that shouldn't be here - it's unavoidable that we'll have some like it if we want to have self-contained functionalities

raychu86:
Can we migrate codecov CI from Github Actions to CircleCI?

PR #1880, opened 2022-08-19
Improve Client node usability 

raychu86:
## Motivation

This PR removes the two hard-coded instances of `LEADER_IP` and instead replaces it with a CLI based declaration. This gives the ability to choose a specified beacon/leader nodes in custom testnets.

Additional (minor) fixes:

- Uses `Ledger` interface instead of directly accessing the `InternalLedger` functions. This should also address - https://github.com/AleoHQ/snarkOS/pull/1876
- Disable fast-syncing `initial_sync_with_network`  in development mode.
- Handle thread loops properly with tokio `Interval`s
- Update incorrect comments.

### DIFF
--- a/snarkos/ledger/mod.rs
+++ b/snarkos/ledger/mod.rs
@@ -103,16 +103,8 @@ impl<N: Network> Ledger<N> {
         })
         .await??;
 
-        // Ensure the given block is a valid next block.
-        self.ledger.read().check_next_block(&next_block)?;
-
         // Add the next block to the ledger.
-        let self_clone = self.clone();
-        let next_block_clone = next_block.clone();
-        if let Err(error) = task::spawn_blocking(move || self_clone.ledger.write().add_next_block(&next_block_clone)).await? {
-            // Log the error.
-            warn!("{error}");
-        }
+        self.add_next_block(&next_block).await?;
 
         // Broadcast the block to all peers.
         let peers = self.peers().read().clone();
@@ -122,6 +114,22 @@ impl<N: Network> Ledger<N> {
 
         // Return the next block.
         Ok(next_block)
+    }
+
+    /// Attempts to add the given block to the ledger.
+    pub(crate) async fn add_next_block(self: &Arc<Self>, next_block: &Block<N>) -> Result<()> {
+        // Ensure the given block is a valid next block.
+        self.ledger.read().check_next_block(next_block)?;
+
+        // Add the next block to the ledger.
+        let self_clone = self.clone();
+        let next_block_clone = next_block.clone();
+        if let Err(error) = task::spawn_blocking(move || self_clone.ledger.write().add_next_block(&next_block_clone)).await? {
+            // Log the error.
+            warn!("{error}");
+        }
+
+        Ok(())
     }
 }
 
--- a/snarkos/ledger/mod.rs
+++ b/snarkos/ledger/mod.rs
@@ -25,7 +25,10 @@ use futures::StreamExt;
 use indexmap::IndexMap;
 use once_cell::race::OnceBox;
 use parking_lot::RwLock;
-use std::{net::SocketAddr, sync::Arc};
+use std::{
+    net::{IpAddr, SocketAddr},
+    sync::Arc,
+};
 use tokio::task;
 
 pub(crate) type InternalLedger<N> = snarkvm::prelude::Ledger<N, BlockDB<N>, ProgramDB<N>>;
@@ -213,7 +216,7 @@ impl<N: Network> Ledger<N> {
 // Internal operations.
 impl<N: Network> Ledger<N> {
     /// Syncs the ledger with the network.
-    pub(crate) async fn initial_sync_with_network(self: &Arc<Self>, leader_ip: &str) -> Result<()> {
+    pub(crate) async fn initial_sync_with_network(self: &Arc<Self>, leader_ip: &IpAddr) -> Result<()> {
         /// The number of concurrent requests with the network.
         const CONCURRENT_REQUESTS: usize = 100;
         /// Url to fetch the blocks from.
--- a/snarkos/ledger/mod.rs
+++ b/snarkos/ledger/mod.rs
@@ -121,15 +121,13 @@ impl<N: Network> Ledger<N> {
 
     /// Attempts to add the given block to the ledger.
     pub(crate) async fn add_next_block(self: &Arc<Self>, next_block: &Block<N>) -> Result<()> {
-        // Ensure the given block is a valid next block.
-        self.ledger.read().check_next_block(next_block)?;
-
         // Add the next block to the ledger.
         let self_clone = self.clone();
         let next_block_clone = next_block.clone();
         if let Err(error) = task::spawn_blocking(move || self_clone.ledger.write().add_next_block(&next_block_clone)).await? {
             // Log the error.
             warn!("{error}");
+            return Err(error);
         }
 
         Ok(())
--- a/snarkos/ledger/mod.rs
+++ b/snarkos/ledger/mod.rs
@@ -71,8 +71,6 @@ impl<N: Network> Ledger<N> {
             .set(Box::new(server))
             .map_err(|_| anyhow!("Failed to save the server"))?;
 
-        // Sync the ledger with the network.
-        ledger.initial_sync_with_network().await?;
         // Return the ledger.
         Ok(ledger)
     }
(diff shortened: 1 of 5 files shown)


snarkos/ledger/mod.rs
```diff
@@ -125,6 +115,22 @@ impl<N: Network> Ledger<N> {
         // Return the next block.
         Ok(next_block)
     }
+
+    /// Attempts to add the given block to the ledger.
+    pub(crate) async fn add_next_block(self: &Arc<Self>, next_block: &Block<N>) -> Result<()> {
+        // Ensure the given block is a valid next block.
+        self.ledger.read().check_next_block(next_block)?;
+
```
```suggestion
```
this operation is redundant, as it's already performed within `Ledger::add_next_block`: https://github.com/AleoHQ/snarkVM/blob/testnet3/vm/compiler/src/ledger/mod.rs#L539

snarkos/ledger/mod.rs
```diff
@@ -207,20 +213,17 @@ impl<N: Network> Ledger<N> {
 // Internal operations.
 impl<N: Network> Ledger<N> {
     /// Syncs the ledger with the network.
-    async fn initial_sync_with_network(self: &Arc<Self>) -> Result<()> {
+    pub(crate) async fn initial_sync_with_network(self: &Arc<Self>, leader_ip: &str) -> Result<()> {
```
```suggestion
    pub(crate) async fn initial_sync_with_network(self: &Arc<Self>, leader_ip: IpAddr) -> Result<()> {
```
the `Display` impl for `IpAddr` makes it possible to use it directly in `format!`

snarkos/network/mod.rs
```diff
@@ -234,23 +234,28 @@ pub async fn handle_listener<N: Network>(listener: TcpListener, ledger: Arc<Ledg
 
 /// Send a ping to all peers every 10 seconds.
 pub async fn send_pings<N: Network>(ledger: Arc<Ledger<N>>) -> Result<(), Box<dyn std::error::Error>> {
```
ljedrz:
```suggestion
pub fn send_pings<N: Network>(ledger: Arc<Ledger<N>>) -> Result<(), Box<dyn std::error::Error>> {
```

snarkos/network/mod.rs
```diff
@@ -234,23 +234,28 @@ pub async fn handle_listener<N: Network>(listener: TcpListener, ledger: Arc<Ledg
...
-    loop {
-        thread::sleep(time::Duration::from_secs(10));
+    tokio::spawn(async move {
```
ljedrz, line 237: we should return this task's handle from this method; even if we don't track task handles yet, we will want to be doing it shortly

snarkos/network/mod.rs
```diff
@@ -234,23 +234,28 @@ pub async fn handle_listener<N: Network>(listener: TcpListener, ledger: Arc<Ledg
...
+            interval.tick().await;
 
-        let peers = ledger.peers().read().clone();
+            let peers = ledger.peers().read().clone();
 
-        for (addr, outbound) in peers.iter() {
-            if let Err(err) = outbound.send(Message::<N>::Ping).await {
-                warn!("Error sending ping {} to {}", err, addr);
+            for (addr, outbound) in peers.iter() {
+                if let Err(err) = outbound.send(Message::<N>::Ping).await {
```
ljedrz:
```suggestion
                if let Err(err) = outbound.try_send(Message::<N>::Ping) {
```
if the channel is full, we likely want to skip that peer

snarkos/ledger/mod.rs
ljedrz: it's easy to miss such things, which is another (besides added safety) argument for introducing something like I mentioned here: https://github.com/AleoHQ/snarkVM/pull/990#issuecomment-1220581592

snarkos/ledger/mod.rs
```diff
@@ -125,6 +115,22 @@ impl<N: Network> Ledger<N> {
         // Return the next block.
         Ok(next_block)
     }
+
+    /// Attempts to add the given block to the ledger.
+    pub(crate) async fn add_next_block(self: &Arc<Self>, next_block: &Block<N>) -> Result<()> {
```
ljedrz, line 123: very much agree with adding a dedicated method for this operation, was planning on doing that myself :+1:

ljedrz (commented):
LGTM in general; left some comments, the most important one being https://github.com/AleoHQ/snarkOS/pull/1880/files#r950670260

snarkos/ledger/mod.rs
ljedrz: if we want the block checks to be more explicit (i.e. outside of `add_next_block`) and to introduce some extra safety, the simplest solution would be something like [this](https://github.com/AleoHQ/snarkVM/compare/testnet3...ljedrz:snarkVM:explicit_block_verification?expand=1). This assumes that the `Block` is the point where all the verification happens (which seems to be the case right now and makes sense).

snarkos/network/mod.rs
```diff
@@ -119,7 +119,7 @@ pub(crate) async fn handle_peer<N: Network>(
                             // Check if the block can be added to the ledger.
                             if block.height() == ledger.ledger().read().latest_height() + 1 {
```
ljedrz, line 120: this check is not needed, as it happens in `add_next_block` (when `check_next_block` is called: https://github.com/AleoHQ/snarkVM/blob/testnet3/vm/compiler/src/ledger/mod.rs#L350)

snarkos/network/mod.rs
```diff
@@ -161,7 +161,7 @@ pub(crate) async fn handle_peer<N: Network>(
...
```
ljedrz, line 162: same as https://github.com/AleoHQ/snarkOS/pull/1880/files#r950697757

snarkos/network/mod.rs
raychu86: Agreed, this is a redundant check. Initially added to clean up logging, since we didn't want to output a "Failed to process block..." warning for every misaligned block and instead declare we are "skipping" it.

snarkos/ledger/mod.rs
raychu86: Great catch! For now, I've removed the duplicate check. We can update this again if we go down the route of passing through explicitly "Verified<Block>"s  in snarkVM.

snarkos/ledger/mod.rs
```diff
@@ -207,20 +214,17 @@ impl<N: Network> Ledger<N> {
 // Internal operations.
 impl<N: Network> Ledger<N> {
     /// Syncs the ledger with the network.
-    async fn initial_sync_with_network(self: &Arc<Self>) -> Result<()> {
+    pub(crate) async fn initial_sync_with_network(self: &Arc<Self>, leader_ip: &IpAddr) -> Result<()> {
```
ljedrz, line 217:
```suggestion
    pub(crate) async fn initial_sync_with_network(self: &Arc<Self>, leader_ip: IpAddr) -> Result<()> {
```
`IpAddr` is `Copy`

snarkos/ledger/mod.rs
```diff
@@ -125,6 +118,20 @@ impl<N: Network> Ledger<N> {
         // Return the next block.
         Ok(next_block)
     }
+
+    /// Attempts to add the given block to the ledger.
+    pub(crate) async fn add_next_block(self: &Arc<Self>, next_block: &Block<N>) -> Result<()> {
```
ljedrz, line 123: if we passed `Block<N>` as value here, we could probably avoid having to clone it for the purpose of moving it into the blocking task

snarkos/network/mod.rs
ljedrz: if we want to keep it (it should be reasonably fast, so it's not a big deal), it'd probably be best to put it in the [`add_next_block`](https://github.com/AleoHQ/snarkOS/pull/1880/files#diff-a965d43412c665e2554b72a2041a4491c9267d2280f9688d349f282487e25b3bR123) method introduced by this PR (since the same is done when handling `BlockBroadcast`)

snarkos/network/mod.rs
ljedrz: as a reference, we're already doing this with the HTTP server's tasks: https://github.com/AleoHQ/snarkOS/blob/1f6c46327ea9db61f5b8ea577645a976c2c8b9c1/snarkos/ledger/server.rs#L72

ljedrz: done in https://github.com/AleoHQ/snarkOS/pull/1884, which is a follow-up to this PR

snarkos/ledger/mod.rs
ljedrz: done in https://github.com/AleoHQ/snarkOS/pull/1884

snarkos/network/mod.rs
ljedrz: though this would complicate the signature a bit; let's skip this for now

snarkos/ledger/mod.rs
ljedrz: done in https://github.com/AleoHQ/snarkOS/pull/1884, which is a follow-up to this PR

ljedrz (approved):
LGTM

snarkos/node.rs
```diff
@@ -41,21 +38,25 @@ impl<N: Network, E: Environment> Node<N, E> {
         // Initialize the ledger.
         let ledger = Ledger::<N>::load(account.private_key()).await?;
 
+        // If the node is not in development mode, perform fast sync with the network.
+        if cli.dev.is_none() {
```
facundoolano, line 42: why is fast sync not done in development mode?

facundoolano (approved):

snarkos/node.rs
ljedrz: dev mode is usually dedicated to running a few nodes locally, so syncing with the actual leader might be undesirable

facundoolano: gotcha, thanks

PR #1885, opened 2022-08-22
Api: re-export snarkVM's Testnet3 struct

niklaslong:
External callers of snarkOS' `Message` won't be able to satisfy the associated trait bounds unless `Testnet3` is also re-exported (snarkVM's won't do, unfortunately). In addition, all fields in `Message` are `Clone` so I've added the `derive` for the type as well.

### DIFF
--- a/snarkos/lib.rs
+++ b/snarkos/lib.rs
@@ -52,7 +52,7 @@ pub use snarkos_environment as environment;
 #[cfg(feature = "rpc")]
 pub use snarkos_rpc as rpc;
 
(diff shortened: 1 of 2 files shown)


ljedrz (approved):

niklaslong:
Note: this may not be needed if snarkVM dependencies are matched to the commit—I'll be checking this shortly.

niklaslong:
Confirm. Closing this for now. The `Clone` implementation for `Message` is still needed however, I will open a new PR.

PR #1886, opened 2022-08-22
Add custom testnet support for client nodes

raychu86:
## Motivation

This PR enables support for client nodes to connect to custom testnets. This is done by requesting the specific genesis block (via the beacon server) for the custom testnet and 

You can specify the specific beacon node using the flags `--dev <id> --connect_to_beacon <IP>`.

Note: There are a few additional things that should be addressed:

 - New Network ID for blocks in custom testnets.
 - Formalized genesis block request and go through Message::BlockRequest instead of the server.
 - Update the storage location. 
   - Currently all networks store blocks in the same `.aleo` database. You will need to run `cargo run --release clean` before spinning up clients for custom networks.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,7 +50,7 @@ version = "2.0.2"
 
 [dependencies.snarkvm]
 git = "https://github.com/AleoHQ/snarkVM.git"
-rev = "7555dc0"
+rev = "a6c838a"
 features = ["circuit", "console", "parallel", "utilities"]
 
 [dependencies.aleo-std]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,7 +50,7 @@ version = "2.0.2"
 
 [dependencies.snarkvm]
 git = "https://github.com/AleoHQ/snarkVM.git"
-rev = "a6c838a"
+rev = "e866f2d"
 features = ["circuit", "console", "parallel", "utilities"]
 
 [dependencies.aleo-std]
--- a/snarkos/ledger/mod.rs
+++ b/snarkos/ledger/mod.rs
@@ -60,7 +60,7 @@ impl<N: Network> Ledger<N> {
         // Initialize the ledger.
         let ledger = Arc::new(Self {
             ledger: RwLock::new(InternalLedger::open()?),
-            peers: RwLock::new(IndexMap::new()),
+            peers: Default::default(),
             server: OnceBox::new(),
             private_key: *private_key,
             view_key,
@@ -102,7 +102,7 @@ impl<N: Network> Ledger<N> {
         // Initialize the ledger.
         let ledger = Arc::new(Self {
             ledger: RwLock::new(internal_ledger),
-            peers: RwLock::new(IndexMap::new()),
+            peers: Default::default(),
             server: OnceBox::new(),
             private_key: *private_key,
             view_key,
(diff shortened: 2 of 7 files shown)


Cargo.toml
```diff
@@ -50,7 +50,7 @@ version = "2.0.2"
 
 [dependencies.snarkvm]
 git = "https://github.com/AleoHQ/snarkVM.git"
-rev = "7555dc0"
+rev = "a6c838a"
```
ljedrz: this commit should first be merged into snarkVM's `testnet3` branch

snarkos/ledger/mod.rs
```diff
@@ -71,8 +74,36 @@ impl<N: Network> Ledger<N> {
...
+        let view_key = ViewKey::try_from(private_key)?;
+        let address = Address::try_from(&view_key)?;
+
+        // Initialize the ledger.
+        let ledger = Arc::new(Self {
+            ledger: RwLock::new(InternalLedger::new_with_genesis(
+                &genesis_block,
+                genesis_block.signature().to_address(),
+            )?),
+            peers: RwLock::new(IndexMap::new()),
```
```suggestion
            peers: Default::default(),
```

ljedrz (commented):
This PR needs rebasing, as it still has commits that are already in `testnet3`

snarkos/network/mod.rs
```diff
@@ -232,6 +232,17 @@ pub async fn handle_listener<N: Network>(listener: TcpListener, ledger: Arc<Ledg
     Ok(())
 }
 
+// TODO (raychu86): Handle this request via `Message::BlockRequest`. This is currently not done,
+//  because the node has not established the leader as a peer.
+/// Request the genesis block from the leader.
+pub(super) async fn request_genesis_block<N: Network>(leader_ip: &SocketAddr) -> Result<Block<N>> {
```
ljedrz:
```suggestion
pub(super) async fn request_genesis_block<N: Network>(leader_ip: IpAddr) -> Result<Block<N>> {
```

snarkos/ledger/mod.rs
```diff
@@ -60,7 +60,49 @@ impl<N: Network> Ledger<N> {
         // Initialize the ledger.
         let ledger = Arc::new(Self {
             ledger: RwLock::new(InternalLedger::open()?),
-            peers: RwLock::new(IndexMap::new()),
+            peers: Default::default(),
+            server: OnceBox::new(),
+            private_key: *private_key,
```
ljedrz: nit: this implies that `PrivateKey` is `Copy`, which means it should be passed by value to this function

ljedrz: also, it seems that we don't need `async` here

snarkos/ledger/mod.rs
```diff
@@ -60,7 +60,49 @@ impl<N: Network> Ledger<N> {
...
+            .server
+            .set(Box::new(server))
+            .map_err(|_| anyhow!("Failed to save the server"))?;
+
+        // Return the ledger.
+        Ok(ledger)
+    }
+
+    /// Initializes a new instance of the ledger.
+    pub(super) async fn new_with_genesis(private_key: &PrivateKey<N>, genesis_block: Block<N>) -> Result<Arc<Self>> {
```
```suggestion
    pub(super) fn new_with_genesis(private_key: PrivateKey<N>, genesis_block: Block<N>) -> Result<Arc<Self>> {
```

ljedrz (approved):

PR #1895, opened 2022-08-26
Upgrading Dockerfile to Ubuntu 22.04

zosorock:
Additionally:
- adding Docker Compose file with environmental variable support
- upgrading CircleCI executor images and sccache version
- upgrading Rust to 1.63
- upgrading sccache to 0.3.0

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -102,7 +102,7 @@ jobs:
 
   # codecov:
   #   machine:
-  #     image: ubuntu-2204:2022.07.1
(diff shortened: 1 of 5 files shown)


rpyell-aleo (approved):
I know the build is failing but it doesn't look like it has anything to do with these changes so LGTM.

howardwu:
Let's queue this PR up for after #1871 is approved and merged

howardwu (approved):

PR #1897, opened 2022-08-30
Implement program deployment and execution endpoints

raychu86:
## Motivation

This PR updates the snarkVM rev to [5888d15](https://github.com/AleoHQ/snarkVM/commit/5888d15) and implements the program deployment and program execute endpoints.

PR #1900, opened 2022-08-30
Add get program and execution endpoints

facundoolano:
This adds a program execution endpoint, similar to the deployment one. The additional fee calculation (assumed to be the same for both) has been extracted to a helper function. This also fixes a couple of stale comments mentioned in another PR.

This is currently pointing to the #1897 branch, I can send it to the testnet3 branch once that's merged if preferred.

This also incorporates the get program endpoint from #1901

### DIFF
--- a/snarkos/ledger/server.rs
+++ b/snarkos/ledger/server.rs
@@ -15,11 +15,12 @@
 // along with the snarkOS library. If not, see <https://www.gnu.org/licenses/>.
 
 use crate::Ledger;
+use snarkvm::prelude::{AdditionalFee, Deployment, Execution, Field, Network, ProgramID, RecordsFilter, Transaction, ViewKey, U64};
+
 use anyhow::Result;
 use core::marker::PhantomData;
 use indexmap::IndexMap;
 use serde_json::json;
-use snarkvm::prelude::{AdditionalFee, Deployment, Execution, Field, Network, ProgramID, RecordsFilter, Transaction, ViewKey, U64};
 use std::sync::Arc;
 use tokio::{sync::mpsc, task::JoinHandle};
 use warp::{http::StatusCode, reject, reply, Filter, Rejection, Reply};
--- a/snarkos/ledger/server.rs
+++ b/snarkos/ledger/server.rs
@@ -351,17 +351,14 @@ impl<N: Network> Server<N> {
         }
     }
 
-    /// Broadcasts the transaction to the ledger.
+    /// Send a program deployment transaction to the ledger
     async fn deploy_program(
         deployment: Deployment<N>,
         ledger: Arc<Ledger<N>>,
         ledger_sender: LedgerSender<N>,
     ) -> Result<impl Reply, Rejection> {
         let additional_fee = Self::execute_additional_fee(ledger)?;
-
-        // Create the transaction.
         let transaction = Transaction::from_deployment(deployment.clone(), additional_fee).or_reject()?;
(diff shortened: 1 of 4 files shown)


snarkos/ledger/server.rs
```diff
@@ -196,14 +196,23 @@ impl<N: Network> Server<N> {
             .and(with(ledger_sender.clone()))
             .and_then(Self::transaction_broadcast);
 
-        // POST /testnet3/program/deploy
+        // POST /testnet3/deploy
         let deploy_program = warp::post()
             .and(warp::path!("testnet3" / "deploy"))
```
raychu86: Let's do "testnet3/program/deploy" and "testnet3/program/execute" when it's convenient to make the switch

snarkos/ledger/server.rs
```diff
@@ -341,17 +351,43 @@ impl<N: Network> Server<N> {
...
+    ) -> Result<impl Reply, Rejection> {
+        let additional_fee = Self::execute_additional_fee(ledger)?;
+        let transaction = Transaction::from_execution(execution.clone(), Some(additional_fee)).or_reject()?;
+        match ledger_sender.send(LedgerRequest::TransactionBroadcast(transaction)).await {
+            Ok(()) => Ok(reply::with_status(reply::json(&json!({ "execution": execution })), StatusCode::OK)),
+            Err(error) => Err(reject::custom(ServerError::Request(format!("{error}")))),
+        }
+    }
+
+    /// Spends the record with the smallest value with at least `1 gate` as the fee
```
raychu86: "as the fee for the execution of a transaction." can be shortened to "as a transaction fee"

snarkos/ledger/server.rs
facundoolano: sorry do you mean keeping it like this or aligning the endpoint url with the old comment?

raychu86: I think we should align it to the old comments.

So the deploy endpoint would be "testnet3/program/deploy" and the execute endpoint would be "testnet3/program/execute"

facundoolano: gotcha, I'll update and talk with the folks to adapt the clients of the endpoint accordingly

snarkos/ledger/server.rs
```diff
@@ -198,12 +198,21 @@ impl<N: Network> Server<N> {
 
         // POST /testnet3/program/deploy
         let deploy_program = warp::post()
-            .and(warp::path!("testnet3" / "deploy"))
+            .and(warp::path!("testnet3" / "program" / "deploy"))
+            .and(warp::body::content_length_limit(10 * 1024 * 1024))
```
ljedrz, line 210: do we know the size limit of programs? that would be the best fit here and in `/testnet3/program/execute`

snarkos/ledger/server.rs
facundoolano: updated

snarkos/ledger/server.rs
facundoolano: Good question, I actually haven't thought about this, just used the same number as the transaction broadcast and deploy endpoint

ljedrz: the same applies to the other types of data that the protocol constrains, including the transaction :+1:

facundoolano: @raychu86 @howardwu any input on those body lengths?

raychu86 (approved):

PR #1907, opened 2022-09-01
Add a get ciphertext endpoint

ilitteri:
## Motivation

This endpoint will be consumed in the `aleo execute` command workflow (https://github.com/AleoHQ/aleo/pull/332). The idea is for this command to be able to get the record to consume in a program execution out of the Record Commitment with the help of the user's View Key.

## Related PRs

This PR branched from #1900 and contains functionality that will be used in https://github.com/AleoHQ/aleo/pull/332.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,7 +50,7 @@ version = "2.0.2"
 
 [dependencies.snarkvm]
 git = "https://github.com/AleoHQ/snarkVM.git"
-rev = "bbc81d0"
+rev = "d6d2f80"
 features = ["circuit", "console", "parallel", "utilities"]
 
 [dependencies.aleo-std]
--- a/snarkos/ledger/server.rs
+++ b/snarkos/ledger/server.rs
@@ -174,13 +174,13 @@ impl<N: Network> Server<N> {
             .and(with(ledger.clone()))
             .and_then(Self::records_unspent);
 
-        // GET /testnet3/ciphertext/unspent/{commitment}
-        let get_unspent_ciphertext = warp::get()
+        // GET /testnet3/ciphertext/{commitment}
+        let get_ciphertext = warp::get()
             .and(warp::path!("testnet3" / "ciphertext" / ..))
             .and(warp::path::param::<Field<N>>())
             .and(warp::path::end())
             .and(with(ledger.clone()))
-            .and_then(Self::get_unspent_ciphertext);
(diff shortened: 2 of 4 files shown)


snarkos/ledger/server.rs
```diff
@@ -162,6 +162,16 @@ impl<N: Network> Server<N> {
             .and(with(ledger.clone()))
             .and_then(Self::records_unspent);
 
+        // GET /testnet3/ciphertext/unspent/{commitment}
+        let get_unspent_ciphertext = warp::get()
```
facundoolano: I'm thinking this shouldn't be its own entity separated from record, perhaps we can make it more uniform since they are different views of the same entity iiuc. we should try to avoid adding a specific endpoint for each particular client needs and favor more flexible general purpose endpoints (this way we can also find opportunities to improve the underlying apis and not bend too much to adjust to the current implementations).

for example:
/testnet3/records to get all records
/testnet3/records?spent=true for spent
/testnet3/records?spent=false for unspent
/testnet3/records/{commitment} for a single record by commitment
/testnet3/records/{commitment}?spent=false if we want the server to do the filtering part and return not found if it's doesn't exist or if it exist but is spent
/testnet3/records/{commitment}?spent=false&decrypt=false if we want the cyphertext

alternatively we could consider making the single records endpoint return both the decrypted data and the cyphertext as an attribute of the response payload, removing the need for a separate endpoint or a filter arg?

maybe the wording or the specific endpoints I'm proposing are not the exact one, but I think we need to try to find a solution in that general direction

facundoolano: @jrchatruc @raychu86 maybe you have input on this

Cargo.toml
```diff
@@ -50,7 +50,7 @@ version = "2.0.2"
 
 [dependencies.snarkvm]
 git = "https://github.com/AleoHQ/snarkVM.git"
-rev = "bbc81d0"
+rev = "d6d2f80"
```
ljedrz, line 53: the huge lockfile diff suggests that maybe another `snarkvm` dependency needs to be updated too?

ilitteri: I think so, I'll check this

ilitteri: I forgot to change the rev in snarkos-environment. Anyways I'll convert this PR to draft until https://github.com/AleoHQ/snarkVM/pull/1015 is merged.

PR #1910, opened 2022-09-02
Get deploy transaction

IAvecilla:
## Motivation

The user of the Aleo CLI tool needs to know the transaction ID associated with the deployment of their programs. This PR includes the transaction_id field on the deployment response to send it to the Aleo CLI tool and show it to the user.

### DIFF
--- a/snarkos/ledger/server.rs
+++ b/snarkos/ledger/server.rs
@@ -373,9 +373,10 @@ impl<N: Network> Server<N> {
     ) -> Result<impl Reply, Rejection> {
         let additional_fee = Self::execute_additional_fee(ledger)?;
         let transaction = Transaction::from_deployment(deployment.clone(), additional_fee).or_reject()?;
+        let transaction_id = transaction.id().to_string();
         match ledger_sender.send(LedgerRequest::TransactionBroadcast(transaction)).await {
             Ok(()) => Ok(reply::with_status(
-                reply::json(&json!({ "deployment": deployment })),
+                reply::json(&json!({ "deployment": deployment , "transaction_id": transaction_id})),
                 StatusCode::OK,
             )),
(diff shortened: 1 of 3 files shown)


snarkos/ledger/server.rs
```diff
@@ -373,9 +373,10 @@ impl<N: Network> Server<N> {
     ) -> Result<impl Reply, Rejection> {
         let additional_fee = Self::execute_additional_fee(ledger)?;
         let transaction = Transaction::from_deployment(deployment.clone(), additional_fee).or_reject()?;
+        let transaction_id = transaction.id().to_string();
         match ledger_sender.send(LedgerRequest::TransactionBroadcast(transaction)).await {
             Ok(()) => Ok(reply::with_status(
-                reply::json(&json!({ "deployment": deployment })),
+                reply::json(&json!({ "deployment": deployment , "transaction_id": transaction_id})),
```
howardwu, line 379: Instead of returning `{ deployment, transaction_id }`, can we return `{ transaction }` since it will already contain the `deployment` and the `ID` (in addition to all other details)?

facundoolano: This sounds good! And it actually makes me wonder, should we have a single endpoint for transactions instead of one for deployment and another for execution? we could distinguish by a type field in the payload if we did that. Perhaps that was the intention of the transaction broadcast endpoint?

jrchatruc: Yeah I agree with both these points, ultimately there should just be a `broadcast_transaction` endpoint, and the conversion from either `Execution` or `Deployment` to `Transaction`, plus the additional fee should be handled by the client.

IAvecilla:
These changes are deprecated with the new versions of the Aleo CLI, snarkOS, and snarkVM. If necessary, these changes will be introduced again in the future.

PR #1949, opened 2022-09-29
Reintroduce formal `Peers` module

raychu86:
## Motivation

This PR re-introduces the formal `Peer` and `Peers` modules from `testnet2`. This enables additional support for organizing and handling peers by separating the peer logic from the ledger.

### DIFF
--- a/snarkos/ledger/mod.rs
+++ b/snarkos/ledger/mod.rs
@@ -210,10 +210,9 @@ impl<N: Network, E: Environment> Ledger<N, E> {
                     let _ = router.send(());
                     // Asynchronously wait for a ledger request.
                     while let Some(request) = ledger_handler.recv().await {
-                        // Hold the ledger write lock briefly, to update the state of the ledger.
                         // Note: Do not wrap this call in a `task::spawn` as `BlockResponse` messages
                         // will end up being processed out of order.
-                        ledger.clone().update(request).await;
+                        ledger.update(request).await;
                     }
                 }),
             );
@@ -311,7 +310,7 @@ impl<N: Network, E: Environment> Ledger<N, E> {
     /// Performs the given `request` to the ledger.
     /// All requests must go through this `update`, so that a unified view is preserved.
     ///
-    pub(super) async fn update(self: Arc<Self>, request: LedgerRequest<N>) {
+    pub(super) async fn update(self: &Arc<Self>, request: LedgerRequest<N>) {
         match request {
             LedgerRequest::BlockResponse(peer_ip, block) => {
                 let block_height = block.height();
--- a/snarkos/ledger/mod.rs
+++ b/snarkos/ledger/mod.rs
@@ -385,12 +385,16 @@ impl<N: Network, E: Environment> Ledger<N, E> {
                 // If the peer is ahead, ask for next block.
                 let latest_height = self.ledger().read().latest_height();
                 if height > latest_height {
-                    if let Err(err) = self
-                        .peers_router()
-                        .send(PeersRequest::MessageSend(peer_ip, Message::<N>::BlockRequest(latest_height + 1)))
-                        .await
-                    {
-                        warn!("[BlockRequest] {}", err);
+                    println!("SENDING A FUCK TON OF THE SAME REQUEST:");
+
+                    for _ in 0..200 {
+                        if let Err(err) = self
+                            .peers_router()
+                            .send(PeersRequest::MessageSend(peer_ip, Message::<N>::BlockRequest(latest_height + 1)))
+                            .await
+                        {
+                            warn!("[BlockRequest] {}", err);
+                        }
                     }
                 }
             }
--- a/snarkos/ledger/mod.rs
+++ b/snarkos/ledger/mod.rs
@@ -14,17 +14,14 @@
 // You should have received a copy of the GNU General Public License
 // along with the snarkOS library. If not, see <https://www.gnu.org/licenses/>.
 
-use crate::{handle_dispatch_error, BlockDB, Data, ProgramDB};
+use crate::{environment::Environment, handle_dispatch_error, BlockDB, Data, Message, Peers, PeersRequest, ProgramDB};
 use snarkvm::prelude::*;
 
 use colored::Colorize;
 use futures::StreamExt;
 use indexmap::IndexMap;
 use parking_lot::RwLock;
-use std::{
-    net::{IpAddr, SocketAddr},
-    sync::Arc,
-};
+use std::{net::IpAddr, sync::Arc};
 use tokio::task;
 use warp::{reply, Filter, Rejection, Reply};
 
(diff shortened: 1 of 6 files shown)


snarkos/ledger/mod.rs
```diff
@@ -204,8 +306,215 @@ impl<N: Network> Ledger<N> {
...
+                }
+            }
+            LedgerRequest::Pong(peer_ip, height, node_type) => {
+                // Ensure the peer has been initialized in the ledger.
+                self.initialize_peer(peer_ip, node_type).await;
+
+                // If the peer is ahead, ask for next block.
+                let latest_height = self.ledger().read().latest_height();
+                if height > latest_height {
+                    println!("SENDING A FUCK TON OF THE SAME REQUEST:");
```
niklaslong:
```suggestion
```
Development log, should be safe to remove now.

snarkos/ledger/mod.rs
```diff
@@ -25,36 +37,81 @@ use std::{
...
 
 pub(crate) type InternalServer<N> = snarkvm::prelude::Server<N, BlockDB<N>, ProgramDB<N>>;
 // pub(crate) type InternalServer<N> = snarkvm::prelude::Server<N, BlockMemory<N>, ProgramMemory<N>>;
 
-pub(crate) type Peers<N> = Arc<RwLock<IndexMap<SocketAddr, crate::Sender<N>>>>;
+pub type PeersState = IndexMap<SocketAddr, Option<NodeType>>;
 
 #[allow(dead_code)]
-pub struct Ledger<N: Network> {
+pub struct Ledger<N: Network, E: Environment> {
```
niklaslong, line 86: We may want to consider not routing through the ledger in future. This would reduce complexity. I'll suggest something alongside the upcoming kadmium + noise additions.

snarkos/ledger/mod.rs
```diff
@@ -25,36 +37,81 @@ use std::{
...
 pub(crate) type InternalLedger<N> = snarkvm::prelude::Ledger<N, BlockDB<N>, ProgramDB<N>>;
 // pub(crate) type InternalLedger<N> = snarkvm::prelude::Ledger<N, BlockMemory<N>, ProgramMemory<N>>;
 
 pub(crate) type InternalServer<N> = snarkvm::prelude::Server<N, BlockDB<N>, ProgramDB<N>>;
 // pub(crate) type InternalServer<N> = snarkvm::prelude::Server<N, BlockMemory<N>, ProgramMemory<N>>;
 
-pub(crate) type Peers<N> = Arc<RwLock<IndexMap<SocketAddr, crate::Sender<N>>>>;
+pub type PeersState = IndexMap<SocketAddr, Option<NodeType>>;
 
 #[allow(dead_code)]
```
ljedrz, line 85: unrelated to this PR, but this is probably not needed anymore?

snarkos/ledger/mod.rs
```diff
@@ -25,36 +37,81 @@ use std::{
...
+    Heartbeat,
+    /// Ping := (peer_ip)
+    Ping(SocketAddr),
+    /// Pong := (peer_ip, block_height, node_type)
+    Pong(SocketAddr, u32, Option<NodeType>),
+    /// UnconfirmedBlock := (peer_ip, block)
+    UnconfirmedBlock(SocketAddr, Block<N>),
+    /// UnconfirmedTransaction := (peer_ip, transaction)
+    UnconfirmedTransaction(SocketAddr, Transaction<N>),
+}
```
ljedrz, line 75: semantics: some of these variants are not _technically_ related to the `Ledger`, which is somewhat confusing; a `LedgerRequest` would ideally only contain `BlockRequest`, `BlockResponse`, `UnconfirmedBlock`, `UnconfirmedTransaction` at this point

snarkos/ledger/mod.rs
```diff
@@ -25,36 +37,81 @@ use std::{
...
+
 pub(crate) type InternalLedger<N> = snarkvm::prelude::Ledger<N, BlockDB<N>, ProgramDB<N>>;
 // pub(crate) type InternalLedger<N> = snarkvm::prelude::Ledger<N, BlockMemory<N>, ProgramMemory<N>>;
 
 pub(crate) type InternalServer<N> = snarkvm::prelude::Server<N, BlockDB<N>, ProgramDB<N>>;
 // pub(crate) type InternalServer<N> = snarkvm::prelude::Server<N, BlockMemory<N>, ProgramMemory<N>>;
 
-pub(crate) type Peers<N> = Arc<RwLock<IndexMap<SocketAddr, crate::Sender<N>>>>;
+pub type PeersState = IndexMap<SocketAddr, Option<NodeType>>;
```
ljedrz, line 83: the peers will likely have a lot more state related to them in the future, so instead of `Option<NodeType>` there should probably be a non-optional `struct PeerState` (or `Peer`, etc.) that only currently holds the `NodeType`

snarkos/ledger/mod.rs
```diff
@@ -25,36 +37,81 @@ use std::{
...
     /// The account view key.
     view_key: ViewKey<N>,
     /// The account address.
     address: Address<N>,
+    /// The map of each peer to their ledger state := node_type.
+    peers_state: RwLock<PeersState>,
+    /// The peers router of the node.
+    peers: Arc<Peers<N, E>>,
+    /// The map of each peer to their failure messages := (failure_message, timestamp).
+    failures: RwLock<IndexMap<SocketAddr, Vec<(String, i64)>>>,
```
ljedrz, line 104: I'm wondering if we need to separate `failures` from `peers_state`; is this split for potential extra performance?

snarkos/ledger/mod.rs
```diff
@@ -112,13 +166,15 @@ impl<N: Network> Ledger<N> {
...
+            async fn get_peers_count<N: Network, E: Environment>(peers: Arc<Peers<N, E>>) -> Result<impl Reply, Rejection> {
+                Ok(reply::json(&peers.number_of_connected_peers().await))
             }
 
             /// Returns the peers connected to the node.
-            async fn get_peers_all<N: Network>(peers: Peers<N>) -> Result<impl Reply, Rejection> {
-                Ok(reply::json(&peers.read().keys().map(|addr| addr.ip()).collect::<Vec<IpAddr>>()))
+            async fn get_peers_all<N: Network, E: Environment>(peers: Arc<Peers<N, E>>) -> Result<impl Reply, Rejection> {
+                Ok(reply::json(
+                    &peers.connected_peers().await.iter().map(|addr| addr.ip()).collect::<Vec<IpAddr>>(),
```
ljedrz, line 176: shouldn't we return the ports too? in the previous testnets there were many non-standard ones

snarkos/ledger/mod.rs
```diff
@@ -127,15 +183,46 @@ impl<N: Network> Ledger<N> {
...
+            let ledger = ledger.clone();
+            let (router, handler) = oneshot::channel();
+            E::resources().register_task(
+                None,
+                task::spawn(async move {
+                    // Notify the outer function that the task is ready.
+                    let _ = router.send(());
+                    // Asynchronously wait for a ledger request.
+                    while let Some(request) = ledger_handler.recv().await {
+                        // Hold the ledger write lock briefly, to update the state of the ledger.
```
```suggestion
                        // Hold the ledger write lock briefly, to update the state of the ledger.
```
there is no lock

snarkos/ledger/mod.rs
```diff
@@ -204,8 +306,215 @@ impl<N: Network> Ledger<N> {
     }
 }
 
+impl<N: Network, E: Environment> Ledger<N, E> {
+    ///
+    /// Performs the given `request` to the ledger.
+    /// All requests must go through this `update`, so that a unified view is preserved.
+    ///
+    pub(super) async fn update(self: Arc<Self>, request: LedgerRequest<N>) {
```
```suggestion
    pub(super) async fn update(self: &Arc<Self>, request: LedgerRequest<N>) {
```

snarkos/ledger/mod.rs
```diff
@@ -127,15 +183,46 @@ impl<N: Network> Ledger<N> {
...
+                None,
+                task::spawn(async move {
+                    // Notify the outer function that the task is ready.
+                    let _ = router.send(());
+                    // Asynchronously wait for a ledger request.
+                    while let Some(request) = ledger_handler.recv().await {
+                        // Hold the ledger write lock briefly, to update the state of the ledger.
+                        // Note: Do not wrap this call in a `task::spawn` as `BlockResponse` messages
+                        // will end up being processed out of order.
+                        ledger.clone().update(request).await;
```
```suggestion
                        ledger.update(request).await;
```

snarkos/network/peers.rs
```diff
@@ -0,0 +1,377 @@
...
+}
+
+///
+/// A list of peers connected to the node server.
+///
+pub struct Peers<N: Network, E: Environment> {
+    /// The peers router of the node.
+    peers_router: PeersRouter<N>,
+    /// The local address of this node.
+    local_ip: SocketAddr,
```
ljedrz, line 75: this should be `Node`-level

snarkos/ledger/mod.rs
```diff
@@ -25,36 +37,81 @@ use std::{
...
-    /// The peers.
-    peers: Peers<N>,
     /// The account private key.
     private_key: PrivateKey<N>,
     /// The account view key.
     view_key: ViewKey<N>,
     /// The account address.
     address: Address<N>,
+    /// The map of each peer to their ledger state := node_type.
+    peers_state: RwLock<PeersState>,
```
ljedrz, line 100: this isn't `Ledger`-related; in `testnet2` we had a `State` (or `NodeState`) struct with different types of state (the ledger, peers etc.) separated

snarkos/ledger/mod.rs
raychu86: At the moment, we don't sanitize the listener ports (since we do not have a formal handshake atm), so the listener ports are the outbound ports of that peer - which usually only exists for that particular connection.

snarkos/ledger/mod.rs
raychu86: The old testnet2 model had failures in the `ledger`, but i believe it should just exist in `Peers`

PR #1954, opened 2022-10-03
Don't keep recreating the reqwest Client while syncing

### LINKED ISSUE #1953: [Feature] Reuse HTTPS connection when syncing the ledger
## 🚀 Feature

Use a proper HTTP connection pool for HTTP syncing phase.

## Motivation

Currently the HTTP syncing code opens 100 concurrent HTTPS connections for syncing, and immediately disconnects after one request is complete. This is a waste of resources as the TCP and TLS handshake takes many round-trips to complete and the syncing code is actually blocking (blocks are added by order, obviously).

EDIT: maybe the most important concern is the server load here as the networking time *could* be negligible when syncing the ledger.

## Implementation

If network performance is a concern, maybe open up to `num_cpus::get()` HTTP connections and do all the requests using existing connections, instead of repeatedly opening new connections.

Only `Ledger::initial_sync_with_network` would be affected.

**Are you willing to open a pull request?** (See [CONTRIBUTING](../../CONTRIBUTING.md))
If the http syncing will stay and the feature will be accepted. I'm not sure what's the good pattern to implement it in Rust though. `futures::stream::iter` seems good enough for a proper queuing of the incoming block data though.

ljedrz:
Fixes https://github.com/AleoHQ/snarkOS/issues/1953.

### DIFF
--- a/snarkos/ledger/mod.rs
+++ b/snarkos/ledger/mod.rs
@@ -293,8 +293,13 @@ impl<N: Network> Ledger<N> {
         // Fetch the ledger height.
         let ledger_height = self.ledger.read().latest_height();
 
+        // Create a Client to maintain a connection pool throughout the sync.
+        let client = reqwest::Client::builder().build()?;
+
         // Fetch the latest height.
-        let latest_height = reqwest::get(format!("http://{leader_ip}/testnet3/latest/height"))
+        let latest_height = client
+            .get(format!("http://{leader_ip}/testnet3/latest/height"))
+            .send()
             .await?
             .text()
             .await?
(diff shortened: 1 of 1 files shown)


snarkos/ledger/mod.rs
```diff
@@ -310,17 +315,21 @@ impl<N: Network> Ledger<N> {
                     trace!("Requesting block {height} of {latest_height}");
 
                     // Download the block with an exponential backoff retry policy.
-                    handle_dispatch_error(move || async move {
-                        // Get the URL for the block download.
-                        let block_url = format!("{TARGET_URL}{height}.block");
+                    let client_clone = client.clone();
```
ljedrz, line 318: this clone is a bit annoying (wasn't able to get around it), but it's only an `Arc` bump

howardwu (approved):

PR #1965, opened 2022-10-10
Add a connection memleak test

ljedrz:
It doesn't seem like we have any leaks there at the moment, but it would be a good idea to preserve this test to ensure this with any future changes to peering.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1580,6 +1580,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b1de2e551fb905ac83f73f7aedf2f0cb4a0da7e35efa24a202a936269f1f18e1"
 
 [[package]]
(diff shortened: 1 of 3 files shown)


ljedrz:
Ah yeah, the test points to in-memory storage; should I make the (snarkOS) ledger more generic, or disable the test more thoroughly?

Zerohedge2020:
> Ah yeah, the test points to in-memory storage; should I make the (snarkOS) ledger more generic, or disable the test more thoroughly?

how thats gonna build up the generic storage though?

ljedrz:
> how thats gonna build up the generic storage though?

It's not for building generic storage; a single type of storage would still be used in the test.

ljedrz:
Closing, as it's no longer applicable; an up-to-date rendition of this test will be filed soon.

PR #1981, opened 2022-10-20
[Fix bug: issue 1926] Check http status before pass http rsp to

NoPointExc:
## Motivation

Trying to fix this bug here:
https://github.com/AleoHQ/snarkOS/issues/1926

## Test Plan

new to this project and I haven't found any dev guide in this repo.

Any suggestions on how could I build & test it locally?

## Related PRs
No


ljedrz:
While this would provide a nicer early return, it would probably be more practical to use `handle_dispatch_error` here instead, in order to not give up after a single failed attempt.

In addition, note that @raychu86 has been planning to eventually change this `reqwest` call to `Message::BlockRequest`, hence the lack of nice error handling there.

NoPointExc:
Oh. I see. Let's abandon this pul request and wait for the TODO then.

PR #1993, opened 2022-10-23
Migration of the ledger and REST server

howardwu:
## Motivation

This PR completes the migration request from https://github.com/AleoHQ/snarkVM/issues/1051.

Its counterpart PR is in https://github.com/AleoHQ/snarkVM/pull/1204.


howardwu:
Early merging as this PR is being combined into `feat/modular-nodes`.

PR #1998, opened 2022-10-28
Introduce a consolidated SnarkOS, Noise and Kadmium codec

niklaslong:
This PR introduces a consolidated SnarkOS, Noise and Kadmium codec with some sanity check tests for snarkOS using the `Noise_XX_25519_ChaChaPoly_BLAKE2s` handshake. If messages exceed the max length allowed by noise, they are broken into chunks encoded and decoded in parallel. 

~~@howardwu it would be most helpful to make all message types `PartialEq` and `Eq`. This would make testing much easier. I've already started this process with messages that don't depend on SnarkVM types but this unfortunately a soft blocker for further testing in that it will limit our coverage. We'll also be able to condense the existing tests once this is in.~~ Fixed in this PR.


howardwu (approved):
LGTM

PR #2000, opened 2022-10-31
Fix(cli): increase compilation recursion limit to `256` 

niklaslong:
This fixes the compilation issue with the cli crate. I understand this crate may be rearchitected; we should consider lowering this limit, if possible, in future.


niklaslong:
This seems to fix the issue on CI as well.

PR #2007, opened 2022-11-02
No need to `Arc` `SocketAddr`, it's `Copy`

niklaslong:
Quick fix for something I found whilst perusing the current network stack.


howardwu (approved):
LGTM

PR #2008, opened 2022-11-02
Consistent genesis block in storage

raychu86:
## Motivation

This PR adds a check (for each non-development node) that the genesis block in storage matches the one defined in the `Network`. In the case that the genesis block is mismatched, the ledger will be cleared.


howardwu (approved):
LGTM

PR #2009, opened 2022-11-03
Use Jemalloc and blocking tasks for expensive operations

raychu86:
## Motivation

This PR uses blocking tasks for `add_unconfirmed_transaction` and block production. Additionally, Jemalloc is set as the `GLOBAL` memory allocator.

### DIFF
--- a/node/src/beacon/router.rs
+++ b/node/src/beacon/router.rs
@@ -76,9 +76,17 @@ impl<N: Network> Inbound<N> for Beacon<N> {
             trace!("Skipping 'UnconfirmedSolution' from '{peer_ip}'");
         } else {
             // Add the unconfirmed solution to the memory pool.
-            if let Err(error) = self.ledger.consensus().write().add_unconfirmed_solution(&solution) {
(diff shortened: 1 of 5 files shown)


node/src/beacon/router.rs
```diff
@@ -76,9 +76,17 @@ impl<N: Network> Inbound<N> for Beacon<N> {
...
+                Some(mut ledger) => {
+                    if let Err(error) = ledger.add_unconfirmed_solution(&solution) {
+                        trace!("[UnconfirmedSolution] {error}");
+                        return true; // Maintain the connection.
+                    }
+                }
+                None => {
+                    trace!("Ignoring an unconfirmed solution (already creating a block).");
+                    return true;
+                }
```
raychu86, line 89: The unconfirmed solutions are valid over an epoch (many blocks), so we shouldn't be ignoring them here. They are still valid as long as you are within the epoch block range (currently 256 blocks).

niklaslong (approved):

ljedrz:
Closing, as we already use `jemalloc`, and blocking tasks were proposed separately.

PR #2017, opened 2022-11-04
Add `MessagePropagateBeacon` to `RouterRequest`

raychu86:
## Motivation

This PR adds support for message propagation to connected beacons (as opposed to every connected peer). 

The beacons will now only propagate unconfirmed solutions to other beacons, which will reduce the number of messages on the network.


howardwu (approved):
LGTM

PR #2020, opened 2022-11-05
Add JWT authorization to record endpoints

raychu86:
## Motivation

This PR implements JWT authorization for the REST api endpoints related to records. 

On node startup, a JWT secret key is sampled and a new JWT token used to authenticated requests is displayed.


howardwu (approved):
LGTM

PR #2022, opened 2022-11-07
perf: optimize DataMap::get_batched

zvolin:
This reduces the cost of `DataMap::get_batched` from `O(n) + O(deserialise)` to `O(1)`.

The cost of whole `start_atomic` + insertions + removals + `finish_atomic` combined should stay similar to the one before the change


niklaslong (approved):

howardwu (approved):
LGTM

PR #2034, opened 2022-11-08
Introduce `last_coinbase_target`

raychu86:
## Motivation

This PR updates the `coinbase_target` calculation to use `last_coinbase_target` instead of the "compounding" effect of using `latest_coinbase_target`.

The `last_coinbase_target` is found by searching the latest blocks for the `last_coinbase_timestamp`.

 An alternative solution is to store the `last_coinbase_target` alongside the `last_coinbase_timestamp` in the block header metadata.


raychu86:
Closed in favor of https://github.com/AleoHQ/snarkOS/pull/2036

PR #2036, opened 2022-11-09
Use `last_coinbase_target` for `coinbase_target` calculation

raychu86:
## Motivation

This PR updates the calculation of `coinbase_targets` using the `last_coinbase_target` attribute in the block header `Metadata`. 

Once https://github.com/AleoHQ/snarkVM/pull/1253 has been included in a new release, the snarkVM dependency here should be updated.


howardwu (approved):
LGTM

PR #2052, opened 2022-11-11
Optimize the generation of `ProverSolution`s 

raychu86:
## Motivation

This PR optimizes the solution generation for provers. The new protocol eliminates unnecessary work by halting production of the prover solution early if the minimum `proof_target` has not been met. 

This will greatly increase the number of proofs a prover can generate per second:
 - Generating a proof that fails to meet the minimum proof target takes approximately half the time it did before.
 - More details in the snarkVM PR listed below.

This PR currently relies on https://github.com/AleoHQ/snarkVM/pull/1259, and the `Cargo.toml` will need to be updated pending a new release.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -45,11 +45,11 @@ dependencies = [
 
 [[package]]
 name = "ahash"
-version = "0.7.6"
+version = "0.8.2"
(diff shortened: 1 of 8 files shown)


howardwu (approved):
LGTM

PR #2062, opened 2022-11-14
Fix: add simple disconnect logic

niklaslong:
Opening this tentative PR, this logic is untested but _should_ trigger a disconnect when a `Disconnect` message is sent.

### DIFF
--- a/node/router/src/lib.rs
+++ b/node/router/src/lib.rs
@@ -474,6 +474,8 @@ impl<N: Network> Router<N> {
                 self.connected_peers.write().await.remove(&peer_ip);
                 // Add an entry for this `Peer` in the candidate peers.
                 self.candidate_peers.write().await.insert(peer_ip);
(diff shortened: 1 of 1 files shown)


zvolin (approved):
LGTM but I'm also partially an author

node/router/src/lib.rs
```diff
@@ -474,6 +474,8 @@ impl<N: Network> Router<N> {
                 self.connected_peers.write().await.remove(&peer_ip);
                 // Add an entry for this `Peer` in the candidate peers.
                 self.candidate_peers.write().await.insert(peer_ip);
+                // Send a disconnect message to the peer.
+                self.handle_send(peer_ip, Message::Disconnect(DisconnectReason::NoReasonGiven.into())).await;
```
howardwu: Why are we attempting to send a disconnect right after we deque'd them from the connected peer list? This message will never be sent.

howardwu: I've removed this logic.

howardwu (approved):

PR #2066, opened 2022-11-15
Add warning in case open files limit is lower than recommended

zvolin:
Warn user if limit of open files for node's process is lower than recommended on unix systems.

image


cli/src/commands/start.rs
```diff
@@ -27,6 +27,19 @@ use rand_chacha::ChaChaRng;
...
+/// Beacon should be able to handle at least 1000 concurrent connections,
+/// each requiring 2 sockets.
+#[cfg(target_family = "unix")]
+const RECOMMENDED_MIN_NOFILES_LIMIT_BEACON: u64 = 2048;
+/// Recommended minimum value of 'open files' limit for validator node.
+///
+/// Beacon should be able to handle at least 500 concurrent connections,
+/// each requiring 2 sockets.
+#[cfg(target_family = "unix")]
+const RECOMMENDED_MIN_NOFILES_LIMIT_VALIDATOR: u64 = 1024;
```
zvolin: @howardwu @niklaslong please help choosing correct values for those constants :)

niklaslong (approved):
LGTM, though I'll defer to @howardwu regarding those limits.

cli/Cargo.toml
```diff
@@ -30,6 +30,9 @@ features = ["derive"]
 [dependencies.colored]
 version = "2"
 
+[dependencies.nix]
```
howardwu: We should cfg guard this dependency

zvolin: good catch! I've just done it

just a note, as all checks have passed then maybe it would be good to add windows build to the ci?

howardwu (approved):
LGTM

howardwu:
Verified CI failure is from an unrelated issue (likely outage)

Merging

PR #2067, opened 2022-11-15
Change HTTP GET Requests with body to HTTP POST Requests

evanmarshall:
## Motivation

We're unable to fetch records from any JS client because HTTP GET Requests should not have a body. Any request that requires a body should be a HTTP POST.

This fixes one of the issues in: https://github.com/AleoHQ/snarkOS/issues/2010

Specifically, this fixes the REST server for JS-based clients: https://xhr.spec.whatwg.org/#the-send()-method


howardwu:
Closing as per #2071

PR #2071, opened 2022-11-17
Remove records endpoints to prevent view key leakage

howardwu:
## Motivation

This PR removes the `/testnet3/records/*` endpoints, to **prevent account view key leakage**.

## Problems

In most cases, **client-side** software should be trial decrypting records by tracking the last block height that the software checked for, and merely scan the delta since the last block height for new records. The current REST endpoints do not protect this user-journey as the endpoints can easily be made accessible by third-parties.

For some software, an initial one-time sync (from genesis to tip) may be desirable to ensure the history of the account has been retrieved. In these instances, users should either 1) trial decrypt client-side, or 2) delegate the request to a third-party in a secure enclave if performance is required. The current REST endpoints are unable to enforce either of these design goals transparently for the user.

Lastly, for some software, they require merely a subset of records that belong to the account. For example, a program may only care about records produced and consumed by the program itself, or a program may only desire records pertaining to a subset of known programs that are already encoded in the program. Again, in this instance, the current REST endpoints are unable to offer filtering with these requirements.

## Solutions

The recommended approach to detecting records is from **client-side** software to fetch blocks and perform trial decryption **client-side**. This can be done by using the `GET blocks` endpoint to fetch blocks, and subsequently check `block.records().filter(|record| record.is_owner(view_key))` on the **client-side** in order to prevent node operators and third-parties from learning user account data.

In order to support the other capabilities that are described as missing in the problems section, Aleo is exploring the development of a secure enclave service to offer users the ability to perform performant scanning. In addition, future work to enable better filtering capabilities will be contemplated as part of the design of state in snarkVM.

### DIFF
--- a/node/rest/src/routes.rs
+++ b/node/rest/src/routes.rs
@@ -177,36 +177,6 @@ impl<N: Network, C: ConsensusStorage<N>> Rest<N, C> {
             .and(with(self.ledger.clone()))
             .and_then(Self::find_transition_id);
 
-        // GET /testnet3/records/all
-        let records_all = warp::get()
-            .and(warp::path!("testnet3" / "records" / "all"))
-            .and(with_auth())
-            .untuple_one()
-            .and(warp::body::content_length_limit(128))
-            .and(warp::body::json())
-            .and(with(self.ledger.clone()))
-            .and_then(Self::records_all);
-
-        // GET /testnet3/records/spent
-        let records_spent = warp::get()
-            .and(warp::path!("testnet3" / "records" / "spent"))
-            .and(with_auth())
-            .untuple_one()
-            .and(warp::body::content_length_limit(128))
-            .and(warp::body::json())
-            .and(with(self.ledger.clone()))
-            .and_then(Self::records_spent);
-
-        // GET /testnet3/records/unspent
-        let records_unspent = warp::get()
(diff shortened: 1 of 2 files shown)


raychu86 (approved):

d0cd (approved):

node/rest/src/routes.rs
```diff
@@ -177,36 +177,6 @@ impl<N: Network, C: ConsensusStorage<N>> Rest<N, C> {
             .and(with(self.ledger.clone()))
             .and_then(Self::find_transition_id);
 
-        // GET /testnet3/records/all
-        let records_all = warp::get()
-            .and(warp::path!("testnet3" / "records" / "all"))
-            .and(with_auth())
```
evanmarshall, line 183: I believe these endpoints were the only ones using the `with_auth` from https://github.com/AleoHQ/snarkOS/blob/testnet3/node/rest/src/helpers/auth.rs . That file as well as the code here: https://github.com/AleoHQ/snarkOS/blob/testnet3/node/rest/src/lib.rs#L118 can probably be deleted.

howardwu: Agreed, we discussed and decided to keep for guarding the existing REST endpoints prior to mainnet.

PR #2078, opened 2022-11-18
[Fix #2068] Allow `$snarkos account new --seed` takes 256-bit numbers as input

### LINKED ISSUE #2068: [Bug] `--seed` parameter can't take 256-bit numbers
## 🐛 Bug Report

If you're offloading randomness during key generation then you might use `account new --seed X`. However, this X must be a decimal number and trying to give it a 256-bit decimal number produces an error.

```shell
seed=$(head --bytes 32 < /dev/urandom | xxd -p | tr -d '\n' | tr '[:lower:]' '[:upper:]')
snarkos account new --seed "$(echo "obase=10;ibase=16;$seed" | bc | tr -d '\\\n')" 
error: Invalid value "38868010450269069756484274649022187108349082664538872491798902858296683054657" for '--seed <SEED>': number too large to fit in target type
```

Ideally `--seed` should just take arbitrary hex-encoded data instead.

Using a smaller seed space works but loses a great deal of security.

NoPointExc:
## Motivation
Taking 32 bytes as input as suggested by @Zwo1in in #2068

## Test Plan
Build and run the cli with different input and now it works for 256bits input.
without this fix:
```
$ snarkos account new -s 388680104502690697564842746490221871083490826645388
error: Invalid value "388680104502690697564842746490221871083490826645388" for '--seed <SEED>': number too large to fit in target type

For more information try --help
```
with this fix:
```
$ ./snarkos account new -s 388680104502690697564842746490221871083490826645388

  Private Key  APrivateKey1zkp7E9K5symeXrs8g7jCm83zoAWwoaZiHYqtp1SsZoPTUqp
     View Key  AViewKey1mG4NQ4e923LA6GBjE6Zb3DcDJi8pcPTQWjPq24HLuHAf
      Address  aleo1dwr53y9n89exe2lsn8xrcpxgzqaf88wejfngtm0uhkljmutlccxqpxqhwd

$./snarkos account new -s 3

  Private Key  APrivateKey1zkpJcCBBJ8fPUUHg1sVcLVH9LFpNf69XZW5E8CoehP6jgZv
     View Key  AViewKey1sQXGxFFSacrjvL5mQZAg8o3tt9VrjHsdCoB6i3EM5np9
      Address  aleo1cq0xhdyrndccrugsp4dr8v6mg4ma77hce9qju8a2u284tfuggvqqv9vqfy
```

## Related PRs
No


cli/Cargo.toml
```diff
@@ -75,3 +75,5 @@ version = "1.0"
 version = "1.21"
 features = ["rt"]
 
+[dependencies.num-bigint]
+version = "0.4.3"
```
zvolin: please keep deps sorted

cli/src/commands/account.rs
```diff
@@ -30,7 +30,7 @@ pub enum Account {
     New {
         /// Seed the RNG with a numeric value
         #[clap(short = 's', long)]
-        seed: Option<u64>,
+        seed: Option<BigInt>,
```
zvolin: I'm wondering whether this shouldn't be BigUint

cli/src/commands/account.rs
```diff
@@ -40,7 +40,14 @@ impl Account {
             Self::New { seed } => {
                 // Sample a new Aleo account.
                 let account = snarkos_account::Account::from(match seed {
-                    Some(seed) => PrivateKey::<Network>::new(&mut ChaChaRng::seed_from_u64(seed))?,
+                    Some(bigint) => {
+                        let mut bytes = bigint.to_bytes_be().1;
+                        bytes.truncate(32);
+                        bytes.resize(32, 0);
```
zvolin: This modifies the original value passed as an argument for numbers taking less amount of bytes than 32.
Since `Vec::resize` is padding with zeros at the end, eg. `[1]` will become `[1, 0, 0, 0, 0, 0, ...]`.

```rust
    let num = 1231275789u64;
    let mut bytes = num.to_be_bytes().to_vec();
    bytes.truncate(32);
    bytes.resize(32, 0);

    assert_eq!(
        num.to_bigint().unwrap(),
        BigInt::from_bytes_be(Sign::Plus, &bytes)
    );
```

```
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `1231275789`,
 right: `7728843391871504577583034288497874377335230004640899776989253074944`', src/main.rs:8:5
```

cli/src/commands/account.rs
```diff
@@ -65,21 +74,45 @@ mod tests {
...
+        expected += &format!(
+            " {:>12}  {}\n",
+            "View Key".cyan().bold(),
+            "AViewKey1cWek62tM1kaHqDqwFBK6dZwYHaKeN8DZ6A82v9AEJiCj"
+        );
+        expected += &format!(
+            " {:>12}  {}",
+            "Address".cyan().bold(),
+            "aleo17h3hthqzctgc5s847pw7c4zmxqce8dxzsmwr3tp7umxcegl9vgrqz2khpg"
+        );
```
zvolin: The goal is to keep this test passing with original values

cli/Cargo.toml
```diff
@@ -75,3 +75,5 @@ version = "1.0"
 version = "1.21"
 features = ["rt"]
 
+[dependencies.num-bigint]
```
zvolin: @howardwu please check if you don't mind an extra dependency

zvolin (dismissed):
Thanks for submitting a PR :) Please try to make algorithm not altering the number's value, otherwise LGTM

cli/src/commands/account.rs
zvolin: `le` instead of `be` would probably work

cli/src/commands/account.rs
```diff
@@ -40,7 +40,14 @@ impl Account {
             Self::New { seed } => {
                 // Sample a new Aleo account.
                 let account = snarkos_account::Account::from(match seed {
-                    Some(seed) => PrivateKey::<Network>::new(&mut ChaChaRng::seed_from_u64(seed))?,
+                    Some(bigint) => {
+                        let mut bytes = bigint.to_bytes_be().1;
+                        bytes.truncate(32);
```
zvolin: From `resize` [docs](https://doc.rust-lang.org/stable/std/vec/struct.Vec.html#method.resize) it seems it will handle truncation by itself. However overflowing the 32bytes isn't likely to be what user wanted to achieve so I would drive towards returning `Err` there if the value provided is bigger than 32bytes. Maybe with `anyhow::ensure!`

cli/Cargo.toml
NoPointExc: Thank you! Do you have any linter recommendations? E.g is there any vscode/vim plugin?

zvolin: So, personally for tomls I use taplo-ls with neovim + nvim-lsp. However, it doesn't sort dependencies, moreover this repo doesn't use it for autoformat so I usually disable it for saving files here :) as per this comment, just move the dep some lines above and it'll be okay

cli/Cargo.toml
howardwu: Thanks for checking. We technically use `num-bigint` to impl `Display` in `snarkvm::utilities::BigInteger256/384`,  so the extra dependency isn't a big deal.

cli/src/commands/account.rs
howardwu: Good catch, this should be `num_bigint::BigUint`

cli/src/commands/account.rs
howardwu: +1 on using `_le` to be consistent with the rest of our stack.

howardwu:
I have updated the logic for recovering from a seed to use `Field::from_str`, which (in `snarkvm::fields::Fp256`) automatically checks that the recovered value is below `P::MODULUS`. Thus the provided seed value is guaranteed to be within our field size.

This approach will also ensure the seed is recovered in little-endian format.

In addition, the `seed` passed in via CLI is now a `String` type, to support field-based seeds.

I have also updated the tests to pin the new expected private key, view key, and address.

With that, this should complete the necessary changes.

howardwu (approved):
LGTM

PR #2082, opened 2022-11-19
Adds TCP stack abstraction

howardwu:
## Motivation

- Adds TCP stack abstraction
- Removes `Executor` module
- Decouples `Router` peers and messages
- Adds `Cache` struct, removes `gc`
- Adds support for verbose `4`


HarukaMa:
Seems the node is not checking if the initiator's listening port is open - is this temporary or permanent?

niklaslong:
@HarukaMa, it will likely be permanent (I'm assuming you mean in the handshake). The rationale is tied to many factors but the tl;dr is this check shouldn't be happening in the handshake (as we're wastefully opening connections) but the peering protocol and the network as a whole should be designed in such a way that there is no incentive to run with the listening port closed (i.e. as they will suffer from poor connectivity and be scored lower by the network).

PR #2094, opened 2022-11-24
Test: handshake with all node combinations (in non-dev mode)

niklaslong:
This PR introduces full coverage for handshake tests. ~~Drafted for now as the validator/beacon pair currently fails, to be investigated.~~ Cases fail because of the phase 2 beacon-related handshake logic. Cases pass if run in dev mode, as expected. 

Test output can be viewed [here](https://app.circleci.com/pipelines/github/AleoHQ/snarkOS/7180/workflows/0eacf84c-809c-4c5b-9d95-08b7a397ca0c/jobs/41042/parallel-runs/0/steps/0-106).


howardwu (approved):
LGTM

PR #2103, opened 2022-11-28
Only log error if sync made progress

howardwu:
## Motivation

Fixes a print log error that could unintentionally mislead the caller.


niklaslong (approved):
LGTM 👍

PR #2107, opened 2022-11-29
Update half life calculation for `coinbase_target`

raychu86:
## Motivation

This PR update the half life selection for the `coinbase_target` re-targeting algorithm. Additionally, a conditional has been added to Consensus to account for the old algorithm when validating previous blocks.


howardwu (approved):
LGTM

PR #2089, opened 2022-11-23
Bumps version from `3` to `4`: Updates message types, adds sync pool

howardwu:
## Motivation

This PR adds support for validators to sync.

**Attention** - This PR performs a version bump of snarkOS. All nodes running `./run-client.sh` or `./run-prover.sh` (as per our instructions) will **automatically** update. All other nodes must manually update themself.

## Changelog

- Bumps version from `3` to `4`
- Adds authenticated handshakes
- Adds `Sync` pool for validators and beacons
- Adds `BlockLocators` object
- Adds `BeaconPropose`, `BeaconTimeout`, `BeaconVote` messages
- Updates `PuzzleResponse` to send `Header` instead of `Block`
- Updates coinbase target algorithm
- Removes `UnconfirmedBlock` messages
- Removes `Status` from all nodes

## Tests

- Adds tests to check that block locators are well-formed
- Adds tests to check sync logic behaves as expected
- Adds tests to check coinbase retargeting is within bounds

PR #2108, opened 2022-11-29
Only show received connection log as the responder

HarukaMa:
## Motivation

The log was showing `Received a connection request from '...'` even if the connection is initiated by the local node. While the actual initiator could be inferred from the surrounding logs, this line itself is misleading.

### DIFF
--- a/node/router/src/handshake.rs
+++ b/node/router/src/handshake.rs
@@ -61,7 +61,9 @@ impl<N: Network> Router<N> {
         if let Err(forbidden_message) = self.ensure_peer_is_allowed(peer_addr) {
             return Err(error(format!("{forbidden_message}")));
         }
-        debug!("Received a connection request from '{peer_addr}'");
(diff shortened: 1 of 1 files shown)


niklaslong:
We might want to expand this change to include a log for both connection directions here? Alternatively, we could move this up a level. Otherwise this looks good to me.

HarukaMa:
Outgoing connections has its own "connecting to" logs, so not sure if an additional log is still needed here. 

However, I have then discovered that there are some warning messages that needs to change as well (they all sounds like the connection is an incoming one right now). Drafting this for now, feel free to supersede this pr if I forget to add those

howardwu:
I see your point, let's fix this once PR #2089 is merged. Thanks for bringing it up

howardwu (approved):
LGTM. @niklaslong lets unify the `Connecting to '{peer_ip}'...` message here in a follow up PR.

PR #2113, opened 2022-11-29
Deps: update Clap to v4.0

niklaslong:
See https://github.com/clap-rs/clap/issues/3234#issuecomment-1271750396 for discussion on the color changes. 

Also supersedes #2109, #2110 and #2111.


howardwu:
Can we wait on this until it has landed?

PR #2118, opened 2022-11-30
Fix(tcp): avoid self-connects

niklaslong:
Should fix the self-connect issue observed on testnet whilst still allowing for testing on loopback.

### DIFF
--- a/node/tcp/src/tcp.rs
+++ b/node/tcp/src/tcp.rs
@@ -171,7 +171,7 @@ impl Tcp {
                         Ok((stream, addr)) => {
                             debug!(parent: node_clone.span(), "tentatively accepted a connection from {}", addr);
 
-                            if !node_clone.can_add_connection() {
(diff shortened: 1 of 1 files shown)


ljedrz (commented):
This should fix the issue, but it would also cause all connections coming from own IP address to be rejected; are `--dev` nodes excluded from this logic?

niklaslong:
Only if these aren't loopback addresses. We likely wouldn't want to bind to an external IP when running with `--dev`?

ljedrz (approved):

ljedrz:
I also wonder if there are users who start multiple nodes on a single machines and intend them all to be able to connect to one another without relying on local IPs. It's probably a niche use case, but it's possible.

In any case, the best solution would be to use non-IP identifiers.

Sharpshop (approved):

howardwu (approved):

howardwu:
LGTM

PR #2121, opened 2022-11-30
Fix start message to only show REST when beacon or validator

### LINKED ISSUE #2120: [Bug] REST server not running for prover
## 🐛 Bug Report

Running a prover has the following log entry early after startup:
```
🌐 Starting the REST server at 0.0.0.0:3033.
```

However, any attempt to connect to that fails:

```
curl localhost:3033/
curl: (56) Recv failure: Connection reset by peer
```

I'm using version https://github.com/AleoHQ/snarkOS/commit/bf9cc9d61cef5893be3a9f62ada6fbfe1ef9c99a

howardwu:
## Motivation

Fixes #2120 

This PR updates the start message to only show REST when beacon or validator.

### DIFF
--- a/cli/src/commands/start.rs
+++ b/cli/src/commands/start.rs
@@ -271,11 +271,14 @@ impl Start {
                 self.node.to_string().bold()
             );
 
-            if let Some(rest_ip) = rest_ip {
-                println!("🌐 Starting the REST server at {}.\n", rest_ip.to_string().bold());
+            // If the node is running a REST server, print the REST IP and JWT.
+            if node_type.is_beacon() || node_type.is_validator() {
+                if let Some(rest_ip) = rest_ip {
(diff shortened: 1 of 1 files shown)


elliot-u410:
Fixes the issue. Might make more sense to put this log in the place where the rest server actually gets bound to a port or something. Then you'd avoid these arbitrary conditionals.

howardwu:
Agreed, the challenge is that if the CDN flag is enabled, then the REST server turns on **after** syncing with the CDN and thus this message is buried in the logs. The server requires the router (in order to render peering state), which similarly would cause the log to be buried amongst `tracing` messages.

elliot-u410 (approved):
Not that I matter, but looks good to me. 😄

raychu86 (approved):

PR #2123, opened 2022-11-30
Add `is_dev` flag to `Consensus`

raychu86:
## Motivation

This PR introduces the `is_dev` flag to `Consensus` so development nodes use the new retargeting algorithm instead of waiting until the `V4_START_HEIGHT`.


howardwu (approved):
LGTM

PR #2142, opened 2022-12-04
Adds vanity address generation for Aleo accounts

howardwu:
## Motivation

This PR adds vanity address generation for Aleo accounts.

The process searches for the first occurrence of an Aleo address that contains the specified vanity string at either the **front** or **back** of the address. The supported vanity characters adhere to the `bech32m` character set.
```
qpzry9x8gf2tvdw0s3jn54khce6mua7l1
```

To try the vanity generator, run:
```
snarkos account new --vanity {VANITY_STRING}
```

## Test Plan

This PR adds basic tests on the `is_in_bech32m_charset` and `has_vanity_string` methods that are introduced for the feature.

## Related PRs

None

### DIFF
--- a/cli/src/commands/account.rs
+++ b/cli/src/commands/account.rs
@@ -20,11 +20,13 @@ use snarkvm::console::{
     types::Field,
 };
 
-use anyhow::{anyhow, Result};
+use anyhow::{anyhow, bail, Result};
 use clap::Parser;
+use colored::Colorize;
 use core::str::FromStr;
 use rand::SeedableRng;
 use rand_chacha::ChaChaRng;
+use rayon::prelude::*;
 
 type Network = snarkvm::prelude::Testnet3;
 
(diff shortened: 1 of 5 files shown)


cli/src/commands/account.rs
```diff
@@ -36,32 +38,110 @@ pub enum Account {
...
-                    None => Field::rand(&mut ChaChaRng::from_entropy()),
-                };
-                // Recover the private key from the seed as a field element.
-                let private_key = PrivateKey::try_from(seed)
-                    .map_err(|_| anyhow!("Failed to convert the seed into a valid private key"))?;
-                // Construct the account.
-                let account = snarkos_account::Account::<Network>::try_from(private_key)?;
-                // Print the new Aleo account.
-                Ok(account.to_string())
+            Self::New { seed, vanity } => {
```
ljedrz, line 50: (this applies to the current state as well) are we expecting variants other than `New` in the future?

ljedrz (approved):
LGTM, left a small comment

cli/src/commands/account.rs
howardwu: Yes, `load` is one planned soon. If others have suggestions, we also open to adding other capabilities.

PR #2144, opened 2022-12-05
Fix/2102

### LINKED ISSUE #2102: [Bug] `Disconnect::handle_disconnect` doesn't appear to be triggering, despite `Disconnect::enable_disconnect` working
## 🐛 Bug Report

This change https://github.com/AleoHQ/snarkOS/pull/2101/commits/7a7797d0d91f736e2fed04beb2435b27f853089b was added in PR #2101 because of a bug whereby the disconnect handler from the `tcp` stack does not appear to be triggering `snarkos_tcp::protocols::Disconnect::handle_disconnect`.

The way I triggered this was by having more than the minimum number of peers, so that triggers the refresh peers logic, and thus the disconnect failure.

Here is a screenshot of the bug in action:

Screen Shot 2022-11-27 at 9 15 56 PM

At `04:49:02`, the first call to disconnect is made. It appears the peer may be disconnected in the `Tcp` stack. However, during the disconnect, `handle_disconnect` doesn't appear to be triggered, since calling `handle_disconnect` will invoke `router.remove_connected_peer(peer_ip)`, which clears the peer from the `Router`.

As a result, the peer remains connected (and becomes what we call a "stale" peer) on the `Router` level (but is already removed on the `Tcp` level), as shown in the following screenshot:

Screen Shot 2022-11-27 at 9 16 13 PM

## Steps to Reproduce

Try running the penultimate commit https://github.com/AleoHQ/snarkOS/pull/2101/commits/cc7db5c7bb1e7237e26ddc6a325a3e0774dd6baa for **at least 30 minutes** with the following command:
```
cargo run --release -- start --nodisplay --prover "APrivateKey1zkpAECeZYpLE5vcwcRdsb8NpRqEe182C4NApb12BonN1fyS" --verbosity 1
```

#### Alternative

If the "production mode" approach doesn't work, I have not tried this, but I would suggest to try running at least 4 dev nodes. I suspect it would also reproduce the issue.

The idea is that you need to have more than the minimum number of peers, then trigger the refresh peers logic, and see the disconnect failure.

Terminal 1:
```
cargo run --release -- start --nodisplay --dev 0 --beacon ""
```
Terminal 2:
```
cargo run --release -- start --nodisplay --dev 1 --validator ""
```
Terminal 3:
```
cargo run --release -- start --nodisplay --dev 2 --validator ""
```
Terminal 4:
```
cargo run --release -- start --nodisplay --dev 3 --prover "" --verbosity 1
```

Similarly, run for at least 30 minutes, and watch the prover (i.e. the 4th terminal window).

## Your Environment

- Run snarkOS @ https://github.com/AleoHQ/snarkOS/pull/2101/commits/cc7db5c7bb1e7237e26ddc6a325a3e0774dd6baa
- Ubuntu 22.04
- Aleo Testnet 3, Phase 2, Production

howardwu:
## Motivation

Fixes #2102

With this change, in the case that the disconnecting peer truly is not connected, the trigger for `handle_disconnect` will still trigger, giving the `Router` an opportunity to perform a clean up of the connection as well.

Based on the `Tcp` protocols, I think the originally intention of this `is_connected` conditional check was to prevent the trigger from being called multiple times. In case this occurs, I can confirm that multiple calls to `handle_disconnect` in the `Router` for a given peer IP is safe.

### DIFF
--- a/node/router/src/lib.rs
+++ b/node/router/src/lib.rs
@@ -148,10 +148,6 @@ impl<N: Network> Router<N> {
             // Disconnect from this peer.
             let _disconnected = router.tcp.disconnect(peer_ip).await;
             debug_assert!(_disconnected);
-            // TODO (howardwu): Revisit this. It appears `handle_disconnect` does not necessarily trigger.
-            //  See https://github.com/AleoHQ/snarkOS/issues/2102.
-            // Remove the peer from the connected peers.
-            router.remove_connected_peer(peer_ip);
         });
     }
 
--- a/node/tcp/src/tcp.rs
+++ b/node/tcp/src/tcp.rs
@@ -287,11 +287,9 @@ impl Tcp {
     /// Disconnects from the provided `SocketAddr`.
     pub async fn disconnect(&self, addr: SocketAddr) -> bool {
         if let Some(handler) = self.protocols.disconnect.get() {
-            if self.is_connected(addr) {
-                let (sender, receiver) = oneshot::channel();
-                handler.trigger((addr, sender));
-                let _ = receiver.await; // can't really fail
-            }
+            let (sender, receiver) = oneshot::channel();
+            handler.trigger((addr, sender));
+            let _ = receiver.await; // can't really fail
         }
 
         let conn = self.connections.remove(addr);


ljedrz:
Multiple calls to `handle_disconnect` are indeed safe; that check is done mostly for performance reasons (as extra disconnect logic might not be lightweight) and clarity (extra associated logs might be confusing).

However, if the main issue is that the extra disconnect logic is not triggering for peers that have not concluded their handshake yet, a better solution would be to either call it within the handshake logic (in case the handshake fails), or to not add the peer (or any other state related to it that needs to be cleaned up upon a disconnect) before the handshake is complete.

Note that the following check in `Tcp::disconnect`:
```
if self.is_connected(addr) {
```
is guaranteed to pass _even if the peer is actually already physically disconnected_, because it relies on the state of the collection of the connected addresses, which only gets updated after that check (L298):
```
let conn = self.connections.remove(addr);
```

ljedrz (commented):
This change is fine, but - as described above - not ideal; I'd recommend double-checking the aforementioned conditions firds.

ljedrz:
The related issue should be resolved by https://github.com/AleoHQ/snarkOS/pull/2168.

PR #2146, opened 2022-12-05
Skip computing block locators if peer is disconnected

howardwu:
## Motivation

This PR improves the efficiency of validators and beacons by checking that the peer is still connected prior to preparing a `Ping` message.

While the call to `send_ping` will already check that the peer is still connected (via logic in `can_send`), it was observed that validators and beacons (due to the number of connections they maintain) frequently prepare block locators that are not used as the peer has already disconnected.

This PR thus significantly reduces the number of calls to `Ledger` for validators and beacons, thereby improving the efficiency in usage of RocksDB for other operations such as syncing blocks.

### DIFF
--- a/node/src/beacon/router.rs
+++ b/node/src/beacon/router.rs
@@ -187,11 +187,14 @@ impl<N: Network, C: ConsensusStorage<N>> Inbound<N> for Beacon<N, C> {
         tokio::spawn(async move {
             // Sleep for the preset time before sending a `Ping` request.
             tokio::time::sleep(Duration::from_secs(Self::PING_SLEEP_IN_SECS)).await;
-            // Retrieve the block locators.
-            match crate::helpers::get_block_locators(&self_clone.ledger) {
-                // Send a `Ping` message to the peer.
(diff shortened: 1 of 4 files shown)


ghostant-1017:
Validator will always get block locators when perform_handshake, however, there's no need to do it when peer is a Prover or 
a Client，right？
https://github.com/AleoHQ/snarkOS/pull/2147

howardwu:
> Validator will always get block locators when perform_handshake, however, there's no need to do it when peer is a Prover or a Client，right？ #2147

I've answered this in PR 2147, thanks

ghostant-1017:
> > Validator will always get block locators when perform_handshake, however, there's no need to do it when peer is a Prover or a Client，right？ #2147
> 
> I've answered this in PR 2147, thanks

Ok, Maybe we can cached the block_locator and update it when new block is coming.
https://github.com/AleoHQ/snarkOS/pull/2152

ljedrz (approved):
LGTM :+1:

PR #2147, opened 2022-12-05
Only get block locators when peer is a beacon or validator

ghostant-1017:
## Motivation

Only get block locators when peer is a beacon or validator


howardwu:
Clients and provers still benefit from receiving block locators as provers need them in order to know who to request puzzle requests from, and clients need it if they would like to ensure their validator and beacon peers are consistent with them.

PR #2148, opened 2022-12-05
Updates router connect and restricted peer logic

howardwu:
## Motivation

This PR updates the `Router` `connect` and `insert_restricted_peer` logic.

## Tests

This PR introduces a test framework for `Router` by implementing a `TestRouter`. Currently, only the `Handshake` and `Disconnect` capabilities are needed in testing, however the `Reading` and `Writing` traits are implemented with `Inbound`, `Outbound`, and `Heartbeat` traits as stubs to support extensibility in the future.

### DIFF
--- a/node/router/tests/common/mod.rs
+++ b/node/router/tests/common/mod.rs
@@ -29,6 +29,19 @@ use snarkos_node_messages::NodeType;
 use snarkos_node_router::Router;
 use snarkvm::prelude::{Block, FromBytes, Network, Testnet3 as CurrentNetwork};
 
+/// A helper macro to print the TCP listening address, along with the connected and connecting peers.
+#[macro_export]
+macro_rules! print_tcp {
+    ($node:expr) => {
+        println!(
+            "{}: Active - {:?}, Pending - {:?}",
+            $node.local_ip(),
+            $node.tcp().connected_addrs(),
+            $node.tcp().connecting_addrs()
+        );
+    };
+}
+
 /// Returns a fixed account.
 pub fn sample_account() -> Account<CurrentNetwork> {
     Account::<CurrentNetwork>::from_str("APrivateKey1zkp2oVPTci9kKcUprnbzMwq95Di1MQERpYBhEeqvkrDirK1").unwrap()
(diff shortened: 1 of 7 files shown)


node/router/tests/disconnect.rs
```diff
@@ -0,0 +1,116 @@
...
+// along with the snarkOS library. If not, see <https://www.gnu.org/licenses/>.
+
+mod common;
+use common::*;
+
+use snarkos_node_tcp::{protocols::Handshake, P2P};
+
+use core::time::Duration;
+
+macro_rules! print_tcp {
```
ljedrz: nit: since this macro is duplicated (macro imports can be annoying), I'd recommend moving it to the `common` module for the tests

node/router/tests/common/mod.rs
```diff
@@ -0,0 +1,123 @@
...
+        .add_directive("tokio_util=off".parse().unwrap())
+        .add_directive("mio=off".parse().unwrap());
+
+    // Initialize tracing.
+    let _ = tracing_subscriber::fmt().with_env_filter(filter).with_target(level == 3).try_init();
+}
+
+/// Initializes a beacon router. Setting the `listening_port = 0` will result in a random port being assigned.
+#[allow(dead_code)]
+pub async fn beacon(listening_port: u16, max_peers: u16) -> TestRouter<CurrentNetwork> {
```
ljedrz, line 76: unless we're planning some more intricate `Router` setups, these functions can probably be deduplicated a bit by adding a common `async fn new_node(node_type: NodeType, listening_port: u16, max_peers: u16) -> TestRouter<CurrentNetwork>` to be called internally by `beacon` etc.

ljedrz (approved):
LGTM in general; left some suggestions that can be applied later too

node/router/tests/disconnect.rs
howardwu: Agreed, I have deduped the test macro into `common`

node/router/tests/common/mod.rs
howardwu: Agreed, I think there will be more intricate logic once `Inbound` and `Outbound` handling rules start to be tested, so to be safe lets leave it as is. If no changes happen then we can always unify them.

PR #2151, opened 2022-12-05
Increase cache size

howardwu:
## Motivation

Increase cache size to `1 << 17`. For those running validators, the cache size should be further increased to `1 << 20` or `1 << 21`.


ljedrz (approved):
As long as we don't see the increase in size causing a degradation in performance, I see no issues here :+1:.

PR #2159, opened 2022-12-06
Feat/puzzle response

ghostant-1017:
## Motivation

Client or Prover should only process the latest PuzzleResponse.


howardwu:
`Sync` already ensures that the client/prover will only request a `PuzzleRequest` with a specific set of peers if it deems them to be further on the canonical chain. Additionally, we enforce that a `PuzzleResponse` message is only processed from a peer if the node sent a `PuzzleRequest` to that peer. So we only process responses which we asked for.

In the case that the network is partitioned or forked, this proposed change introduces risk that the provers will not target the rebased ledger (i.e. the new canonical chain).

ghostant-1017:
> `Sync` already ensures that the client/prover will only request a `PuzzleRequest` with a specific set of peers if it deems them to be further on the canonical chain. Additionally, we enforce that a `PuzzleResponse` message is only processed from a peer if the node sent a `PuzzleRequest` to that peer. So we only process responses which we asked for.
> 
> In the case that the network is partitioned or forked, this proposed change introduces risk that the provers will not target the rebased ledger (i.e. the new canonical chain).

It seems reasonable when the network is stable. However, I push the Prover current_height metrics to exporter, it seems like sometimes the current_height decline sharply。

ghostant-1017:
> `Sync` already ensures that the client/prover will only request a `PuzzleRequest` with a specific set of peers if it deems them to be further on the canonical chain. Additionally, we enforce that a `PuzzleResponse` message is only processed from a peer if the node sent a `PuzzleRequest` to that peer. So we only process responses which we asked for.
> 
> In the case that the network is partitioned or forked, this proposed change introduces risk that the provers will not target the rebased ledger (i.e. the new canonical chain).

We will modify the MessageVersion in order to force provers to target the rebased target？

ghostant-1017:
image

ghostant-1017:
Prover/Client will not process BlockRequest and BlockResponse now, and never update itself's canon chain, If I'm not mistaken, It only update_peer_locators in processing Ping.
Please check this logic, thanks.

ghostant-1017:
image
@howardwu

howardwu:
I see your point here, let me think about it. While the drop in height isn't great, it is frankly the correct behavior.

We should be optimizing for higher quality peers, as opposed to artificially preserving (stale) block locators.

ghostant-1017:
> 
All right. Not as you say "drop in height isn't great", it's even worse now, more frequent, greater height. 
Scoring mechanism is needed.
There's no reward for Provers to broadcast proofs to theirs connected peers, but just drop the proofs is benefit.

node/src/client/router.rs
```diff
@@ -150,6 +150,15 @@ impl<N: Network, C: ConsensusStorage<N>> Inbound<N> for Client<N, C> {
         let epoch_number = serialized.epoch_challenge.epoch_number();
         // Retrieve the block height.
         let block_height = header.height();
+        if let Some(block_header) = self.latest_block_header.read().as_ref() {
+            let latest_height = block_header.height();
+            if latest_height >= block_height {
+                info!(
```
raychu86: This should be a `trace` or a `debug` rather than an `info`.

node/src/client/router.rs
```diff
@@ -150,6 +150,15 @@ impl<N: Network, C: ConsensusStorage<N>> Inbound<N> for Client<N, C> {
...
+                    "Received stale PuzzleResponse from {peer_ip}, Current Height {block_height}, Received {latest_height}"
+                );
+                return true;
```
raychu86:
```suggestion
        // Determine if the puzzle response is stale.
        if let Some(latest_block_header) = self.latest_block_header.read().as_ref() {
            if latest_block_header.height() >= block_height {
                trace!(
                    "Received stale PuzzleResponse from {peer_ip}, Current Height {block_height}, Received {latest_height}"
                );
                return true;
```

raychu86:
Logic is mostly sound, left some nits.

kaimast:
This PR still looks useful, but would need to be tested again before we could consider merging it. I will mark it as draft until we have some documentation of what problem it solves and how to test it.

PR #2164, opened 2022-12-06
Enforce user provides account private key to startup scripts

raychu86:
## Motivation

This PR removes the default private key from `run-prover.sh` and `run-validator.sh` and enforces that the user manually provides a private key.


howardwu (approved):
LGTM

PR #2166, opened 2022-12-07
Allow the node to start listening at any given time

ljedrz:
This PR allows the node to begin listening for connection requests at a specified point in time (rather than at the point of creation of `Tcp`), allowing it to have greater control over its startup. This also allows `Tcp::new` to no longer be `async` or fallible.

`Router::initialize_routing` was chosen as the point where the node starts listening; more specifically, this now happens right after all the TCP protocols are enabled.

In order to reflect the fact that listening can start at any point in time, some objects that used to be able to rely on a predefined listening address now need to instead rely on `Tcp` (which can be cloned around freely).

### DIFF
--- a/node/router/src/helpers/sync.rs
+++ b/node/router/src/helpers/sync.rs
@@ -14,6 +14,7 @@
 // You should have received a copy of the GNU General Public License
 // along with the snarkOS library. If not, see <https://www.gnu.org/licenses/>.
 
+use crate::Tcp;
 use snarkos_node_messages::BlockLocators;
 use snarkvm::prelude::{Block, Network};
 
@@ -67,8 +68,9 @@ impl Hash for PeerPair {
 /// - When a request is timed out, the `requests`, `request_timestamps`, and `responses` map remove the entry for the request height;
 #[derive(Clone, Debug)]
 pub struct Sync<N: Network> {
-    /// The listener IP of this node.
-    local_ip: SocketAddr,
+    /// The TCP stack of this node.
+    #[cfg_attr(test, allow(dead_code))]
+    tcp: Tcp,
     /// The canonical map of block height to block hash.
     /// This map is a linearly-increasing map of block heights to block hashes,
     /// updated solely from the ledger and candidate blocks (not from peers' block locators, to ensure there are no forks).
@@ -95,9 +97,9 @@ pub struct Sync<N: Network> {
 
 impl<N: Network> Sync<N> {
     /// Initializes a new instance of the sync pool.
-    pub fn new(local_ip: SocketAddr) -> Self {
+    pub fn new(tcp: Tcp) -> Self {
         Self {
-            local_ip,
+            tcp,
             canon: Default::default(),
             locators: Default::default(),
             common_ancestors: Default::default(),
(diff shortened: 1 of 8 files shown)


node/tcp/src/tcp.rs
```diff
@@ -64,7 +66,7 @@ pub struct InnerTcp {
     /// The node's configuration.
     config: Config,
     /// The node's listening address.
-    listening_addr: Option<SocketAddr>,
+    listening_addr: OnceCell<SocketAddr>,
```
howardwu, line 69: A question more for my curiosity:

`OnceCell` will work well for our use case here, but I'm curious if `OnceBox` would be a better choice (and whether it makes a difference or not). Again, more for my learning than an actual change request here.

node/router/src/helpers/sync.rs
```diff
@@ -67,8 +68,9 @@ impl Hash for PeerPair {
 /// - When a request is timed out, the `requests`, `request_timestamps`, and `responses` map remove the entry for the request height;
 #[derive(Clone, Debug)]
 pub struct Sync<N: Network> {
-    /// The listener IP of this node.
-    local_ip: SocketAddr,
+    /// The TCP stack of this node.
```
howardwu, line 71: Can you help explain the rationale behind the code changes in this file? I'm not following why this would make testing bettter.

My preference would be for the `Sync` module to be a subchild of `Router` and adhere strictly to `Router`'s state instead of carrying on `Tcp` state so that we ensure network abstracts remain pure to `Router`.

howardwu:
Sorry, I just realized CI was not checking the `tcp` crate, and as this PR modifies it, I have added it in here.

For those reading this, I have previously been running all tests locally for each PR in addition to relying on CI, so we are safe.

node/tcp/src/tcp.rs
ljedrz: It's a bit more strict, and since I hadn't digested all the new logic yet, I wanted to make sure that the cell doesn't get initialized twice with these changes.

Now that we've confirmed that this doesn't happen, it could indeed be `OnceBox` instead (though it's unlikely that we'll see any difference).

node/router/src/helpers/sync.rs
```diff
@@ -95,9 +97,9 @@ pub struct Sync<N: Network> {
 
 impl<N: Network> Sync<N> {
     /// Initializes a new instance of the sync pool.
-    pub fn new(local_ip: SocketAddr) -> Self {
+    pub fn new(tcp: Tcp) -> Self {
```
howardwu: (I can see this is a tough design request) Is there a way to preserve the original interface of merely passing in the `local_ip` since the `Sync` pool itself does not utilize `Tcp` functionalities?

howardwu (changes requested):
LGTM pending the `Sync` pool interface change discussion

node/router/src/helpers/sync.rs
ljedrz: The fact that it needs to know the listening address already makes it dependent on the network stack. If it's desirable for it to be decoupled from it, it would be best if it could instead not use that address at all, delegating the related logic to a higher-level object.

howardwu: The only use of `local_ip` in `Sync` is for it to self identify itself in the `HashMap` of common ancestors. We could perhaps use a dummy IP for all cases of "self" then.

ljedrz:
I updated the PR; it was rebased, I added some missing calls to `enable_listener` in `Router` tests (it seems that those were just added), and I proposed an alternative way of handling the `local_ip` field in `Sync` that doesn't require it to hold `Tcp`.

niklaslong (approved):

PR #2167, opened 2022-12-07
Guard against "two-way" connections

ljedrz:
In some rare cases, two nodes can attempt to connect to one another at almost the exact same time, which can result in an undesirable "two-way" connection. This PR introduces additional safeguards to avoid this.

### DIFF
--- a/node/router/src/handshake.rs
+++ b/node/router/src/handshake.rs
@@ -52,17 +52,34 @@ impl<N: Network> Router<N> {
         peer_side: ConnectionSide,
         genesis_header: Header<N>,
     ) -> io::Result<(SocketAddr, Framed<&mut TcpStream, MessageCodec<N>>)> {
-        if peer_side == ConnectionSide::Initiator {
+        // If this is an inbound connection, we log it, but don't know the listening address yet.
+        // Otherwise, we can immediately register the listening address and check for its uniqueness.
+        let mut peer_ip = if peer_side == ConnectionSide::Initiator {
             debug!("Received a connection request from '{peer_addr}'");
+            None
+        } else if !self.connecting_peers.lock().insert(peer_addr) {
+            warn!("Already accepting a connection from '{peer_addr}', rejecting the duplicate handshake attempt");
+            return Err(io::ErrorKind::ConnectionRefused.into());
+        } else {
+            Some(peer_addr)
+        };
+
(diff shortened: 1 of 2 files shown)


ljedrz:
Actually, there's one insertion missing; I'll add it shortly.

Edit: done.

howardwu:
I'm a big fan of the `handshake_inner` abstraction here. Thank you for decoupling it, very nice approach to handing `connecting_peers`. It is elegant and straightforward.

howardwu:
I made one logical change from the original PR change.

Before this PR, we always called `ensure_peer_is_allowed`. This PR made the call to `ensure_peer_is_allowed` only when `ConnectionSide::Initiator`. I changed it back, however in the process I noticed a bug on my part.

In `ensure_peer_is_allowed`, `Cache::insert_inbound_connection` was being called from both sides, so I have guarded that. In addition, as we now have confidence that the `peer_ip` is correct, I am checking the full `SocketAddr` instead of the `IpAddr`, which should help the case when multiple nodes are running on the same IP.

ljedrz:
The reason why I removed the `ensure_peer_is_allowed` check in case of outgoing connections was mostly due to the fact that it was designed with the listening address (the current unique peer identifier) in mind, but there is another reason - the checks on outbound connections should - ideally - be handled differently than those on the inbound ones; note that, for the outbound connections:
- the max peers check and the one for a preexisting connection are already done internally (and before the connection is even started), by `Tcp`
- the inbound connection spam check is inapplicable (as you later adjusted)
- outbound connection checks should be done before the stream is even started - it's much faster that way

My recommendation would be to only use `ensure_peer_is_allowed` for inbound connection checks (and only once the listening address is provided during the handshake), and to add a separate `check_outbound_connection` (or something like that) method that would be called in `Router::connect` as the very first operation.

ljedrz:
One more thing:
> In addition, as we now have confidence that the peer_ip is correct, I am checking the full SocketAddr instead of the IpAddr, which should help the case when multiple nodes are running on the same IP.

Note that a potential spammer could take advantage of this and send a fake listening port every time it attempts to connect; since it's a somewhat niche feature to run multiple nodes from a singe IP, if anything, I'd be even stricter and use `IpAddr` in the list of restricted peers, and possibly add a configurable whitelist to exclude the desired addresses from that check.

niklaslong (approved):

PR #2168, opened 2022-12-07
Resolve the addresses for the purposes of handle_disconnect

ljedrz:
This is a simple bugfix: the address provided in `Disconnect::handle_disconnect` is always the _true_ connected address, so it potentially (i.e. in case of inbound connections) needs to be resolved to the listener address which is used for all the higher-level peer-related accounting.


howardwu:
Amazing catch!

howardwu (approved):
LGTM

PR #2180, opened 2022-12-13
Correctly resolve the addresses when disconnecting directly

ljedrz:
While looking into potentially introducing strong typing to avoid IP addresses from being mixed up (though that would be a large change, so it might need to wait), I found 3 more mix-ups:
- `Router::disconnect`, like all other "higher-level" (higher than `Tcp`) objects, should use the listening address as the unique connection identifier, and resolve to the actually connected address internally
- the `Disconnect` impl for the `TestRouter` was not resolving the address (missed this one last time, as no test was hitting it - I'll post such a test case soon)
- `Reading::process_message` was not resolving the address before calling `Outbound::send`, though it was accidentally calling `Router::disconnect` correctly (though technically with the "wrong" address type, which the 1st point refers to)

This should fix the remaining occurrences of https://github.com/AleoHQ/snarkOS/issues/2102.

### DIFF
--- a/node/router/src/lib.rs
+++ b/node/router/src/lib.rs
@@ -142,13 +142,11 @@ impl<N: Network> Router<N> {
     pub fn disconnect(&self, peer_ip: SocketAddr) {
         let router = self.clone();
         tokio::spawn(async move {
-            // Disconnect from this peer.
-            let _disconnected = router.tcp.disconnect(peer_ip).await;
-            debug_assert!(_disconnected);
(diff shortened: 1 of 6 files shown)


ljedrz:
Hmm, some weird `git`-related mishap? Let me retry...

ljedrz:
Apparently this was a GitHub outage; reopened as https://github.com/AleoHQ/snarkOS/pull/2181.

PR #2181, opened 2022-12-13
Correctly resolve addresses when calling Router::disconnect directly

ljedrz:
While looking into potentially introducing strong typing to avoid IP addresses from being mixed up (though that would be a large change, so it might need to wait), I found 3 more mix-ups:
- `Router::disconnect`, like all other "higher-level" (higher than `Tcp`) objects, should use the listening address as the unique connection identifier, and resolve to the actually connected address internally
- the `Disconnect` impl for the `TestRouter` was not resolving the address (missed this one last time, as no test was hitting it - I'll post such a test case soon)
- `Reading::process_message` (in the unhappy path) was not resolving the address before calling `Outbound::send`, though it was accidentally calling `Router::disconnect` correctly (though technically with the "wrong" address type, which the 1st point refers to)

This should fix the remaining occurrences of https://github.com/AleoHQ/snarkOS/issues/2102.

### DIFF
--- a/node/router/src/lib.rs
+++ b/node/router/src/lib.rs
@@ -137,13 +137,11 @@ impl<N: Network> Router<N> {
     pub fn disconnect(&self, peer_ip: SocketAddr) {
         let router = self.clone();
         tokio::spawn(async move {
-            // Disconnect from this peer.
-            let _disconnected = router.tcp.disconnect(peer_ip).await;
-            debug_assert!(_disconnected);
-            // TODO (howardwu): Revisit this. It appears `handle_disconnect` does not necessarily trigger.
-            //  See https://github.com/AleoHQ/snarkOS/issues/2102.
-            // Remove the peer from the connected peers.
-            router.remove_connected_peer(peer_ip);
+            if let Some(peer_addr) = router.resolve_to_ambiguous(&peer_ip) {
+                // Disconnect from this peer.
+                let _disconnected = router.tcp.disconnect(peer_addr).await;
+                debug_assert!(_disconnected);
+            }
         });
     }
 
--- a/node/src/beacon/router.rs
+++ b/node/src/beacon/router.rs
@@ -104,13 +104,15 @@ impl<N: Network, C: ConsensusStorage<N>> Reading for Beacon<N, C> {
     }
 
     /// Processes a message received from the network.
-    async fn process_message(&self, peer_ip: SocketAddr, message: Self::Message) -> io::Result<()> {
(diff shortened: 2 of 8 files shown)


ljedrz:
GitHub seems to be experiencing some kind of an outage; will retry soon.

ljedrz:
The one failing test passes for me locally, so it might just be a fluke; will retry shortly.

Update: it failed again, but at a different spot, so I'll bump the deadlines.

node/router/src/lib.rs
```diff
@@ -142,13 +142,11 @@ impl<N: Network> Router<N> {
...
         let router = self.clone();
         tokio::spawn(async move {
-            // Disconnect from this peer.
-            let _disconnected = router.tcp.disconnect(peer_ip).await;
-            debug_assert!(_disconnected);
-            // TODO (howardwu): Revisit this. It appears `handle_disconnect` does not necessarily trigger.
-            //  See https://github.com/AleoHQ/snarkOS/issues/2102.
-            // Remove the peer from the connected peers.
-            router.remove_connected_peer(peer_ip);
+            if let Some(peer_addr) = router.resolve_to_ambiguous(&peer_ip) {
```
niklaslong, line 140: Note: perhaps we can rename `resolve_to_ambiguous` to something clearer in future? Perhaps `resolve_connection_addr`?

node/src/beacon/router.rs
```diff
@@ -104,13 +104,15 @@ impl<N: Network, C: ConsensusStorage<N>> Reading for Beacon<N, C> {
...
     /// Processes a message received from the network.
-    async fn process_message(&self, peer_ip: SocketAddr, message: Self::Message) -> io::Result<()> {
+    async fn process_message(&self, peer_addr: SocketAddr, message: Self::Message) -> io::Result<()> {
         // Process the message. Disconnect if the peer violated the protocol.
-        if let Err(error) = self.inbound(peer_ip, message).await {
-            warn!("Disconnecting from '{peer_ip}' - {error}");
-            self.send(peer_ip, Message::Disconnect(DisconnectReason::ProtocolViolation.into()));
-            // Disconnect from this peer.
-            self.router().disconnect(peer_ip);
+        if let Err(error) = self.inbound(peer_addr, message).await {
```
niklaslong, line 109: Note: `inbound` could also be clearer, perhaps `process_inbound` or something similar.

niklaslong (approved):

ljedrz:
Rebased.

PR #2195, opened 2022-12-19
Make a new warp-related compiler warning go away

ljedrz:
As suggested in https://github.com/seanmonstar/warp/issues/1012.


niklaslong (approved):
Nice 👍

PR #2219, opened 2023-01-23
Add a connection cleanup test

ljedrz:
This PR contains a test ensuring that our connections don't leak memory. Other than aiding the earlier disconnect-related fixes, it already helped detect the two following issues in `staging`:
- `Router::spawn` should **not** be used for short, "self-destructing" tasks (at least not in its current shape); this was reverted
- the accounting of connecting peers needs to be tighter for outbound connections; this was adjusted as a drive-by

`cargo update` was also run, because this PR alters the lockfile and there's already a pending dependabot PR which this will take care of.

### DIFF
--- a/node/router/src/lib.rs
+++ b/node/router/src/lib.rs
@@ -143,7 +143,7 @@ impl<N: Network> Router<N> {
         }
 
         let router = self.clone();
-        self.spawn(async move {
+        tokio::spawn(async move {
             // Attempt to connect to the candidate peer.
             debug!("Connecting to {peer_ip}...");
             match router.tcp.connect(peer_ip).await {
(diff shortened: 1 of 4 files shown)


node/router/src/lib.rs
```diff
@@ -177,13 +176,17 @@ impl<N: Network> Router<N> {
         if self.is_restricted(&peer_ip) {
             bail!("Dropping connection attempt to '{peer_ip}' (restricted)")
         }
+        // Ensure the node is not already connecting to this peer.
+        if !self.connecting_peers.lock().insert(peer_ip) {
```
niklaslong, line 180: I was wondering if we should seperate this logic but so long as it's the last block in this function, it's ok. Splitting it out runs the risk of check and insert not operating on the same state.

niklaslong (approved):

PR #2221, opened 2023-01-24
Make the handshake asymmetric

### LINKED ISSUE #2149: Unify the `Connecting to '{peer_ip}'...` message in `Router::handshake`
Unify the `Connecting to '{peer_ip}'...` message in `Router::handshake`.

_Originally posted by @howardwu in https://github.com/AleoHQ/snarkOS/pull/2108#pullrequestreview-1203846755_

### LINKED ISSUE #2184: [Proposal] Make the handshakes asymmetric
## 💥 Proposal

The current snarkOS handshake works roughly as follows:
1. the initiator connects to the responder
2. both send the challenge request
3. both verify the received challenge request and respond with challenge responses
4. both verify the received challenge response

(total number of messages sent: 2 from both sides)

My proposal is to change this to the following:
1. the initiator connects to the responder
2. the initiator sends the challenge request
3. the responder verifies the received challenge request and replies with a single message containing the challenge response and the challenge request
4. the initiator verifies the received challenge response **first**, then the bundled challenge request, and then replies with the challenge response

(total number of messages sent: 2 from the initiator, 1 from the responder)

This already reduces the number of messages exchanged, but the primary benefit (with the responder being the primary beneficiary) is extra resistance to malicious connection spam or noncompliant peers, for the following reasons:
- the connection initiator sends more messages, i.e. the handshake is more "lightweight" for the responder
- the responder doesn't need to do anything until it receives a challenge request (which reduces the amount of work when being spammed with malicious connection attempts that don't even provide any messages)
- the initiator is unable to "tailor" its challenge request based on the challenge request of the responder (e.g. set a compatible `version`)
- the initiator verifies the responder's challenge response before verifying the challenge request, reducing the work on its side in case the responder is noncompliant

Such an asymmetric setup is also much more predictable, i.e. simpler to test and debug. The downside is that it would be a breaking change.

ljedrz:
This PR addresses https://github.com/AleoHQ/snarkOS/issues/2184 with one difference - the messages are not bundled together (this can happen later on if desired). In addition, the handshake message size is now restricted to 1MiB, which could probably be made even smaller.

As a drive-by, some logs are disabled in tests (as they are only useful for debugging), and the connection initiation logs are unified.

Closes #2184.
Closes #2149.

### DIFF
--- a/node/router/src/handshake.rs
+++ b/node/router/src/handshake.rs
@@ -43,8 +43,40 @@ impl<N: Network> P2P for Router<N> {
     }
 }
 
+/// A macro unwrapping the expected handshake message or returning an error for unexpected messages.
+#[macro_export]
+macro_rules! expect_message {
+    ($msg_ty:path, $framed:expr, $peer_addr:expr) => {
+        match $framed.try_next().await? {
+            // Received the expected message, proceed.
+            Some($msg_ty(data)) => {
+                trace!("Received '{}' from '{}'", data.name(), $peer_addr);
+                data
+            }
+            // Received a disconnect message, abort.
+            Some(Message::Disconnect(reason)) => {
+                return Err(error(format!("'{}' disconnected: {reason:?}", $peer_addr)))
+            }
+            // Received an unexpected message, abort.
+            _ => return Err(error(format!("'{}' did not follow the handshake protocol", $peer_addr))),
+        }
+    };
+}
+
+/// A macro for cutting a handshake short if message verification fails.
+#[macro_export]
+macro_rules! handle_verification {
+    ($result:expr, $framed:expr, $peer_addr:expr) => {
+        if let Some(reason) = $result {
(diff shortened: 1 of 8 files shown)


node/router/src/handshake.rs
```diff
@@ -43,8 +43,40 @@ impl<N: Network> P2P for Router<N> {
...
+            Some($msg_ty(data)) => {
+                tracing::trace!("Received '{}' from '{}'", snarkos_node_messages::MessageTrait::name(&data), $peer_addr);
+                data
+            },
+            // Received a disconnect message, abort.
+            Some(Message::Disconnect(reason)) => {
+                return Err(error(format!("'{}' disconnected: {reason:?}", $peer_addr)))
+            }
+            // Received an unexpected message, abort.
+            _ => return Err(error(format!("'{}' did not send a challenge bundle", $peer_addr))),
```
niklaslong: I think this error message needs updating?

niklaslong (approved):

PR #2225, opened 2023-01-27
Avoid cloning PuzzleResponse/EpochChallenge

ljedrz:
While there are some inbound messages where it's beneficial to preserve their (semi-)serialized form for the purposes of relaying them further, this is not needed with `PuzzleResponse`; this was found via memory profiling, where the associated clone operation turned out to be quite significant.

### DIFF
--- a/node/router/src/inbound.rs
+++ b/node/router/src/inbound.rs
@@ -23,12 +23,11 @@ use snarkos_node_messages::{
     PeerResponse,
     Ping,
     Pong,
-    PuzzleResponse,
     UnconfirmedSolution,
     UnconfirmedTransaction,
 };
 use snarkos_node_tcp::protocols::Reading;
-use snarkvm::prelude::{Block, Header, Network, ProverSolution, Transaction};
+use snarkvm::prelude::{Block, EpochChallenge, Header, Network, ProverSolution, Transaction};
 
 use anyhow::{bail, ensure, Result};
 use std::{net::SocketAddr, time::Instant};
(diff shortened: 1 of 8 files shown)


ljedrz:
In addition, I wrapped the `EpochChallenge` contained in the `Prover` in an (extra)  `Arc` which allows us to reduce the memory consumption of a prover node by ~5MB (~13% with a small number of peers).

node/router/src/inbound.rs
```diff
@@ -202,15 +201,13 @@ pub trait Inbound<N: Network>: Reading + Outbound<N> {
...
                 self.router().cache.decrement_outbound_puzzle_requests(peer_ip);
 
-                // Clone the serialized message.
-                let serialized = message.clone();
                 // Perform the deferred non-blocking deserialization of the block header.
                 let header = match message.block_header.deserialize().await {
                     Ok(header) => header,
                     Err(error) => bail!("[PuzzleResponse] {error}"),
                 };
                 // Process the puzzle response.
```
niklaslong, line 209: Perhaps this doc needs some updating?

node/router/src/inbound.rs
```diff
@@ -383,7 +380,7 @@ pub trait Inbound<N: Network>: Reading + Outbound<N> {
     fn puzzle_request(&self, peer_ip: SocketAddr) -> bool;
 
     /// Handles a `PuzzleResponse` message.
```
niklaslong, line 382: Ditto?

niklaslong (approved):
LGTM 👍

node/router/src/inbound.rs
ljedrz: Actually, it should still be ok, as the arguments consist of the contents of that message type.

node/router/src/inbound.rs
ljedrz: Same here - a few other methods also don't take the full message as a parameter.

niklaslong: Ah, there's precedent—alright 😄

PR #2236, opened 2023-02-02
Feat: introduce the `metrics` crate

niklaslong:
In the spirit of keeping PRs small and incremental, here's an initial impl for the `metrics` crate. I haven't plugged it into the node yet but that should be a follow-up (and carefully done so as not to bother the ongoing consensus work) along with testing. 

In addition, this PR fixes a clippy lint on latest `nightly` and fixes a build warning related to compiling snarkVM without default features in subcrates, which is incompatible with the workspace definition of the dependency.


Cargo.toml
```diff
@@ -30,6 +30,7 @@ members = [
     "node/router",
     "node/store",
     "node/tcp",
+    "metrics",
```
ljedrz: I'm wondering if this shouldn't be `node/metrics` :thinking:

metrics/Cargo.toml
```diff
@@ -0,0 +1,30 @@
...
+
+[dependencies.metrics-exporter-prometheus]
+version = "0.11"
+
+[dependencies.tokio]
+version = "1.23"
+features = ["rt"]
+
+
+
```
ljedrz:
```suggestion
```

ljedrz (approved):
LGTM as a shim, left one comment

Cargo.toml
niklaslong: done

PR #2253, opened 2023-02-14
Merge the staging branch

ljedrz:
This is a rollup containing the following changes, pending a burn-in on the devnet:

- https://github.com/AleoHQ/snarkOS/pull/2221
- https://github.com/AleoHQ/snarkOS/pull/2223
- https://github.com/AleoHQ/snarkOS/pull/2225
- https://github.com/AleoHQ/snarkOS/pull/2234
- https://github.com/AleoHQ/snarkOS/pull/2236

The most notable change is the asymmetrical handshake, which is accompanied by a `VERSION` bump; other than that, there's an assortment of small refactorings and performance tweaks.

It's ready to be merged, but I'll mark it as a draft until the devnet burn-in is complete.

This branch will not be updated with any further PRs; instead, further work will proceed in the [`staging-next`](https://github.com/AleoHQ/snarkOS/tree/staging-next) branch based on it.

Cc @zosorock


ljedrz:
Rebased and updated snarkVM to `0.9.16`.

Cargo.toml
```diff
@@ -51,6 +52,9 @@ version = "1"
 version = "3.2"
 features = ["derive"]
 
+[dependencies.tikv-jemallocator]
```
howardwu: nit: alphabetical order

node/messages/src/helpers/codec.rs
```diff
@@ -21,6 +21,9 @@ use ::bytes::{BufMut, BytesMut};
 use core::marker::PhantomData;
 use tokio_util::codec::{Decoder, Encoder, LengthDelimitedCodec};
 
+/// The maximum size of a message that can be transmitted during the handshake.
+const MAXIMUM_HANDSHAKE_MESSAGE_SIZE: usize = 1024 * 1024; // 1 MiB
```
howardwu, line 25: How is this being used? I only see it used as the Default for the MessageCodec

node/router/src/helpers/cache.rs
```diff
@@ -180,39 +177,40 @@ impl<N: Network> Cache<N> {
         key: K,
         interval_in_secs: i64,
     ) -> usize {
+        // Fetch the current timestamp.
+        let now = OffsetDateTime::now_utc();
+
```
howardwu, line 182: Couldn't `now` become stale if placed at the top of this method (and before the `.write()` call)?

node/messages/src/helpers/codec.rs
ljedrz: Its purpose is to limit the size of handshake messages (after which it gets raised to `MAXIMUM_MESSAGE_SIZE` via `MessageCodec::update_max_message_len`) - otherwise any new peer could send us some other large `Message` (e.g. a big `Block`) as one of the handshake messages and we'd be wasting time deserializing it before deciding that we should drop it. It's an anti-DoS measure.

node/router/src/helpers/cache.rs
ljedrz: That value is actually closer to the timestamp of the appearance of the new key, making it more accurate; also, the difference would be miniscule, but any locking time - especially writing - should be limited to a minimum.

Cargo.toml
ljedrz: reordered :+1:

howardwu (approved):
LGTM

PR #2265, opened 2023-02-22
[Fix] Migrate integration test to run on a schedule to unblock CI

iamalwaysuncomfortable:
## Motivation

The integration test is currently configured to run on every push, which is timing out and causing ci to fail. 

Although this test SHOULD be run regularly, it is more of a healthcheck than a unit test and thus, more suited to run at regular intervals instead of on every push.

## Changelog
* Moved integration test to run on scheduled runs twice daily
* Created a `- run_serial_long` command with a timeout of `300m` designed for tests that are known/designed to be long running. This param could've been changed in `run_serial` directly, but it would mean that any tests not designed to be long running would hang for hours, hurting iteration time against PRs & costing ci time.

## Test Plan
* The integration test should be triggered in SnarkOS to ensure it runs correctly

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -101,6 +101,26 @@ commands:
       - clear_environment:
           cache_key: << parameters.cache_key >>
 
+  run_serial_long:
+    description: "Build and run long running tests"
+    parameters:
+      workspace_member:
+        type: string
+      cache_key:
+        type: string
+      flags:
+        type: string
+        default: ""
+    steps:
+      - checkout
+      - setup_environment:
+          cache_key: << parameters.cache_key >>
+      - run:
+          no_output_timeout: 300m
+          command: cd << parameters.workspace_member >> && RUST_MIN_STACK=67108864 cargo test << parameters.flags >>
+      - clear_environment:
+          cache_key: << parameters.cache_key >>
+
   run_parallel:
     description: "Build and run tests (in parallel)"
     parameters:
@@ -139,7 +159,7 @@ jobs:
       - image: cimg/rust:1.65
     resource_class: 2xlarge
     steps:
-      - run_serial:
+      - run_serial_long:
           workspace_member: .integration
           cache_key: snarkos-integration-cache
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -139,7 +139,6 @@ jobs:
       - image: cimg/rust:1.65
     resource_class: 2xlarge
     steps:
-      - no_output_timeout: 300m
       - run_serial:
           workspace_member: .integration
           cache_key: snarkos-integration-cache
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -101,6 +101,26 @@ commands:
       - clear_environment:
           cache_key: << parameters.cache_key >>
 
+  run_serial_long:
```
iamalwaysuncomfortable, line 104: Key change 1: Add a test run command for long-running tests

.circleci/config.yml
```diff
@@ -139,7 +159,7 @@ jobs:
       - image: cimg/rust:1.65
     resource_class: 2xlarge
     steps:
-      - run_serial:
+      - run_serial_long:
```
iamalwaysuncomfortable, line 162: Key change 2: Add long running run config to the integration test

.circleci/config.yml
```diff
@@ -339,3 +358,14 @@ workflows:
                 display,
                 node
               ]
+
+  scheduled-workflow:
```
iamalwaysuncomfortable, line 362: Key change 3: Add integration to a new a workflow for running regularly scheduled tests

ljedrz:
Where can the logs for the scheduled runs be seen?

Also, in the [CircleCI run for this PR](https://app.circleci.com/pipelines/github/AleoHQ/snarkOS?branch=pull%2F2265) I saw the following errors:
- [error 1](https://app.circleci.com/pipelines/github/AleoHQ/snarkOS/7641)
- [error 2](https://app.circleci.com/pipelines/github/AleoHQ/snarkOS/7642)

iamalwaysuncomfortable:
> Where can the logs for the scheduled runs be seen?

In the regular list of ci runs. None have run yet because it's scheduled to only run on testnet3.

There is also a slightly updated way to do this via triggers, but I don't currently have enough access to create a trigger for SnarkOS. If that's elevated we can do it this way: https://youtu.be/x3ruGpx6SEI?t=555

> Also, in the [CircleCI run for this PR](https://app.circleci.com/pipelines/github/AleoHQ/snarkOS?branch=pull%2F2265) I saw the following errors:
> 
> * [error 1](https://app.circleci.com/pipelines/github/AleoHQ/snarkOS/7641)
> * [error 2](https://app.circleci.com/pipelines/github/AleoHQ/snarkOS/7642)

Those errors occurred originally when the PR was in draft, and occurred because the `no_output_timeout` had been put in the `job` scope instead of the `command` scope. It was moved to the appropriate place before the PR was set to green.

ljedrz (approved):
I'm not an expert in CircleCI config, but since this job is currently failing, there is little risk involved even if we need to tweak it later on.

howardwu (approved):
LGTM

PR #2281, opened 2023-03-09
Update message queue depth to 1024

raychu86:
## Motivation

This PR increases the message queue depth for `reading` and `writing` from `64` to `1024`. Though we may want to tune the number down, as `1024` might be a little high.

This change was necessary as we saw 
`ERROR tcp{name=\"0\"}: snarkos_node_tcp::protocols::reading: can't process a message from <IP_ADDRESS>: no available capacity` messages when the number of incoming coinbase solutions spiked.


ljedrz (approved):
This should be fine, though if it turns out to not be enough, it will strongly suggest that there is some performance issue involved (likely a missing blocking task).

PR #2291, opened 2023-03-17
Update decrypt.rs

lizatkinson:
possible error in the 'parse()' function. Specifically, the format!() call in the last line returns a formatted string, but the string is not returned from the function. Instead, it should be returned like this:

Ok(format!("✅ Created deployment transaction for '{}'", self.program_id.to_string().bold())) Without the Ok(), the formatted string will not be returned as a Result<String>, which is the expected return type of the function.

## Motivation

(Write your motivation here)

## Test Plan

## Related PRs

(Link any related PRs here)


howardwu:
Can you confirm this is to fix the `Deploy` command and not the `Decrypt` command?

PR #2311, opened 2023-04-11
modified atomic::ordering about batch_in_progress

wang384670111:
https://github.com/AleoHQ/snarkOS/blob/8cfa1657746bcdc372207f788f093b3060aac350/node/store/src/rocksdb/map.rs#L95
I think the use of ordering here is irregular. In the start_atomic function, batch_in_progress is not synchronized with other variables, so Relaxed can be used for batch_in_progress.

https://github.com/AleoHQ/snarkOS/blob/8cfa1657746bcdc372207f788f093b3060aac350/node/store/src/rocksdb/map.rs#L116
https://github.com/AleoHQ/snarkOS/blob/8cfa1657746bcdc372207f788f093b3060aac350/node/store/src/rocksdb/map.rs#L149
In the finish_atomic and abort_atomic functions, batch_in_progress is used to synchronize the atomic_batch variable. Therefore, the read operation of batch_in_progress needs to see the changes of the atomic_batch variable.  Using Ordering::SeqCst is unnecessary and may impact program performance, I think just Acquire/Release needs to be used here to ensure the correctness of the program.

### DIFF
--- a/node/store/src/rocksdb/map.rs
+++ b/node/store/src/rocksdb/map.rs
@@ -44,7 +44,7 @@ impl<
     ///
     fn insert(&self, key: K, value: V) -> Result<()> {
         // Determine if an atomic batch is in progress.
-        let is_batch = self.batch_in_progress.load(Ordering::SeqCst);
+        let is_batch = self.batch_in_progress.load(Ordering::Acquire);
 
         match is_batch {
             // If a batch is in progress, add the key-value pair to the batch.
@@ -68,7 +68,7 @@ impl<
     ///
     fn remove(&self, key: &K) -> Result<()> {
         // Determine if an atomic batch is in progress.
-        let is_batch = self.batch_in_progress.load(Ordering::SeqCst);
+        let is_batch = self.batch_in_progress.load(Ordering::Acquire);
 
         match is_batch {
             // If a batch is in progress, add the key to the batch.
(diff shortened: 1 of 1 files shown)


ljedrz:
Thanks! I think it makes sense and this "loosening" is something that has always been the plan; let me just go over it once more with a fresh set of eyes tomorrow :+1:.

wang384670111:
> Thanks! I think it makes sense and this "loosening" is something that has always been the plan; let me just go over it once more with a fresh set of eyes tomorrow 👍.

Great! I also found some other improper usage of atomic types in snarkOS, and I will submit some PRs later.

ljedrz:
Thanks for the further adjustments to atomic ordering!

I have a request: could you collect them all in a single pull request (it could be this one) and re-categorize them a little bit? The categories (which would be in their own commits) are:
- `batch_in_progress` (the same commit as in this PR)
- `shutdown` (collected from all the PRs that adjust it in different node types)
- `SEQUENTIAL_NODE_ID`
- `puzzle_instances`
- `block_generation_time`

In other words, in the end there would be a single pull request with 5 commits, each containing one of the listed categories, and outlining that the atomics that don't require synchronization are changed to `Relaxed`, and ones that do are changed to the `Acquire`-`Release` "sandwich".

wang384670111:
> 

Sure! I will resubmit a new pull request that combines all the previous 5 pull requests uploaded.

ljedrz:

PR #2313, opened 2023-04-11
modified atomic::ordering about shutdown

wang384670111:
https://github.com/AleoHQ/snarkOS/blob/8cfa1657746bcdc372207f788f093b3060aac350/node/src/validator/mod.rs#L139


ljedrz:

PR #2314, opened 2023-04-11
modified atomic::ordering about SEQUENTIAL_NODE_ID

wang384670111:
https://github.com/AleoHQ/snarkOS/blob/8cfa1657746bcdc372207f788f093b3060aac350/node/tcp/src/tcp.rs#L89
I think the use of ordering here is irregular, SEQUENTIAL_NODE_ID is used here for counting, not to synchronize access to other shared variables. Although Ordering::SeqCst ensures the correctness of the program, it affects the performance of the program. Therefore, just Ordering::Relaxed needs to be used here to ensure the correctness of the program.


ljedrz:

PR #2315, opened 2023-04-11
modified atomic::ordering about shutdown and puzzle_instances

wang384670111:
https://github.com/AleoHQ/snarkOS/blob/8cfa1657746bcdc372207f788f093b3060aac350/node/src/prover/mod.rs#L123
https://github.com/AleoHQ/snarkOS/blob/8cfa1657746bcdc372207f788f093b3060aac350/node/src/prover/mod.rs#L247
https://github.com/AleoHQ/snarkOS/blob/8cfa1657746bcdc372207f788f093b3060aac350/node/src/prover/mod.rs#L252
https://github.com/AleoHQ/snarkOS/blob/8cfa1657746bcdc372207f788f093b3060aac350/node/src/prover/mod.rs#L259
Here puzzle_instance is used here for counting, not to synchronize access to other shared variables. Although Ordering::SeqCst ensures the correctness of the program, it affects the performance of the program. Therefore, just Ordering::Relaxed needs to be used here to ensure the correctness of the program.

### DIFF
--- a/node/src/prover/mod.rs
+++ b/node/src/prover/mod.rs
@@ -120,7 +120,7 @@ impl<N: Network, C: ConsensusStorage<N>> NodeInterface<N> for Prover<N, C> {
 
         // Shut down the coinbase puzzle.
         trace!("Shutting down the coinbase puzzle...");
-        self.shutdown.store(true, Ordering::SeqCst);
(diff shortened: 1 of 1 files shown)


ljedrz:

PR #2316, opened 2023-04-12
modified atomic::ordering about shutdown and block_generation_time

wang384670111:
https://github.com/AleoHQ/snarkOS/blob/8cfa1657746bcdc372207f788f093b3060aac350/node/src/beacon/mod.rs#L186
https://github.com/AleoHQ/snarkOS/blob/8cfa1657746bcdc372207f788f093b3060aac350/node/src/beacon/mod.rs#L234
https://github.com/AleoHQ/snarkOS/blob/8cfa1657746bcdc372207f788f093b3060aac350/node/src/beacon/mod.rs#L252
I think block_generation_time needs to be synchronized with consensus. The read operation of produce_next_block needs to see the next block produced by the produce_next_block function. Using Ordering::SeqCst is unnecessary and may impact program performance, I think just Acquire/Release needs to be used here to ensure the correctness of the program.

### DIFF
--- a/node/src/beacon/mod.rs
+++ b/node/src/beacon/mod.rs
@@ -183,7 +183,7 @@ impl<N: Network, C: ConsensusStorage<N>> NodeInterface<N> for Beacon<N, C> {
 
         // Shut down block production.
         trace!("Shutting down block production...");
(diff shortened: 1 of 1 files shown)


ljedrz:

PR #2317, opened 2023-04-12
Fine-tune the Ordering for all the atomics

wang384670111:
https://github.com/AleoHQ/snarkOS/blob/8cfa1657746bcdc372207f788f093b3060aac350/node/src/beacon/mod.rs#L234
https://github.com/AleoHQ/snarkOS/blob/8cfa1657746bcdc372207f788f093b3060aac350/node/src/beacon/mod.rs#L252
I think block_generation_time needs to be synchronized with consensus. The read operation of produce_next_block needs to see the next block produced by the produce_next_block function. Using Ordering::SeqCst is unnecessary and may impact program performance, I think just Acquire/Release needs to be used here to ensure the correctness of the program.

https://github.com/AleoHQ/snarkOS/blob/8cfa1657746bcdc372207f788f093b3060aac350/node/src/prover/mod.rs#L247
https://github.com/AleoHQ/snarkOS/blob/8cfa1657746bcdc372207f788f093b3060aac350/node/src/prover/mod.rs#L252
https://github.com/AleoHQ/snarkOS/blob/8cfa1657746bcdc372207f788f093b3060aac350/node/src/prover/mod.rs#L259
Here puzzle_instance is used here for counting, not to synchronize access to other shared variables. Although Ordering::SeqCst ensures the correctness of the program, it affects the performance of the program. Therefore, just Ordering::Relaxed needs to be used here to ensure the correctness of the program.

https://github.com/AleoHQ/snarkOS/blob/8cfa1657746bcdc372207f788f093b3060aac350/node/tcp/src/tcp.rs#L89
I think the use of ordering here is irregular, SEQUENTIAL_NODE_ID is used here for counting, not to synchronize access to other shared variables. Although Ordering::SeqCst ensures the correctness of the program, it affects the performance of the program. Therefore, just Ordering::Relaxed needs to be used here to ensure the correctness of the program.

https://github.com/AleoHQ/snarkOS/blob/8cfa1657746bcdc372207f788f093b3060aac350/node/store/src/rocksdb/map.rs#L95
I think the use of ordering here is irregular. In the start_atomic function, batch_in_progress is not synchronized with other variables, so Relaxed can be used for batch_in_progress.

https://github.com/AleoHQ/snarkOS/blob/8cfa1657746bcdc372207f788f093b3060aac350/node/store/src/rocksdb/map.rs#L116
https://github.com/AleoHQ/snarkOS/blob/8cfa1657746bcdc372207f788f093b3060aac350/node/store/src/rocksdb/map.rs#L149
In the finish_atomic and abort_atomic functions, batch_in_progress is used to synchronize the atomic_batch variable. Therefore, the read operation of batch_in_progress needs to see the changes of the atomic_batch variable.  Using Ordering::SeqCst is unnecessary and may impact program performance, I think just Acquire/Release needs to be used here to ensure the correctness of the program.

https://github.com/AleoHQ/snarkOS/blob/8cfa1657746bcdc372207f788f093b3060aac350/node/src/beacon/mod.rs#L186
https://github.com/AleoHQ/snarkOS/blob/8cfa1657746bcdc372207f788f093b3060aac350/node/src/prover/mod.rs#L123

### DIFF
--- a/node/src/beacon/mod.rs
+++ b/node/src/beacon/mod.rs
@@ -183,7 +183,7 @@ impl<N: Network, C: ConsensusStorage<N>> NodeInterface<N> for Beacon<N, C> {
 
         // Shut down block production.
         trace!("Shutting down block production...");
-        self.shutdown.store(true, Ordering::SeqCst);
+        self.shutdown.store(true, Ordering::Relaxed);
 
         // Abort the tasks.
         trace!("Shutting down the beacon...");
@@ -231,7 +231,7 @@ impl<N: Network, C: ConsensusStorage<N>> Beacon<N, C> {
 
                 // Do not produce a block if the elapsed time has not exceeded `ROUND_TIME - block_generation_time`.
                 // This will ensure a block is produced at intervals of approximately `ROUND_TIME`.
-                let time_to_wait = ROUND_TIME.saturating_sub(beacon.block_generation_time.load(Ordering::SeqCst));
+                let time_to_wait = ROUND_TIME.saturating_sub(beacon.block_generation_time.load(Ordering::Acquire));
                 trace!("Waiting for {time_to_wait} seconds before producing a block...");
                 if elapsed_time < time_to_wait {
                     if let Err(error) = timeout(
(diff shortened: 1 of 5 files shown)


ljedrz (approved):
LGTM :+1:

howardwu (approved):
LGTM

PR #2319, opened 2023-04-14
Categorize DataMap IDs, introduce MapID

ljedrz:
This PR introduces a new `enum MapID` and a few associated `enum`s in order to split the now-private lower-level `enum DataID` into more readable "chunks" whose order can be changed if need be.

The rules regarding `DataID` remain the same - the order of its variants - once a storage has been initialized - should never change, and new ones should only ever be appended to the end.

### DIFF
--- a/node/store/src/lib.rs
+++ b/node/store/src/lib.rs
@@ -47,6 +47,25 @@ pub enum MapID {
     Transaction(TransactionMap),
     Transition(TransitionMap),
     Program(ProgramMap),
+    #[cfg(test)]
+    Test(TestMap),
+}
+
+impl From<MapID> for u16 {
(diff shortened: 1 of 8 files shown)


node/store/src/lib.rs
raychu86, line 1: We should move `DataID:DeploymentReverseFeeMap` up to its relative position (instead of at the end, which we had to tack on for deploy/execute). 

This will invalidate current databases, however it'll be nice to have an initially organized `DataID` enum

ljedrz: Done; as long as this branch is only used in fresh runs, we can reorder the `DataID` enum as we see fit :+1:.

PR #2323, opened 2023-04-15
Categorize DataMap IDs, introduce MapID

ljedrz:
This is just https://github.com/AleoHQ/snarkOS/pull/2319 reopened against `testnet3`; the changes are the same:

This PR introduces a new enum MapID and a few associated enums in order to split the now-private lower-level enum DataID into more readable "chunks" whose order can be changed if need be.

The rules regarding DataID remain the same - the order of its variants - once a storage has been initialized - should never change, and new ones should only ever be appended to the end.

### DIFF
--- a/node/store/src/block.rs
+++ b/node/store/src/block.rs
@@ -16,7 +16,8 @@
 
 use crate::{
     rocksdb::{self, DataMap, Database},
-    DataID,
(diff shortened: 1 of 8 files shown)


howardwu (approved):
LGTM

PR #2332, opened 2023-04-18
Add fee to beacon transfers

raychu86:
## Motivation

This PR adds fees to the transfer functions created by the beacon. 

Currently the fee is set to the minimum of 1 microcredit, however will need to be updated once rules for execution fees are established.


raychu86:
Closed in favor of #2333

PR #2333, opened 2023-04-18
Change Beacon transaction from `transfer` to `mint`

raychu86:
## Motivation

This PR changes the beacon transaction from a `transfer` to a `mint`. 

Consensus was also changed to support these coinbase transactions (mint) and that the mint was attributed to an authorized beacon address.


howardwu (approved):
LGTM

PR #2337, opened 2023-04-19
Switch from warp to axum

ljedrz:
Note: this PR is against the `staging` branch.

This PR replaces `warp` with `axum` in the node's HTTP server.


ghost:
From @howardwu:

The intention of this PR is to make it exactly 1:1 to what we currently have but this should be tested separately in a dev net to sanity check that it does match what we expect.

This will give us a bunch more extensibility and will provide better middleware support which streamlines and increases usability.

ljedrz:
The intention is 1:1 parity indeed; I've done some local testing and everything seemed to work fine, but I always welcome additional tests :+1:.

niklaslong (approved):

niklaslong:
LGTM, I also tested the endpoints when initially implementing this but a sanity check would be most welcome.

PR #2342, opened 2023-04-19
Add enforcements for `Execution` fee

raychu86:
## Motivation

This PR adds an enforcement for execution fees where the fee in microcredits must be at least the execution size in bytes.


howardwu (approved):
LGTM

PR #2343, opened 2023-04-20
Update the frame size more cleanly

ljedrz:
A simple fix that should take case of the occasional `frame size too big` error.


howardwu (approved):
LGTM

PR #2345, opened 2023-04-20
Correctly set the post-handshake codec limit

ljedrz:
While the change in https://github.com/AleoHQ/snarkOS/pull/2343 is fine, it was insufficient, as the original issue was more complex.

We are not persisting the handshake codec, so the ones created for the purposes of reading and writing messages need to have their maximum frame size adjusted separately.


howardwu (approved):
LGTM

PR #2346, opened 2023-04-20
Updates the logic for `cdn_height` method

howardwu:
## Motivation

Updates the logic for `cdn_height` method.

The new approach decouples `snarkOS` from `vm.aleo.org` completely, now only relying purely on the `CDN` when needed.


raychu86 (approved):
LGTM

PR #2349, opened 2023-04-24
Use generic instead of hard coded MapID for `open_map`

raychu86:
## Motivation

This PR updates the `mapID` type in `RocksDB::open_map` to a generic. This purpose of this, is that external repos that use the `DataMap` and `RocksDB` can define their own custom MapIDs instead of using the hard-coded snarkOS ones.


ljedrz (approved):
LGTM :+1:

howardwu (approved):

PR #2352, opened 2023-04-25
Bump snarkVM to `v0.10.3`

raychu86:
## Motivation

This PR bumps the snarkVM version to `v0.10.3` and makes the relevant changes associated with the version bump.

 - Do not enforce fee amount for `credits.aleo/split` executions
 - Update `test_ledger_execute_many` to ensure that the above check is written correctly.
 - Update `snarkos developer` CLI to enforce fees (except for `credits.aleo/split`)
 - Remove use of "gates" in favor of "microcredits"
 
 
 NOTE: This relaxation of fee enforcement for `splits` is **TEMPORARY**, and will be removed once batched transactions are integrated in snarkVM. The intention here is to allow users on Testnet 3 Phase 3 to start executing/deploying programs without requiring multiple records initially.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2960,9 +2960,9 @@ dependencies = [
 
 [[package]]
 name = "snarkvm"
-version = "0.10.2"
+version = "0.10.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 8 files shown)


howardwu (approved):
LGTM

PR #2358, opened 2023-04-26
Introduce OnConnect for automatic post-handshake actions

ljedrz:
This feature is designed to "tighten" the handshake and to allow the `Disconnect` actions to kick in if an initial action fails; this currently doesn't happen, which can sometimes cause a discrepancy between the low-level and high-level connected peers count

### DIFF
--- a/node/tcp/src/protocols/on_connect.rs
+++ b/node/tcp/src/protocols/on_connect.rs
@@ -1,0 +1,80 @@
+// Copyright (C) 2019-2023 Aleo Systems Inc.
+// This file is part of the snarkOS library.
+
+// The snarkOS library is free software: you can redistribute it and/or modify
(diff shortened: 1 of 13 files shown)


node/tcp/src/protocols/on_connect.rs
```diff
@@ -0,0 +1,80 @@
...
+use tracing::*;
+
+use crate::{protocols::ProtocolHandler, P2P};
+#[cfg(doc)]
+use crate::{
+    protocols::{Reading, Writing},
+    Connection,
+};
+
+/// Can be used to automatically perform some initial actions when the connection with a peer is
```
niklaslong:
```suggestion
/// Can be used to automatically perform some initial actions once the connection with a peer is
```

niklaslong (approved):

PR #2362, opened 2023-04-28
Register execution environment information

ljedrz:
This PR will allow us to get a range of information on the execution environment of snarkOS. This will be useful for the dev/testnets, as well as bug reports. At this point it is exposed as a new HTTP API route.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2698,6 +2698,7 @@ version = "2.0.2"
 dependencies = [
  "anyhow",
  "clap",
+ "once_cell",
  "rusty-hook",
  "snarkos-account",
  "snarkos-cli",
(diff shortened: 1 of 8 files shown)


ljedrz:
it seems that the `disconnect` tests can be flaky; I'll look into it shortly

niklaslong (commented):
LGTM so far 👍

joske:
So you get this info and store it in the struct. Then what? Or am I missing something?

ljedrz:
@joske you're 100% right, hence the draft - the PR is pending feedback regarding how this info is to be obtainable (HTTP API / runtime argument / both).

joske (approved):

ljedrz:
Merged as part of a `staging` proposal.

PR #2382, opened 2023-05-10
Return prettified JSON objects in REST API

raychu86:
## Motivation

This PR updates the REST API responses to return prettified JSON objects for better human readability.


howardwu (approved):
LGTM

PR #2397, opened 2023-05-18
Enforce a 1s raw TCP connection timeout

ljedrz:
While we already impose a limit on the duration of the handshake, we currently rely on the OS to impose a time limit on raw TCP connections, and that time can be rather long.

This PR introduces a way to configure it, and sets it to a default value of 1 second.


joske (approved):

joske:
LGTM

PR #2399, opened 2023-05-21
change license from GPL to Apache 2.0

apruden2008:
## Motivation

The GPL license is somewhat less permissive than the Apache 2.0 license. Switching to Apache 2.0 will make it easier for community members to contribute and build on top of the Aleo protocol.

## Test Plan

N/A

## Related PRs

N/A

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,596 +1,202 @@
-GNU General Public License
-==========================
-
-Version 3, 29 June 2007
-
-Copyright © 2007 Free Software Foundation, Inc. &lt;<https://fsf.org/>&gt;
-
-Everyone is permitted to copy and distribute verbatim copies of this license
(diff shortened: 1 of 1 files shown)


ljedrz:
Note: the license headers in the source files would also need to be adjusted to reflect this. What is more, with Apache 2.0 there is no need to include the license in every source file ([source](https://lists.apache.org/thread/tj0frc009q5yswxq84yd0h8r9ombx8d1)). Happy to perform this adjustment here and in the other affected repositories.

SplittyDev:
Note: There's a badge in the README that says GPL, that should probably be changed also. The license is hard-coded into the badge, it's not inferred from the GitHub repo.

howardwu:
I'll address all of these this week, merging to keep things moving.

PR #2408, opened 2023-05-24
Update the MSRV

ljedrz:
Due to the use of `BTreeMap::last_key_value` in snarkVM, the MSRV of snarkOS has increased as well.

In addition to updating the README, the `rust-version` key is added to the `Cargo.toml` of the project in order for the users to immediately know if their compiler is unable to compile it.


howardwu (approved):
LGTM

PR #2410, opened 2023-05-24
Update clap to 4.3

ljedrz:
This is the snarkOS counterpart to https://github.com/AleoHQ/snarkVM/pull/1583.

### DIFF
--- a/cli/src/commands/mod.rs
+++ b/cli/src/commands/mod.rs
@@ -27,11 +27,19 @@ pub use start::*;
 mod update;
 pub use update::*;
 
+use anstyle::{AnsiColor, Color, Style};
 use anyhow::Result;
-use clap::Parser;
+use clap::{builder::Styles, Parser};
+
+const HEADER_COLOR: Option<Color> = Some(Color::Ansi(AnsiColor::Yellow));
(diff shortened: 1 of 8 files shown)


joske (approved):

ljedrz:
Updated with 3 test cases for arg parsing.

cli/src/commands/mod.rs
```diff
@@ -78,4 +86,99 @@ mod tests {
         use clap::CommandFactory;
         CLI::command().debug_assert()
     }
+
+    #[test]
+    fn clap_snarkos_start() {
```
howardwu: Move these to their respective files so their fields don't need to be `pub` exposed

ljedrz: They would still need to be `pub` there, as the modules would be separate (the `tests` module vs. the parent module); however, I don't see a reason why those fields wouldn't all be `pub` (like in `struct Start`), as they are only used to parse commands, so there is not much reason to not expose them.

howardwu: In my experience, they won't need `pub` if they're in the same file.

ljedrz: You're right, I must have mis-pasted something last time, which gave me import errors; updated and rebased :+1:.

howardwu (approved):
LGTM

PR #2411, opened 2023-05-24
fix: reduce log level

joske:
This reduces log level in consensus from info! to debug! in a few cases. 

Rationale: we don't want to have 'per message' logging in info level to avoid log polution.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,5 @@ wasm/Cargo.lock
 **/build
 **.ledger-*
 **/committee/.dev/
+.bft-storage-3/
+validator-*.log
--- a/node/bft-consensus/src/state.rs
+++ b/node/bft-consensus/src/state.rs
@@ -66,7 +66,7 @@ impl<N: Network, C: ConsensusStorage<N>> ExecutionState for BftExecutionState<N,
         let mut validator_id = self.primary_pub.to_string();
         validator_id.truncate(8);
 
-        info!(
+        debug!(
             "Consensus (id: {}) output for round {}: {} batches, leader: {}",
             validator_id,
             consensus_output.sub_dag.leader.header.round,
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -6,3 +6,5 @@ wasm/Cargo.lock
 **/build
 **.ledger-*
 **/committee/.dev/
+.bft-storage-3/
+validator-*.log
```
ljedrz, line 10: these are unrelated, no?

joske: not really related to reducing the log level no

node/consensus/src/lib.rs
```diff
@@ -359,7 +359,7 @@ impl<N: Network, C: ConsensusStorage<N>> Consensus<N, C> {
             self.memory_pool.clear_invalid_solutions(self);
         }
 
-        info!("Advanced to block {}", block.height());
+        debug!("Advanced to block {}", block.height());
```
ljedrz: don't we want to maintain this level? we should be able to see that a block was advanced in every node

.gitignore
joske: can remove if you prefer this separately

ljedrz: the `validator` bit we don't need anymore IIRC, the other one I'm not sure, so we might want to remove those

ljedrz (approved):
LGTM with a couple of comments

.gitignore
joske: if you do `run_bft.sh` it creates the `validator-x.log` files.

node/consensus/src/lib.rs
joske: Well, I'm not sure, you can still see it if you want (set verbosity to 1). Not a big deal to me though, this is just one log line per 30 seconds or so.

ljedrz: yeah, I think I'd keep this as `info` as it's not spammy and it's an interesting piece of information; all the others LGTM

PR #2422, opened 2023-06-02
Staking test cases

niklaslong:
This PR tests generic staking type transactions within Narwhal/Bullshark. It also adds an id to the test transaction types to allow deduplication logic and makes some adjustements to the test helpers.


joske (approved):

node/bft-consensus/tests/common/state.rs
```diff
@@ -38,24 +40,29 @@ pub type Amount = u64;
 
 // A simple state for BFT consensus tests.
 pub struct TestBftExecutionState {
+    pub committee: SharedCommittee,
     pub balances: Mutex<HashMap<Address, Amount>>,
-    pub processed_txs: AtomicUsize,
+    pub processed_tx_count: AtomicUsize,
+    pub processed_txs: Mutex<HashSet<u64>>,
```
ljedrz: do we still need the count if we store the tx ids now?

node/bft-consensus/tests/common/state.rs
```diff
@@ -108,10 +113,15 @@ impl TestBftExecutionState {
         let mut balances = self.balances.lock();
 
         for transaction in transactions {
-            self.processed_txs.fetch_add(1, Ordering::Relaxed);
+            self.processed_tx_count.fetch_add(1, Ordering::Relaxed);
+
+            // Skip transactions that have already been processed to avoid corrupting state.
+            if !self.processed_txs.lock().insert(transaction.id()) {
+                continue;
+            }
```
ljedrz, line 105: is this even possible?

node/bft-consensus/tests/common/transaction.rs
```diff
@@ -22,12 +23,30 @@ pub const MAX_TRANSFER_AMOUNT: u64 = 10_000;
...
     pub from: Address,
     pub to: Address,
     pub amount: Amount,
 }
+
+#[derive(Serialize, Deserialize)]
+pub struct StakeChange {
+    pub id: u64,
+    pub pub_key: PublicKey,
+    pub stake: i64,
```
ljedrz:
```suggestion
    pub change: i64,
```

node/bft-consensus/tests/staking.rs
```diff
@@ -0,0 +1,75 @@
...
+
+    // Use a deterministic Rng.
+    let mut rng = TestRng::default();
+
+    for i in 0..NUM_STAKE_CHANGES {
+        // 1. Generate a stake change for a random authority.
+        let (pub_key, _authority) = state.committee.load().authorities.clone().into_iter().choose(&mut rng).unwrap();
+
+        // Generate a random stake change.
+        let stake = rng.gen_range(1..=10);
```
ljedrz:
```suggestion
        let stake = rng.gen_range(-10..=10);
```

ljedrz: though this would likely require a bit more tinkering, so that no one ends up with a negative value

node/bft-consensus/tests/staking.rs
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    // Wait for a while to allow the transfers to be processed.
+    sleep(Duration::from_secs(10)).await;
+
+    // 3. Check the state matches across the network.
+    let first_state = running_consensus_instances[0].state.committee.load_full();
+    for state in running_consensus_instances.iter().skip(1).map(|rci| &rci.state) {
+        assert_eq!(first_state, state.committee.load_full());
+    }
```
ljedrz, line 74: shouldn't we compare the stake of all the primaries instead of just the first one?

node/bft-consensus/tests/staking.rs
niklaslong: It should work already actually, it's a saturating add/sub.

node/bft-consensus/tests/staking.rs
niklaslong: This will break if any state isn't identical to all the others, it's mostly meant as a sanity check, we're not asserting on precise values (though we should in future).

node/bft-consensus/tests/common/state.rs
niklaslong: It is, since txs are never deduped before this. If we submit the tx to a single worker it's not an issue but if we submit it to multiple, the node won't recognise the duplicates.

node/bft-consensus/tests/common/state.rs
niklaslong: not really

node/bft-consensus/tests/staking.rs
niklaslong: confirmed and added

node/bft-consensus/tests/common/transaction.rs
niklaslong: fixed

node/bft-consensus/tests/common/state.rs
niklaslong: fixed

niklaslong:
Circleci seems to be having some issues today, rerunning. 
Edit: that's a little better, `node` failures are expected and `clippy` is updated in #2421.

niklaslong:
Closing as these aren't relevant anymore.

PR #2427, opened 2023-06-06
Updates snarkVM and migrates consensus logic to snarkVM

howardwu:
## Motivation

Updates snarkVM and migrates consensus logic to snarkVM.

This PR bumps the message version to `7`.

This PR's cousin PR is here: https://github.com/AleoHQ/snarkVM/pull/1628


howardwu:
@ljedrz @niklaslong any chance you can take a look at the `node-router` CI failure?

(the `node-cdn` failure is expected, it is because this is a breaking version change)

niklaslong:
@howardwu I've updated the router tests by lengthening the first `sleep` in each case a little, this seems to be sufficient. I'm fairly confident this is benign since the router doesn't at all depend on the consensus. 

I can also reproduce the consensus failure locally fwiw, but I notice this was in your inital changeset so haven't touched it so far.

howardwu:
It's odd because this PR doesn't touch the `router` in any way. And the failing test does not call any code related to the changeset here either.

ljedrz (approved):

niklaslong (approved):

PR #2429, opened 2023-06-07
Install on Windows FAQ added

liolikus:
I added Windows FAQ with the most common cases without `jemalloc feature` due to https://github.com/AleoHQ/snarkOS/pull/2423


howardwu:
Clsoing as this is no longer a concern as of #2427. `jemalloc` is no longer enabled by default.

PR #2430, opened 2023-06-08
[Feature] Finalize Fees.

d0cd:
This PR adds logic to `check_transaction_basic` checking that the fee is enough the cover the cost of executing all finalize blocks associated with the transaction.


howardwu:
@d0cd can you move this into `snarkvm-ledger`? The `snarkos-consensus` crate has been refactored in preparation for next week's reset in #2427

PR #2438, opened 2023-06-13
Read the node's private key from the filesystem

### LINKED ISSUE #1583: [Bug] password should be passed from a file
## 🐛 Bug Report

The password is read from a value in the config or command line, it should be read from a file. This way you can pass it to a systemd service from load credentials or to docker from docker secrets. (usually secrets should be read from files).

I would be happy to make a Proposal PR if you are open to the change.

(I filed it as a bug because this is potentially a security risk, let me know if you want to classify it differently)

### LINKED ISSUE #2169: [Proposal] entering private key in plaintext raise concerns
## 💥 Proposal

Entering private key in plaintext in terminal and yields potential threats, instead a signature derived from private key file would be much more secure I suppose.

ljedrz:
This PR is a proposal to address the community requests to allow the node's private key to be loaded from the filesystem in addition to being passed as a runtime argument or input from the command line.

The main change is as follows:
```sh
# Using a private key string
--beacon --private-key "{private_key_string}"
--validator --private-key "{private_key_string}"
--prover --private-key "{private_key_string}"
--client --private-key "{private_key_string}"

# Using a private key file
--beacon --private-key-file "{file_path}"
--validator --private-key-file "{file_path}"
--prover --private-key-file "{file_path}"
--client --private-key-file "{file_path}"
```

This setup, while a bit more verbose, provides better optionality for the user.

### DIFF
--- a/cli/src/commands/start.rs
+++ b/cli/src/commands/start.rs
@@ -61,8 +61,8 @@ pub struct Start {
     /// Specify the node's account private key.
     #[clap(long = "private-key")]
     pub private_key: Option<String>,
-    /// Specify the path to the node's account private key.
-    #[clap(long = "private-key-path")]
+    /// Specify the path to a file containing the node's account private key.
+    #[clap(long = "private-key-file")]
     pub private_key_path: Option<PathBuf>,
 
     /// Specify the IP address and port for the node server
@@ -170,6 +170,11 @@ impl Start {
 
     /// Read the private key directly from an argument or from a filesystem location.
     fn parse_private_key(&mut self) -> Result<()> {
+        // Ensure only one private key flag is provided to the CLI.
+        if self.private_key.is_some() && self.private_key_path.is_some() {
+            bail!("Both the private key string and file path flags were specified, please pick only one");
+        }
+
         // If the private key is provided directly, don't do anything else.
         if self.private_key.is_some() {
             return Ok(());
--- a/cli/src/commands/start.rs
+++ b/cli/src/commands/start.rs
@@ -45,18 +45,25 @@ pub struct Start {
     #[clap(default_value = "3", long = "network")]
     pub network: u16,
 
-    /// Specify this node as a beacon, with the account private key as an argument
+    /// Specify this node as a beacon
     #[clap(long = "beacon")]
-    pub beacon: Option<String>,
-    /// Specify this node as a validator, with the account private key as an argument
+    pub beacon: bool,
+    /// Specify this node as a validator
     #[clap(long = "validator")]
-    pub validator: Option<String>,
-    /// Specify this node as a prover, with the account private key as an argument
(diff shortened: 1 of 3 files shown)


joske (approved):

niklaslong (approved):

howardwu:
^This was Github. Sorry, I don't know why these "auto-closes" keep happening. In snarkVM, the branches usually retarget, rather than closing.

ljedrz:
Updated the PR to maintain the current node type parameter setup; with the new changes, the nodes should be started with:
```
--beacon --private-key <PRIVATE_KEY>
```
or
```
--validator --private-key-path <PATH>
```
That way the private key can be provided directly (as before) or from the filesystem.

cli/src/commands/start.rs
```diff
@@ -45,18 +45,25 @@ pub struct Start {
...
+    pub prover: bool,
+    /// Specify this node as a client
     #[clap(long = "client")]
-    pub client: Option<String>,
+    pub client: bool,
+
+    /// Specify the node's account private key.
+    #[clap(long = "private-key")]
+    pub private_key: Option<String>,
+    /// Specify the path to the node's account private key.
```
howardwu: We may want to provide more clarity to the user that this is the `file path` and perhaps the `absolute file path` (or relative if that is the case)

cli/src/commands/start.rs
```diff
@@ -45,18 +45,25 @@ pub struct Start {
...
+    #[clap(long = "private-key-path")]
+    pub private_key_path: Option<PathBuf>,
```
howardwu: nit: perhaps we rename this to `private-key-file` for more clarify? (as a `path` can denote a `file` or `folder`)

cli/src/commands/start.rs
```diff
@@ -160,6 +168,29 @@ impl Start {
         }
     }
 
+    /// Read the private key directly from an argument or from a filesystem location.
+    fn parse_private_key(&mut self) -> Result<()> {
```
howardwu, line 172: We should add a check that both cannot be provided:
For example:
```rs
// Ensure only one private key flag is provided to the CLI.
if self.private_key.is_some() && self.private_key_path.is_some() {
    bail!("Both the private key string and file path flags were specified, please pick only one")
}
```

howardwu:
We can also make the `--private-key*` variants optional on the `--dev` mode commands.

howardwu:
We should also forbid `--private-key ""` since the flag itself is optional now.

ljedrz:
> We can also make the --private-key* variants optional on the --dev mode commands.

This is already the case; `--dev` runs can omit those params, in which case a random key is sampled.

> We should also forbid --private-key "" since the flag itself is optional now.

Done in https://github.com/AleoHQ/snarkOS/pull/2465.

PR #2439, opened 2023-06-14
REST rate limiting

### LINKED ISSUE #1597: [Feature] RPC rate limiter
## 🚀 Feature

1. Add controls that allow snarkOS to adjust 2 rates:
    a. The rate that RPC requests from **all** IP addresses are processed
    b. The rate that RPC requests from a single IP address is processed

2. Add a "restricted_peers" list for RPC IP addresses that exceed the rate.

## Motivation

Some IP addresses are spamming the sync node RPC endpoints.

ljedrz:
This PR introduces a rate limiting service (with a limit of 5 reqs/s) to the REST server.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4574,6 +4574,7 @@ dependencies = [
  "pin-project",
  "pin-project-lite",
  "tokio",
+ "tokio-util",
  "tower-layer",
  "tower-service",
  "tracing",
--- a/node/rest/Cargo.toml
+++ b/node/rest/Cargo.toml
@@ -81,6 +81,7 @@ version = "1"
 
 [dependencies.tower]
 version = "0.4"
+features = ["buffer", "limit"]
 
 [dependencies.tower_governor]
 version = "0.3"
--- a/node/rest/src/lib.rs
+++ b/node/rest/src/lib.rs
@@ -36,18 +36,21 @@ use snarkvm::{
 use anyhow::Result;
 use axum::{
     body::Body,
+    error_handling::HandleErrorLayer,
     extract::{ConnectInfo, DefaultBodyLimit, Path, Query, State},
     http::{header::CONTENT_TYPE, Method, Request, StatusCode},
     middleware,
     middleware::Next,
     response::Response,
     routing::{get, post},
+    BoxError,
     Json,
 };
 use axum_extra::response::ErasedJson;
 use parking_lot::Mutex;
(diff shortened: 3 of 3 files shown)


joske (approved):

howardwu:
Can this be per IP?

howardwu:
^This was Github. Sorry, I don't know why these "auto-closes" keep happening. In snarkVM, the branches usually retarget, rather than closing.

niklaslong (dismissed):

niklaslong:
(pending the per IP Q)

ljedrz:
> Can this be per IP?

There is a [crate](https://crates.io/crates/tower_governor) that facilitates this, but it doesn't seem like it's widely used or well-maintained; the current version is `0.0.4` and the last commit is from 6 months ago.

howardwu:
Would it be inefficient/unwise for us to implement our own?

It seems we can use an LRUMap on 100K IPs, and track each IP's last `n` request timestamps, where `n` is `5 * 60` for `5 req/s` averaged over `60 seconds` (to give tolerance for bursty activity, potentially from network latency or scheduler latency)

howardwu:
Args, this is once again an auto-close by Github. A unique issue to only this repo.

joske:
and again :-D

howardwu:
@ljedrz I think we have RPS rate limiting now. Is this PR tangential?

howardwu:
#2942 says that this PR may be complementary. Can you add some color?

ljedrz:
The rate-limiting in this PR is not related to the source IP - it's a limit applicable to the entire REST server, to avoid a DDoS angle. However, if we are content with the limits we impose per IP, it may be redundant. We can always return to this if we have a stress test that executes many REST requests from unique source IPs, and determine that it's needed.

howardwu:
@ljedrz can you resolve the merge conflicts and set this limit to be at 100 requests per second?

For reader's context: snarkOS includes a **per IP** rate limiter that can be set as a CLI argument. By default, this is set to 10 requests per second (RPS) on a **per IP** basis. This PR is mean to limit the **total requests across all IPs** as an upper bound.

ljedrz:
Cc https://github.com/AleoHQ/snarkOS/issues/3190 for that one unrelated CI failure.

vicsn:
Closing because there is no need, it can disrupt operations without careful design, and can be fixed outside of snarkOS

Issue #2444, opened 2023-06-15
[Bug]snarkos transfer command not work


AlexZhao6666:
## 🐛 Bug Report
the transfer command for snarkos is not work,with prompting 'Function 'transfer' does not exist in program 'credits.aleo''

## Steps to Reproduce

update latest snarkos with branch 'testnet3'

## Expected Behavior

success transfer

## Your Environment

- 
snarkos: 2.1.1
- 
rust:1.70.0
- 
 macos 12.4
<img width="1470" alt="1" src="https://github.com/AleoHQ/snarkOS/assets/136443781/1b23815a-c555-48bd-b364-7541efcaec09">

reason:
snarkvm change:
<img width="502" alt="2" src="https://github.com/AleoHQ/snarkOS/assets/136443781/32e396ce-9fe9-4133-aca0-36a2f6cbc8b0">

howardwu:
The functions have been changed to `transfer_public` and `transfer_private`. @raychu86

PR #2446, opened 2023-06-15
Update `snarkos developer` to use `transfer_private`

raychu86:
## Motivation

This PR updates the `snarkos developer transfer` command to `snarkos developer transfer-private` in accordance with the new changes to `credits.aleo`.

This should address #2444


howardwu (approved):
LGTM

PR #2449, opened 2023-06-17
ignore: using for reviewing

howardwu:
## Motivation

(Write your motivation here)

## Test Plan

## Related PRs

(Link any related PRs here)

### DIFF
--- a/node/src/validator/router.rs
+++ b/node/src/validator/router.rs
@@ -14,27 +14,112 @@
 
 use super::*;
 
+use snarkos_node_bft_consensus::{batched_transactions, sort_transactions};
 use snarkos_node_messages::{
     BlockRequest,
     BlockResponse,
+    ConsensusId,
     Data,
     DataBlocks,
     DisconnectReason,
     Message,
     MessageCodec,
+    NewBlock,
     Pong,
     UnconfirmedTransaction,
 };
 use snarkos_node_tcp::{Connection, ConnectionSide, Tcp};
 use snarkvm::prelude::{error, EpochChallenge, Network, Transaction};
 
-use std::{io, net::SocketAddr, time::Duration};
-use tokio::task::spawn_blocking;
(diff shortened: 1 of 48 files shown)


node/src/validator/router.rs
```diff
@@ -38,6 +46,77 @@ impl<N: Network, C: ConsensusStorage<N>> P2P for Validator<N, C> {
...
+
+        // 1.
+        // BFT must be set here.
+        // TODO: we should probably use something else than the public key, potentially interactive, since this could
+        // be copied and reused by a malicious validator.
+        let public_key = self.primary_keypair.public();
+        let signature = self
+            .primary_keypair
+            .private()
+            .sign_bytes(public_key.to_bytes().as_slice(), &mut rand::thread_rng())
```
howardwu, line 72: Seems we should be signing the `last_executed_sub_dag_index`, otherwise run the risk of a MITM attack as well.

howardwu:
There doesn't appear to be any penalties for double signing or being offline.

node/src/validator/router.rs
```diff
@@ -38,6 +46,77 @@ impl<N: Network, C: ConsensusStorage<N>> P2P for Validator<N, C> {
...
+        // their public key.
+        if !consensus_id.signature.verify_bytes(&consensus_id.public_key, &consensus_id.public_key.to_bytes()) {
+            return Err(error(format!("'{peer_ip}' couldn't verify their identity")));
+        }
+
+        // 3.
+        // Track the committee member.
+        // TODO: in future we could error here if it already exists in the collection but that
+        // logic is probably best implemented when dynamic committees are being considered.
+        self.connected_committee_members.write().insert(peer_ip, consensus_id.public_key);
```
howardwu, line 107: ^Read the TODO first.

Need to dedup the public key. Otherwise, a compromised validator would be able to sabotage the actual validator.

PR #2454, opened 2023-06-20
Early signal handling

### LINKED ISSUE #1996: [Bug] snarkOS client has no SIGTERM/SIGINT handlers
## 🐛 Bug Report

snarkOS client does not respond to SIGTERM/SIGKILL when running as the root process in a Docker container. This is because in Linux, a process running with PID 1 will ignore SIGTERM/SIGINT unless a signal handler is installed. This means that running `docker stop` takes longer than necessary because it waits until the process exit timeout before sending SIGKILL. This gives the process no chance for graceful shutdown.

## Steps to Reproduce

1. Run the snarkOS client in a Docker container
2. Run `docker stop -t 60`
3. Twiddle your thumbs for 60 seconds while your container happily chugs along

## Expected Behavior

The snarkOS process should gracefully exit immediately

## Your Environment

- 
- 
-

ljedrz:
This is a new rendition of https://github.com/AleoHQ/snarkOS/pull/1997; it enables the handling of SIGTINT/SIGTERM signals earlier than we do right now, including the ledger loading and block syncing phases.

### DIFF
--- a/node/Cargo.toml
+++ b/node/Cargo.toml
@@ -44,6 +44,9 @@ version = "1"
 [dependencies.num_cpus]
 version = "1"
 
+[dependencies.once_cell]
+version = "1"
+
 [dependencies.parking_lot]
 version = "0.12"
 
--- a/node/src/beacon/mod.rs
+++ b/node/src/beacon/mod.rs
@@ -98,6 +98,9 @@ impl<N: Network, C: ConsensusStorage<N>> Beacon<N, C> {
     ) -> Result<Self> {
         let timer = timer!("Beacon::new");
 
+        // Initialize the signal handler.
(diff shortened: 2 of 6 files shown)


node/Cargo.toml
```diff
@@ -44,6 +44,9 @@ version = "1"
 [dependencies.num_cpus]
 version = "1"
 
+[dependencies.once_cell]
```
niklaslong, line 47: Can we use the now stabilised std version instead?

ljedrz: I've considered this before, but only a small subset has been stabilized so far - we would still need to use `once_cell` for now.

niklaslong (approved):

joske (approved):

PR #2455, opened 2023-06-20
Logging improvements

vvp:
This PR introduces few minor improvements on logging. Potentially user-space breaking change is that `RUST_LOG` environment variable now overrides the `--verbosity` setting (rationale is that `RUST_LOG` allows more fine-grained log level setting for debug/development purposes).

### DIFF
--- a/cli/src/helpers/logger.rs
+++ b/cli/src/helpers/logger.rs
@@ -12,10 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use crate::helpers::LogWriter;
+use std::{fs::File, io, path::Path};
 
 use crossterm::tty::IsTty;
-use std::{fs::File, io, path::Path};
 use tokio::sync::mpsc;
 use tracing_subscriber::{
     layer::{Layer, SubscriberExt},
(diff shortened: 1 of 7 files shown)


ljedrz (approved):
LGTM, though I'd squash these changes into a single commit for easier rebases later on

cli/src/helpers/logger.rs
```diff
@@ -12,25 +12,31 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use crate::helpers::LogWriter;
+use std::{env::VarError, fs::File, io, path::Path};
```
ljedrz: `clippy` complains that `env::VarError` is unused

node/src/validator/mod.rs
```diff
@@ -258,8 +258,9 @@ impl<N: Network, C: ConsensusStorage<N>> Validator<N, C> {
 
         // Load the primary's keys.
         let primary_key_file = format!("{base_path}.primary-{primary_id}-key.json");
-        let primary_keypair =
-            read_authority_keypair_from_file(primary_key_file).expect("Failed to load the node's primary keypair");
+        let primary_keypair = read_authority_keypair_from_file(primary_key_file).unwrap_or_else(|_| {
+            panic!("Failed to load the node's primary keypair {base_path}.primary-{primary_id}-key.json")
```
joske: why not add the actual error?

vvp: Oh good point, I was using `except(format!(...))` before and just blindly took clippy's advice for this replacement. Will add the error logging too.

PR #2456, opened 2023-06-20
Guard against overallocation when deserializing with bincode

ljedrz:
This change applies to network messages that involve collections.


howardwu (approved):
LGTM

PR #2466, opened 2023-06-22
feat: all validators produce blocks

joske:
* all validators create their own block
* disable sync
* broadcast to non-validators (clients & provers)

### DIFF
--- a/node/src/validator/mod.rs
+++ b/node/src/validator/mod.rs
@@ -15,7 +15,21 @@
 mod router;
 
 use crate::traits::NodeInterface;
+use anyhow::Result;
+use narwhal_config::{Committee, Import};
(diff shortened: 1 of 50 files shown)


node/src/validator/mod.rs
```diff
@@ -142,7 +142,7 @@ impl<N: Network, C: ConsensusStorage<N>> Validator<N, C> {
             node.rest = Some(Rest::start(rest_ip, Some(consensus), ledger, Arc::new(node.clone()))?);
         }
         // Initialize the sync pool.
-        node.initialize_sync()?;
+        // node.initialize_sync()?;
```
howardwu, line 145: For now this is okay, however we should keep this operational in case a node's quic goes down, or in case there are byzantine actors - the sync protocol will help to disconnect the byzantine validators.

joske:
no longer needed.

PR #2469, opened 2023-06-23
Use a custom function to deserialize the IndexMaps in BlockLocators

ljedrz:
A follow-up to https://github.com/AleoHQ/snarkOS/pull/2456.

This was found via simple fuzzing; it appears that `with_limit` doesn't fully guard us against overallocations, so introduce and use a custom method to deserialize the `IndexMap`s in `BlockLocators`.

Filing as a draft, as I'm not sure if we don't want to approach this some other way (e.g. restricting the number of entries etc.).

### DIFF
--- a/node/messages/src/helpers/block_locators.rs
+++ b/node/messages/src/helpers/block_locators.rs
@@ -16,8 +16,18 @@ use snarkvm::prelude::{has_duplicates, Network};
 
 use anyhow::{bail, ensure, Result};
 use indexmap::{indexmap, IndexMap};
-use serde::{Deserialize, Serialize};
-use std::collections::{btree_map::IntoIter, BTreeMap};
+use serde::{
+    de::{MapAccess, Visitor},
+    Deserialize,
+    Deserializer,
+    Serialize,
+};
+use std::{
+    collections::{btree_map::IntoIter, BTreeMap},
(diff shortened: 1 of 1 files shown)


howardwu:
What's the perf hit of using this custom deserializer?

Also, shouldn't this be brought up to the `indexmap` crate for fixing?

ljedrz:
> What's the perf hit of using this custom deserializer?

As far as I'm aware, there shouldn't be any; also, this implementation is basically the same as the one in `IndexMap`, just doesn't pre-allocate.

> shouldn't this be brought up to the indexmap crate for fixing?

AFAICT this would be a `serde` issue. Actually, I just looked at it and can see that there have been some related changes recently, so perhaps it would no longer be needed (i.e. it should be checked again).

Ultimately, though, it would be best to just use the `impl<N: Network> FromBytes for BlockLocators<N>` that was merged into the `narwhal` branch today, in order to have full control over deserialization. If that is the intent, this PR can be closed.

howardwu:
Let's switch out all of the Router messages to `To/FromBytes` just like the Gateway events.

ljedrz:
Superseded by https://github.com/AleoHQ/snarkOS/pull/2731.

PR #2480, opened 2023-06-29
Filters out fee transactions from the unconfirmed pool

howardwu:
## Motivation

This PR filters out fee transactions from the unconfirmed pool, as by protocol rules, we do not allow these.

Currently the protocol is safe even without this change, however it takes until reaching `propose_advance_to_next_block` to detect this. With this PR, we can detect it much sooner.


raychu86 (approved):
lgtm

PR #2493, opened 2023-07-04
Narwhal

joske:
* adjust some logging
* sign our own proposed_batch so the quorum is met also when only 3 validators (out of 4 present). This does lead to WARN about unknown batch when the 4th signature is received

### DIFF
--- a/node/narwhal/src/primary.rs
+++ b/node/narwhal/src/primary.rs
@@ -155,8 +155,19 @@ impl<N: Network> Primary<N> {
         // Sign the batch.
         let batch = Batch::new(private_key, round, transmissions, previous_certificates, &mut rng)?;
 
-        // Set the proposed batch.
-        *self.proposed_batch.write() = Some((batch.clone(), Default::default()));
+        // Initialize an RNG.
+        let rng = &mut rand::thread_rng();
+        // Generate a timestamp.
+        let timestamp = now();
+        // Sign the batch ID.
+        let signature = self.gateway.account().sign(&[batch.batch_id(), Field::from_u64(timestamp as u64)], rng)?;
+
+        // Set the proposed batch. The signature and timestamp need to be precomputed to include own
+        // stake in quorum threshold computations as BatchPropose and BatchSignature events are not
+        // broadcasted locally.
+        let mut signature_map = IndexMap::new();
+        signature_map.insert(signature, timestamp);
+        *self.proposed_batch.write() = Some((batch.clone(), signature_map));
 
         // Broadcast the batch to all validators for signing.
         self.gateway.broadcast(Event::BatchPropose(BatchPropose::new(Data::Object(batch.to_header()?))));
@@ -248,6 +259,7 @@ impl<N: Network> Primary<N> {
             for signature in signatures.keys() {
                 stake = stake.saturating_add(self.committee.get_stake(&signature.to_address()));
             }
+            debug!("Stake {stake}/{}", self.committee.quorum_threshold()?);
             // Check if the batch has reached the threshold.
             if stake >= self.committee.quorum_threshold()? {
                 info!("Quorum threshold reached, preparing to certify the batch");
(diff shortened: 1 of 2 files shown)


node/narwhal/src/primary.rs
```diff
@@ -155,8 +155,19 @@ impl<N: Network> Primary<N> {
         // Sign the batch.
         let batch = Batch::new(private_key, round, transmissions, previous_certificates, &mut rng)?;
 
-        // Set the proposed batch.
-        *self.proposed_batch.write() = Some((batch.clone(), Default::default()));
+        // Generate a timestamp.
+        let timestamp = now();
+        // Sign the batch ID.
+        let signature =
```
howardwu, line 161: Our signature is already in the `Batch` struct. This step should be skipped - the batch certificate will treat this as double signing.

Instead, when checking quorum threshold, one needs to pull out the signature there.

node/narwhal/src/primary.rs
```diff
@@ -286,7 +298,7 @@ impl<N: Network> Primary<N> {
         // Broadcast the certified batch to all validators.
         self.gateway.broadcast(Event::BatchCertified(event));
 
-        info!("\n\n\n\n\nOur batch for round {} has been certified!\n\n\n\n", self.committee.round());
+        info!("\n\n\n\n\nOur batch {batch_id} for round {} has been certified!\n\n\n\n", self.committee.round());
```
howardwu, line 301: What's the rationale for printing the `batch_id`? (It is very verbose and uninformative)

node/narwhal/src/primary.rs
```diff
@@ -248,6 +259,7 @@ impl<N: Network> Primary<N> {
             for signature in signatures.keys() {
                 stake = stake.saturating_add(self.committee.get_stake(&signature.to_address()));
             }
+            debug!("Stake {stake}/{}", self.committee.quorum_threshold()?);
```
howardwu, line 262: nit: Switch to `trace!`

node/narwhal/src/primary.rs
```diff
@@ -237,7 +248,7 @@ impl<N: Network> Primary<N> {
         if let Some((_, signatures)) = self.proposed_batch.write().as_mut() {
             // Add the signature to the batch.
             signatures.insert(signature, timestamp);
-            debug!("Added a batch signature from peer '{peer_ip}'");
+            debug!("Added a batch signature {} from peer '{peer_ip}'", batch_id);
```
howardwu, line 251: ditto: what's the rationale for `batch_id`?

node/narwhal/src/primary.rs
howardwu: I have fixed this in the `narwhal` branch here:
https://github.com/AleoHQ/snarkOS/commit/34813e31f6f226922b1f331b982384eabb839432

howardwu: Thanks for pointing this out

node/narwhal/src/primary.rs
howardwu: Actually, on second thought, I think this is okay, it could really go either way. Up to you!

node/narwhal/src/primary.rs
joske: I don't agree on uninformative, this helps to correlate the different log lines. Otherwise it's not clear what has been certified (although it should maybe rather be debug level).

node/narwhal/src/primary.rs
joske: Oh, if there's a better fix, sure. First thing I tried was shutting down one validator, and immediately the quorum stopped.

node/narwhal/src/primary.rs
joske: This was instrumental in finding the race condition for the WARN. But now that you have a better fix for the quorum threshold, I think most of this PR can just be abandoned. Do you agree @vvp ?

node/narwhal/src/primary.rs
joske: ditto above, this helped establish the order of the log lines.

node/narwhal/src/primary.rs
vvp: Yeah, I tested the latest fixes and they work better than this PR so let's just abandon this one.

joske:
superseded by a better fix.

PR #2495, opened 2023-07-05
fix: test_certificate_insert_remove

joske:
* fix this test
* implement Storage::remove_transmission

### DIFF
--- a/node/narwhal/src/helpers/storage.rs
+++ b/node/narwhal/src/helpers/storage.rs
@@ -296,7 +296,10 @@ impl<N: Network> Storage<N> {
         self.certificates.write().remove(&certificate_id);
         // Remove the batch ID.
         self.batch_ids.write().remove(&batch_id);
-        // TODO (howardwu): Remove the transmissions.
+        // Remove the transmissions.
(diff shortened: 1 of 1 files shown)


node/narwhal/src/helpers/storage.rs
```diff
@@ -296,7 +296,10 @@ impl<N: Network> Storage<N> {
         self.certificates.write().remove(&certificate_id);
         // Remove the batch ID.
         self.batch_ids.write().remove(&batch_id);
-        // TODO (howardwu): Remove the transmissions.
+        // Remove the transmissions.
```
howardwu, line 299: This case is not so easy.

We can only remove the transmission if this certificate is the **last** instance that references the transmission ID.

This was why it was initially left as a TODO to resolve.

joske: Oh, then we'll just keep this as a draft for now.

howardwu:
I've fixed it in this commit: https://github.com/AleoHQ/snarkOS/commit/16cae7a39c8a9b694fecd24e8d801a86d65b0cf4

joske:
superseded

PR #2502, opened 2023-07-05
Initial property test support

vvp:
This PR adds the `test-strategy` and `proptest` dependencies for convenient property testing, and initial proptests for `Committee` and `Gateway` structs. More tests will be added in following PRs, so this is ready for review.

Note that `test-strategy` has not yet officially released the support for async proptests, so the dependency points directly to Github for now (I haven't bumped into any quality issues).

### DIFF
--- a/node/narwhal/Cargo.toml
+++ b/node/narwhal/Cargo.toml
@@ -107,3 +107,8 @@ features = [ "fs", "trace" ]
 [dev-dependencies.tracing-subscriber]
 version = "0.3"
 features = [ "env-filter" ]
+
+[dev-dependencies]
+test-strategy = "0.3.0"
+proptest = "1.0.0"
+rand_xorshift = "0.3.0"
--- a/node/narwhal/Cargo.toml
+++ b/node/narwhal/Cargo.toml
@@ -109,6 +109,5 @@ version = "0.3"
 features = [ "env-filter" ]
 
 [dev-dependencies]
-test-strategy = "0.3.0"
+test-strategy = { git = "https://github.com/frozenlib/test-strategy/", branch = "master"}
 proptest = "1.0.0"
-rand_xorshift = "0.3.0"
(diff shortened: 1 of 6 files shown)


node/narwhal/Cargo.toml
```diff
@@ -107,3 +107,7 @@ features = [ "fs", "trace" ]
 [dev-dependencies.tracing-subscriber]
 version = "0.3"
 features = [ "env-filter" ]
+
+[dev-dependencies]
+test-strategy = { git = "https://github.com/frozenlib/test-strategy/", branch = "master"}
```
howardwu, line 112: Will this impact our ability to `cargo publish` this crate?

niklaslong: There are workarounds, see: https://github.com/rust-lang/cargo/issues/4242 but it's not ideal.

niklaslong: I tested this in another repo (publish dry run) and it seems to work fine, I think we can green-light this PR.

node/narwhal/src/gateway.rs
```diff
@@ -793,3 +794,128 @@ impl<N: Network> Gateway<N> {
...
+        };
+        let tcp_config = gateway.tcp().config();
+        assert_eq!(tcp_config.max_connections, MAX_COMMITTEE_SIZE);
+        assert_eq!(gateway.account().address(), address);
+    }
+
+    #[proptest(async = "tokio")]
+    async fn gateway_start(#[filter(|x| x.dev.is_some())] input: GatewayInput) {
+        let Some(dev) = input.dev else { unreachable!() };
+        println!("{input:?}");
```
joske: may want to remove the println!

vvp: :grinning: woops, good catch, removed now

node/narwhal/Cargo.toml
SplittyDev: @niklaslong I believe this behavior is also properly documented here: https://github.com/rust-lang/cargo/pull/7333. This is default behavior since 1.40 for dev-dependencies without a specified `version`.

PR #2507, opened 2023-07-06
[Narhwal] initial integration test tooling

niklaslong:
Copies over and adapts the setup code in `simple_node.rs` for integration test purposes.

### DIFF
--- a/node/narwhal/tests/common/primary.rs
+++ b/node/narwhal/tests/common/primary.rs
@@ -1,0 +1,126 @@
+// Copyright (C) 2019-2023 Aleo Systems Inc.
+// This file is part of the snarkOS library.
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at:
+// http://www.apache.org/licenses/LICENSE-2.0
+
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use snarkos_account::Account;
+use snarkos_node_narwhal::{
+    helpers::{init_primary_channels, Committee, PrimarySender, Storage},
+    Primary,
+    MAX_GC_ROUNDS,
+    MEMORY_POOL_PORT,
+};
+
+use tracing::*;
+
+use anyhow::Result;
+
+use indexmap::IndexMap;
+use parking_lot::RwLock;
+use rand::SeedableRng;
+use std::{net::SocketAddr, str::FromStr, sync::Arc};
+
+use crate::common::CurrentNetwork;
+
+/// Starts the primary instance.
+pub async fn start_primary(
+    node_id: u16,
+    num_nodes: u16,
+) -> Result<(Primary<CurrentNetwork>, PrimarySender<CurrentNetwork>)> {
+    // Sample a account.
+    let account = Account::new(&mut rand_chacha::ChaChaRng::seed_from_u64(node_id as u64))?;
+    println!("\n{account}\n");
+
+    // Initialize a map for the committee members.
+    let mut members = IndexMap::with_capacity(num_nodes as usize);
+    // Add the validators as members.
+    for i in 0..num_nodes {
+        // Sample the account.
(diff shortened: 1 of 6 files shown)


node/narwhal/tests/common/primary.rs
```diff
@@ -0,0 +1,102 @@
...
+    const INITIAL_STAKE: u64 = 1000;
+
+    let mut accounts = Vec::with_capacity(n as usize);
+    let mut members = IndexMap::with_capacity(n as usize);
+    for i in 0..n {
+        // Sample the account.
+        let account = Account::new(&mut rand_chacha::ChaChaRng::seed_from_u64(i as u64)).unwrap();
+
+        // TODO(nkls): use tracing instead.
+        println!("  Validator {}: {}", i, account.address());
```
ljedrz: looks like `tracing` is already imported and used, any reason why it's not used here yet?

node/narwhal/tests/common/utils.rs
```diff
@@ -0,0 +1,148 @@
...
+        // This RNG samples *different* fake solutions for each node.
+        let mut unique_rng = rand_chacha::ChaChaRng::seed_from_u64(node_id as u64);
+
+        // A closure to generate a commitment and solution.
+        fn sample(mut rng: impl Rng) -> (PuzzleCommitment<CurrentNetwork>, Data<ProverSolution<CurrentNetwork>>) {
+            // Sample a random fake puzzle commitment.
+            // TODO (howardwu): Use a mutex to bring in the real 'proof target' and change this sampling to a while loop.
+            let commitment = PuzzleCommitment::<CurrentNetwork>::from_g1_affine(rng.gen());
+            // Sample random fake solution bytes.
+            let solution = Data::Buffer(Bytes::from((0..1024).map(|_| rng.gen::<u8>()).collect::<Vec<_>>()));
```
ljedrz: instead of `gen`, `try_fill` would be more efficient; it would look sth like this:
```
let mut vec = vec![0u8; 1024];
vec.try_fill(&mut rng).unwrap();
let solution = Data::Buffer(Bytes::from(vec));
```
or
```
let mut vec = vec![0u8; 1024];
rng.fill(&mut vec);
let solution = Data::Buffer(Bytes::from(vec));
```

node/narwhal/tests/common/utils.rs
```diff
@@ -0,0 +1,148 @@
...
+        let mut unique_rng = rand_chacha::ChaChaRng::seed_from_u64(node_id as u64);
+
+        // A closure to generate an ID and transaction.
+        fn sample(
+            mut rng: impl Rng,
+        ) -> (<CurrentNetwork as Network>::TransactionID, Data<Transaction<CurrentNetwork>>) {
+            // Sample a random fake transaction ID.
+            let id = Field::<CurrentNetwork>::rand(&mut rng).into();
+            // Sample random fake transaction bytes.
+            let transaction = Data::Buffer(Bytes::from((0..1024).map(|_| rng.gen::<u8>()).collect::<Vec<_>>()));
```
ljedrz: same note on `try_fill`

node/narwhal/tests/e2e.rs
```diff
@@ -0,0 +1,42 @@
...
+        // Fire unconfirmed solutions.
+        fire_unconfirmed_solutions(&sender, id);
+        // Fire unconfirmed transactions.
+        fire_unconfirmed_transactions(&sender, id);
+    }
+
+    // TODO(nkls): the easiest would be to assert on the anchor or bullshark's output, once
+    // implemented.
+
+    // std::future::pending::<()>().await;
```
ljedrz: unless the test is `#[ignore]`, we won't want this

ljedrz (approved):

node/narwhal/tests/e2e.rs
niklaslong: Indeed, I left it in only to be able to run the nodes for a while and inspect the nodes until we figure out what the proper assertions should be.

node/narwhal/tests/common/utils.rs
niklaslong: done, used `fill_bytes` in the end

node/narwhal/tests/common/utils.rs
niklaslong: ditto

node/narwhal/tests/common/primary.rs
niklaslong: good catch, done

PR #2508, opened 2023-07-06
[Narwhal] test duplicates in ready queue

niklaslong:
Adds a test case to cover a duplicate insert in the ready queue. This gives the module decently complete test coverage.

### DIFF
--- a/node/narwhal/src/helpers/ready.rs
+++ b/node/narwhal/src/helpers/ready.rs
@@ -202,10 +202,13 @@ mod tests {
 
     #[test]
     fn test_ready_duplicate() {
+        use rand::RngCore;
         let rng = &mut TestRng::default();
 
         // Sample random fake bytes.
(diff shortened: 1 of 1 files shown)


node/narwhal/src/helpers/ready.rs
```diff
@@ -199,4 +199,28 @@ mod tests {
                 .collect::<IndexMap<_, _>>()
         );
     }
+
+    #[test]
+    fn test_ready_duplicate() {
+        let rng = &mut TestRng::default();
+
+        // Sample random fake bytes.
+        let data = |rng: &mut TestRng| Data::Buffer(Bytes::from((0..512).map(|_| rng.gen::<u8>()).collect::<Vec<_>>()));
```
ljedrz: nit: it would likely be more efficient to use [Fill](https://docs.rs/rand/latest/rand/trait.Fill.html) to generate many bytes at once

ljedrz (approved):

node/narwhal/src/helpers/ready.rs
niklaslong: done

howardwu (approved):

PR #2511, opened 2023-07-06
test: resolver

joske:


node/narwhal/src/helpers/resolver.rs
```diff
@@ -89,3 +89,39 @@ impl<N: Network> Resolver<N> {
...
     }
 }
+
+#[cfg(test)]
+pub mod tests {
+
+    use super::*;
+
+    type CurrentNetwork = snarkvm::prelude::Testnet3;
+    #[test]
```
ljedrz:
```suggestion
    type CurrentNetwork = snarkvm::prelude::Testnet3;
    
    #[test]
```

node/narwhal/src/helpers/resolver.rs
```diff
@@ -89,3 +89,39 @@ impl<N: Network> Resolver<N> {
...
+    fn test_resolver() {
+        let resolver = Resolver::<CurrentNetwork>::new();
+        let listener_ip = SocketAddr::from(([127, 0, 0, 1], 1234));
+        let peer_addr = SocketAddr::from(([127, 0, 0, 1], 4321));
+        let mut rng = rand::thread_rng();
```
ljedrz: nit: if possible, it's better to use `TestRng` (it's faster and provides the seed in case of failure), but in this case it's unlikely to make a difference

ljedrz (approved):

howardwu (approved):
LGTM

PR #2513, opened 2023-07-06
[Narwhal] some write lock tweaks

niklaslong:
A single write is usually faster than write-read-write, even if the last write isn't triggered.


howardwu (approved):
LGTM

PR #2516, opened 2023-07-07
Precompute total stake on committee changes 

vvp:
This PR changes the internal behavior of `Committee` by precomputing `total_stake` only on membership changes and storing it in memory. This saves us from redundant computing and overflow checking, and removes a couple of `Result`s from the API.

### DIFF
--- a/node/narwhal/src/helpers/committee.rs
+++ b/node/narwhal/src/helpers/committee.rs
@@ -124,6 +124,7 @@ impl<N: Network> Committee<N> {
 pub mod prop_tests {
     use crate::helpers::Committee;
     use snarkos_account::Account;
+    use std::collections::HashSet;
 
     use anyhow::Result;
     use indexmap::IndexMap;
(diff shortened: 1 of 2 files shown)


node/narwhal/src/helpers/committee.rs
```diff
@@ -17,12 +17,37 @@ use snarkvm::console::{prelude::*, types::Address};
...
+}
+
 #[derive(Clone, Debug, PartialEq)]
 pub struct Committee<N: Network> {
     /// The current round number.
     round: u64,
-    /// A map of `address` to `stake`.
-    members: IndexMap<Address<N>, u64>,
+    /// The current committee members in this `round`
+    committee: CommitteeMembers<N>,
```
joske: Maybe better `members`? Otherwise you get `Committee::committee` which looks a bit weird (to me at least :sweat:)

vvp: I tried that name first, but then in the functions like `is_committee_member` the code reads like `self.members.members.contains_key(...)` which is weird too :grin:

vvp: Folded `CommitteeMembers` into `Committee`, so naming is no longer an issue

joske (approved):

howardwu (approved):
LGTM

howardwu:
Superseded by #2522

howardwu:
Resolved merge conflicts, but lack push access.

PR #2526, opened 2023-07-10
Feat/ledgerservice

joske:
Add a LedgerService.

Open questions: 
* what operations to include (also consensus output?)
* what to do if ledger already has the certificate_id

### DIFF
--- a/node/narwhal/src/primary.rs
+++ b/node/narwhal/src/primary.rs
@@ -30,6 +30,7 @@ use crate::{
     CertificateResponse,
     Event,
     Gateway,
+    LedgerService,
     Worker,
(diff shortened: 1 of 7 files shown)


node/narwhal/src/primary.rs
```diff
@@ -95,6 +103,7 @@ impl<N: Network> Primary<N> {
             proposed_batch: Default::default(),
             pending: Default::default(),
             handles: Default::default(),
+            ledger_service: Arc::new(ledger_service),
```
ljedrz:
```suggestion
            ledger_service: Arc::from(ledger_service),
```

howardwu (approved):
LGTM

PR #2529, opened 2023-07-10
[Narwhal] first set of e2e tests 

niklaslong:
Tests:
- basic quorum threshold (connect nodes until quorum is reached)
- basic quorum break (break enough connections so that quorum is lost)

These cases unfortunately require the use of timeouts, we might be able to further cut down on run time but I opted for safe defaults initially. Logging is commented out for now but available at the top of each case if need be. We may also be able to improve ergonomics but I'll leave that to a follow up.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3090,6 +3090,7 @@ dependencies = [
  "axum-extra",
  "bincode 1.3.3",
  "bytes",
+ "deadline",
  "futures",
  "indexmap 2.0.0",
  "itertools",
--- a/node/narwhal/Cargo.toml
+++ b/node/narwhal/Cargo.toml
@@ -89,6 +89,9 @@ version = "0.7.4"
 default-features = false
 features = [ "erased-json" ]
 
+[dev-dependencies.deadline]
+version = "0.2"
+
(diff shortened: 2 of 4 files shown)


niklaslong:
Rebased.

node/narwhal/src/helpers/proposal.rs
```diff
@@ -114,7 +114,7 @@ impl<N: Network> Proposal<N> {
     /// Returns `true` if the quorum threshold has been reached for the proposed batch.
     pub fn is_quorum_threshold_reached(&self) -> bool {
         // Construct an iterator over the signers.
-        let signers = self.signatures.keys().chain([self.batch.signature()].into_iter()).map(Signature::to_address);
+        let signers = self.signatures.keys().chain([self.batch.signature()]).map(Signature::to_address);
```
ljedrz: is this needed? the code had compiled without this

niklaslong: oh, I think this was a clippy autofix, I'll remove it.

ljedrz (approved):

howardwu (approved):
LGTM

PR #2532, opened 2023-07-11
[Narwhal] A range of perf tweaks

ljedrz:
An initial batch found via code analysis and/or profiling. In addition there are a few small drive-by refactorings.

### DIFF
--- a/node/narwhal/src/helpers/committee.rs
+++ b/node/narwhal/src/helpers/committee.rs
@@ -20,6 +20,7 @@ use snarkvm::console::{
 };
 
 use indexmap::IndexMap;
+use std::collections::HashSet;
 
 #[derive(Clone, Debug, PartialEq)]
 pub struct Committee<N: Network> {
--- a/node/narwhal/src/helpers/committee.rs
+++ b/node/narwhal/src/helpers/committee.rs
@@ -169,7 +169,7 @@ impl<N: Network> Committee<N> {
     /// Note: This ensures the method returns a deterministic result that is SNARK-friendly.
     fn sorted_members(&self) -> indexmap::map::IntoIter<Address<N>, u64> {
         let members = self.members.clone();
-        members.sorted_by(|address1, stake1, address2, stake2| {
+        members.sorted_unstable_by(|address1, stake1, address2, stake2| {
             // Sort by stake in decreasing order.
             let cmp = stake2.cmp(stake1);
             // If the stakes are equal, sort by x-coordinate in decreasing order.
--- a/node/narwhal/src/helpers/committee.rs
+++ b/node/narwhal/src/helpers/committee.rs
@@ -91,12 +91,12 @@ impl<N: Network> Committee<N> {
 
     /// Returns `true` if the combined stake for the given addresses reaches the quorum threshold.
     /// This method takes in a `HashSet` to guarantee that the given addresses are unique.
-    pub fn is_quorum_threshold_reached(&self, addresses: impl IntoIterator<Item = Address<N>>) -> bool {
+    pub fn is_quorum_threshold_reached(&self, addresses: &HashSet<Address<N>>) -> bool {
         let mut stake = 0u64;
         // Compute the combined stake for the given addresses.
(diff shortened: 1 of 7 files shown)


ljedrz:
Marking as a draft, as I seem to need to update my `nightly` toolchain ~and double-check 2 tests :male_detective:~. The `node-narwhal` test failures appear to be unrelated, as I can reproduce them in the current `narwhal` branch. As far as I understand, https://github.com/AleoHQ/snarkOS/pull/2521 fixes them.

node/narwhal/src/helpers/committee.rs
```diff
@@ -162,7 +161,7 @@ impl<N: Network> Committee<N> {
     /// Note: This ensures the method returns a deterministic result that is SNARK-friendly.
     fn sorted_members(&self) -> Vec<(Address<N>, u64)> {
         let mut members = self.members.iter().map(|(address, stake)| (*address, *stake)).collect::<Vec<_>>();
-        members.sort_by(|(address1, stake1), (address2, stake2)| {
+        members.sort_unstable_by(|(address1, stake1), (address2, stake2)| {
```
ljedrz: this assumes that equal elements may be reordered

niklaslong (approved):
LGTM, pending that SNARK-friendliness Q with the unstable sort.

joske (approved):

node/narwhal/src/helpers/committee.rs
```diff
@@ -91,12 +91,12 @@ impl<N: Network> Committee<N> {
 
     /// Returns `true` if the combined stake for the given addresses reaches the quorum threshold.
     /// This method takes in a `HashSet` to guarantee that the given addresses are unique.
-    pub fn is_quorum_threshold_reached(&self, addresses: &HashSet<Address<N>>) -> bool {
+    pub fn is_quorum_threshold_reached(&self, addresses: impl IntoIterator<Item = Address<N>>) -> bool {
```
howardwu: This change should be reverted. See documentation above.

ljedrz: hmm, since we only need to ensure that the provided addresses are unique (which is currently the case, as all the sources come from `IndexMap`s and `IndexSet`s), perhaps we could verify the uniqueness with something like [itertools::unique](https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.unique)?

howardwu: snarkvm_utilities has a uniqueness check. Either way it is O(n) cost

node/narwhal/src/helpers/committee.rs
```diff
@@ -169,7 +169,7 @@ impl<N: Network> Committee<N> {
     /// Note: This ensures the method returns a deterministic result that is SNARK-friendly.
     fn sorted_members(&self) -> indexmap::map::IntoIter<Address<N>, u64> {
         let members = self.members.clone();
-        members.sorted_by(|address1, stake1, address2, stake2| {
+        members.sorted_unstable_by(|address1, stake1, address2, stake2| {
```
howardwu, line 172: Is this safe? Can you provide more context on the implications (I understand what `unstable` means here)

node/narwhal/src/helpers/committee.rs
ljedrz: indeed, so this would only be a possibly nicer API; I wonder if this is something we could `debug_assert` against in order to avoid runtime penalties (since right now all the use-cases are unique), but I'll revert this for now

node/narwhal/src/helpers/committee.rs
ljedrz: unstable sorting is just faster, so it's preferable if the order of equal elements doesn't matter; in our case, the `members` are initially sorted by insertion order and already deduplicated by the `Address` (since it's the key used for hashing), so unless we could end up with identical X coordinates for different `Address`es _and_ would still want to retain the insertion order for such equal sets (which would also need to have equal stakes), we are free to do unstable sorting.

howardwu (approved):
LGTM

PR #2533, opened 2023-07-12
[Narwhal] cache tests

niklaslong:
This PR extends the current cache test coverage to all fields.

### DIFF
--- a/node/narwhal/src/helpers/cache.rs
+++ b/node/narwhal/src/helpers/cache.rs
@@ -135,150 +135,72 @@ mod tests {
 
     type CurrentNetwork = Testnet3;
 
-    #[test]
-    fn test_inbound_connection() {
-        let cache = Cache::<CurrentNetwork>::default();
-        let peer_ip = IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1));
-
-        // Check that the cache is empty.
-        assert_eq!(cache.seen_inbound_connections.read().len(), 0);
-
-        // Insert a connection.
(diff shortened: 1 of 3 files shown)


node/narwhal/src/helpers/cache.rs
```diff
@@ -131,47 +131,74 @@ mod tests {
...
+                    }
+                }
+            )*
+        }
+    }
 
-        // Check that the cache still contains the transmission.
-        assert_eq!(cache.seen_inbound_transmissions.read().len(), 1);
+    test_cache_fields! {
+       inbound_connection with IpAddr,
```
niklaslong, line 196: Note: using `paste` to construct the macro means the common name subset (field + insert func) needs to be passed in here as opposed to the full field name.

node/narwhal/src/helpers/cache.rs
```diff
@@ -131,47 +131,76 @@ mod tests {
...
+                    }
+                }
+            )*
+        }
+    }
 
-        // Check that the cache still contains the transmission.
-        assert_eq!(cache.seen_inbound_transmissions.read().len(), 1);
+    test_cache_fields! {
+       inbound_connection,
```
niklaslong, line 198: Note: using paste (additive only) to construct the macro means the common name subset (field + insert func) needs to be passed in here as opposed to the full field name.

ljedrz (approved):

howardwu (approved):
LGTM

PR #2534, opened 2023-07-12
Feat/dag test

joske:
add a few tests for DAG. Due to non-determinism of `Certificate.round`, it's harder to create more elaborate scenarios. We should have a function in snarkVM to sample a `BatchCertificate` for a given round, not a random one.


ljedrz (commented):
LGTM, though I'm not sure about the retain change

howardwu (approved):
LGTM

PR #2537, opened 2023-07-13
[Narwhal] proptest design iteration

vvp:
This PR changes snarkos-narwhal's proptesting approach towards more composable way. Instead of writing a new struct that mimics the input data of test target, only to be able to `#[derive(Arbitrary)]` - we now use more context-specific approach:

- compose module-internal strategy-fns for failure cases and specify them in proptest arguments e.g.  `#[strategy(too_small_committee())]`
- compose public strategy-fns for success cases, so that other strategies depending on valid structs can be composed as well; if there's a single type of valid struct that makes sense, or if too much flexibility hurts the shrinking performance, then implement `Arbitrary` as a canonical strategy for the type.

The biggest benefits are the improved composability and not having to use `#[filter(TestInput::is_valid())]` everywhere. This PR also adds few more proptests to validate this. 

Few limitations and workarounds I ran into: 
- as we cannot implement `Arbitrary` for foreign types (e.g. the ones in `snarkvm-narwhal`) we then need to fallback to strategy-fns or wrap them into local types. Both have their use cases.
- `Strategy` requires that the generated value implements `Debug`, this isn't the case in all our test targets so a custom test implementation may be needed.
- In some cases, we want to make sure that generated values share some input value (e.g. use the same `Storage`) - the natural way would be using `any_with(input_value)` strategy, but that requires that input value implements `Default`.


node/narwhal/src/helpers/committee.rs
```diff
@@ -324,118 +324,174 @@ pub mod prop_tests {
...
-            }
+    #[derive(Debug, Clone)]
+    pub struct ValidatorSet(pub HashSet<Validator>);
+
+    impl Default for ValidatorSet {
+        fn default() -> Self {
+            ValidatorSet(
+                (0..=4u64)
+                    .map(|i| Validator {
+                        account: Account::new(&mut rand_chacha::ChaChaRng::seed_from_u64(i)).unwrap(),
```
ljedrz:
```suggestion
                        account: Account::new(&mut TestRng::fixed(i)).unwrap(),
```

node/narwhal/src/helpers/committee.rs
```diff
@@ -324,118 +324,174 @@ pub mod prop_tests {
...
+            .boxed()
+    }
+
+    fn invalid_stake_validator() -> BoxedStrategy<Validator> {
+        (0..MIN_STAKE, any_valid_account()).prop_map(|(stake, account)| Validator { stake, account }).boxed()
+    }
+
+    pub fn any_valid_account() -> BoxedStrategy<Account<CurrentNetwork>> {
+        any::<u64>()
+            .prop_map(|seed| match Account::new(&mut rand_chacha::ChaChaRng::seed_from_u64(seed)) {
```
ljedrz:
```suggestion
            .prop_map(|seed| match Account::new(&mut TestRng::fixed(seed)) {
```

node/narwhal/src/helpers/committee.rs
```diff
@@ -324,118 +324,174 @@ pub mod prop_tests {
...
+
+    fn validator_set<T: Strategy<Value = Validator>>(
+        element: T,
+        size: impl Into<SizeRange>,
+    ) -> impl Strategy<Value = ValidatorSet> {
+        hash_set(element, size).prop_map(ValidatorSet)
+    }
+
+    fn invalid_round_committee() -> BoxedStrategy<Result<Committee<CurrentNetwork>>> {
+        (Just(0), validator_set(any_valid_validator(), size_range(4..=MAX_COMMITTEE_SIZE as usize)))
```
ljedrz:
```suggestion
        (Just(0), validator_set(any_valid_validator(), size_range(3..=MAX_COMMITTEE_SIZE as usize)))
```
at least marked with a TODO if it doesn't work yet

node/narwhal/src/helpers/committee.rs
```diff
@@ -324,118 +324,174 @@ pub mod prop_tests {
...
+            .prop_map(to_committee)
+            .boxed()
+    }
+
+    fn too_small_committee() -> BoxedStrategy<Result<Committee<CurrentNetwork>>> {
+        (1u64.., validator_set(any_valid_validator(), 0..4)).prop_map(to_committee).boxed()
```
ljedrz:
```suggestion
        (1u64.., validator_set(any_valid_validator(), 0..3)).prop_map(to_committee).boxed()
```

node/narwhal/src/helpers/committee.rs
```diff
@@ -324,118 +324,174 @@ pub mod prop_tests {
...
-                Ok(account) => account,
-                Err(err) => panic!("Failed to create account {err}"),
-            }
+    #[derive(Debug, Clone)]
+    pub struct ValidatorSet(pub HashSet<Validator>);
+
+    impl Default for ValidatorSet {
+        fn default() -> Self {
+            ValidatorSet(
+                (0..=4u64)
```
ljedrz: just to double-check: this means the default set consists of **5** validators

node/narwhal/src/helpers/committee.rs
```diff
@@ -444,4 +500,28 @@ pub mod prop_tests {
...
+        if let Err(err) = committee {
+            assert_eq!(err.to_string().as_str(), "All members must have sufficient stake");
+        }
+    }
+
+    #[proptest]
+    fn invalid_member_count(#[strategy(too_small_committee())] committee: Result<Committee<CurrentNetwork>>) {
+        assert!(committee.is_err());
+        if let Err(err) = committee {
+            assert_eq!(err.to_string().as_str(), "Committee must have at least 4 members");
```
ljedrz:
```suggestion
            assert_eq!(err.to_string().as_str(), "Committee must have at least 3 members");
```
see the earlier comment

node/narwhal/src/helpers/committee.rs
```diff
@@ -444,4 +500,28 @@ pub mod prop_tests {
...
+        }
+    }
+
+    #[proptest]
+    fn invalid_round(#[strategy(invalid_round_committee())] committee: Result<Committee<CurrentNetwork>>) {
```
ljedrz: if the round can't be zero, why not change its type within `Committee` to `NonZeroU64`? that would make this test any and related checks redundant

node/narwhal/src/helpers/storage.rs
```diff
@@ -817,22 +817,258 @@ pub mod tests {
...
+    use rand::{CryptoRng, Error, Rng, RngCore};
+
+    use snarkvm::{
+        ledger::narwhal::{Batch, Data},
+        prelude::Signature,
+    };
+
+    // The `proptest::TestRng` doesn't implement `rand_core::CryptoRng` trait which is required in snarkVM, so we use a wrapper
+    #[derive(Debug)]
+    pub struct CryptoTestRng(TestRng);
```
ljedrz, line 871: why not use the `TestRng` from `snarkvm::prelude`? that one impls `CryptoRng`; this applies to all the other spots where I suggest the use of `TestRng`

node/narwhal/src/helpers/storage.rs
```diff
@@ -817,22 +817,258 @@ pub mod tests {
...
+        fn arbitrary_with(_: Self::Parameters) -> Self::Strategy {
+            any_transmission_id().prop_map(AnyTransmissionID).boxed()
+        }
+    }
+
+    fn any_transmission() -> BoxedStrategy<Transmission<CurrentNetwork>> {
+        prop_oneof![
+            (collection::vec(any::<u8>(), 512..=512))
+                .prop_map(|bytes| Transmission::Solution(Data::Buffer(Bytes::from(bytes)))),
+            (collection::vec(any::<u8>(), 2048..=2048))
```
ljedrz, line 929: why the fixed sizes? if we don't care about the contents at all, the buffer might as well be empty or have 1 byte

node/narwhal/src/helpers/storage.rs
```diff
@@ -817,22 +817,258 @@ pub mod tests {
...
+    impl ValidatorSet {
+        fn sign_batch_header<R: Rng + CryptoRng>(
+            &self,
+            batch_header: &BatchHeader<CurrentNetwork>,
+            rng: &mut R,
+        ) -> IndexMap<Signature<CurrentNetwork>, i64> {
+            let mut signatures = IndexMap::with_capacity(self.0.len());
+            for validator in self.0.iter() {
+                let private_key = validator.account.private_key();
+                let timestamp = time::OffsetDateTime::now_utc().unix_timestamp();
```
ljedrz: perhaps it would be preferable to obtain the timestamp just once before the loop? or do we want all of the validators to have a (very) slightly different one?

node/narwhal/src/helpers/storage.rs
```diff
@@ -817,22 +817,258 @@ pub mod tests {
...
+                let timestamp_field = Field::from_u64(timestamp as u64);
+                signatures
+                    .insert(private_key.sign(&[batch_header.batch_id(), timestamp_field], rng).unwrap(), timestamp);
+            }
+            signatures
+        }
     }
+    use crate::helpers::committee::prop_tests::{CommitteeContext, ValidatorSet};
+    use proptest::sample::{size_range, Selector};
```
ljedrz: those should be moved to the top of the file

node/narwhal/src/helpers/storage.rs
```diff
@@ -597,7 +597,7 @@ pub mod tests {
     type CurrentNetwork = snarkvm::prelude::Testnet3;
 
     /// Asserts that the storage matches the expected layout.
-    fn assert_storage<N: Network>(
+    pub fn assert_storage<N: Network>(
```
ljedrz, line 601: it's not from this PR, but all of the params to this function should be references - there is zero need to pass them by value, which requires cloning in tests

node/narwhal/src/worker.rs
```diff
@@ -365,40 +365,32 @@ impl<N: Network> Worker<N> {
...
-            }
+    fn invalid_worker_id(
+        #[strategy(MAX_WORKERS..)] id: u8,
+        gateway: Gateway<CurrentNetwork>,
+        storage: Storage<CurrentNetwork>,
+    ) {
+        let worker = Worker::new(id, gateway, storage, Default::default());
+        assert!(worker.is_err());
+        if let Err(error) = worker {
+            assert_eq!(error.to_string().as_str(), format!("Invalid worker ID '{}'", id));
```
ljedrz:
```suggestion
            assert_eq!(error.to_string(), format!("Invalid worker ID '{}'", id));
```
this should be redundant?

ljedrz (commented):
Left some questions and suggestions. I like the overall design.

node/narwhal/src/gateway.rs
```diff
@@ -834,65 +834,132 @@ impl<N: Network> Gateway<N> {
...
+        let gateway = Gateway::new(account, storage, dev);
+        assert!(gateway.is_ok());
+        let gateway = gateway.unwrap();
+        let tcp_config = gateway.tcp().config();
+        assert_eq!(tcp_config.listener_ip, Some(IpAddr::V4(Ipv4Addr::LOCALHOST)));
+        assert_eq!(tcp_config.desired_listening_port, Some(MEMORY_POOL_PORT + dev.unwrap()));
+
+        let tcp_config = gateway.tcp().config();
+        assert_eq!(tcp_config.max_connections, MAX_COMMITTEE_SIZE);
+        assert_eq!(gateway.account().address(), address);
```
niklaslong: I think there's some unnecessary variable shadowing in this case? I'd perhaps write it like this? 
```suggestion
        let gateway = Gateway::new(account, storage, dev);
        let gateway = gateway.unwrap();
        assert_eq!(gateway.account().address(), address);
        
        let tcp_config = gateway.tcp().config();
        assert_eq!(tcp_config.listener_ip, Some(IpAddr::V4(Ipv4Addr::LOCALHOST)));
        assert_eq!(tcp_config.desired_listening_port, Some(MEMORY_POOL_PORT + dev.unwrap()));
        assert_eq!(tcp_config.max_connections, MAX_COMMITTEE_SIZE);
```

node/narwhal/src/gateway.rs
```diff
@@ -834,65 +834,132 @@ impl<N: Network> Gateway<N> {
...
+    }
+
+    #[proptest]
+    fn gateway_prod_initialization(#[strategy(any_valid_prod_gateway())] input: GatewayInput) {
```
niklaslong, line 954: Ditto

node/narwhal/src/gateway.rs
```diff
@@ -834,65 +834,132 @@ impl<N: Network> Gateway<N> {
...
+    #[proptest(async = "tokio")]
+    async fn gateway_start(
+        #[strategy(any_valid_dev_gateway())] input: GatewayInput,
+        #[strategy(0..MAX_WORKERS)] workers_count: u8,
+    ) {
+        let (storage, _, account, dev) = input;
+        let worker_storage = storage.clone();
+        let gateway = Gateway::new(account, storage, dev);
+        assert!(gateway.is_ok());
+        let gateway = gateway.unwrap();
```
niklaslong:
```suggestion
        let gateway = gateway.unwrap();
```
(The test will also panic when this is an err.)

node/narwhal/src/helpers/proposal.rs
```diff
@@ -163,3 +163,49 @@ impl<N: Network> Proposal<N> {
...
+            now(),
+            transmission_map.clone(),
+            Default::default(),
+            &mut rng,
+        )
+        .unwrap();
+
+        let proposal = Proposal::new(committee, batch.clone());
+        assert!(proposal.is_ok());
+        let proposal = proposal.unwrap();
```
niklaslong:
```suggestion
        let proposal = proposal.unwrap();
```
Same as earlier, I think the `assert` is not needed?

node/narwhal/src/worker.rs
```diff
@@ -365,40 +365,32 @@ impl<N: Network> Worker<N> {
...
-            self.id < MAX_WORKERS
-        }
+    #[proptest]
+    fn worker_initialization(
+        #[strategy(0..MAX_WORKERS)] id: u8,
+        gateway: Gateway<CurrentNetwork>,
+        storage: Storage<CurrentNetwork>,
+    ) {
+        let worker = Worker::new(id, gateway, storage, Default::default());
+        assert!(worker.is_ok());
```
niklaslong:
```suggestion
```

node/narwhal/src/worker.rs
```diff
@@ -365,40 +365,32 @@ impl<N: Network> Worker<N> {
...
-                assert_eq!(e.to_string().as_str(), format!("Invalid worker ID '{}'", input.id));
-            }
+    fn invalid_worker_id(
+        #[strategy(MAX_WORKERS..)] id: u8,
+        gateway: Gateway<CurrentNetwork>,
+        storage: Storage<CurrentNetwork>,
+    ) {
+        let worker = Worker::new(id, gateway, storage, Default::default());
+        assert!(worker.is_err());
+        if let Err(error) = worker {
```
niklaslong, line 482: I think you could flatten this with https://doc.rust-lang.org/std/result/enum.Result.html#method.unwrap_err (and remove the `assert`)?

node/narwhal/src/helpers/committee.rs
vvp: Actually the original is correct, as we are only generating cases here where the round is invalid (so committee size must be valid).

node/narwhal/src/helpers/committee.rs
vvp: The original is correct, we want to test all committee sizes smaller than the minimum. `0..=3`  would be communicate the intent better I guess.

node/narwhal/src/helpers/committee.rs
vvp: Ah good catch - however I only implement `Default` because of the type bounds in `Arbitrary::Parameters`, we don't actually use it anywhere.

node/narwhal/src/helpers/committee.rs
vvp: I would've used `TestRng::fixed` otherwise, but it prints out the seed everytime, which in proptesting results in tons of spam in the test output, especially if shrinking.

node/narwhal/src/helpers/committee.rs
vvp: https://github.com/AleoHQ/snarkOS/pull/2537#discussion_r1264311767 and https://github.com/AleoHQ/snarkOS/pull/2537#discussion_r1264310224

node/narwhal/src/helpers/committee.rs
vvp: Yep I totally agree that would be a better solution :+1: It's out of the scope of this PR though so let's fix that in another PR.

node/narwhal/src/helpers/storage.rs
vvp: Two main reasons:
- the `proptest` crate implements its own `rand::RngCore`  which it uses internally (e.g. `prop_perturb` returns a strategy where rng is always identical during shrinking). If we wanted to use our `TestRng` instead, we would need to add explicit seed generation into all places where we generate anything that needs `TestRng` to keep shrinking proceed deterministically - so more boilerplate. Also, we might miss out on future improvements on shrinking that `proptest`'s `rand::RngCore` impl is able to use because they control the rng state internally.
- `TestRng::fixed` outputs the seed, which is a lot of spam e.g. during shrinking.

node/narwhal/src/helpers/storage.rs
vvp: Good question. I basically converted the [existing  `sample_transmission`](https://github.com/AleoHQ/snarkOS/blob/5261addfaf42ec28ac0e20abbbb6376b8d87ccd9/node/narwhal/src/helpers/storage.rs#L620-L631)- function into a strategy. 

I can check with @howardwu what's the minimum amount of random data needed to make this random data generation a bit faster, and submit the changes in a separate PR.

node/narwhal/src/helpers/storage.rs
vvp: As this fn is strictly for testing purposes, the slightly different signature timestamp for each validator is the one that reflects the target system behavior better. That said, proptests would be a better fit for testing the `BatchCertificate` and related logic for any combinations of timestamp variation.

node/narwhal/src/worker.rs
vvp: Oh yeah, that's much better :+1:

node/narwhal/src/helpers/storage.rs
vvp: Yeah, good catch. I can change that in separate PR.

node/narwhal/src/worker.rs
vvp: Good catch, fixing it with https://github.com/AleoHQ/snarkOS/pull/2537#discussion_r1262901582

node/narwhal/src/worker.rs
vvp: ah, `Worker` doesn't implement `Debug` yet so `unwrap_err` doesn't work. I'll remove the `assert` though.

node/narwhal/src/helpers/committee.rs
ljedrz: hmm, in that case perhaps we could use [XorShiftRng](https://docs.rs/rand_xorshift/0.3.0/rand_xorshift/struct.XorShiftRng.html) instead?

vvp: Good idea. Let's do that in a separate PR.

vvp:
Note to reviewers:
1. @niklaslong @ljedrz your review comments were addressed in commit https://github.com/AleoHQ/snarkOS/pull/2537/commits/8fa5f2fdd406c4cc95e7a6cebbfcdaa5d1353229
2. resolving the merge to `narwhal` wasn't straight-forward due to #2543 and the removal of `Batch` struct in `snarkVM` - @joske can you check the merge commit https://github.com/AleoHQ/snarkOS/pull/2537/commits/8fa5f2fdd406c4cc95e7a6cebbfcdaa5d1353229 if it's done along your ideas 
3. As there were weird, unfixable clippy warnings, I decided to upgrade the `test-strategy` crate to a freshly released version which included fixes (also got rid of the git-dep :partying_face: ) - however, then clippy started warning about dead_code when using `#[strategy(fn)]` on function parameter, but _only_ in `node/narwhal/committee/src/prop_tests.rs` for some reason.

joske (approved):

ljedrz (approved):

node/narwhal/committee/src/prop_tests.rs
```diff
@@ -133,3 +196,27 @@ fn committee_members(input: CommitteeInput) {
...
+        assert_eq!(err.to_string().as_str(), "Committee must have at least 4 members");
+    }
+}
+
+#[proptest]
+fn invalid_round(#[strategy(invalid_round_committee())] committee: Result<Committee<CurrentNetwork>>) {
+    assert!(committee.is_err());
+    if let Err(err) = committee {
+        assert_eq!(err.to_string().as_str(), "Round must be nonzero");
+    }
```
niklaslong: You might be able to do:
```suggestion
    assert!(matches!(committee, Err(e) if e.to_string().as_str() == "Round must be nonzero"))
```

node/narwhal/src/worker.rs
niklaslong: You could also maybe use a match (see above)

niklaslong (approved):

vvp:
Ok merging this finally, thanks everybody :pray:

PR #2539, opened 2023-07-14
[Narwhal] improve e2e test tooling

niklaslong:
This PR introduces a few structures to improve the ergonomics of writting e2e tests, notably`TestNetworkConfig` which can be used to construct a `TestNetwork`. Task handles (for the cannons and logging) are tracked on a per-primary basis, which will be helpful for cases requiring primary restarts. I've also introduced a few convenience methods to check for network halts and quorum reaching a certain round.

### DIFF
--- a/node/narwhal/tests/common/primary.rs
+++ b/node/narwhal/tests/common/primary.rs
@@ -43,7 +43,7 @@ use tracing::*;
 #[derive(Clone)]
 pub struct TestNetworkConfig {
     pub num_nodes: u16,
-    pub initiate_connections: bool,
+    pub connect_all: bool,
     pub fire_cannons: bool,
     pub log_level: Option<u8>,
     pub log_connections: bool,
@@ -87,7 +87,6 @@ impl TestPrimary {
     }
 
     pub fn log_connections(&mut self) {
-        // let node = self.clone();
         let self_clone = self.clone();
         self.handles.write().push(tokio::task::spawn(async move {
             loop {
@@ -103,6 +102,7 @@ impl TestPrimary {
 }
 
 impl TestNetwork {
+    // Creates a new test network with the given configuration.
     pub fn new(config: TestNetworkConfig) -> Self {
         if let Some(log_level) = config.log_level {
             initialize_logger(log_level);
@@ -123,6 +123,7 @@ impl TestNetwork {
         Self { config, primaries }
     }
 
+    // Starts each node in the network.
     pub async fn start(&mut self) {
         for primary in self.primaries.values_mut() {
             // Setup the channels and start the primary.
@@ -139,29 +140,37 @@ impl TestNetwork {
             }
         }
 
-        if self.config.initiate_connections {
-            initiate_connections(&self.primaries).await;
+        if self.config.connect_all {
+            self.connect_all().await;
         }
     }
 
+    // Starts the solution and trasnaction cannons for node.
     pub fn fire_cannons(&mut self, id: u16) {
         self.primaries.get_mut(&id).unwrap().fire_cannons();
     }
 
+    // Connects a node to another node.
     pub async fn connect(&self, id: u16, peer_id: u16) {
         let primary = self.primaries.get(&id).unwrap();
         let peer_ip = self.primaries.get(&peer_id).unwrap().gateway().local_ip();
         primary.gateway().connect(peer_ip);
         // Give the connection time to be established.
         sleep(Duration::from_millis(100)).await;
-
-        //  // TODO(nkls): maybe deadline could be used here instead?
-        //  let primary_clone = primary.clone();
-        //  deadline::deadline!(std::time::Duration::from_millis(100), move || {
-        //      primary_clone.gateway().is_connected(peer_ip)
-        //  });
     }
 
+    // Connects all nodes to each other.
+    pub async fn connect_all(&self) {
+        for (primary, other_primary) in self.primaries.values().tuple_combinations() {
+            // Connect to the node.
+            let ip = other_primary.gateway().local_ip();
+            primary.gateway().connect(ip);
+            // Give the connection time to be established.
+            tokio::time::sleep(std::time::Duration::from_millis(100)).await;
+        }
+    }
+
+    // Disconnects a node from all other nodes.
     pub async fn disconnect(&self, num_nodes: u16) {
         for id in 0..num_nodes {
             let primary = self.primaries.get(&id).unwrap();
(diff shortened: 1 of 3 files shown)


node/narwhal/tests/common/primary.rs
```diff
@@ -12,23 +12,193 @@
...
+};
+
 use indexmap::IndexMap;
 use itertools::Itertools;
-use std::collections::HashMap;
+use parking_lot::RwLock;
+use tokio::{task::JoinHandle, time::sleep};
 use tracing::*;
 
+#[derive(Clone)]
```
ljedrz:
```suggestion
#[derive(Clone, Copy, Debug)]
```

node/narwhal/tests/common/primary.rs
```diff
@@ -12,23 +12,193 @@
...
+    fn deref(&self) -> &Self::Target {
+        &self.primary
+    }
+}
+
+impl DerefMut for TestPrimary {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        &mut self.primary
+    }
+}
```
ljedrz: I'm not convinced about these `Deref` impls; recall what the `std` [docs](https://doc.rust-lang.org/std/ops/trait.Deref.html) say about them:
> the rules regarding Deref and DerefMut were designed specifically to accommodate smart pointers. Because of this, Deref should only be implemented for smart pointers to avoid confusion.

node/narwhal/tests/common/primary.rs
```diff
@@ -12,23 +12,193 @@
...
+    pub config: TestNetworkConfig,
+    pub primaries: HashMap<u16, TestPrimary>,
+}
+
+#[derive(Clone)]
+pub struct TestPrimary {
+    pub id: u16,
+    pub primary: Primary<CurrentNetwork>,
+    pub sender: Option<PrimarySender<CurrentNetwork>>,
+    pub handles: Arc<RwLock<Vec<JoinHandle<()>>>>,
```
ljedrz:
```suggestion
    pub handles: Arc<Mutex<Vec<JoinHandle<()>>>>,
```
these won't be read that much

node/narwhal/tests/common/primary.rs
```diff
@@ -12,23 +12,193 @@
...
+    pub connect_all: bool,
+    pub fire_cannons: bool,
+    pub log_level: Option<u8>,
+    pub log_connections: bool,
+}
+
+#[derive(Clone)]
+pub struct TestNetwork {
+    pub config: TestNetworkConfig,
+    pub primaries: HashMap<u16, TestPrimary>,
```
ljedrz: if the test IDs are consecutive and start at `0`, this might as well be a `Vec<TestPrimary>`

node/narwhal/tests/common/primary.rs
```diff
@@ -12,23 +12,193 @@
...
+        sleep(Duration::from_millis(100)).await;
+    }
+
+    // Checks a quorum of nodes have reached the given round.
+    pub fn is_round_reached(&self, round: u64) -> bool {
+        let quorum_threshold = self.primaries.len() / 2 + 1;
+        self.primaries.values().filter(|p| p.current_round() >= round).count() >= quorum_threshold
+    }
+
+    // Checks all nodes have stopped progressing.
```
ljedrz:
```suggestion
    // Checks if all the nodes have stopped progressing.
```

node/narwhal/tests/common/primary.rs
```diff
@@ -12,23 +12,193 @@
...
+            }
+        }
+
+        if self.config.connect_all {
+            self.connect_all().await;
+        }
+    }
+
+    // Starts the solution and trasnaction cannons for node.
+    pub fn fire_cannons(&mut self, id: u16) {
```
ljedrz:
```suggestion
    pub fn fire_cannons_at(&mut self, id: u16) {
```

node/narwhal/tests/common/primary.rs
```diff
@@ -12,23 +12,193 @@
...
+        self.primaries.get_mut(&id).unwrap().fire_cannons();
+    }
+
+    // Connects a node to another node.
+    pub async fn connect(&self, id: u16, peer_id: u16) {
```
ljedrz:
```suggestion
    pub async fn connect_primaries(&self, first_id: u16, second_id: u16) {
```

ljedrz: this API is a bit confusing; I was first expecting a broad network-related object to connect 2 nodes, but it seems that this treats `self` as belonging to a node already... but then its ID still needs to be provided?

node/narwhal/tests/common/primary.rs
```diff
@@ -12,23 +12,193 @@
...
+        for (primary, other_primary) in self.primaries.values().tuple_combinations() {
+            // Connect to the node.
+            let ip = other_primary.gateway().local_ip();
+            primary.gateway().connect(ip);
+            // Give the connection time to be established.
+            tokio::time::sleep(std::time::Duration::from_millis(100)).await;
+        }
+    }
+
+    // Disconnects a node from all other nodes.
```
ljedrz: if this disconnects from *all* other nodes, then why the `num_nodes` param? either the comment or the API needs to be adjusted

node/narwhal/tests/common/primary.rs
```diff
@@ -12,23 +12,193 @@
...
+            for peer_ip in primary.gateway().connected_peers().read().iter() {
+                primary.gateway().disconnect(*peer_ip);
+            }
+        }
+
+        // Give the connections time to be closed.
+        sleep(Duration::from_millis(100)).await;
+    }
+
+    // Checks a quorum of nodes have reached the given round.
```
ljedrz: this comment could use some polish

ljedrz (commented):
The overall direction LGTM, left some comments and suggestions.

node/narwhal/tests/common/primary.rs
niklaslong: Yes, although it'll be nicer to have a `HashMap` when killing and restarting nodes?

node/narwhal/tests/common/primary.rs
niklaslong: That's fair, mostly wanted to do it for ergonomic reasons, is there a major downside?

node/narwhal/tests/common/primary.rs
niklaslong: The idea was to make it easy to disconnect N nodes at a time (e.g. 2 in a 4 node network to break quorum). I'll adjust the doc.

node/narwhal/tests/common/primary.rs
ljedrz: it's just unexpected; when I see `.deref()` or `.as_mut()`, I'm used to different behavior

node/narwhal/tests/common/primary.rs
niklaslong: Happy to go with another setup ofc, just haven't found anything better yet.

node/narwhal/tests/common/primary.rs
ljedrz: that's fair :+1:

node/narwhal/tests/common/primary.rs
ljedrz: as long as the docs are adjusted, it should be fine

SplittyDev (approved):

PR #2540, opened 2023-07-14
Correctly clear the solutions when new epoch starts

ghostant-1017:
## Motivation

Correctly clear the solutions when new epoch starts

### DIFF
--- a/node/consensus/src/lib.rs
+++ b/node/consensus/src/lib.rs
@@ -146,13 +146,14 @@ impl<N: Network, C: ConsensusStorage<N>> Consensus<N, C> {
     /// Advances the ledger to the next block.
     pub fn advance_to_next_block(&self, block: &Block<N>) -> Result<()> {
         // Adds the next block to the ledger.
+        let old_epoch = self.ledger.latest_epoch_number();
         self.ledger.advance_to_next_block(block)?;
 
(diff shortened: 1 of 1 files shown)


node/consensus/src/lib.rs
```diff
@@ -146,13 +146,14 @@ impl<N: Network, C: ConsensusStorage<N>> Consensus<N, C> {
     /// Advances the ledger to the next block.
     pub fn advance_to_next_block(&self, block: &Block<N>) -> Result<()> {
         // Adds the next block to the ledger.
+        let old_epoch = self.ledger.latest_epoch_number();
         self.ledger.advance_to_next_block(block)?;
```
ghostant-1017, line 150: When `advance_to_next_block` is Ok, `block.epoch_number()` will always equal to `self.ledger.latest_epoch_number()`. Thanks to the beacon's logic: `produce_next_block` then `check_block_basic` before `advance_to_next_block`, this doesn't cause any security problem now, otherwise, it will cause chain fork

ljedrz (approved):
The value returned by `Ledger::latest_epoch_number` is based on the current block height so, based on the doc comments, if we want to know if the new block started a new epoch, we should indeed register the former value before calling `Ledger::advance_to_next_block` and compare against that one.

howardwu (approved):
LGTM

PR #2541, opened 2023-07-14
fix: gateway proptest

joske:
Fixed the failing Gateway proptest. The proptests are putting a too large value for the dev parameter. This causes an overflow when 5000 + dev is done (u16).

Constrain the `dev` field to < 10_000. (I tried a lower value, but then proptest complains there are too many rejects for this field).


ljedrz (approved):

howardwu (approved):
LGTM

PR #2542, opened 2023-07-14
feat: improve DAG test

joske:
Added the cases with different author and lower and higher round than the committed one. 

To be able to test with same author, also the author field should be settable on the certificate via some `sample_batch_certificate_` method.

### DIFF
--- a/node/narwhal/src/helpers/dag.rs
+++ b/node/narwhal/src/helpers/dag.rs
@@ -131,7 +131,7 @@ impl<N: Network> DAG<N> {
 mod tests {
     use super::*;
     use snarkvm::{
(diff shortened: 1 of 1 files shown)


ljedrz (approved):

node/narwhal/src/helpers/dag.rs
```diff
@@ -149,9 +149,10 @@ mod tests {
         let mut rng = TestRng::default();
         let mut dag = DAG::<Testnet3>::new();
 
-        let certificate = sample_batch_certificate(&mut rng);
+        let certificate = sample_batch_certificate_for_round(2, &mut rng);
         dag.insert(certificate.clone());
         let round = certificate.round();
+        assert_eq!(round, 2);
```
howardwu, line 155: What property does this test?

PR #2550, opened 2023-07-17
[Narwhal] Another round of perf tweaks

ljedrz:
This is another perf-oriented PR with a few drive-bys in the mix. Each commit is self-contained and describes the specific change.

There are also a few `TODO`s pertaining to future perf improvements that may require some snarkVM-side adjustments first or that can't be verified just yet.

### DIFF
--- a/node/narwhal/src/gateway.rs
+++ b/node/narwhal/src/gateway.rs
@@ -990,10 +990,9 @@ pub mod prop_tests {
                 // Construct the worker channels.
                 let (tx_worker, rx_worker) = init_worker_channels();
                 // Construct the worker instance.
-                let ledger = Box::new(MockLedgerService::new());
+                let ledger = Arc::new(MockLedgerService::new());
                 let worker =
-                    Worker::new(id, gateway.clone(), worker_storage.clone(), Arc::new(ledger), Default::default())
-                        .unwrap();
+                    Worker::new(id, gateway.clone(), worker_storage.clone(), ledger, Default::default()).unwrap();
                 // Run the worker instance.
                 worker.run(rx_worker);
 
--- a/node/narwhal/src/gateway.rs
+++ b/node/narwhal/src/gateway.rs
@@ -355,6 +355,8 @@ impl<N: Network> Gateway<N> {
     }
 
     /// Broadcasts the given event to all connected peers.
+    // TODO(ljedrz): the event should be checked for the presence of Data::Object, and
+    // serialized in advance if it's there.
     pub(crate) fn broadcast(&self, event: Event<N>) {
         // Ensure there are connected peers.
         if self.number_of_connected_peers() > 0 {
--- a/node/narwhal/src/helpers/dag.rs
+++ b/node/narwhal/src/helpers/dag.rs
@@ -101,7 +101,8 @@ impl<N: Network> DAG<N> {
         let author = certificate.author();
 
         // Update the last committed round for the author.
-        self.last_committed_authors
+        let last_committed_round = self
+            .last_committed_authors
             .entry(author)
             .and_modify(|last_committed_round| {
                 if certificate_round > *last_committed_round {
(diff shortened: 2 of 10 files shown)


ljedrz:
The `clippy` job failure is unrelated.

niklaslong (approved):
LGTM 👍

node/narwhal/src/helpers/dag.rs
```diff
@@ -101,7 +101,8 @@ impl<N: Network> DAG<N> {
         let author = certificate.author();
 
         // Update the last committed round for the author.
-        self.last_committed_authors
+        let last_committed_author = self
```
howardwu: nit: this should read `last_committed_round` (the type is a round, not an author)

node/narwhal/src/helpers/dag.rs
```diff
@@ -111,8 +112,7 @@ impl<N: Network> DAG<N> {
             .or_insert(certificate_round);
 
         // Update the last committed round.
-        // Note: The '.unwrap()' here is guaranteed to be safe.
-        self.last_committed_round = *self.last_committed_authors.values().max().unwrap();
+        self.last_committed_round = *last_committed_author;
```
howardwu: I would guard this check with an `if` check that the new `last_committed_round` is indeed greater than the last committed round, since we are no longer scanning the entire `HashSet` anymore to be sure.

howardwu: Just for the record: My preference remains to perform `.max()` on the existing `HashSet` (which we know to be a reasonable size, i.e. the number of validators), to ensure safety of this logic over time. It is unclear how this `DAG` struct will evolve over time.

howardwu: It would be in our interest to include a `debug_assert!()` that the committed round `==` the `.max()` result here.

node/narwhal/src/helpers/storage.rs
```diff
@@ -567,16 +567,17 @@ impl<N: Network> Storage<N> {
...
-                // Remove the entry for the transmission ID.
-                transmissions.remove(transmission_id);
+            match transmissions.entry(*transmission_id) {
+                Entry::Occupied(mut occupied_entry) => {
+                    let (_, certificate_ids) = occupied_entry.get_mut();
+                    // Remove the certificate ID for the transmission ID.
+                    certificate_ids.remove(&certificate_id);
+                    // If there are no more certificate IDs for the transmission ID, remove the transmission.
+                    if certificate_ids.is_empty() {
+                        occupied_entry.remove();
```
howardwu, line 578:
```suggestion
                        // Remove the entry for the transmission ID.
                        occupied_entry.remove();
```

node/narwhal/src/primary.rs
```diff
@@ -259,6 +259,7 @@ impl<N: Network> Primary<N> {
         // rebroadcast the batch header to the non-signers, and return early.
         if let Some(proposal) = self.proposed_batch.read().as_ref() {
             // Construct the event.
+            // FIXME(ljedrz): the BatchHeader should be serialized only once in advance before being sent to non-signers.
```
howardwu: nit: switch to `TODO (ljedrz)` so that these will be found in our search.

node/narwhal/src/gateway.rs
```diff
@@ -355,6 +355,8 @@ impl<N: Network> Gateway<N> {
     }
 
     /// Broadcasts the given event to all connected peers.
+    // FIXME(ljedrz): the event should be checked for the presence of Data::Object, and
```
howardwu: nit: switch to `TODO (ljedrz)` so that these will be found in our search.

node/narwhal/src/primary.rs
```diff
@@ -74,9 +74,9 @@ pub struct Primary<N: Network> {
     /// The storage.
     storage: Storage<N>,
     /// The ledger service.
-    ledger: Arc<Ledger<N>>,
+    ledger: Ledger<N>,
     /// The workers.
-    workers: Arc<Vec<Worker<N>>>,
+    workers: Arc<[Worker<N>]>,
```
howardwu, line 79: TIL this is possible (now), thanks for the change!

node/narwhal/src/helpers/dag.rs
ljedrz: if this structure might still evolve over time, I'll revert that bit :+1:

ljedrz:
I addressed the review comments.

howardwu (approved):
LGTM

PR #2551, opened 2023-07-18
[Narwhal] set up BFT e2e test tooling

niklaslong:
This PR adapts the existing test tooling to optionally spin up BFT instances on top of Narwhal. 

The only non-test change is adding `Clone` to the `BFTSender`.

### DIFF
--- a/node/narwhal/src/helpers/channels.rs
+++ b/node/narwhal/src/helpers/channels.rs
@@ -58,7 +58,7 @@ pub fn init_consensus_channels<N: Network>() -> (ConsensusSender<N>, ConsensusRe
     (sender, receiver)
 }
 
-#[derive(Debug)]
+#[derive(Clone, Debug)]
 pub struct BFTSender<N: Network> {
     pub tx_primary_round: mpsc::Sender<(u64, oneshot::Sender<Result<()>>)>,
     pub tx_primary_certificate: mpsc::Sender<(BatchCertificate<N>, oneshot::Sender<Result<()>>)>,
--- a/node/narwhal/tests/common/primary.rs
+++ b/node/narwhal/tests/common/primary.rs
@@ -36,35 +36,53 @@ use parking_lot::Mutex;
 use tokio::{task::JoinHandle, time::sleep};
 use tracing::*;
 
+/// The configuration for the test network.
 #[derive(Clone, Copy, Debug)]
 pub struct TestNetworkConfig {
+    /// The number of nodes to spin up.
     pub num_nodes: u16,
+    /// If this is set to `true`, the BFT protocol is started on top of Narwhal.
     pub bft: bool,
+    /// If this is set to `true`, all nodes are connected to each other (when they're first
+    /// started).
     pub connect_all: bool,
-    pub fire_cannons: bool,
+    /// If `Some(i)` is set, the cannons will fire every `i` milliseconds.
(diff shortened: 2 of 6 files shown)


node/narwhal/src/helpers/channels.rs
```diff
@@ -58,7 +58,7 @@ pub fn init_consensus_channels<N: Network>() -> (ConsensusSender<N>, ConsensusRe
     (sender, receiver)
 }
 
-#[derive(Debug)]
+#[derive(Clone, Debug)]
```
niklaslong, line 61: This is actually not needed for the code to function after all but there's no reason not to impl it afaict.

SplittyDev (approved):
Looks good, works good. 10/10 PR, would review again

node/narwhal/tests/common/primary.rs
```diff
@@ -107,10 +118,16 @@ impl TestNetwork {
...
     pub async fn start(&mut self) {
         for validator in self.validators.values_mut() {
-            // Setup the channels and start the primary.
-            let (sender, receiver) = init_primary_channels();
-            validator.sender = Some(sender.clone());
-            validator.primary.run(sender.clone(), receiver, None).await.unwrap();
+            let (primary_sender, primary_receiver) = init_primary_channels();
+            if let Some(bft) = &mut validator.bft {
+                // Setup the channels and start the bft.
+                validator.primary_sender = Some(primary_sender.clone());
```
ljedrz: this line could be moved before the `if let`...

node/narwhal/tests/common/primary.rs
```diff
@@ -107,10 +118,16 @@ impl TestNetwork {
...
+                bft.run(primary_sender, primary_receiver, None).await.unwrap();
+            } else {
+                // Setup the channels and start the primary.
+                validator.primary_sender = Some(primary_sender.clone());
```
ljedrz: ...after which this one can be removed

node/narwhal/tests/common/primary.rs
```diff
@@ -38,6 +39,7 @@ use tracing::*;
 #[derive(Clone, Copy, Debug)]
 pub struct TestNetworkConfig {
     pub num_nodes: u16,
+    pub bft: bool,
```
ljedrz, line 45: this one needs a doc describing what it does; the others could use some too

node/narwhal/tests/common/primary.rs
```diff
@@ -76,7 +79,7 @@ impl TestValidator {
                 for connection in connections {
                     debug!("  {}", connection);
                 }
-                sleep(Duration::from_secs(10)).await;
+                sleep(Duration::from_secs(5)).await;
```
ljedrz, line 100: perhaps this could be a part of `TestNetworkConfig`? `fire_cannons` could be changed from `bool` to `Option<u64>`, indicating the interval that goes here; the name of the member might need to be adjusted for clarity

ljedrz (approved):
LGTM in general, left a few comments/suggestions.

node/narwhal/tests/common/primary.rs
niklaslong: This is the connection timeout but if you mean the cannon sleep, I'm onboard 👍

howardwu (approved):

PR #2553, opened 2023-07-19
feat: bind on wildcard if an IP address was given

joske:
This changes the logic for selecting the binding ip/port so that if an IP address was given, bind to the wildcard (0.0.0.0) and given port. This is to facilitate devnet efforts.

### DIFF
--- a/node/consensus/src/lib.rs
+++ b/node/consensus/src/lib.rs
@@ -89,7 +89,8 @@ impl<N: Network, C: ConsensusStorage<N>> Consensus<N, C> {
         // Initialize the ledger service.
         let ledger_service = Arc::new(CoreLedgerService::<N, C>::new(ledger.clone()));
         // Initialize the BFT.
-        let bft = BFT::new(account, storage, ledger_service, ip, dev)?;
+        // Note that ip is always passed in as None by the validator.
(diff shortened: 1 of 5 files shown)


ljedrz (commented):
Why not just provide `0.0.0.0` as the IP address in the devnet? Seems like a simpler solution. Also, seems that this change breaks some proptest.

niklaslong:
As mentioned in slack, there are two requirements we need to meet here:
- we must be able to bind to `0.0.0.0` or a passed in ip (or localhost when in dev mode)
- we must always be able to pass in an IP so the node can recognise itself and avoid self-connects (except in dev mode)

joske:
The peers file must have the external IPs for the other nodes. Binding on 0.0.0.0 should be fine to allow the other nodes to connect. I don't expect this to clobber other interfaces, you can choose the port anyway. They will be running nothing else on these nodes anyway.

howardwu:
So why don't we change this back to passing in just the `PORT` instead of the entire `SocketAddr`?

joske:
~~The nodes need to know the IP addresses/ports of all the peers if not running locally.~~

Edit: Sorry, misunderstood your comment (read it late at night 😅). Addressing.

joske:
obsolete

PR #2558, opened 2023-07-19
[Narwhal] Leader election consistency test

SplittyDev:
This PR adds an end-to-end test that checks for bft leader election consistency

### DIFF
--- a/node/narwhal/tests/bft_e2e.rs
+++ b/node/narwhal/tests/bft_e2e.rs
@@ -156,17 +156,29 @@ async fn test_leader_election_consistency() {
         let cloned_network = network.clone();
         deadline!(Duration::from_secs(20), move || { cloned_network.is_round_reached(target_round) });
 
-        // Get all leaders
-        let leaders = network
-            .validators
-            .values()
+        // Get all validators in the network
+        let validators = network.validators.values().collect_vec();
+
+        // Get all validators in sync with the current round
+        let validators_for_round =
+            validators.iter().filter(|validator| validator.primary.current_round() == target_round).collect_vec();
+
+        // Get all leaders for the current round
+        let leaders = validators_for_round
+            .iter()
             .flat_map(|v| v.bft.clone().map(|bft| bft.leader()))
             .flatten()
             .collect::<Vec<_>>();
-        println!("Found {} validators with a leader (out of {})", leaders.len(), network.validators.values().count());
+
+        println!(
+            "Found {} validators with a leader (out of {}, {} out of sync)",
+            leaders.len(),
+            validators_for_round.len(),
+            validators.len() - validators_for_round.len()
+        );
 
         // Assert that we have N leaders
-        assert_eq!(leaders.len(), N as usize);
+        assert_eq!(leaders.len(), validators_for_round.len());
 
         // Assert that all leaders are equal
         assert!(leaders.iter().all_equal());
--- a/node/narwhal/tests/bft_e2e.rs
+++ b/node/narwhal/tests/bft_e2e.rs
@@ -142,7 +142,7 @@ async fn test_leader_election_consistency() {
         connect_all: true,
         fire_cannons: Some(CANNON_INTERVAL_MS),
         // Set this to Some(0..=4) to see the logs.
-        log_level: Some(1),
+        log_level: None,
         log_connections: true,
     });
     network.start().await;
--- a/node/narwhal/tests/bft_e2e.rs
+++ b/node/narwhal/tests/bft_e2e.rs
@@ -129,11 +129,11 @@ async fn test_quorum_break() {
 #[ignore = "long-running e2e test"]
 async fn test_leader_election_consistency() {
     // The round at which to start checking for leader consistency.
-    const STARTING_ROUND: u64 = 9;
+    const STARTING_ROUND: u64 = 2;
 
     // The minimum and maximum rounds to check for leader consistency.
-    const MIN_ROUND: u64 = 10;
-    const MAX_ROUND: u64 = 100;
+    const MIN_ROUND: u64 = 4;
+    const MAX_ROUND: u64 = 30;
 
     // Start N nodes, connect them and start the cannons for each.
     const N: u16 = 4;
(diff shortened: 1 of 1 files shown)


node/narwhal/tests/bft_e2e.rs
```diff
@@ -123,3 +124,50 @@ async fn test_quorum_break() {
...
+        fire_cannons: Some(CANNON_INTERVAL_MS),
+        // Set this to Some(0..=4) to see the logs.
+        log_level: Some(1),
+        log_connections: true,
+    });
+    network.start().await;
+
+    // Wait for starting round to be reached
+    let cloned_network = network.clone();
+    deadline!(Duration::from_secs(60), move || { cloned_network.is_round_reached(STARTING_ROUND) });
```
ljedrz, line 152: shouldn't this test be `#[ignore]`d due to its runtime?

node/narwhal/tests/bft_e2e.rs
```diff
@@ -123,3 +124,50 @@ async fn test_quorum_break() {
     // Check the nodes have stopped advancing through the rounds.
     assert!(network.is_halted().await);
 }
+
+#[tokio::test(flavor = "multi_thread")]
+async fn test_leader_election_consistency() {
+    // The round at which to start checking for leader consistency.
```
ljedrz: is this the minimum possible round to start the checks? if so, it should be commented on

node/narwhal/tests/bft_e2e.rs
```diff
@@ -123,3 +124,50 @@ async fn test_quorum_break() {
...
+    const STARTING_ROUND: u64 = 2;
+
+    // The minimum and maximum rounds to check for leader consistency.
+    const MIN_ROUND: u64 = 4;
```
ljedrz: sooo is this one or `STARTING_ROUND` the one to start checking for leader consistency :thinking:? this should be clarified

node/narwhal/tests/bft_e2e.rs
```diff
@@ -123,3 +124,50 @@ async fn test_quorum_break() {
...
+    const MAX_ROUND: u64 = 30;
```
ljedrz: I'm assuming that this is an arbitrary value; it could use a comment

node/narwhal/tests/bft_e2e.rs
```diff
@@ -123,3 +124,50 @@ async fn test_quorum_break() {
...
+        log_connections: true,
+    });
+    network.start().await;
+
+    // Wait for starting round to be reached
+    let cloned_network = network.clone();
+    deadline!(Duration::from_secs(60), move || { cloned_network.is_round_reached(STARTING_ROUND) });
+
+    // Check that validators agree about leaders in every even round
+    for target_round in (MIN_ROUND..MAX_ROUND).step_by(2) {
```
ljedrz:
```suggestion
    for target_round in (MIN_ROUND..=MAX_ROUND).step_by(2) {
```
seems to me that we want to check the maximum value too, no? especially if we only consider evens

node/narwhal/tests/bft_e2e.rs
```diff
@@ -123,3 +124,50 @@ async fn test_quorum_break() {
...
+        let cloned_network = network.clone();
+        deadline!(Duration::from_secs(20), move || { cloned_network.is_round_reached(target_round) });
+
+        // Get all leaders
```
ljedrz: all leaders for the specific round or thus far? this could use a bit of clarification

ljedrz (approved):
LGTM and a fine addition; left some comments/suggestions

node/narwhal/tests/bft_e2e.rs
SplittyDev: The runtime is around 69 seconds on my device, Niklas wanted it to run within a timeframe that's somewhat acceptable as part of the normal tests

node/narwhal/tests/bft_e2e.rs
SplittyDev: I tried starting on 0 and got some inconsistency issues, probably starting from round 2 might already be fine. I just added a tiny bit of buffer.

node/narwhal/tests/bft_e2e.rs
```diff
@@ -123,3 +124,50 @@ async fn test_quorum_break() {
...
+    // Start N nodes, connect them and start the cannons for each.
+    const N: u16 = 4;
+    const CANNON_INTERVAL_MS: u64 = 10;
+    let mut network = TestNetwork::new(TestNetworkConfig {
+        num_nodes: N,
+        bft: true,
+        connect_all: true,
+        fire_cannons: Some(CANNON_INTERVAL_MS),
+        // Set this to Some(0..=4) to see the logs.
+        log_level: Some(1),
```
niklaslong: This should probably be set to `None` before we merge.

node/narwhal/tests/bft_e2e.rs
```diff
@@ -123,3 +124,63 @@ async fn test_quorum_break() {
...
+            .collect::<Vec<_>>();
+
+        println!(
+            "Found {} validators with a leader (out of {}, {} out of sync)",
+            leaders.len(),
+            validators_for_round.len(),
+            validators.len() - validators_for_round.len()
+        );
+
+        // Assert that we have N leaders
```
niklaslong: Nit: overloading `N` here is a little confusing.

node/narwhal/tests/bft_e2e.rs
niklaslong: Can we get it below 60 without loosing too much quality?

node/narwhal/tests/bft_e2e.rs
SplittyDev: I clarified the comment, it now says `Assert that we have x leaders for x validators that have reached the target round`

node/narwhal/tests/bft_e2e.rs
SplittyDev: We could probably set the `MAX_ROUND` to 20 instead of 30, that should cut down test times quite a bit (well, by 1/3rd). But that would also mean we're effectively only evaluating 8 rounds. Right now, it's 13 rounds.

niklaslong: I think that's reasonable 👍

joske (approved):

PR #2559, opened 2023-07-19
test: worker tests

joske:
Split out some of the `Gateway` API into a trait so that we can mock it.

Add tests for `Worker`

### DIFF
--- a/node/narwhal/src/gateway.rs
+++ b/node/narwhal/src/gateway.rs
@@ -74,6 +74,13 @@ const MAX_CONNECTION_ATTEMPTS: usize = 10;
 /// The maximum interval to restrict a peer.
 const RESTRICTED_INTERVAL: i64 = (MAX_CONNECTION_ATTEMPTS as u64 * MAX_BATCH_DELAY / 1000) as i64; // seconds
 
+/// Part of the Gateway API that deals with networking.
+/// This is a separate trait to allow for easier testing/mocking.
+pub trait Transport<N: Network>: Send + Sync {
+    fn send(&self, peer_ip: SocketAddr, event: Event<N>);
+    fn broadcast(&self, event: Event<N>);
+}
+
 #[derive(Clone)]
 pub struct Gateway<N: Network> {
     /// The account of the node.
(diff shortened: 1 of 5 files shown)


node/narwhal/src/gateway.rs
```diff
@@ -74,6 +74,11 @@ const MAX_CONNECTION_ATTEMPTS: usize = 10;
 /// The maximum interval to restrict a peer.
 const RESTRICTED_INTERVAL: i64 = (MAX_CONNECTION_ATTEMPTS as u64 * MAX_BATCH_DELAY / 1000) as i64; // seconds
 
+pub trait ReliableTransport<N: Network>: Send + Sync {
```
ljedrz: this trait could use a description; why makes it reliable? should the methods perform retransmission on failure? should there be some ACK mechanism? if reliability is a factor, why are the methods infallible?

ljedrz (changes requested):
Since it's more than just tests, I'm missing a description, and the creation of a new trait and the related refactoring should be in its own commit.

The use of a trait object in the `Worker` could either be a part of that or - preferably - a separate commit as well.

node/narwhal/src/worker.rs
```diff
@@ -477,10 +479,228 @@ mod prop_tests {
...
+    mock! {
+        Gateway<N: Network> {}
+        impl<N:Network> ReliableTransport<N> for Gateway<N> {
+            fn broadcast(&self, event: Event<N>);
+            fn send(&self, peer_ip: SocketAddr, event: Event<N>);
+        }
+    }
+
+    mock! {
+        LedgerService<N: Network> {}
```
ljedrz: I'd prefer this to be called either `Ledger` or `LedgerProvider`, so as not to have `impl LedgerService for LedgerService`

node/narwhal/src/worker.rs
```diff
@@ -477,10 +479,228 @@ mod prop_tests {
...
+        worker.finish_transmission_request(peer_ip, TransmissionResponse {
+            transmission_id,
+            transmission: Transmission::Solution(Data::Buffer(Bytes::from(vec![0; 512]))),
+        });
+        // Check the transmission was removed from the pending set.
+        assert!(!worker.pending.contains(transmission_id));
+    }
+
+    #[tokio::test]
+    async fn test_process_solution_ok() {
```
ljedrz, line 590: applies to this one and the rest of the tests - they need some newlines that at least separate the setup part from the actual test bits

node/narwhal/src/gateway.rs
joske: Naming is hard. I'm open to another name. It's just to facilitate testing/mocking.

ljedrz (approved):
LGTM :ok_hand:

howardwu (approved):
LGTM

PR #2561, opened 2023-07-19
Revert "feat: increase max committee size to 100"

howardwu:
Reverts AleoHQ/snarkOS#2554

We need to introduce a `max_committee` variable in `Gateway` and change all constants in `Gateway` to be computed configurations if this constant is changed arbitrarily like this. Otherwise, the rate limiters are broken and we have minimal control on rate of flow for inbound and outbound messages.


ljedrz (commented):
This is likely no longer needed with https://github.com/AleoHQ/snarkOS/pull/2562. However, if we decide against that one, the revert is trivial enough and LGTM.

PR #2562, opened 2023-07-20
Dynamic rate limiting based on current committee size

niklaslong:
Adjusts the rate-limiting thresholds to be dynamic based on the current committee size. I also adjusted the naming slightly but happy to revert or change it to something else if it is desired. 

Note, the queue depths for the tcp stack need to be constant and as such still depend on the static `MAX_COMMITTEE_SIZE` but we should consider making this much smaller than it currently is.

### DIFF
--- a/node/narwhal/src/gateway.rs
+++ b/node/narwhal/src/gateway.rs
@@ -74,15 +74,6 @@ const CACHE_EVENTS_INTERVAL: i64 = (MAX_BATCH_DELAY / 1000) as i64; // seconds
 /// The maximum interval of requests to cache.
 const CACHE_REQUESTS_INTERVAL: i64 = (MAX_BATCH_DELAY / 1000) as i64; // seconds
 
-/// The maximum number of events to cache.
-const CACHE_EVENTS: usize = CACHE_TRANSMISSIONS;
-/// The maximum number of certificate requests to cache.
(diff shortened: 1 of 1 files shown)


niklaslong:
Drafted until we decide what the message queue depth should be.

ljedrz (approved):

niklaslong:
Rebased.

node/narwhal/src/gateway.rs
```diff
@@ -108,6 +99,34 @@ pub struct Gateway<N: Network> {
     handles: Arc<Mutex<Vec<JoinHandle<()>>>>,
 }
 
+// Dynamic rate limiting.
+impl<N: Network> Gateway<N> {
+    /// The current maxiumum committee size.
+    fn max_committee_size(&self) -> usize {
+        self.storage.current_committee().num_members()
```
howardwu: This will incur a notable perf hit, mainly because it hits the db to fetch the committee

howardwu: Perhaps we just set a hardcoded number for now? Not ideal, but perhaps `200` will be acceptable

niklaslong:
Rebased. We still rely on `MAX_COMMITEE_SIZE` as a fallback value (and it's used to set the max peers on the gateway) but the committee should exist when `max_committee_size` is called for dynamic rate limiting purposes.

howardwu (approved):

PR #2565, opened 2023-07-20
[Narwhal] Return a &'static str from EventTrait::name

ljedrz:
`EventTrait::name` doesn't need to return allocated strings, it can just use static ones instead.


niklaslong (approved):

howardwu (approved):
LGTM, good catch

PR #2570, opened 2023-07-21
Bump `snarkVM` to `v0.14.5`

raychu86:
## Motivation

This PR bumps the snarkVM version to `v0.14.5`. In addition, there are a few updates to few of the `snarkOS developer` CLI commands:

 - Deploy now uses the `Package::deploy` from snarkVM
 - Record inputs now support ciphertexts as well as plaintext. If a ciphertext is provided, it will be decrypted by the view key of the caller.
 - Add recursive loading of program imports in `execute`
 - Removal of redundant logic

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -931,6 +931,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "dotenvy"
+version = "0.15.7"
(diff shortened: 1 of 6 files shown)


howardwu (approved):
LGTM

PR #2560, opened 2023-07-19
[Narwhal] make tx cannons configurable from cli

niklaslong:
This PR introduces a basic `clap`-based CLI to make the `simple_nodes` example more configurable for devnet purposes. The tx cannons can now be enabled/disabled and configured.

PR #2571, opened 2023-07-21
[Narwhal] e2e storage coherence cases

niklaslong:
This PR adds a e2e test case to check storage coherence over mulitple rounds in the network (narwhal only). It verifies the committee each node has the same committee for each round and verifies the certificates, batches and authors are coherent over the network. 

I have also renamed `fire_cannons` to `fire_transmissions` to be consistent with the changes in #2560.

### DIFF
--- a/node/narwhal/tests/common/primary.rs
+++ b/node/narwhal/tests/common/primary.rs
@@ -210,12 +210,29 @@ impl TestNetwork {
         self.validators.values().all(|v| v.primary.current_round() <= halt_round)
     }
 
+    // Checks if the committee is coherent in storage for all nodes (not quorum).
     pub fn is_committee_coherent<T>(&self, rounds_range: T) -> bool
     where
         T: RangeBounds<u64> + IntoIterator<Item = u64>,
     {
         rounds_range.into_iter().fold(true, |acc, round| {
             acc && self.validators.values().map(|v| v.primary.storage().get_committee(round).unwrap()).dedup().count()
+                == 1
+        })
+    }
+
+    // Checks if the certificates are coherent in storage for all nodes (not quorum).
+    pub fn is_certificate_round_coherent<T>(&self, rounds_range: T) -> bool
+    where
+        T: RangeBounds<u64> + IntoIterator<Item = u64>,
+    {
+        rounds_range.into_iter().fold(true, |acc, round| {
+            acc && self
+                .validators
+                .values()
+                .map(|v| v.primary.storage().get_certificates_for_round(round))
+                .dedup()
+                .count()
                 == 1
         })
     }
--- a/node/narwhal/tests/common/primary.rs
+++ b/node/narwhal/tests/common/primary.rs
@@ -47,7 +47,7 @@ pub struct TestNetworkConfig {
     /// started).
     pub connect_all: bool,
     /// If `Some(i)` is set, the cannons will fire every `i` milliseconds.
-    pub fire_cannons: Option<u64>,
+    pub fire_transmissions: Option<u64>,
     /// The log level to use for the test.
     pub log_level: Option<u8>,
     /// If this is set to `true`, the number of connections is logged every 5 seconds.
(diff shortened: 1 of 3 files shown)


node/narwhal/tests/common/primary.rs
```diff
@@ -209,6 +209,33 @@ impl TestNetwork {
         sleep(Duration::from_millis(MAX_BATCH_DELAY * 2)).await;
         self.validators.values().all(|v| v.primary.current_round() <= halt_round)
     }
+
+    // Checks if the committee is coherent in storage for all nodes (not quorum).
+    pub fn is_committee_coherent<T>(&self, rounds_range: T) -> bool
```
niklaslong, line 215: I made these functions since they might end up being useful for other e2e tests. Note they also verify the conditions for all nodes in the network, not just the subset making up quorum. It may be better to implement the latter in future if/when we write cases that include byzantine nodes. For now, I'm assuming all the nodes are honest.

niklaslong: Alternatively, we could go the other way and check that all nodes have reached the round before asserting on state (with a new helper similar to `is_round_reached`, perahps `is_round_reached_by_all`).

node/narwhal/tests/common/primary.rs
```diff
@@ -209,6 +209,28 @@ impl TestNetwork {
...
     }
+
+    // Checks if the committee is coherent in storage for all nodes (not quorum) over a range of
+    // rounds.
+    pub fn is_committee_coherent<T>(&self, rounds_range: T) -> bool
+    where
+        T: RangeBounds<u64> + IntoIterator<Item = u64>,
+    {
+        rounds_range.into_iter().all(|round| {
+            self.validators.values().map(|v| v.primary.storage().get_committee(round).unwrap()).dedup().count() == 1
```
ljedrz, line 220: are the entries consecutive, so that `dedup` will work?

node/narwhal/tests/common/primary.rs
```diff
@@ -209,6 +209,28 @@ impl TestNetwork {
...
+    }
+
+    // Checks if the certificates are coherent in storage for all nodes (not quorum) over a range
+    // of rounds.
+    pub fn is_certificate_round_coherent<T>(&self, rounds_range: T) -> bool
+    where
+        T: RangeBounds<u64> + IntoIterator<Item = u64>,
+    {
+        rounds_range.into_iter().all(|round| {
+            self.validators.values().map(|v| v.primary.storage().get_certificates_for_round(round)).dedup().count() == 1
```
ljedrz, line 231: ditto

node/narwhal/tests/narwhal_e2e.rs
```diff
@@ -128,3 +129,36 @@ async fn test_quorum_break() {
...
+
+    // Check the nodes have started advancing through the rounds.
+    const TARGET_ROUND: u64 = 12;
+    // Note: cloning the network is fine because the primaries it wraps are `Arc`ed.
+    let network_clone = network.clone();
+    deadline!(Duration::from_secs(40), move || { network_clone.is_round_reached(TARGET_ROUND) });
+
+    // Check the committee is coherent across the network up to the target round. We skip the
+    // genesis round.
+    assert!(network.is_committee_coherent(1..TARGET_ROUND));
```
ljedrz, line 157: just making sure that this shouldn't be `..=`

ljedrz (approved):
LGTM with some comments

node/narwhal/tests/common/primary.rs
niklaslong: Currently yes, committees are static—we may need to tweak this once dynamic committees are in.

node/narwhal/tests/common/primary.rs
niklaslong: I haven't encountered issues but perhaps ~~it would be safer to sort.~~ no need, it is backed by an `IndexMap`.

node/narwhal/tests/narwhal_e2e.rs
niklaslong: It can be currently but with dynamic committees, the update might happen just after the round is incremented, in which case this would be too strict.

howardwu (approved):
LGTM

niklaslong:
Rebased.

PR #2576, opened 2023-07-25
[Narwhal] Proptests for BFT events

vvp:
This PR adds property tests to BFT events in `snarkos-node-narwhal`-crate. Most of the event implementation is deserialization and serialization logic, which lends itself well for proptesting.

### DIFF
--- a/node/narwhal/committee/src/prop_tests.rs
+++ b/node/narwhal/committee/src/prop_tests.rs
@@ -108,7 +108,8 @@ impl Arbitrary for ValidatorSet {
     type Strategy = BoxedStrategy<ValidatorSet>;
 
     fn arbitrary_with(_: Self::Parameters) -> Self::Strategy {
-        validator_set(any_valid_validator(), size_range(4..=MAX_COMMITTEE_SIZE as usize)).boxed()
+        // use minimal validator set to speed up tests that require signing from the committee members
+        validator_set(any_valid_validator(), size_range(4..=4usize)).boxed()
     }
 }
 
--- a/node/narwhal/src/event/batch_signature.rs
+++ b/node/narwhal/src/event/batch_signature.rs
@@ -82,12 +82,12 @@ mod prop_tests {
     }
 
     #[proptest]
-    fn serialize_deserialize(#[strategy(any_batch_signature())] signature: BatchSignature<CurrentNetwork>) {
+    fn serialize_deserialize(#[strategy(any_batch_signature())] original: BatchSignature<CurrentNetwork>) {
         let mut buf = BytesMut::with_capacity(64).writer();
-        BatchSignature::serialize(&signature, &mut buf).unwrap();
+        BatchSignature::serialize(&original, &mut buf).unwrap();
 
-        let deserialized_signature: BatchSignature<CurrentNetwork> =
+        let deserialized: BatchSignature<CurrentNetwork> =
             BatchSignature::deserialize(buf.get_ref().clone()).unwrap();
-        assert_eq!(signature, deserialized_signature);
+        assert_eq!(original, deserialized);
     }
 }
--- a/node/narwhal/src/event/batch_signature.rs
+++ b/node/narwhal/src/event/batch_signature.rs
@@ -57,7 +57,7 @@ impl<N: Network> EventTrait for BatchSignature<N> {
 }
 
 #[cfg(test)]
-mod prop_tests {
+pub mod prop_tests {
     use crate::{
         event::{
             certificate_request::prop_tests::any_field,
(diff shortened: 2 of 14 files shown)


node/narwhal/committee/src/prop_tests.rs
```diff
@@ -108,7 +108,8 @@ impl Arbitrary for ValidatorSet {
     type Strategy = BoxedStrategy<ValidatorSet>;
 
     fn arbitrary_with(_: Self::Parameters) -> Self::Strategy {
-        validator_set(any_valid_validator(), size_range(4..=MAX_COMMITTEE_SIZE as usize)).boxed()
+        // use minimal validator set to speed up tests that require signing from the committee members
+        validator_set(any_valid_validator(), size_range(4..=4usize)).boxed()
```
ljedrz, line 112: maybe we should at least test 1 more number? or is the current setup hardcoded to a 4-member committee only?

node/narwhal/src/event/batch_signature.rs
```diff
@@ -55,3 +55,38 @@ impl<N: Network> EventTrait for BatchSignature<N> {
...
+        BatchSignature,
+    };
+    use bytes::{BufMut, BytesMut};
+    use proptest::prelude::{BoxedStrategy, Just, Strategy};
+    use test_strategy::proptest;
+
+    type CurrentNetwork = snarkvm::prelude::Testnet3;
+
+    pub fn any_batch_signature() -> BoxedStrategy<BatchSignature<CurrentNetwork>> {
+        (any_field(), any_signature(), Just(now()), -10..10i64)
```
ljedrz, line 77: is this the only drift range that we accept? seems a bit narrow at first glance

node/narwhal/src/event/disconnect.rs
```diff
@@ -63,3 +63,44 @@ impl EventTrait for Disconnect {
...
+            Disconnect::serialize(&disconnect, &mut buf).unwrap();
+
+            let disconnect = Disconnect::deserialize(buf.get_ref().clone()).unwrap();
+            assert_eq!(reason, &disconnect.reason);
+        }
+    }
+
+    #[test]
+    fn deserializing_empty_defaults_no_reason() {
+        let buf = BytesMut::with_capacity(64).writer();
```
ljedrz:
```suggestion
        let buf = BytesMut::default().writer();
```

node/narwhal/src/event/mod.rs
```diff
@@ -178,3 +178,94 @@ impl<N: Network> Event<N> {
...
+#[cfg(test)]
+mod tests {
+    use crate::Event;
+    use bytes::{BufMut, BytesMut};
+    type CurrentNetwork = snarkvm::prelude::Testnet3;
+
+    #[test]
+    #[should_panic(expected = "Missing event ID")]
+    fn deserializing_empty_defaults_no_reason() {
+        let buf = BytesMut::with_capacity(64).writer();
```
ljedrz:
```suggestion
        let buf = BytesMut::default().writer();
```

node/narwhal/src/event/mod.rs
```diff
@@ -178,3 +178,94 @@ impl<N: Network> Event<N> {
...
+        Event::<CurrentNetwork>::deserialize(buf.get_ref().clone()).unwrap();
+    }
+
+    #[test]
+    fn deserializing_invalid_data_panics() {
+        let mut buf = BytesMut::with_capacity(64).writer();
+        let invalid_id = 11u16;
```
ljedrz:
```suggestion
        let invalid_id = u16::MAX;
```
nit: this is the usual convention when looking for an impossible variant value

ljedrz (approved):
LGTM, left a few comments/suggestions

node/narwhal/committee/src/prop_tests.rs
vvp: This is the `Arbitrary` implementation for `ValidatorSet`, so it generates the "standard" generation strategy for use in external proptests. There are additional (internal) strategies for testing different `ValidatorSet` setups.

node/narwhal/src/event/batch_signature.rs
vvp: Perhaps, however none of the code in `src/event` does anything with those timestamps. Once we expand proptesting to `Gateway::inbound` and other call sites, we can generate additional strategies for wider ranges.

node/narwhal/src/event/disconnect.rs
```diff
@@ -63,3 +63,44 @@ impl EventTrait for Disconnect {
...
+
+    #[test]
+    fn serialize_deserialize() {
+        // TODO switch to an iteration method that doesn't require manually updating this vec if enums are added
+        let all_reasons = vec![
+            DisconnectReason::ProtocolViolation,
+            DisconnectReason::NoReasonGiven,
+            DisconnectReason::InvalidChallengeResponse,
+            DisconnectReason::OutdatedClientVersion,
+        ];
```
SplittyDev, line 80: Could we implement `strum::EnumIter` for `DisconnectReason`?

SplittyDev (approved):

node/narwhal/src/event/disconnect.rs
vvp: We could, but wasn't sure if we should :sweat_smile:

vvp: Ok, after thinking this, let's fix this in another PR if more enum-iterating needs arise.

vvp:
CI failures are unrelated. Since this is a test-only PR, I'll merge this now. Thank you all for reviewing :pray:

PR #2577, opened 2023-07-25
[Narwhal] Event serialization tweaks

ljedrz:
This PR reduces the number of allocations when serializing `Event` objects.

Similar improvements are likely possible in `Message`, I'm going to look at those next.

### DIFF
--- a/node/messages/src/beacon_propose.rs
+++ b/node/messages/src/beacon_propose.rs
@@ -43,10 +43,10 @@ impl<N: Network> MessageTrait for BeaconPropose<N> {
     /// Serializes the message into the buffer.
     #[inline]
     fn serialize<W: Write>(&self, writer: &mut W) -> Result<()> {
-        writer.write_all(&self.version.to_le_bytes())?;
-        writer.write_all(&self.round.to_le_bytes())?;
-        writer.write_all(&self.block_height.to_le_bytes())?;
-        writer.write_all(&self.block_hash.to_bytes_le()?)?;
+        self.version.write_le(&mut *writer)?;
+        self.round.write_le(&mut *writer)?;
+        self.block_height.write_le(&mut *writer)?;
+        self.block_hash.write_le(&mut *writer)?;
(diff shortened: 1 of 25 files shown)


niklaslong (approved):

SplittyDev (approved):
Nice 👍

node/messages/src/beacon_propose.rs
```diff
@@ -36,17 +38,17 @@ impl<N: Network> BeaconPropose<N> {
...
     }
 
     /// Serializes the message into the buffer.
     #[inline]
     fn serialize<W: Write>(&self, writer: &mut W) -> Result<()> {
-        writer.write_all(&self.version.to_le_bytes())?;
-        writer.write_all(&self.round.to_le_bytes())?;
-        writer.write_all(&self.block_height.to_le_bytes())?;
-        writer.write_all(&self.block_hash.to_bytes_le()?)?;
+        self.version.write_le(&mut *writer)?;
```
howardwu, line 48: For all instances in this PR, instead of:
```
&mut *writer
```
I believe you can just write:
```
writer
```

ljedrz: sadly, while it would be intuitively true, this isn't the case, due to the generics involved; the use of `&mut *` is necessary to work around this:
```
error[E0382]: use of moved value: `writer`
  --> node/messages/src/beacon_propose.rs:49:29
   |
47 |     fn serialize<W: Write>(&self, writer: &mut W) -> Result<()> {
   |                                   ------ move occurs because `writer` has type `&mut W`, which does not implement the `Copy` trait
48 |         self.version.write_le(writer)?;
   |                               ------ value moved here
49 |         self.round.write_le(writer)?;
   |                             ^^^^^^ value used here after move
```

joske (approved):

ljedrz:
These changes provide the following performance improvements:
- allocations down ~9%
- temporary allocations down ~4%

PR #2578, opened 2023-07-27
test: add tests for some BFT functions

joske:
I tried to add tests for some more, but they required too much setup to be of much value.


howardwu (approved):
LGTM

PR #2598, opened 2023-08-15
[Narwhal] `Cache` now cleans up the expired entries from sub-caches more reliably

vvp:
This PR changes the internal data structures in `Cache` struct from `IndexMap<K, VecDeque<OffsetDateTime>>` to `BTreeMap<OffsetDateTime, K>`, to remove expired entries from the cache and fix the memory leak without too much overhead. 

Profiling the memory impact of this PR with `heaptrack` showed following improvements (4-node network, 2000+ rounds = 26 minute run):
- Peak heap memory: before 13.9MB, after 9.7MB
- Total memory leaked: before 12.7MB, after 6.3MB
- `Cache::retain_and_insert()` peak heap usage: before 5.5MB, after 3.5kB
- `Cache::retain_and_insert()` peak leaked memory: before 5.5MB, after 3.5kB

The implementation partly relies on the convention that `Cache` call sites are using the same expiration interval for a given sub-cache (e.g. `Cache::insert_inbound_transmission()` is always called with `snarkos_node_narwhal::gateway::CACHE_REQUESTS_INTERVAL`- interval). With this assumption, all the expired entries can be safely and efficiently dropped from the same timestamp-ordered cache on every `Cache::retain_and_insert()` call - even the entries unrelated to the `key`.


vvp:
Hmm, CI failure on `node-narwhal` seems to be related to this PR and especially the commit https://github.com/AleoHQ/snarkOS/pull/2598/commits/deefc6d8c8979e52dbfdd4badab5ca6b894d9d0b which definitely made the `test_quorum_threshold` flakier. 

I can reproduce it locally around 40% of runs, though making tx cannons slightly slower makes the test pass always, so it is a workaround. I'll continue the investigations.

node/narwhal/src/helpers/cache.rs
```diff
@@ -12,33 +12,32 @@
...
-use parking_lot::RwLock;
 use std::{
-    collections::VecDeque,
+    collections::{BTreeMap, HashMap},
     net::{IpAddr, SocketAddr},
 };
+
+use parking_lot::RwLock;
+use snarkvm::{console::types::Field, ledger::narwhal::TransmissionID, prelude::Network};
 use time::{Duration, OffsetDateTime};
```
howardwu: nit: incorrect ordering of import declarations

node/narwhal/src/helpers/cache.rs
```diff
@@ -105,33 +104,39 @@ impl<N: Network> Cache<N> {
...
 
 #[cfg(test)]
 mod tests {
-    use super::*;
+    use std::net::Ipv4Addr;
+
     use snarkvm::prelude::Testnet3;
 
-    use std::net::Ipv4Addr;
+    use super::*;
```
howardwu: nit: incorrect ordering of import declarations

ljedrz:
I updated the PR with a reduced (down to 1s) time resolution; the test workaround should no longer be needed.

node/narwhal/src/helpers/cache.rs
ljedrz: fixed

howardwu (approved):

howardwu:
LGTM. Merging (failures unrelated)

PR #2610, opened 2023-08-17
[Narwhal] Assorted performance tweaks

ljedrz:
These changes reduce the number of allocations by ~8% and temporary allocations by 16% in BFT runs.

### DIFF
--- a/node/narwhal/src/helpers/storage.rs
+++ b/node/narwhal/src/helpers/storage.rs
@@ -33,6 +33,17 @@ use std::{
     },
 };
 
+#[derive(Clone, Debug)]
+pub struct Storage<N: Network>(Arc<StorageInner<N>>);
+
+impl<N: Network> std::ops::Deref for Storage<N> {
+    type Target = Arc<StorageInner<N>>;
+
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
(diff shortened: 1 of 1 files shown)


joske (approved):

howardwu:
@ljedrz can you rebase?

ljedrz:
Rebased.

node/narwhal/src/bft.rs
```diff
@@ -560,6 +561,8 @@ impl<N: Network> BFT<N> {
                 // Insert the previous certificate into the buffer.
                 buffer.push(previous_certificate);
             }
+            // Insert the certificate into the map.
+            commit.entry(certificate_round).or_default().insert(certificate);
```
howardwu: Is there a reason to move this down? This logic is quite sensitive.

ljedrz: This was done so that the clone can be avoided (which wouldn't be possible otherwise), and while the overall logic should remain intact, since it is sensitive, this change can wait until we have more proptests etc.

I'll remove that commit with the next rebase and if this is still noticeable in heap profiles, we can always return to it in the future.

ljedrz:
Rebased and dropped the last commit for now.

joske (approved):

howardwu (approved):

PR #2618, opened 2023-08-22
[Narwhal] first primary unit tests

niklaslong:
Sets up some basic test helpers and test cases for the primary's first round. Testing N rounds requires a more complex state setup. Some key considerations:
- the core structures are seeded with the necessary state to compute the various methods tested in each case,
- the gateway is largely ignored, though its resolver needs to be seeded with the committee's socket addresses, otherwise some checks will fail in various methods,
- I've deduplicated some of the core state setup code, perhaps this can be further reduced in future, when the primary is refactored... just food for thought.


node/narwhal/src/primary.rs
```diff
@@ -1053,3 +1053,317 @@ impl<N: Network> Primary<N> {
...
+    use rand::RngCore;
+
+    type CurrentNetwork = snarkvm::prelude::Testnet3;
+
+    // Returns a primary and a list of accounts in the configured committee.
+    async fn primary_without_handlers(
+        rng: &mut TestRng,
+    ) -> (Primary<CurrentNetwork>, Vec<(SocketAddr, Account<CurrentNetwork>)>) {
+        // Create a committee containing the primary's account.
+        let (accounts, committee) = {
```
joske, line 1087: Is there a reason you can't use `snarkos_node_narwhal_committee::test_helpers::sample_committee(rng)`?

node/narwhal/src/primary.rs
```diff
@@ -1053,3 +1053,317 @@ impl<N: Network> Primary<N> {
...
+        let transmission_ids = [solution_commitment.into(), (&transaction_id).into()].into();
+        let transmissions = [
+            (solution_commitment.into(), Transmission::Solution(solution)),
+            ((&transaction_id).into(), Transmission::Transaction(transaction)),
+        ]
+        .into();
+        // Prepare the certificate IDs.
+        let certificate_ids = Default::default();
+        // Sign the batch header.
+        let batch_header =
```
joske, line 1178: There is also a `snarkos_node_narwhal:test_helpers::sample_batch_header(rng);`. Can that be used?

node/narwhal/src/primary.rs
niklaslong: Yes, I need the accounts as well (private keys) but I could maybe refactor this code to make use of the helper (or alternatively expand the helper) 🤔

node/narwhal/src/primary.rs
niklaslong: I can find `sample_batch_certificate` but not `sample_batch_header`?

joske: Sorry, it's in `snarkVM/ledger/src/batch_header/mod.rs:160`

niklaslong: Also can't use it unfortunately, I need the transmissions to create a proposal, not just the transmission IDs, good shout though 👍

node/narwhal/src/primary.rs
```diff
@@ -1053,3 +1053,317 @@ impl<N: Network> Primary<N> {
...
+        rng: &mut TestRng,
+    ) -> (Primary<CurrentNetwork>, Vec<(SocketAddr, Account<CurrentNetwork>)>) {
+        // Create a committee containing the primary's account.
+        let (accounts, committee) = {
+            const COMMITTEE_SIZE: usize = 4;
+            let mut accounts = Vec::with_capacity(COMMITTEE_SIZE);
+            let mut members = IndexMap::new();
+
+            for i in 0..COMMITTEE_SIZE {
+                let socket_addr = format!("127.0.0.1:{i}").parse().unwrap();
```
ljedrz: shouldn't the port range start somewhere later? some (perhaps all of these) of the low ports are reserved

niklaslong: Didn't have issue funnily enough, good catch!

node/narwhal/src/primary.rs
```diff
@@ -1053,3 +1053,317 @@ impl<N: Network> Primary<N> {
...
+
+    // Creates a mock solution.
+    fn sample_unconfirmed_solution(
+        rng: &mut TestRng,
+    ) -> (PuzzleCommitment<CurrentNetwork>, Data<ProverSolution<CurrentNetwork>>) {
+        // Sample a random fake puzzle commitment.
+        let affine = rng.gen();
+        let commitment = PuzzleCommitment::<CurrentNetwork>::from_g1_affine(affine);
+        // Sample random fake solution bytes.
+        let mut vec = vec![0u8; 1024];
```
ljedrz:
```suggestion
        let mut vec = vec![0u8; 1024];
```
should we perhaps have different lengths here, just in case we were to hit some edge case?

node/narwhal/src/primary.rs
```diff
@@ -1053,3 +1053,317 @@ impl<N: Network> Primary<N> {
...
+    }
+
+    // Creates a mock transaction.
+    fn sample_unconfirmed_transaction(
+        rng: &mut TestRng,
+    ) -> (<CurrentNetwork as Network>::TransactionID, Data<Transaction<CurrentNetwork>>) {
+        // Sample a random fake transaction ID.
+        let id = Field::<CurrentNetwork>::rand(rng).into();
+        // Sample random fake transaction bytes.
+        let mut vec = vec![0u8; 1024];
```
ljedrz:
```suggestion
        let mut vec = vec![0u8; 1024];
```
ditto

ljedrz (approved):
LGTM

node/narwhal/src/primary.rs
niklaslong: done

node/narwhal/src/primary.rs
niklaslong: done

node/narwhal/src/primary.rs
niklaslong: aaaand done

node/narwhal/src/primary.rs
```diff
@@ -1079,7 +1079,7 @@ mod tests {
             let mut members = IndexMap::new();
 
             for i in 0..COMMITTEE_SIZE {
-                let socket_addr = format!("127.0.0.1:{i}").parse().unwrap();
+                let socket_addr = format!("127.0.0.1:500{i}").parse().unwrap();
```
ljedrz: what if we have a committee size greater than 10 :detective:?

niklaslong: I thought that might be unlikely in tests—fixed just in case!

joske (approved):

howardwu (approved):

PR #2622, opened 2023-08-24
fix: avoid falling back to random port if desired port not free

joske:
If you ask for a desired port and it's not available, the TCP stack should not silently choose a random port, but return an error.

Some tests needed to be fixed for this, as they were all trying to bind to the same port, making parallel test execution impossible.

### DIFF
--- a/node/src/beacon/mod.rs
+++ b/node/src/beacon/mod.rs
@@ -70,9 +70,11 @@ pub struct Beacon<N: Network, C: ConsensusStorage<N>> {
 
 impl<N: Network, C: ConsensusStorage<N>> Beacon<N, C> {
     /// Initializes a new beacon node.
+    #[allow(clippy::too_many_arguments)]
     pub async fn new(
         node_ip: SocketAddr,
         rest_ip: Option<SocketAddr>,
+        bft_ip: Option<SocketAddr>,
         account: Account<N>,
         trusted_peers: &[SocketAddr],
         genesis: Block<N>,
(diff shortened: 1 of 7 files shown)


node/src/beacon/mod.rs
```diff
@@ -70,13 +70,15 @@ pub struct Beacon<N: Network, C: ConsensusStorage<N>> {
...
     /// Initializes a new beacon node.
+    #[allow(clippy::too_many_arguments)]
     pub async fn new(
         node_ip: SocketAddr,
         rest_ip: Option<SocketAddr>,
         account: Account<N>,
         trusted_peers: &[SocketAddr],
         genesis: Block<N>,
         cdn: Option<String>,
+        bft_ip: Option<SocketAddr>,
```
niklaslong: I'd maybe group it with the rest of the IPs?

node/tcp/src/helpers/config.rs
```diff
@@ -84,7 +84,7 @@ impl Default for Config {
             name: None,
             listener_ip: default_ip(),
             desired_listening_port: None,
-            allow_random_port: true,
+            allow_random_port: false,
```
ljedrz, line 87: this belongs in a separate commit IMO, together with the necessary test changes

joske: Not sure what you mean? All the rest ARE the necessary test changes. I've split up this change from the test changes. Is that what you meant?

ljedrz: sorry for not being clear; I meant the extension of the node creation methods, but I'm guessing those were necessary for the test adjustments

joske: Indeed. Still it makes sense to split of the actual fix from the test fixes.

howardwu:
I would request we take a different approach to handling this issue.

For instance, one could use serial_test or for the test cases add a cfg on a /tmp counter to help assist in this.

joske:
The reason I opened this PR is a conceptual one: it seems wrong to fallback to a random port when you ask for a specific one. 

But, if we fix that, then the tests break, because they rely on the random port fallback.

I'm fine to leave this as is, if you think the fallback problem is not an issue.

PR #2628, opened 2023-08-25
[Narwhal] noise network encryption

niklaslong:
This PR implements noise encryption for the gateway. I've also added proptests testing codec roundtrips for all events. E2e tests (and anything using the gateway) cover the noise handshake followed by protocol messaging.


node/narwhal/src/gateway.rs
```diff
@@ -743,10 +792,33 @@ impl<N: Network> Gateway<N> {
         peer_addr: SocketAddr,
         peer_ip: &mut Option<SocketAddr>,
         stream: &'a mut TcpStream,
-    ) -> io::Result<(SocketAddr, Framed<&mut TcpStream, EventCodec<N>>)> {
+    ) -> io::Result<(SocketAddr, Framed<&mut TcpStream, NoiseCodec<N>>)> {
+        /* Noise codec setup */
+
+        let params: NoiseParams = "Noise_XX_25519_ChaChaPoly_BLAKE2s".parse().unwrap();
```
SplittyDev: I think this should be a constant somewhere

node/narwhal/src/helpers/codec.rs
```diff
@@ -84,3 +84,311 @@ impl<N: Network> Decoder for EventCodec<N> {
...
+
+pub enum NoiseState {
+    Handshake(Box<HandshakeState>),
+    PostHandshake(PostHandshakeState),
+}
+
+impl Clone for NoiseState {
+    fn clone(&self) -> Self {
+        match self {
+            Self::Handshake(..) => unimplemented!(),
```
ljedrz:
```suggestion
            Self::Handshake(..) => unreachable!(),
```
IIRC

node/narwhal/src/helpers/codec.rs
```diff
@@ -84,3 +84,311 @@ impl<N: Network> Decoder for EventCodec<N> {
...
+    }
+}
+
+impl NoiseState {
+    pub fn into_post_handshake_state(self) -> Self {
+        if let Self::Handshake(noise_state) = self {
+            let noise_state = noise_state.into_stateless_transport_mode().expect("handshake isn't finished");
+            Self::PostHandshake(PostHandshakeState { state: Arc::new(noise_state), tx_nonce: 0, rx_nonce: 0 })
+        } else {
+            panic!()
```
ljedrz:
```suggestion
            unreachable!()
```

node/narwhal/src/helpers/codec.rs
```diff
@@ -84,3 +84,311 @@ impl<N: Network> Decoder for EventCodec<N> {
...
+                    // Don't allow message sending before the noise handshake has completed.
+                    EventOrBytes::Event(_) => unimplemented!(),
+                    EventOrBytes::Bytes(bytes) => {
+                        let mut buffer = [0u8; MAX_MESSAGE_LEN + 1];
+                        let len = noise
+                            .write_message(&bytes, &mut buffer[1..])
+                            .map_err(|e| Self::Error::new(io::ErrorKind::InvalidInput, e))?;
+
+                        // Set the message type flag.
+                        buffer[0] = message_type as u8;
```
ljedrz: this can prolly go just below the `buffer` binding

node/narwhal/src/helpers/codec.rs
```diff
@@ -84,3 +84,311 @@ impl<N: Network> Decoder for EventCodec<N> {
...
+                    }
+                }
+            }
+
+            NoiseState::PostHandshake(ref mut noise) => {
+                // Encode the message using the event codec.
+                let mut bytes = BytesMut::new();
+                match message_or_bytes {
+                    // Don't allow sending raw bytes after the noise handshake has completed.
+                    EventOrBytes::Bytes(_) => unimplemented!(),
```
ljedrz:
```suggestion
                    EventOrBytes::Bytes(_) => panic!("Unsupported post-handshake"),
```

node/narwhal/src/helpers/codec.rs
```diff
@@ -84,3 +84,311 @@ impl<N: Network> Decoder for EventCodec<N> {
...
+                            .write_message(noise.tx_nonce + nonce_offset as u64, plaintext_chunk, &mut buffer)
+                            .map_err(|e| Self::Error::new(io::ErrorKind::InvalidInput, e))?;
+
+                        buffer.truncate(len);
+
+                        Ok(buffer)
+                    })
+                    .collect();
+
+                let mut buffer = BytesMut::new();
```
ljedrz:
```suggestion
                let mut buffer = BytesMut::with_capacity(1 + encrypted_chunks.iter().map(|c| c.len()).sum::<usize>());
```

node/narwhal/src/helpers/codec.rs
```diff
@@ -84,3 +84,311 @@ impl<N: Network> Decoder for EventCodec<N> {
...
+                let len = noise.read_message(&bytes, &mut buffer).map_err(|_| io::ErrorKind::InvalidData)?;
+
+                Some(EventOrBytes::Bytes(Bytes::copy_from_slice(&buffer[..len])))
+            }
+
+            NoiseState::PostHandshake(ref mut noise) => {
+                // Ignore raw bytes after the noise handshake has completed.
+                if let MessageType::Bytes = flag {
+                    return Ok(None);
+                }
```
ljedrz: hmm, now that I think about it, do we even need this flag if we always send bytes during the handshake and events post-handshake?

ljedrz: we would still need the `enum EventOrBytes`, but not necessarily the leading byte or its `enum`

node/narwhal/src/helpers/codec.rs
```diff
@@ -84,3 +84,311 @@ impl<N: Network> Decoder for EventCodec<N> {
...
+
+                // Noise decryption.
+                let chunked_encrypted_msg: Vec<_> = bytes.chunks(MAX_MESSAGE_LEN).collect();
```
ljedrz: couldn't we instead use `bytes.par_chunks()` for the purposes of parallel processing?

node/narwhal/src/helpers/codec.rs
```diff
@@ -84,3 +84,311 @@ impl<N: Network> Decoder for EventCodec<N> {
...
+                }
+
+                // Chunk the payload if necessary.
+                //
+                // A Noise transport message is simply an AEAD ciphertext that is less than or
+                // equal to 65535 bytes in length, and that consists of an encrypted payload plus
+                // 16 bytes of authentication data.
+                //
+                // See: https://noiseprotocol.org/noise.html#the-handshakestate-object
+                let chunked_plaintext_msg: Vec<_> = bytes.chunks(MAX_MESSAGE_LEN - 16).collect();
```
ljedrz: same note as on decryption - check if `par_chunks` could be used here to avoid the intermediate vector

node/narwhal/src/helpers/codec.rs
```diff
@@ -84,3 +84,311 @@ impl<N: Network> Decoder for EventCodec<N> {
...
+                let num_chunks = chunked_plaintext_msg.len() as u64;
+
+                // Encrypt the resulting bytes with Noise.
+                let encrypted_chunks: Vec<io::Result<Vec<u8>>> = chunked_plaintext_msg
```
ljedrz: maybe we could collect into an `io::Result<Vec<Vec<_>>>` instead? it should still short-circuit

ljedrz (commented):
Very solid work overall, will do another pass once the current comments are address.

niklaslong:
I think that's all the comments addressed 🙏

node/narwhal/src/helpers/codec.rs
```diff
@@ -84,3 +92,244 @@ impl<N: Network> Decoder for EventCodec<N> {
...
+
+                // Chunk the payload if necessary and encrypt with Noise.
+                //
+                // A Noise transport message is simply an AEAD ciphertext that is less than or
+                // equal to 65535 bytes in length, and that consists of an encrypted payload plus
+                // 16 bytes of authentication data.
+                //
+                // See: https://noiseprotocol.org/noise.html#the-handshakestate-object
+                let encrypted_chunks = bytes
+                    .par_chunks(MAX_MESSAGE_LEN - 16)
```
ljedrz:
```suggestion
                    .par_chunks(MAX_MESSAGE_LEN - TAG_LEN)
```
we should add this `const` so that it's clear where the `16` comes from

node/narwhal/src/helpers/codec.rs
```diff
@@ -84,3 +92,244 @@ impl<N: Network> Decoder for EventCodec<N> {
...
+                    buffer.extend_from_slice(&chunk);
+                    noise.tx_nonce += 1;
+                }
+
+                buffer
+            }
+        };
+
+        // Encode the resulting ciphertext using the length-delimited codec.
+        self.codec.encode(ciphertext.freeze(), dst)
```
ljedrz, line 216: I missed this before, but I think we might instead want to have this call within the branches in order to write directly into the output buffer, without going through intermediate bytes

node/narwhal/src/helpers/codec.rs
```diff
@@ -84,3 +92,244 @@ impl<N: Network> Decoder for EventCodec<N> {
...
+                // Decrypt the ciphertext in handshake mode.
+                let mut buffer = [0u8; MAX_MESSAGE_LEN];
+                let len = noise.read_message(&bytes, &mut buffer).map_err(|_| io::ErrorKind::InvalidData)?;
+
+                Some(EventOrBytes::Bytes(Bytes::copy_from_slice(&buffer[..len])))
+            }
+
+            NoiseState::PostHandshake(ref mut noise) => {
+                // Noise decryption.
+                let decrypted_chunks: Vec<io::Result<Vec<u8>>> = bytes
```
ljedrz:
```suggestion
                let decrypted_chunks: io::Result<Vec<Vec<u8>>> = bytes
```
(also needs an adjustment when chunking)

ljedrz (commented):
Just a few extra suggestions and I think it's good to go :ok_hand:.

node/narwhal/src/helpers/codec.rs
niklaslong: As discussed, we shouldn't do this because of the length delimited codec, we'd be writing a `u32` per chunk to the stream.

ljedrz (approved):
LGTM :+1:

SplittyDev (approved):
Nice 👍

joske (approved):

howardwu (approved):

PR #2649, opened 2023-09-11
[Testnet3] Update README.md

web3deadline:
2.1.7 has new private-key and private-key-file params and some explanations in readme need to be update, delete wrong param

## Motivation

make it right and easy to use snarkOS

## Test Plan

no need

## Related PRs

none


web3deadline:
@howardwu

howardwu (approved):

PR #2654, opened 2023-09-12
Election test race

vbar:
## Motivation

Fixing an intermittent test failure.

## Test Plan

Changes are only in test code.


joske (approved):

howardwu:
@vbar Please use the precommit hook.

PR #2657, opened 2023-09-13
fix: make it more elegant

web3deadline:
## Motivation

I found the "// FIXME: this can probably be done more elegantly", and try to fix it

## Test Plan

No need

## Related PRs

None

### DIFF
--- a/node/tcp/src/helpers/mod.rs
+++ b/node/tcp/src/helpers/mod.rs
@@ -26,24 +26,20 @@ pub use stats::Stats;
 
 use tracing::{debug_span, error_span, info_span, trace_span, warn_span, Span};
 
-// FIXME: this can probably be done more elegantly
 /// Creates the Tcp's tracing span based on its name.
 pub fn create_span(tcp_name: &str) -> Span {
     let mut span = trace_span!("tcp", name = tcp_name);
-    if !span.is_disabled() {
-        return span;
-    } else {
+    if span.is_disabled() {
         span = debug_span!("tcp", name = tcp_name);
     }
-    if !span.is_disabled() {
-        return span;
-    } else {
+    if span.is_disabled() {
         span = info_span!("tcp", name = tcp_name);
     }
-    if !span.is_disabled() {
-        return span;
-    } else {
+    if span.is_disabled() {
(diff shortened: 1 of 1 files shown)


ljedrz (approved):
This is not exactly what I had in mind (I'm mostly looking for a way to remove the chain of `if`s), but hey, this _is_ a bit more elegant :ok_hand:. Thanks!

web3deadline:
> This is not exactly what I had in mind (I'm mostly looking for a way to remove the chain of `if`s), but hey, this _is_ a bit more elegant 👌. Thanks!

If we want to remove the 'if', maybe we can do like this, define the levels and use an iter. I'm a green hand in rust, so I can't judge whether it's suitable. I think 'if' easier to read, but 'iter' looks more elegant. How do u feel about the new one? If it make sense, I'll change my commit.
```rust
pub fn create_span(tcp_name: &str) -> Span {
    let log_levels = [
        trace_span!("tcp", name = tcp_name),
        debug_span!("tcp", name = tcp_name),
        info_span!("tcp", name = tcp_name),
        warn_span!("tcp", name = tcp_name),
        error_span!("tcp", name = tcp_name),
    ];

    // default "error"
    let mut selected_span = error_span!("tcp", name = tcp_name);

    for span in log_levels.iter() {
        if !span.is_disabled() {
            selected_span = span.clone();
            break;
        }
    }

    selected_span
}
```

ljedrz:
Thanks, though this still doesn't resolve the issue of possibly having to go over all possible span levels instead of just picking the "highest" one available; I'm hoping that one day it will be made more ergonomic by `tracing`.

web3deadline:
> Thanks, though this still doesn't resolve the issue of possibly having to go over all possible span levels instead of just picking the "highest" one available; I'm hoping that one day it will be made more ergonomic by `tracing`.

So just use if this time, I haven't found other place where init the span level.🤔

web3deadline:
> > Thanks, though this still doesn't resolve the issue of possibly having to go over all possible span levels instead of just picking the "highest" one available; I'm hoping that one day it will be made more ergonomic by `tracing`.
> > So just use if this time, I haven't found other place where init the span level.🤔

howardwu (approved):

PR #2659, opened 2023-09-14
WIP: Warnings

joske:
## Motivation

stream line the logging: 
* remove `warn!` where safe
  * safely skipping rounds: this seems totally safe, it even says so in the message
* disambiguate identical logging lines

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3163,6 +3163,7 @@ dependencies = [
  "snarkvm",
  "snow",
  "test-strategy",
+ "thiserror",
  "time",
  "tokio",
  "tokio-stream",
--- a/node/narwhal/Cargo.toml
+++ b/node/narwhal/Cargo.toml
@@ -19,6 +19,9 @@ edition = "2021"
 [dependencies.anyhow]
 version = "1.0.70"
 
+[dependencies.thiserror]
+version = "1.0.48"
+
 [dependencies.async-recursion]
 version = "1.0"
 
--- a/node/narwhal/src/helpers/proposal.rs
+++ b/node/narwhal/src/helpers/proposal.rs
@@ -153,7 +153,7 @@ impl<N: Network> Proposal<N> {
         }
         // Ensure the signer is new.
         if self.signers().contains(&signer) {
-            bail!("Signature is from a duplicate peer '{signer}'")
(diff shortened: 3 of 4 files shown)


node/narwhal/src/bft.rs
```diff
@@ -706,7 +712,10 @@ mod tests {
         // Ensure this call fails on an even round.
         let result = bft.is_leader_quorum_or_nonleaders_available(2);
         assert!(result.is_err());
-        assert_eq!(result.unwrap_err().to_string(), "BFT storage reference is out of sync with the current round");
+        assert_eq!(
+            result.unwrap_err().to_string(),
+            "is_leader_quorum_or_nonleaders_available: BFT storage reference is out of sync with the current round"
```
vbar: I wouldn't re-use the same error message - even in test, it's going to confuse people grepping for it...

node/narwhal/src/bft.rs
```diff
@@ -792,7 +801,10 @@ mod tests {
         // Ensure this call succeeds on an even round.
         let result = bft.update_leader_certificate_to_even_round(6);
         assert!(result.is_err());
-        assert_eq!(result.unwrap_err().to_string(), "BFT storage reference is out of sync with the current round");
+        assert_eq!(
+            result.unwrap_err().to_string(),
+            "update_leader_certificate_to_even_round: BFT storage reference is out of sync with the current round"
```
vbar: ditto

node/narwhal/src/bft.rs
joske: what do you suggest?

vbar: I'd probably leave it unchanged - then the message is unique, and the test framework will show which test it's failing in...

joske: I don't understand what you mean. This text needs to match the actual error returned.

node/narwhal/src/bft.rs
vbar: Oh, sorry, then it's me who doesn't understand...

node/narwhal/src/bft.rs
vbar: Oh, sorry, then it's me who doesn't understand...

vbar (approved):

PR #2673, opened 2023-09-20
[TOB] Improve peer address filtering

ljedrz:
This approach allows us to be stricter and more clear about the filtering of addresses received from other peers while not restricting addresses the node may want to connect to itself.

Finding: TOB-ALEO-11

### DIFF
--- a/node/router/src/lib.rs
+++ b/node/router/src/lib.rs
@@ -47,14 +47,7 @@ use anyhow::{bail, Result};
 use core::str::FromStr;
 use indexmap::{IndexMap, IndexSet};
 use parking_lot::{Mutex, RwLock};
-use std::{
-    collections::HashSet,
-    future::Future,
-    net::{IpAddr, SocketAddr},
-    ops::Deref,
-    sync::Arc,
-    time::Instant,
-};
+use std::{collections::HashSet, future::Future, net::SocketAddr, ops::Deref, sync::Arc, time::Instant};
 use tokio::task::JoinHandle;
 
 #[derive(Clone)]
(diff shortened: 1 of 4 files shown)


ljedrz:
An additional improvement would be to merge `Router::is_peer_ip_valid` with `is_bogon_address` so that all peer address filtering would happen at the same place; I'll propose this in another commit.

node/router/src/lib.rs
```diff
@@ -210,10 +213,12 @@ impl<N: Network> Router<N> {
...
-    /// Returns `true` if the given IP is this node.
-    pub fn is_local_ip(&self, ip: &SocketAddr) -> bool {
-        *ip == self.local_ip()
-            || (ip.ip().is_unspecified() || ip.ip().is_loopback()) && ip.port() == self.local_ip().port()
+    /// Returns `true` if the IP provided by a peer is a valid candidate peer address.
+    pub fn is_peer_ip_valid(&self, ip: &SocketAddr) -> bool {
+        *ip != self.local_ip()
+            && !ip.ip().is_unspecified()
+            && !ip.ip().is_loopback()
+            && if let IpAddr::V4(ip) = ip.ip() { !ip.is_broadcast() } else { true }
```
joske: so V6 addresses are always valid?

joske: These also have multicast. And I guess many more features (but I don't know them).

ljedrz: no, they just can't be broadcast, so the final condition is ignored for them; it might not be clear enough, it should be more apparent in the next commit

ljedrz: > These also have multicast.

indeed, there are some quirks involved there, but I don't think we've seen any issues with them in the past

ljedrz:
Ultimately the peer address filtering is still twofold, as at the point of `Inbound` we can't check for our local address or whether we're already connected to a given peer. However, those are now the only two points where the checks are being performed, and the checks are made more clear.

node/tcp/src/lib.rs
```diff
@@ -43,3 +43,11 @@ pub fn is_bogon_address(ip: IpAddr) -> bool {
         IpAddr::V6(ipv6) => ipv6.is_loopback(),
     }
 }
+
+/// Checks if the given IP address is unspecified or broadcast.
+pub fn is_unspecified_address(ip: IpAddr) -> bool {
+    match ip {
+        IpAddr::V4(ipv4) => ipv4.is_unspecified() || ipv4.is_broadcast(),
+        ipv6 => ipv6.is_unspecified(),
```
joske, line 51: I see that all checks pass, but how can this compile? Can you leave out the variant if there is only one left? I didn't know that.

ljedrz: it just acts as a wildcard; I might as well have written:
```
_ => ip.is_unspecified(),
```

joske (approved):

niklaslong (approved):

PR #2674, opened 2023-09-20
Expand timestamp liveness check

niklaslong:
The liveness check verifies the timestamp equals or is larger than the timestamp for the previous round which is the median timestamp of all its batches. This is implemented through calls to storage.


howardwu:
Superseded by https://github.com/AleoHQ/snarkOS/pull/3187

PR #2683, opened 2023-09-22
[Narwhal] detect duplicate noise state in handshake

niklaslong:
This PR checks if an attempt is made to store a second noise state for a given listener IP and, if so, drops the handshake without mutating existing state. This prevents double-connects from corrupting internal state and causing issues with the pre-existing connections. Tested locally, though the issue is hard to reproduce to begin with.

### DIFF
--- a/node/narwhal/src/gateway.rs
+++ b/node/narwhal/src/gateway.rs
@@ -891,6 +891,13 @@ impl<N: Network> Handshake for Gateway<N> {
         // Remove the address from the collection of connecting peers if the handshake got to the point where it's known,
         // even if the entire process has failed in the end.
         if let Some(ip) = peer_ip {
+            // If the handshake has failed AND it's not a duplicate handshake attempt, remove the noise state.
+            // FIXME: we should deduplicate connections earlier in the handshake to avoid intricate logic
+            if let Err(e) = &handshake_result {
+                if e.kind() != io::ErrorKind::AlreadyExists {
+                    self.noise_states.write().remove(&ip);
+                }
+            }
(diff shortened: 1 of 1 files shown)


node/narwhal/src/gateway.rs
```diff
@@ -877,30 +888,34 @@ impl<N: Network> Handshake for Gateway<N> {
...
-                warn!("{CONTEXT} Handshake with '{peer_ip}' failed: noise state already exists");
-                return Err(io::ErrorKind::AlreadyExists.into());
-            }
-            Vacant(entry) => {
-                // Otherwise, we store the noise state for reuse by the Reading and Writing impls
-                // and proceed.
+            Occupied(mut entry) => {
+                // A connection with this peer listener has already occured, cancel the handshake.
+                if entry.get().is_some() {
+                    self.connecting_peers.lock().shift_remove(peer_ip);
```
ljedrz: this should be unnecessary, as it's done in `perform_handshake` in case of an early return here

ljedrz: nvm, I mixed the sites :ok_hand:

ljedrz (approved):
LGTM :+1:

ljedrz:
Drafting again to make sure all the cleanups work correctly.

niklaslong (commented):
LGTM too, but let's give this another test before undrafting 👍

ljedrz:
Closing until `noise` is brought back. @niklaslong if you intend to have some `noise` branch tracking the default one in the background, these commits should be cherry-picked into it.

PR #2684, opened 2023-09-22
Adds validators request and response events

howardwu:
## Motivation

Adds validators request and response events.

This PR also introduces additional checks on the validator address. The gateway now enforces that the same address does not connect twice within the gateway network. Note that this deviates from what the router enforces.


raychu86 (approved):
lgtm

PR #2685, opened 2023-09-24
Add synchronization methods in storage

howardwu:
## Motivation

Add synchronization methods in storage

### DIFF
--- a/node/narwhal/src/helpers/storage.rs
+++ b/node/narwhal/src/helpers/storage.rs
@@ -15,16 +15,20 @@
 use crate::helpers::check_timestamp_for_liveness;
 use snarkos_node_narwhal_ledger_service::LedgerService;
 use snarkvm::{
-    ledger::narwhal::{BatchCertificate, BatchHeader, Transmission, TransmissionID},
+    ledger::{
+        block::Block,
+        narwhal::{BatchCertificate, BatchHeader, Transmission, TransmissionID},
+    },
     prelude::{bail, ensure, Address, Field, Network, Result},
 };
 
 use indexmap::{map::Entry, IndexMap, IndexSet};
-use parking_lot::RwLock;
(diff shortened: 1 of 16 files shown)


node/narwhal/src/helpers/storage.rs
```diff
@@ -534,6 +569,131 @@ impl<N: Network> Storage<N> {
...
+                        Some(solution) => missing_transmissions.insert(*transmission_id, (*solution).into()),
+                        None => {
+                            error!("Missing solution {puzzle_commitment} in block {}", block.height());
+                            continue;
+                        }
+                    };
+                }
+                TransmissionID::Transaction(transaction_id) => {
+                    // Retrieve the transaction.
+                    match block.get_transaction(transaction_id) {
```
raychu86, line 686: Note: This will not be be able to fetch the original transmission for rejected transactions. 

The transmissionID won't match the transaction id of the `ConfirmedTransaction` in the block.

node/narwhal/src/helpers/storage.rs
```diff
@@ -534,6 +569,131 @@ impl<N: Network> Storage<N> {
...
+        self.sync_storage_with_block(latest_block)
+    }
+
+    /// Syncs the storage with the block.
+    pub fn sync_storage_with_block(&self, block: Block<N>) -> Result<()> {
+        // Retrieve the block height.
+        let block_height = block.height();
+        // Determine the earliest height, conservatively set to the block height minus the max GC rounds.
+        // By virtue of the BFT protocol, we can guarantee that all GC range blocks will be loaded.
+        let gc_height = block_height.saturating_sub(u32::try_from(self.max_gc_rounds)?);
```
raychu86, line 596: There could be many rounds between blocks, so this is treating `max_gc_rounds` more as `max_gc_block_heights`.

PR #2688, opened 2023-09-25
[Narwhal] revert to plaintext codec until noise is stabilised

niklaslong:
Reverts the network to use the `EventCodec` until rayon-enabled Noise can be troubleshooted with larger message sizes.

### DIFF
--- a/node/narwhal/events/src/helpers/codec.rs
+++ b/node/narwhal/events/src/helpers/codec.rs
@@ -99,7 +99,7 @@ impl<N: Network> Decoder for EventCodec<N> {
 /* NOISE CODEC */
 
 // The maximum message size for noise messages. If the data to be encrypted exceedes it, it is chunked.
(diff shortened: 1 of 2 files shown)


joske (approved):

vvp (commented):
Perhaps we should clean all Noise related code in this rollback? (e.g. code in node/narwhal/events/src/helpers/codec.rs)

ljedrz:
> Perhaps we should clean all Noise related code in this rollback?

I'd rather keep it, it will be easier to enable it again later on.

ljedrz (commented):
I'd revert the `MAX_MESSAGE_LEN` to the protocol value in this PR as well, so as not to forget about it later on.

niklaslong:
Rebased and
> I'd revert the MAX_MESSAGE_LEN

done 👍

howardwu (approved):
LGTM

PR #2690, opened 2023-09-25
Delay peer status promotion

ljedrz:
This PR aims to resolve a rare issue where the high-level peer collection indicates that a peer is connected while the TCP stack is not aware of it, requiring a workaround in `Router::disconnect`. Based on the logs I've analyzed, it seems that we need to make sure that the high-level collection of peers is updated only once all the lower-level TCP work is complete; the way to do it is via the `OnConnect` protocol.

The workaround remains in place for now so that we can make sure that we don't see it being used anymore.

Filing as a draft to do some more testing and consider additional improvements to the logic.

### DIFF
--- a/node/router/src/handshake.rs
+++ b/node/router/src/handshake.rs
@@ -29,7 +29,7 @@ use snarkvm::prelude::{block::Header, error, Address, Network};
 use anyhow::{bail, Result};
 use futures::SinkExt;
 use rand::{rngs::OsRng, Rng};
-use std::{io, net::SocketAddr};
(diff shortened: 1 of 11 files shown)


niklaslong (approved):

ljedrz:
I think it's ready for review now, but I'd recommend going though the devnet before merging this, as it's pretty delicate code and the workaround is not a big issue (especially if this PR fixes the root cause).

ljedrz:
Superseded by https://github.com/AleoHQ/snarkOS/pull/2729.

PR #2691, opened 2023-09-25
[Narwhal] fuzz target for chunked noise codec

niklaslong:
This PR adds a fuzz target for the noise codec and a prop test case with larger transmissions to exercise the rayon-based chunked message encoding and decoding. It also reverts the `MAX_MESSAGE_LEN` constant back to the value defined by the noise specification. 

```
#4194304	pulse  cov: 1889 ft: 2911 corp: 80/191Kb lim: 57739 exec/s: 1982 rss: 652Mb
```


ljedrz:
I think we can apply the same treatment here as in https://github.com/AleoHQ/snarkOS/pull/2683#issuecomment-1734124319.

niklaslong:
Indeed, I'll cherry pick these into another branch 👍

PR #2692, opened 2023-09-25
Use previous committee for quorum and membership checks

raychu86:
## Motivation

This PR updates the quorum checks and committee membership checks for each certificate and proposal to the the previous certificate. This is required because if we use the current committee instead of the previous, new committee changes could cause invalid certificate state.

Example:
If a committee is changed in round 5, then certificates from round 5 must be checked with a committee in round 4.  In addition, round 6 must also use the committee in round 4, because round 5 can't be committed without round 6. 

Previously we were not testing committee changes, so using the use of `current_committee` was fine because the quorum requirements would be the same.

### DIFF
--- a/node/narwhal/ledger-service/src/ledger.rs
+++ b/node/narwhal/ledger-service/src/ledger.rs
@@ -135,6 +135,7 @@ impl<N: Network, C: ConsensusStorage<N>> LedgerService<N> for CoreLedgerService<
     }
 
     /// Returns the previous committee for the given round.
+    /// If the previous round is in the future, then the current committee is returned.
     fn get_previous_committee_for_round(&self, round: u64) -> Result<Committee<N>> {
         // Get the round number for the previous committee. Note, we subtract 2 from odd rounds,
         // because committees are updated in even rounds.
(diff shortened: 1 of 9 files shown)


node/narwhal/ledger-service/src/ledger.rs
```diff
@@ -134,6 +134,20 @@ impl<N: Network, C: ConsensusStorage<N>> LedgerService<N> for CoreLedgerService<
         }
     }
 
+    /// Returns the previous committee for the given round.
+    /// If the previous round is in the future, then the current committee is returned.
+    fn get_previous_committee_for_round(&self, round: u64) -> Result<Committee<N>> {
+        // Get the round number for the previous committee. Note, we subtract 2 from even rounds,
+        // because committees are updated in odd rounds.
+        let previous_round = match round % 2 == 0 {
```
raychu86, line 142: Technically, `let previous_round = round.saturating_sub(2);` should suffice as well. Since an `odd_round - 2` and `odd_round - 1` would reference the same committee.

howardwu (approved):
LGTM

PR #2696, opened 2023-09-26
Handle `sync_certificate_with_block` for rejected transactions

raychu86:
## Motivation

This PR uses the `unconfirmed_transactions` when processing transmissions in `sync_certificate_with_block`. This allows us to boot and process rejected transactions. 

This is required because the if a transaction is rejected, the ID will be different from its original transmission ID. Thus, requiring the use of `ConfirmedTransaction::unconfirmed_transaction` to derive the original transmission.

NOTE: This does not cover the case of `aborted` transmission transactions.

### DIFF
--- a/node/narwhal/src/helpers/storage.rs
+++ b/node/narwhal/src/helpers/storage.rs
@@ -607,6 +607,14 @@ impl<N: Network> Storage<N> {
         }
         // Retrieve the transmissions for the certificate.
         let mut missing_transmissions = HashMap::new();
+
+        // Reconstruct the unconfirmed transactions.
+        let mut unconfirmed_transactions = block
+            .transactions()
+            .iter()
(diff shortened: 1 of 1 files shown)


howardwu (approved):

node/narwhal/src/helpers/storage.rs
```diff
@@ -607,6 +607,14 @@ impl<N: Network> Storage<N> {
         }
         // Retrieve the transmissions for the certificate.
         let mut missing_transmissions = HashMap::new();
+
+        // Reconstruct the unconfirmed transactions.
+        let mut unconfirmed_transactions = block
+            .transactions()
+            .iter()
```
howardwu: Perhaps use `cfg_iter!`?

PR #2698, opened 2023-09-26
Ensures the core primary events originate from the validator

howardwu:
## Motivation

Ensures the core primary events originate from the validator


raychu86 (approved):
LGTM

PR #2700, opened 2023-09-26
Re-enable noise encryption

niklaslong:
This PR reintroduces Noise encryption along with some tweaks to the handshaking logic to detect double-connects.
I've also included a fuzz target for the codec.

Testing: I created another branch that includes #2692 and ran a small devnet and successfully bonding new validators. The implementation seems quite stable in normal operations. #2690 should also increase stability.


niklaslong:
Rebased and updated fuzz targets.

niklaslong:
Deprioritised for mainnet, closing for now.

PR #2706, opened 2023-09-27
Switch to a more elegant detection handler for self connect

howardwu:
## Motivation

Switch to a more elegant detection handler for self connect


raychu86 (approved):
lgtm

PR #2709, opened 2023-09-27
Switch HDD check to println

raychu86:
## Motivation

This PR changes the HDD check on validator bootup to a println instead of a painc


howardwu (approved):
LGTM

PR #2711, opened 2023-09-27
[Golden Master] AleoBFT Release

howardwu:
## Motivation

[Golden Master] AleoBFT Release


HarukaMa:
the disk type check (`snarkOS/cli/src/helpers/mod.rs`) feels strange to me - if a disk is called `sda` then it most likely is not NVMe as all devices under NVMe driver should be called something like `nvme0`; the machine might have both SATA disks and NVMe disks, and the node data folder might very well reside in the NVMe disk (in fact for some servers the `sda` might be just the usb stick used to store `/boot`); for VMs, depending on the drive type configured, the virtual disk might be marked as a spinning disk while the underlying storage device is actually NVMe SSDs.

Also "Ubuntu required" message feels strange as well. (I personally avoid Ubuntu as best as I can)

howardwu:
Good point, we'll remove the disk checks.

howardwu:
^I've removed the drive checks

PR #2712, opened 2023-09-28
Fix syncing edge-case

raychu86:
## Motivation

This PR changes updates `Sync::is_synced` to return false if the gateway is not connected to any peers. This also addresses an edge-case where the node initially fails to sync properly.


joske (approved):

Pjrich1313 (commented):
Rerun nobe debug

PR #2717, opened 2023-09-28
Add additional check for `has_unconfirmed_transaction`

raychu86:
## Motivation

This PR adds an additional check that batches can't be proposed unless it contains at least one unconfirmed transaction.


howardwu (approved):
LGTM

PR #2722, opened 2023-10-02
Automatic tests upgrade

vbar:
## Motivation

`cargo test` fails in the node/narwhal package

## Test Plan

Changes are mostly in test code; the one change in sync skips sending of a message the recipient doesn't handle.


node/narwhal/tests/common/trans.rs
```diff
@@ -0,0 +1,168 @@
...
+        committee::Committee,
+        narwhal::{Data, Subdag, Transmission, TransmissionID},
+        store::ConsensusStorage,
+        Ledger,
+    },
+    prelude::{narwhal::BatchCertificate, Field, Network, Result},
+};
+use std::{fmt, ops::Range};
+
+pub struct TranslucentLedgerService<N: Network, C: ConsensusStorage<N>> {
```
howardwu: nit: can you move this under `ledger-service` with a `cfg(feature = "test")` guard and `test = [ ]` feature flag?

node/narwhal/tests/common/primary.rs
```diff
@@ -237,10 +262,24 @@ impl TestNetwork {
...
+                            if first != committee {
+                                return false;
+                            }
+                        }
+                    }
+                }
+            }
+        }
+
+        return true;
```
ljedrz:
```suggestion
        true
```

node/narwhal/tests/common/trans.rs
vbar: moved

howardwu (approved):

PR #2729, opened 2023-10-03
[TOB] Clean up the list of connecting peers in case of handshake timeout

ljedrz:
Supersedes https://github.com/AleoHQ/snarkOS/pull/2690 as a much simpler and tighter solution.

I verified that this works by manually reducing the handshake timeout to 1ms and checking the number of connecting peers in 2 nodes after a timed out handshake between them (which is non-zero in `testnet3` and zero with this PR). While adding a regular test case for this would be possible, it would be quite verbose (as one of the nodes would need to use a custom handshake impl), so I didn't include one.

This change should obviate the current extra `Disconnect` check, but I'm leaving it in for now just to be sure we're no longer hitting it.

Finding: TOB-ALEO-22

### DIFF
--- a/node/bft/src/gateway.rs
+++ b/node/bft/src/gateway.rs
@@ -394,15 +394,17 @@ impl<N: Network> Gateway<N> {
     }
 
     /// Ensure the peer is allowed to connect.
-    fn ensure_peer_is_allowed(&self, peer_ip: SocketAddr) -> Result<()> {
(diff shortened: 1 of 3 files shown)


ljedrz:
Rebased and applied the same fix for the `Gateway`.

howardwu:
It probably works but isn't a great design.

ljedrz:
Alternatives are difficult, because there is no high-level callsite where an inbound connection timeout can be handled; having that would break the modularity of the network stack, and would be significantly more complex than this proposal.

zosorock:
Superseded by #3418

PR #2732, opened 2023-10-03
Update `UnconfirmedSolution` propagation

raychu86:
## Motivation

This PR updates the propagation rules of UnconfirmedSolutions.

- Provers propagate to all connected peers.

- Clients propagate to all connected peers.

- Validators only propagate solutions to other validators.


howardwu (approved):
LGTM

PR #2741, opened 2023-10-06
Prevent committing the subdag certificates on block failures

raychu86:
## Motivation

This PR ensures that we don't prematurely commit subdag certificates to the DAG if the subdag failed to be committed in a block. 

This is important because, if the block production fails for any reason, but we already assume that certain certificates are canonicalized in the DAG, then subsequent block production will create corrupted state.


howardwu (approved):
LGTM

PR #2744, opened 2023-10-09
Update snarkVM rev; Removes `Result` from `update_to_next_round`

howardwu:
## Motivation

Removes `Result` from `update_to_next_round`.

The goal here is to improve ergonomics and to catch any case where failure is unpredictable.

Note: I have tested this code change on my machine and it performs unchanged.

In addition, to support the test cases, snarkVM is updated to `0.15.4`.


raychu86 (approved):
LGTM

PR #2750, opened 2023-10-11
Update README.md about dev node clean up

web3deadline:
Add dev clean up method

## Motivation

after reset, dev node need to be cleaned up


howardwu:
The instruction to remove state is already there. By running the command, we can ensure all future devs remove all necessary state. Just because it is 1 folder now does not mean it will always stay 1 folder.

PR #2752, opened 2023-10-11
fix: improve proptest: don't generate same block every time

joske:
improve proptest: don't generate same block every time


ljedrz (approved):
LGTM :ok_hand:

howardwu (approved):

PR #2754, opened 2023-10-11
Update run-client.sh

collinc97:
Add `--client` to  `./run-client.sh` script.


raychu86 (approved):
lgtm.

The node should default to running as a client without this `--client` flag, but I suspect it fails now due to the new private key parsing.

PR #2757, opened 2023-10-13
Optimize for single lock read when check proposed_batch expiration

ghostant-1017:
## Motivation

Check the proposed batch in single read.


ghostant-1017:
Further, Do we need to hold a write lock for the duration of the function? It will make sense in some race conditions.

ljedrz (approved):
LGTM :ok_hand:; there are likely several more improvements like this to be had in the codebase.

howardwu (approved):

PR #2759, opened 2023-10-13
Speed up Cache::retain_and_insert

ljedrz:
I noticed `Cache::retain_and_insert` popping up in heap profiles due to the use of `BTreeMap::split_off`, and after some investigating I realized that most of those calls can be avoided, because usually all of the contents of the cache are retained due to the cutoff time being lower (i.e. "older") than all of the timestamps in the cache.

In addition, there is an improvement to `max_cache_duplicates`, which currently costs a double database lookup and double allocation; with this trivial change, only one of each is needed to calculate it.

These changes reduced the number of allocations by ~14% in my test run.

### DIFF
--- a/node/narwhal/src/gateway.rs
+++ b/node/narwhal/src/gateway.rs
@@ -218,7 +218,7 @@ impl<N: Network> Gateway<N> {
 
     /// The maximum number of duplicates for any particular request.
     fn max_cache_duplicates(&self) -> usize {
-        self.max_committee_size() * self.max_committee_size()
+        self.max_committee_size().pow(2)
     }
 }
 
--- a/node/narwhal/src/helpers/cache.rs
+++ b/node/narwhal/src/helpers/cache.rs
@@ -135,15 +135,27 @@ impl<N: Network> Cache<N> {
         let mut map_write = map.write();
         // Insert the new timestamp and increment the frequency for the key.
         *map_write.entry(now).or_default().entry(key).or_default() += 1;
-        // Extract the subtree after interval (i.e. non-expired entries)
-        let retained = map_write.split_off(&now.saturating_sub(interval_in_secs));
-        // Clear all the expired entries.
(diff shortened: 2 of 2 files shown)


node/narwhal/src/gateway.rs
```diff
@@ -218,7 +218,7 @@ impl<N: Network> Gateway<N> {
 
     /// The maximum number of duplicates for any particular request.
     fn max_cache_duplicates(&self) -> usize {
-        self.max_committee_size() * self.max_committee_size()
+        self.max_committee_size().pow(2)
```
joske, line 221: wouldn't the compiler produce the same code?

Edit: checked with compiler explorer, no it doesn't, in fact it generates a lot more code in the `pow(2)` case (error checking and a call to pow - which ultimately does `imul` again). And with `opt-level = 2` it generates the exact same machine code. 

So not sure why this change makes sense?

ljedrz: I don't think so, as it would have no way of knowing whether the second call results in the same output

howardwu: can we revert since its the same?

ljedrz: I'll double-check it, but I'm pretty sure this is not the same, and the current form causes double calculation

howardwu: I see the concern about possible having a double lookup hit, I think we can keep the `pow`

howardwu (approved):

node/narwhal/src/gateway.rs
ljedrz: I did a quick measurement of runtime performance of both variants under `--release`, and the current version takes ~500ns (with spikes of 1us) on average on my PC, while this change takes it down to ~250ns; both are very quick with a fresh node with an empty db, but this will become slower over time

PR #2762, opened 2023-10-16
Print account in dev mode and address only in product mode

ghostant-1017:
## Motivation

More convenient in dev mode


howardwu:
We now print out the private key when running in dev mode.

If you feel this feature is still warranted, feel free to reopen it and we can consider it.

Thanks @ghostant-1017 !

PR #2776, opened 2023-10-18
Message proptests

ljedrz:
This PR introduces rundtrip proptests for all variants of the `Message`.

### DIFF
--- a/node/router/messages/Cargo.toml
+++ b/node/router/messages/Cargo.toml
@@ -68,3 +68,21 @@ features = [ "codec" ]
 
 [dependencies.tracing]
 version = "0.1"
+
+[dev-dependencies.snarkos-node-sync-locators]
+path = "../../sync/locators"
+version = "=2.1.7"
+features = [ "test" ]
+
+[dev-dependencies.snarkvm]
+workspace = true
+features = [ "algorithms" ]
+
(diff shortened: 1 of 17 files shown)


node/router/messages/Cargo.toml
```diff
@@ -68,3 +68,21 @@ features = [ "codec" ]
 
 [dependencies.tracing]
 version = "0.1"
+
+[dev-dependencies.snarkos-node-sync-locators]
+path = "../../sync/locators"
```
howardwu, line 73: remove

node/router/messages/Cargo.toml
```diff
@@ -68,3 +68,21 @@ features = [ "codec" ]
...
+version = "=2.1.7"
+features = [ "test" ]
+
+[dev-dependencies.snarkvm]
+workspace = true
+features = [ "algorithms" ]
+
+[dev-dependencies.snarkvm-ledger-test-helpers]
```
howardwu: please use `snarkvm` with `test-helpers` flag

ljedrz: just did; however, this doesn't work with the currently pinned snarkVM (it does work with `snarkVM/staging`, though

PR #2778, opened 2023-10-19
Added test_partially_connected

vbar:
## Motivation

Validators should be able to handle missing connections to other validators (as long as they have
some connections left).

## Test Plan

Changes are only in test code.

## Related PRs

This is rebased #2769


howardwu:
@vbar what are these failing CIs?

vbar:
> @vbar what are these failing CIs?

Sorry, didn't notice - and I can't see them now (presumably because the PR is closed)... Should I try opening it again?

PR #2789, opened 2023-10-20
Trim the logs in the temporary logfile

ljedrz:
This was a lot more tricky than I anticipated, but now this PR not only performs log trimming, but should also make logging to the filesystem faster.

We might also want to consider using [RollingFileAppender](https://docs.rs/tracing-appender/0.2.2/tracing_appender/rolling/struct.RollingFileAppender.html), which can work in tandem with `NonBlocking` used in this PR.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -729,6 +729,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "crossbeam-channel"
+version = "0.5.8"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a33c2bf77f2df06183c3aa30d1e96c0695a313d4f9c453cc3762a6db39f99200"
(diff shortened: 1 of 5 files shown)


ljedrz:
The CI failures are unrelated.

joske (approved):

kaimast:
Is this PR still needed? I am not 100% sure what problems it solves.

ljedrz:
IIRC, this is related to the snarkOS log file stored at `/tmp` - I think it was requested by @howardwu. There may also have been a performance angle, judging by the contents.

CC @vicsn

howardwu:
Honestly, can scrap, unclear what it was needed for at this point.

kaimast:
Okay. I'll close it for now.

PR #2807, opened 2023-10-27
allowing repeated BatchPropose for the same batch

vbar:
## Motivation

The network shouldn't stall on malformed or missing `BatchSignature` event. Currently its recipient handles it correctly, by asking for the signature again, but the sender has an extra check dropping such repeated requests, in `process_batch_propose_from_peer`:

``` 
// If the round and batch ID matches, then skip signing the batch a second time.
            if signed_round == batch_header.round() && signed_batch_id == batch_header.batch_id() {
                debug!("Skipping a proposal for round {signed_round} from '{peer_ip}' {}", "(already signed)".dimmed());
                return Ok(());
            }
```

This is wrong and should be removed, so that `process_batch_propose_from_peer` can proceed to send the `BatchSignature` again.

## Test Plan

This was causing problems in a test setup on 2 machines, running 11 of 16 validators (i.e. just the quorum) which occasionally produces malformed `BatchSignature` events (still not clear why). The check above reliably stops progress after the first "Signature verification failed" warning. Without the check,
the same setup runs for much longer, logging repeated  "Signature verification failed" warnings but not stopping
on them.

## Related PRs

This reverts a part of  #2792.

### DIFF
--- a/node/bft/src/primary.rs
+++ b/node/bft/src/primary.rs
@@ -504,11 +504,6 @@ impl<N: Network> Primary<N> {
                 self.gateway.disconnect(peer_ip);
                 bail!("Malicious peer - proposed another batch for the same round ({signed_round})");
             }
-            // If the round and batch ID matches, then skip signing the batch a second time.
-            if signed_round == batch_header.round() && signed_batch_id == batch_header.batch_id() {
-                debug!("Skipping a proposal for round {signed_round} from '{peer_ip}' {}", "(already signed)".dimmed());
-                return Ok(());
-            }
         }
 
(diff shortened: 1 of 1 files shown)


howardwu:
I took a similar but different approach to this problem. Instead of re-signing a new signature, I cached the signature and resent the existing signature.

You can find it in this commit here:
- [Add ability to resend cached signature, prepare to tokio spawn this handler](https://github.com/AleoHQ/snarkOS/commit/3ea8076d3d2e32ef642a1202f9e01b71f6924b50)

PR #2812, opened 2023-10-30
fix: cli should return non-zero exit code on error

joske:
When a CLI command fails, it should return non-zero exit code. 

This facilitates scripting of the commands.

### DIFF
--- a/snarkos/main.rs
+++ b/snarkos/main.rs
@@ -12,11 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::process::exit;
-
 use snarkos_cli::{commands::CLI, helpers::Updater};
 
 use clap::Parser;
+use std::process::exit;
 #[cfg(feature = "jemalloc")]
 use tikv_jemallocator::Jemalloc;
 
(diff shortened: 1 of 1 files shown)


ljedrz (approved):

niklaslong (approved):

snarkos/main.rs
```diff
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::process::exit;
+
 use snarkos_cli::{commands::CLI, helpers::Updater};
 
 use clap::Parser;
```
howardwu:
```suggestion
use snarkos_cli::{commands::CLI, helpers::Updater};

use clap::Parser;
use std::process::exit;
```

snarkos/main.rs
```diff
@@ -15,6 +15,7 @@
...
+use std::process::exit;
 #[cfg(feature = "jemalloc")]
 use tikv_jemallocator::Jemalloc;
```
howardwu:
```suggestion
use std::process::exit;

#[cfg(feature = "jemalloc")]
use tikv_jemallocator::Jemalloc;
```

PR #2813, opened 2023-10-30
Non-reentrant propose_batch

vbar:
## Motivation

The `propose_batch` function is called 1) from a dedicated task started in `start_handlers`, and 2) from `try_increment_to_the_next_round` - so potentially from different threads. If those threads start it in parallel,
it might send different batches for the same round, breaking the protocol. This patch adds a lock enforcing
that doesn't happen.

## Test Plan

Without this patch, the recipient produces a warning on the duplicate batches:
```
2023-10-30T11:33:02.170998Z  WARN snarkos_node_bft::primary: Cannot sign a batch from '10.0.0.36:5002' - Malicious peer - proposed another batch for the same round (5)
```

With the fix, the warning should never appear.

### DIFF
--- a/node/bft/src/primary.rs
+++ b/node/bft/src/primary.rs
@@ -66,7 +66,10 @@ use std::{
     sync::Arc,
     time::Duration,
 };
-use tokio::{sync::OnceCell, task::JoinHandle};
+use tokio::{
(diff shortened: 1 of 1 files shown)


howardwu:
I don't know why Github keeps reclosing these...

PR #2815, opened 2023-11-01
Deprecate `snarkOS developer` CLI commands

raychu86:
## Motivation

This PR deprecates the `snarkOS developer` CLI commands. 

For `deployment` and `execution` tooling, refer to the Aleo SDK (https://github.com/AleoHQ/sdk), which is also live at https://aleo.tools/.


howardwu (approved):
Attention: this PR was reverted. These commands will be deprecated in the future, after a migration path is announced.

PR #2816, opened 2023-11-01
Serialize invocations of propose_batch

vbar:
## Motivation

The propose_batch function is called 1) from a dedicated task started in start_handlers, and 2) from try_increment_to_the_next_round - so potentially from different threads. If those threads start it in parallel,
it might send different batches for the same round, breaking the protocol. This patch adds a lock enforcing
that doesn't happen.

## Test Plan

Without this patch, the recipient produces a warning on the duplicate batches:

2023-10-30T11:33:02.170998Z  WARN snarkos_node_bft::primary: Cannot sign a batch from '10.0.0.36:5002' - Malicious peer - proposed another batch for the same round (5)

With the fix, the warning should never appear.

## Related PRs

Unlike #2813 , this patch waits on a lock, rather than using try_lock - the difference shouldn't matter to the dedicated task, but other threads call try_increment_to_the_next_round while handling an incoming network event, and propose_batch erroring out leads to that event being dropped, which is probably less efficient that waiting...

### DIFF
--- a/node/bft/src/primary.rs
+++ b/node/bft/src/primary.rs
@@ -431,7 +431,7 @@ impl<N: Network> Primary<N> {
             return Ok(());
         }
         // Ditto if the batch had already been proposed.
-        assert!(round > 0);
+        ensure!(round > 0, "Round 0 cannot have transaction batches");
         if *lock_guard == round {
             warn!("Primary is safely skipping a batch proposal - round {round} already proposed");
             return Ok(());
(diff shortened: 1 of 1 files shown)


node/bft/src/primary.rs
```diff
@@ -421,6 +430,14 @@ impl<N: Network> Primary<N> {
             debug!("Primary is safely skipping a batch proposal {}", "(no unconfirmed transactions)".dimmed());
             return Ok(());
         }
+        // Ditto if the batch had already been proposed.
+        assert!(round > 0);
```
raychu86:
```suggestion
```

howardwu (approved):

PR #2822, opened 2023-11-03
Add --dev-tx-interval-ms CLI option

vvp:
This PR adds a new CLI option for controlling transaction generation in development mode. For example `--dev-tx-interval-ms 1000` makes a dev node execute and broadcast dummy transactions on a one-second interval. This setting works for other nodes besides `--dev 0` too and provides a simple way to increase load in a devnet for testing purposes.

If the `--dev-tx-interval-ms` is not set, then node behaves similarly as before - only node `--dev 0` generates transactions every 500 msec. `--dev-tx-interval-ms 0` will disable the transaction generation (also with `--dev 0`).


joske (approved):

vvp:
Closing this as obsolete after #3132 was already merged.

PR #2823, opened 2023-11-07
[TOB] Correct an off-by-one message count limit

ljedrz:
In addition, introduce 2 related `const`s.

Finding: TOB-ALEO-10

### DIFF
--- a/node/router/src/inbound.rs
+++ b/node/router/src/inbound.rs
@@ -57,8 +57,8 @@ pub trait Inbound<N: Network>: Reading + Outbound<N> {
             None => bail!("Unable to resolve the (ambiguous) peer address '{peer_addr}'"),
         };
 
-        // Drop the peer, if they have sent more than MESSAGE_LIMIT messages
-        // in the last MESSAGE_LIMIT_TIME_FRAME_IN_SECS seconds.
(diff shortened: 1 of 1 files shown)


joske (approved):

node/router/src/inbound.rs
```diff
@@ -53,9 +57,10 @@ pub trait Inbound<N: Network>: Reading + Outbound<N> {
             None => bail!("Unable to resolve the (ambiguous) peer address '{peer_addr}'"),
         };
 
-        // Drop the peer, if they have sent more than 500 messages in the last 5 seconds.
-        let num_messages = self.router().cache.insert_inbound_message(peer_ip, 5);
-        if num_messages >= 500 {
+        // Drop the peer, if they have sent more than MESSAGE_LIMIT messages
+        // in the last MESSAGE_LIMIT_TIME_FRAME_IN_SECS seconds.
```
howardwu:
```suggestion
        // Drop the peer, if they have sent more than `MESSAGE_LIMIT` messages
        // in the last `MESSAGE_LIMIT_TIME_FRAME_IN_SECS` seconds.
```

howardwu (approved):

PR #2825, opened 2023-11-07
[TOB] allow the restriction of CORS origins

ljedrz:
Cc @zosorock, as this is probably most applicable to services run atop the REST server.

Finding: TOB-ALEO-23

### DIFF
--- a/node/rest/src/lib.rs
+++ b/node/rest/src/lib.rs
@@ -36,7 +36,7 @@ use snarkvm::{
 use anyhow::Result;
 use axum::{
     extract::{ConnectInfo, DefaultBodyLimit, Path, Query, State},
-    http::{header::CONTENT_TYPE, Method, Request, StatusCode},
+    http::{header::CONTENT_TYPE, HeaderValue, Method, Request, StatusCode},
     middleware,
     middleware::Next,
     response::Response,
@@ -48,7 +48,7 @@ use parking_lot::Mutex;
 use std::{net::SocketAddr, sync::Arc};
 use tokio::task::JoinHandle;
 use tower_http::{
-    cors::{Any, CorsLayer},
+    cors::{AllowOrigin, CorsLayer},
     trace::TraceLayer,
 };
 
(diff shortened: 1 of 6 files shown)


node/rest/src/lib.rs
```diff
@@ -95,9 +96,15 @@ impl<N: Network, C: ConsensusStorage<N>, R: Routing<N>> Rest<N, C, R> {
...
 
 impl<N: Network, C: ConsensusStorage<N>, R: Routing<N>> Rest<N, C, R> {
-    fn spawn_server(&mut self, rest_ip: SocketAddr) {
+    fn spawn_server(&mut self, rest_ip: SocketAddr, allowed_origins: Vec<String>) {
+        let allowed_origins = if allowed_origins.is_empty() {
+            AllowOrigin::any()
+        } else {
+            let origins = allowed_origins.iter().map(|s| HeaderValue::from_str(s).expect("Invalid CORS origin"));
+            AllowOrigin::list(origins)
+        };
```
SplittyDev, line 105: IIRC this doesn't parse `*` correctly. From one of my own projects I remember having to check for `*` explicitly and return `AllowOrigin::any()` in that case. This is totally anecdotal though.

SplittyDev: Snippet from my code where I parse the origins from a comma-separated string and handle the wildcard case explicitly. Sadly I didn't put a comment there explaining the necessity of handling this case specifically, so I can't be 100% sure, but if I remember correctly the origins didn't work correctly when parsing `*` as a `HeaderValue`.

CleanShot 2023-11-07 at 13 35 00@2x

ljedrz: in this case it is expected to leave the vector empty in order to accept `*`, the filtering is opt-in

vvp:
Alternatively, instead of this PR we could remove the direct CORS support from REST API and let node operators run a reverse proxy for snarkOS in which the CORS is configured (e.g. nginx), if they need it. 

This option would simplify our codebase and allow more flexible CORS configurations, but it would also add that extra step and operational overhead when integrating snarkOS into websites via REST (unsure how common use case that'll be).

howardwu:
This is already guarded using firewalls today.

vicsn:
> This is already guarded using firewalls today.

Closing because the above policy is recommended to node operators in snarkOS documentation

PR #2827, opened 2023-11-07
[TOB] add WARNs for saturating cache operations

ljedrz:
IMO adding warnings should suffice here, as we're not expecting either to happen, but would like to know if it ever does.

Finding: TOB-ALEO-27

### DIFF
--- a/node/bft/src/helpers/cache.rs
+++ b/node/bft/src/helpers/cache.rs
@@ -166,6 +166,9 @@ impl<N: Network> Cache<N> {
         let mut map_write = map.write();
         // Load the entry for the key, and increment the counter.
         let entry = map_write.entry(key).or_default();
+        if *entry == u16::MAX {
+            warn!("A cache counter was maxed out");
+        }
         *entry = entry.saturating_add(1);
         // Return the updated counter.
         *entry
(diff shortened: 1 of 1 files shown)


node/bft/src/helpers/cache.rs
```diff
@@ -166,6 +166,9 @@ impl<N: Network> Cache<N> {
         let mut map_write = map.write();
         // Load the entry for the key, and increment the counter.
         let entry = map_write.entry(key).or_default();
+        if *entry == u16::MAX {
+            warn!("A cache counter was maxed out");
```
joske, line 170: Is this a candidate for a metric?

node/bft/src/helpers/cache.rs
```diff
@@ -166,6 +166,9 @@ impl<N: Network> Cache<N> {
...
+        }
         *entry = entry.saturating_add(1);
         // Return the updated counter.
         *entry
```
joske, line 174: nobody seems to do anything with this number?

node/bft/src/helpers/cache.rs
ljedrz: hmm, probably not, as it's per-peer and keeping track of all could quickly become too verbose

node/bft/src/helpers/cache.rs
ljedrz: indeed (at least when returned from this method), though perhaps it's a plan for the future? in any case, IMO it's not an issue

node/bft/src/helpers/cache.rs
ljedrz: although, if summed up for all the peers could be a useful metric

node/bft/src/helpers/cache.rs
joske: sure just for my understanding

PR #2831, opened 2023-11-09
Enable `snarkos developer scan` for development networks

raychu86:
## Motivation

This PR updates the `snarkos developer scan` CLI to work for development networks by disabling the CDN in those cases. We determine if we need to use the CDN by checking if the genesis block of the network we are querying via `--endpoint` is equivalent to the live network's genesis block.


vicsn (approved):
LGTM!

howardwu (approved):

PR #2847, opened 2023-11-14
Add validation for incoming `TransmissionResponse` messages

raychu86:
## Motivation

This PR adds validation for incoming `TransmissionResponse` messages. Previously without this check, malicious nodes could send a transmission completely unrelated to the `TransmissionID` being requested, and the node would blindly accept it. This discrepancy leads to incorrect state and impacts block production.


ljedrz (approved):
LGTM

PR #2852, opened 2023-11-16
Add `InputID` filtering to transmission selection

raychu86:
## Motivation

This PR adds additional filtering to the transmissions we decide to commit via `BFT::commit_leader_certificate`. We now check against the transaction `InputID`s to prevent block production failure when multiple transactions attempt to spend the same record.

We've also added snarkVM level safety checks - https://github.com/AleoHQ/snarkVM/pull/2229

### DIFF
--- a/node/bft/ledger-service/src/ledger.rs
+++ b/node/bft/ledger-service/src/ledger.rs
@@ -149,6 +149,11 @@ impl<N: Network, C: ConsensusStorage<N>> LedgerService<N> for CoreLedgerService<
         self.get_committee_for_round(previous_round)
     }
 
+    /// Returns `true` if the ledger contains the given input ID.
+    fn contains_input_id(&self, input_id: &Field<N>) -> Result<bool> {
+        self.ledger.contains_input_id(input_id)
+    }
+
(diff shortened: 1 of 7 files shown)


ghostant-1017:
I wonder if the safety check is enough here, since we check the transaction until `try_advance_to_next_block`, the transacation itself may be invalid not only the input_id.

ghostant-1017:
I mean maybe two nearby transmissions may contain the same input ids, and it will pass the check here.
If we were to do these checks at the snarkOS level, it would become complex because there are so many scenarios to consider. Whereas if we just hand over all the transmissons to snarkVM for processing, I think it will become simpler because each Transaction is a change of state and we just consider whether the change of state is legitimate or not.

As AleoBFT just focus on the logic to make the transmissions in order.
The snarkVM should input the sequenced transmissions, and according to the global state in vm, we execute the transmission one by one, if meet some invalid transaction, just skip.

raychu86:
Closed in favor of lower level snarkVM check - https://github.com/AleoHQ/snarkVM/pull/2229

PR #2872, opened 2023-11-27
Custom ledger location

### LINKED ISSUE #2870: [Feature] Customize the ledger data directory
## 🚀 Feature

Customize the directory where the ledger data is stored to a desired location, including using mounted disks.

## Motivation

In almost blockchains, ledger data continues to increase. if the ledger data stored in the root partition in the node, it can be caused ran out of the root partition. so, It is recommended that ledger data be stored in additional mounted disks separate from the root partition.
but, currently, snarkos stores its ledger data only in the default path. `~/.aleo/`. 
It would be helpful for us, if we can choose specific ledger data path.
## Implementation

Should be able to set the `aleo_ledger_dir()` through user input argument from CLI.

https://github.com/AleoHQ/aleo-std/blob/main/storage/src/lib.rs#L45

**Are you willing to open a pull request?** (See [CONTRIBUTING](../../CONTRIBUTING.md))

ljedrz:
This is the third of 3 PRs needed to close https://github.com/AleoHQ/snarkOS/issues/2870. It requires both:
- an update of `aleo-std` (after https://github.com/AleoHQ/aleo-std/pull/10 is merged)
- an update of `snarkVM` (after https://github.com/AleoHQ/snarkVM/pull/2210 is merged)

in order to work.

I've tested it with `start` and `clean`, both of which worked fine locally,

### DIFF
--- a/cli/src/commands/clean.rs
+++ b/cli/src/commands/clean.rs
@@ -15,6 +15,7 @@
 use anyhow::{bail, Result};
 use clap::Parser;
 use colored::Colorize;
+use std::path::PathBuf;
 
 /// Cleans the snarkOS node storage.
 #[derive(Debug, Parser)]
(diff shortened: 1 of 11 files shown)


ljedrz:
Superseded by https://github.com/AleoHQ/snarkOS/pull/3044.

PR #2833, opened 2023-11-10
[TOB-11] Adds `is_valid_peer_ip` filtering

howardwu:
## Motivation

[TOB-11] Adds `is_valid_peer_ip` filtering

Supersedes #2673

PR #2878, opened 2023-11-29
Accept peers with local/private network IPs when in dev mode

vvp:
This PR relaxes the peer IP validity requirements when running in dev mode. The stricter requirements were introduced in #2833 but they caused problems when running `snarkOS` in devnets locally or remotely in private IP range.


node/bft/src/gateway.rs
```diff
@@ -713,10 +720,18 @@ impl<N: Network> Gateway<N> {
...
-                            if !self_.is_valid_peer_ip(validator_ip) {
-                                continue;
+                            if self_.dev.is_some() {
+                                // Ensure the validator IP is not this node.
+                                if self_.is_local_ip(validator_ip) {
+                                    continue;
+                                }
+                            } else {
+                                // Ensure the validator IP is not this node and is well-formed.
+                                if !self_.is_valid_peer_ip(validator_ip) {
```
ljedrz, line 742: `clippy` didn't want to collapse this one into `else if` :detective:?

node/router/src/inbound.rs
```diff
@@ -256,7 +257,12 @@ pub trait Inbound<N: Network>: Reading + Outbound<N> {
         // Retrieve the connected peers.
         let peers = self.router().connected_peers();
         // Filter out invalid addresses.
-        let peers = peers.into_iter().filter(|ip| self.router().is_valid_peer_ip(ip)).collect();
+        let peers = if self.router().is_dev() {
+            // In dev mode, relax the validity requirements to make running devnets more flexible.
+            peers.into_iter().filter(|ip| !is_bogon_ip(ip.ip())).collect()
```
ljedrz: do we want to filter out _all_ bogon addresses here? I could see private network addresses being practical in dev mode

node/router/src/inbound.rs
```diff
@@ -265,7 +271,13 @@ pub trait Inbound<N: Network>: Reading + Outbound<N> {
     /// Handles a `PeerResponse` message.
     fn peer_response(&self, _peer_ip: SocketAddr, peers: &[SocketAddr]) -> bool {
         // Filter out invalid addresses.
-        let peers = peers.iter().copied().filter(|ip| self.router().is_valid_peer_ip(ip)).collect::<Vec<_>>();
+        let peers = if self.router().is_dev() {
+            // In dev mode, relax the validity requirements to make running devnets more flexible.
+            peers.iter().copied().filter(|ip| !is_bogon_ip(ip.ip())).collect::<Vec<_>>()
```
ljedrz: ditto

node/router/src/inbound.rs
vvp: Good question, for this PR I basically cherry-picked what we had in place before https://github.com/AleoHQ/snarkOS/pull/2833 (and tested that it still works).

howardwu (approved):

PR #2897, opened 2023-12-05
Round quorum before advancement

### LINKED ISSUE #2876: [Bug] Leader sending certificate with round far in the future halts the network
## 🐛 Bug Report

If a malicious leader sends a certificate with a round set far into the future, the receiving nodes will blindly pass that through `try_increment_to_the_next_round(...)`, repeatedly calling `self.storage.increment_to_next_round(...)` as the attacker desires.

## Steps to Reproduce

We should make sure to have all related malicious leader tests in one branch. For now you can check out [this branch](https://github.com/AleoHQ/snarkOS/tree/vicsn/fix/stress_test_5) and run `devnet.sh` with a network of size 4.

```
2023-11-28T14:58:01.379402Z ERROR is_behind_schedule: false, is_peer_far_in_future: true
2023-11-28T14:58:01.379419Z DEBUG Primary is not ready to propose the next round                          
2023-11-28T14:58:01.379439Z  WARN Cannot store a certificate from '127.0.0.1:5001' - Missing certificates for the previous round 57 
in storage (gc = 0)
...
2023-11-28T15:10:36.799805Z  WARN Cannot sign a batch from '127.0.0.1:5001' - Unable to fetch batch certificate - (timeout) deadline
 has elapsed                                                                                                                        
2023-11-28T15:10:36.800164Z DEBUG BFT (timer expired) - Checking for quorum threshold (without the leader)                          
```

## Expected Behavior
The malicious node's certificates should be ignored.

## Your Environment

cargo 1.71.1 (7f1d04c00 2023-07-29)
macOS

vicsn:
## Motivation

Closes https://github.com/AleoHQ/snarkOS/issues/2876

## Open questions
- we can consider merging `AddressWithCoordinate<N>` upstream into snarkVM

## Performance
In the tradition of Lukasz brilliance, I'm using [Vecs+BinarySearch instead of maps](https://github.com/AleoHQ/snarkVM/pull/2193).

In [this repo](https://github.com/vicsn/round_cache_test) you can find some benchmarks of different constructions. If we don't check quorum, we can do 250 updates per millisecond, if we do check quorum each time because we get increased round numbers, we can do 9 updates per millisecond. I expect on average an update will cost closer to 1/250 milliseconds. Moreover, an update will only be done if any peer actually sends a very high round number.

Also note that currently computing quorum requires us to collect into a HashSet, we could add a version which just uses an iterator (checking along the way if all elements are unique).

## Persistence
I'm not storing the cache in the database, we should expect to receive a quorum's worth of updates from other nodes every round.


node/bft/src/primary.rs
```diff
@@ -1297,15 +1301,21 @@ impl<N: Network> Primary<N> {
             bail!("Round {batch_round} is too far in the past")
         }
 
+        // TODO: should we get the current or previous committee?
```
niklaslong: I think the current committee is likely best? After all this represents the latest valid state update the node is aware of?

node/bft/src/helpers/round_cache.rs
```diff
@@ -0,0 +1,276 @@
...
+    fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+#[derive(Debug)]
+pub struct RoundCache<N: Network> {
+    /// The current highest round which has (stake-weighted) quorum
+    last_highest_round_with_quorum: u64,
+    /// A list of (round, Vec<Address>), indicating the last seen highest round for each address
```
niklaslong:
```suggestion
    /// A list of (round, Vec<AddressCoordinates>), indicating the last seen highest round for each pair of address coordinates
```

node/bft/src/helpers/round_cache.rs
```diff
@@ -0,0 +1,276 @@
...
+        Ok(())
+    }
+
+    /// Update based on a new (round, address) pair seen in the wild. This does two things:
+    /// - If the round is higher than a previous one from this address, set it in highest_rounds
+    /// - Keep incrementing `last_highest_round_with_quorum` as long as it passes a stake-weighted quorum
+    /// We ignore the case where tomorrow's stake-weighted quorum round is *lower* than the current one
+    pub fn update(&mut self, round: u64, validator_address: Address<N>, committee: &Committee<N>) -> Result<u64> {
+        ensure!(committee.members().contains_key(&validator_address), "Address is not a member of the committee");
+        let validator = AddressCoordinates::from_address(validator_address);
```
niklaslong: Done for perf reasons, I assume? Or is there another motivation?

niklaslong:
Nice benches, I'm almost surprised `BTreeMap` wasn't the fastest with multiple inserts.

node/bft/src/helpers/round_cache.rs
vicsn: Indeed, inferring the coordinates costs comparatively a lot of time, so I'm just doing it once ahead of time.

node/bft/src/helpers/round_cache.rs
vicsn: Changed and fixed in [9159ede](https://github.com/AleoHQ/snarkOS/pull/2897/commits/9159ede39ecf9141fb35e753ad0df3c94ff9aaf3)

node/bft/src/primary.rs
vicsn: Yeah agreed, let's see what the test suite thinks

node/bft/src/helpers/mod.rs
```diff
@@ -15,6 +15,10 @@
 pub mod cache;
 pub use cache::*;
 
+// TODO: rename so it is right next to cache
```
niklaslong: The `Cache` on `Gateway` is mostly metrics-oriented at the moment. I don't think it would be unreasonable to consolidate but eventually metrics and caching should be separate (with metrics being exposed with e.g. Grafana). Either way, the Gateway will need some refactoring in future, perhaps consolidating to reduce complexity/fragmentation in the meantime isn't a bad idea.

node/bft/tests/narwhal_e2e.rs
```diff
@@ -150,7 +150,7 @@ async fn test_storage_coherence() {
     const TARGET_ROUND: u64 = 12;
     // Note: cloning the network is fine because the primaries it wraps are `Arc`ed.
     let network_clone = network.clone();
-    deadline!(Duration::from_secs(40), move || { network_clone.is_round_reached(TARGET_ROUND) });
+    deadline!(Duration::from_secs(60), move || { network_clone.is_round_reached(TARGET_ROUND) });
```
niklaslong: I don't think this is strictly tied to business logic or timeouts at the moment, just based on repeated runs but this is a substantial increase?

niklaslong:
> Do we usually run all ignore = tests manually, if so how?

Most ignored e2e tests are long running (e.g. manual testing) or are too costly to perform on CI. We run them locally from time to time as a sanity check but ideally, our coverage should be complete with regular tests.

node/bft/src/helpers/round_cache.rs
```diff
@@ -0,0 +1,250 @@
...
+use anyhow::{ensure, Result};
+use snarkvm::{
+    console::types::{Address, Field},
+    ledger::committee::Committee,
+    prelude::Network,
+};
+use std::collections::{HashSet, VecDeque};
+
+#[derive(Copy, Clone, Debug)]
+struct AddressWithCoordinate<N: Network> {
```
niklaslong: Just riffing but:
```suggestion
struct AddressWithAbscissa <N: Network> {
```
or 
```suggestion
struct AddressWithXCoordinate<N: Network> {
```

niklaslong: Not sure these are better than your original suggestion overall.

node/bft/tests/narwhal_e2e.rs
vicsn: Indeed it is a scary big increase. Note however that we are measuring a very different thing now: instead of our target node hitting `TARGET_ROUND`, `quorum` nodes need to hit the round ánd share those certificates with target node.

node/bft/src/helpers/cache_round.rs
```diff
@@ -0,0 +1,250 @@
...
+        Ok(())
+    }
+
+    /// Find and prune validators which are no longer in the committee
+    fn prune_stale_validators(&mut self, committee: &Committee<N>) -> Result<()> {
+        let addresses_to_prune = self
+            .address_rounds
+            .iter()
+            .filter_map(|(a, _)| (!committee.members().contains_key(&a.address)).then_some(a.x))
+            .collect::<Vec<_>>();
```
ljedrz, line 95: do we need to collect here?

node/bft/src/helpers/cache_round.rs
```diff
@@ -0,0 +1,250 @@
...
+            address_rounds: Default::default(),
+        }
+    }
+
+    /// Insert a validator at a round
+    fn insert_validator_at_round(&mut self, round: u64, validator: AddressWithCoordinate<N>) {
+        match self.highest_rounds.binary_search_by_key(&round, |(r, _)| *r) {
+            Ok(new_address_index) => self.highest_rounds[new_address_index].1.push(validator),
+            Err(new_address_index) => {
+                let mut validators = Vec::with_capacity(200);
```
ljedrz: nit: it would be cool to be able to get the actual committee size

ljedrz: but ultimately I don't think we should preallocate here, not unless we're confident that we'll have all the members of the committee in this vec

node/bft/src/helpers/cache_round.rs
vicsn: Yes otherwise `self.address_rounds` is borrowed in `addresses_to_prune` and we can't subsequently remove from it.

node/bft/src/helpers/cache_round.rs
vicsn: Good point 8049d6ee1

node/bft/src/primary.rs
```diff
@@ -1304,12 +1308,25 @@ impl<N: Network> Primary<N> {
...
-        if is_behind_schedule || is_peer_far_in_future {
+
+        // If our primary is behind shedule, update our committee to the batch round.
+        if is_behind_schedule {
             // If the batch round is greater than the current committee round, update the committee.
             self.try_increment_to_the_next_round(batch_round).await?;
+        // If our peer is far ahead, check if a quorum of peers is ahead and consider updating our committee.
+        } else if is_peer_far_in_future {
+            // Get the highest round seen from a quorum of the current committee
+            let committee = self.ledger.get_committee_for_round(self.current_round())?;
```
raychu86: This should be the `get_previous_committee_for_round`.

In the `mainnet` branch this should be changed to `get_committee_lookback_for_round`.

node/bft/src/helpers/cache_round.rs
```diff
@@ -0,0 +1,248 @@
...
+            // Confirm we did not cache more validators than the current committee size
+            ensure!(self.address_rounds.len() <= committee.num_members());
+            // Confirm we did not cache more validators than the current committee size
+            ensure!(self.highest_rounds.iter().map(|(_, a)| a.len()).sum::<usize>() <= committee.num_members());
+        }
+
+        // Check if we reached quorum on a new round
+        if inserted {
+            while committee.is_quorum_threshold_reached(&self.validators_in_support(committee)?) {
+                self.last_highest_round_with_quorum += 1;
```
raychu86, line 154: Should this be `*self.last_highest_round_with_quorum = round`?

node/bft/src/helpers/cache_round.rs
```diff
@@ -0,0 +1,248 @@
...
+            }
+            // Confirm we did not cache more validators than the current committee size
+            ensure!(self.address_rounds.len() <= committee.num_members());
+            // Confirm we did not cache more validators than the current committee size
+            ensure!(self.highest_rounds.iter().map(|(_, a)| a.len()).sum::<usize>() <= committee.num_members());
+        }
+
+        // Check if we reached quorum on a new round
+        if inserted {
+            while committee.is_quorum_threshold_reached(&self.validators_in_support(committee)?) {
```
raychu86, line 153: The committee being used here may changed based on the round you are checking.

node/bft/src/helpers/cache_round.rs
```diff
@@ -0,0 +1,248 @@
...
+    pub fn new() -> Self {
+        Self {
+            last_highest_round_with_quorum: Default::default(),
+            highest_rounds: Default::default(),
+            address_rounds: Default::default(),
+        }
+    }
+
+    /// Insert a validator at a round
+    fn insert_validator_at_round(&mut self, round: u64, validator: AddressWithCoordinate<N>) {
```
raychu86: Why doesn't this add the validator to `address_rounds`?

node/bft/src/helpers/cache_round.rs
vicsn: I think its fine to use the single given committee (from `get_previous_committee_for_round`) to update completely, because who cares whether old outdated committees have high rounds..

Sidenote: if we were to only update by 1 round at a time based on the given committee, instead of always taking just `<quorum>` batch proposals to update to the correct round, we will then need `<quorum> + i` batch proposals to increase our round by `i`.  If `<quorum> > max_gc_rounds`, then we might never catch up...

node/bft/src/helpers/cache_round.rs
vicsn: The function is badly named, we are only updating because we saw a new round for this validator: ef446e9f2

node/bft/src/helpers/cache_round.rs
vicsn: Though you marked it as resolved, some extra context:

The while loop ensures we keep incrementing as long as we reached quorum. 

I do realize upon reading again that the increments might take a *significant* amount of compute if we have to do so a million times (e.g. because this case is only triggered after months or years, requiring us to move up a million rounds).

Fixed here: c8fff2947

node/bft/src/primary.rs
vicsn: e53bdb0e4

Let me know if you need me to create a new PR targeting mainnet.

node/bft/src/helpers/cache_round.rs
raychu86: Is there any case where you won't ever be able to see quorum because the fixed committee you are using doesn't include newly bonded validators?

vicsn: Great question, it seems so yes... Because we use our old outdated round to determine the committee: `self.ledger.get_previous_committee_for_round(self.current_round())`.

The only alternative I see now is to use the peer's advertised `batch_round` - but one question on that: is it possible that we don't know the committee yet for rounds far in the future so that `get_previous_committee_for_round(batch_round)?` will keep failing? Or do we expect that to succeed at some point?

node/bft/src/helpers/cache_round.rs
```diff
@@ -0,0 +1,267 @@
...
+    fn from(address: Address<N>) -> Self {
+        Self { address, x: address.to_group().to_x_coordinate() }
+    }
+}
+
+#[derive(Debug)]
+pub struct RoundCache<N: Network> {
+    /// The current highest round which has (stake-weighted) quorum
+    last_highest_round_with_quorum: u64,
+    /// A list of (round, Vec<AddressWithCoordinate<N>>), indicating the last seen highest round for each address
```
ljedrz:
```suggestion
    /// A sorted list of (round, Vec<AddressWithCoordinate<N>>), indicating the last seen highest round for each address
```

node/bft/src/helpers/cache_round.rs
```diff
@@ -0,0 +1,267 @@
...
+    /// - If the round is higher than a previous one from this address, set it in `highest_rounds`
+    /// - Keep incrementing `last_highest_round_with_quorum` as long as it passes a stake-weighted quorum
+    /// We ignore the case where tomorrow's stake-weighted quorum round is *lower* than the current one
+    pub fn update(&mut self, round: u64, validator_address: Address<N>, committee: &Committee<N>) -> Result<u64> {
+        ensure!(committee.members().contains_key(&validator_address), "Address is not a member of the committee");
+        let validator = AddressWithCoordinate::from(validator_address);
+
+        // Determine if validator was inserted into the cache.
+        let mut inserted = false;
+        // Only consider updating the cache. if we see a high round.
```
ljedrz:
```suggestion
        // Only consider updating the cache if we see a high round.
```

node/bft/src/primary.rs
```diff
@@ -96,6 +97,8 @@ pub struct Primary<N: Network> {
     handles: Arc<Mutex<Vec<JoinHandle<()>>>>,
     /// The lock for propose_batch.
     propose_lock: Arc<TMutex<u64>>,
+    /// A cache for which batch rounds have been seen
```
```suggestion
    /// A cache for which batch rounds have been seen.
```

ljedrz (approved):
I can't test it right now, but LGTM code-wise :+1:

howardwu:
@vicsn in light of major code drift from `testnet3`, can you rebase this PR onto `mainnet-staging` and test this PR on a network to confirm its validity/necessity?

vicsn:
Yes, will coordinate and help Michel to first get this related issue over the finish line to see if it resolves the attack: https://github.com/AleoHQ/snarkOS/pull/3119

node/bft/src/helpers/cache_round.rs
```diff
@@ -0,0 +1,267 @@
...
+        Self { address, x: address.to_group().to_x_coordinate() }
+    }
+}
+
+#[derive(Debug)]
+pub struct RoundCache<N: Network> {
+    /// The current highest round which has (stake-weighted) quorum
+    last_highest_round_with_quorum: u64,
+    /// A sorted list of (round, Vec<AddressWithCoordinate<N>>), indicating the last seen highest round for each address
+    highest_rounds: Vec<(u64, Vec<AddressWithCoordinate<N>>)>,
```
joske, line 40: Why does this `Vec` have to be sorted? It seems only to quickly fetch an item? If that is true, why not use `HashMap`? This gives you O(1) lookups and O(1) inserts. Now you have O(n) inserts and O(log n) lookups.

vicsn: Will come back to this if we decide to keep this PR

ljedrz: it's faster and more lightweight, unless the collection can become arbitrarily large

vicsn:
Deprecating this in favour of https://github.com/AleoHQ/snarkOS/pull/3119 - which has been tested and works

PR #2901, opened 2023-12-07
Add LRU cache to mitigate flood traffic

iamalwaysuncomfortable:
## Motivation

Large volumes of unconfirmed_transactions introduce a state where the network significantly slows due to states on individual nodes becoming very far out of sync with each other. 

The exact reason why isn't %100 known, but evidence from logs indicate it might be because too many async tasks prevent the `Tokio` executor from allowing tasks to proceed such as batch signing that are key to advancing Consensus rounds - and once this happens, the network takes some time to Establish a normal cadence of block production.

An LRU cache is introduced to limit the number of transactions that can be in the `transaction_queue` object.

## Test Plan & Performance

* This fix is currently being tested on a 40-node network where transaction volume will reach roughly 2000tx/minute to validate that the nodes simply processing a lower volume of transaction via usage of the LRU cache works. 

* Should this experiment prove to work, a cache in the primary for transactions might considered wherein `unconfirmed_transactions` are pulled at a cadence which doesn't slow other critical Consensus processes.

### DIFF
--- a/node/consensus/src/lib.rs
+++ b/node/consensus/src/lib.rs
@@ -63,7 +63,7 @@ pub struct Consensus<N: Network> {
     /// The primary sender.
     primary_sender: Arc<OnceCell<PrimarySender<N>>>,
     /// The unconfirmed solutions queue.
-    solutions_queue: Arc<Mutex<IndexMap<PuzzleCommitment<N>, ProverSolution<N>>>>,
+    solutions_queue: Arc<Mutex<LruCache<PuzzleCommitment<N>, ProverSolution<N>>>>,
     /// The unconfirmed transactions queue.
     transactions_queue: Arc<Mutex<LruCache<N::TransactionID, Transaction<N>>>>,
     /// The recently-seen unconfirmed solutions.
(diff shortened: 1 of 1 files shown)


ljedrz (approved):
LGTM; as long as we can allow the least recently used entries to be dropped in case of flooding, it makes a lot of sense to restrict the number of items we store before processing them.

vicsn:
[This new PR](https://github.com/AleoHQ/snarkOS/pull/2977) changes the transactions_queue to a sorted map so we can order by priority_fee. Moreover, limiting just based on # of transactions makes us still at risk of running out of memory when someone spams large numbers of large deployments. So in that PR I added a byte size limit for the queue. If you have any thoughts for refinement would be great [to discuss in the thread there](https://github.com/AleoHQ/snarkOS/pull/2977#discussion_r1450344333).

> Should this experiment prove to work, a cache in the primary for transactions might considered wherein unconfirmed_transactions are pulled at a cadence which doesn't slow other critical Consensus processes.

Want to make a new issue for this if its still relevant in the new design?

howardwu (approved):

PR #2902, opened 2023-12-07
Reject unsolicited `PeerResponse` messages

### LINKED ISSUE #2894: [Bug] A malicious peer can directly send PeerResponse to other peers with high frequency and flood network with fake peer info
# https://hackerone.com/reports/2272999

## Summary:
The  router does not check if the PeerResponse is a direct response of a previous PeerRequest  https://github.com/AleoHQ/snarkOS/blob/testnet3/node/router/src/inbound.rs#L120

A malicious peer can directly send `PeerResponse` to other peers with high frequency. In this way, the other peers' `candidate_peers` will be full of malicious peer. Gradually, the good peer's connected peers are all malicious. Finally the good peers are disjoint with each oher and whole network is controlled by the malicious peers.

## Steps To Reproduce:
1. Clone https://github.com/AleoHQ/snarkOS and checkout `testnet3`
2. Add the following code at: node/router/src/heartbeat.rs `heartbeat` function

```
    /// Handles the heartbeat request.
    fn heartbeat(&self) {
        self.safety_check_minimum_number_of_peers();
        self.log_connected_peers();

        //////////////////// ADD CODE HERE ////////////////////
        // We can directly send PeerResponse to other peers (without previous PeerRequest). We can send PeerResponse with high frequency. Attacker can leverage this to attck the whole network.
        for peer_ip in self.router().connected_peers().into_iter() {
            let ip = "1.1.1.1:111";
            let malicious_addr: SocketAddr = ip
                .parse()
                .expect("Unable to parse socket address");
            self.send(peer_ip, Message::PeerResponse( PeerResponse { peers: vec![malicious_addr; 100] }));
        }
        /////////////////// ADD CODE HERE ////////////////////

        // Remove any stale connected peers.
        self.remove_stale_connected_peers();
        // Remove the oldest connected peer.
       ....
```
3. add log at node/router/src/inbound.rs `peer_response` function

```
    fn peer_response(&self, _peer_ip: SocketAddr, peers: &[SocketAddr]) -> bool {
        // Filter out invalid addresses.
        //////////////////// ADD LOG HERE ////////////////////
        warn!("received peer_response {:?}", peers);
        //////////////////// ADD LOG HERE ////////////////////
        let peers = peers.iter().copied().filter(|ip| self.router().is_valid_peer_ip(ip)).collect::<Vec<_>>();
        // Adds the given peer IPs to the list of candidate peers.
        self.router().insert_candidate_peers(&peers);
        true
    }
```

4. run ./devnet.sh
5. check log, we will find something like
```
WARN received peer_response [1.1.1.1:111, 1.1.1.1:111, 1.1.1.1:111, 1.1.1.1:111
```

6. Attack success: Though the good peer haven't made `PeerRequest`, it received malicious `PeerResponse`.

## Proof-of-Concept (PoC)

1. The malicious peer starts sending a high frequency of PeerResponse messages directly to other peers. The PeerResponse messages contain information about the other peers which are all controlled by the hacker.

2. The other peers receive these PeerResponse messages and update their candidate_peers list with the information provided by the malicious peer.

3. Due to the high frequency of messages and the continuous updates, the candidate_peers list of the other peers becomes populated mainly or entirely with malicious peer entries.

4. As the good peers' candidate_peers list becomes filled with malicious peers, the chances of connecting to other good peers decrease significantly.

5. Over time, the good peers' connections are predominated by malicious peers, reducing the opportunities for good peers to communicate with each other.

6. Eventually, the good peers become disjointed from each other, as their connections are primarily with malicious peers.

7. With the network effectively controlled by malicious peers, they can manipulate communication, block transactions, tamper with data, or perform other malicious activities without detection or intervention from the good peers.

screenshot-20231205-014632

## Impact

This bug have signigicant impact on all kinds of node: Prover, Validator and Client. The bug allows a malicious peer to flood the network with fake peer information, causing good peers to connect primarily with malicious peers. This gives the malicious peer control over the network and disrupts communication among the good peers. The malicious peers can tamper with transactions, manipulate data, and potentially launch further attacks. This undermines trust in the network and compromises its security and reliability.

## Fix Suggestions:

In the short term, verify every `PeerResponse ` is a real response of `PeerRequest`. In the middle or long term, consider formulating specification about the network layer. Also, consider using battle tested framework like `devp2p` and `libp2p`.

niklaslong:
This PR naively caches sent peer requests until a response is received, the goal being to detect when an unsolicited `PeerResponse` message is sent in order to influence a node's peer lists. 

TBD:

- [x] ~~protocol violating nodes are currently disconnected and added to the candidate peers. Might be good to restrict them instead?~~ (follow up)
- [x] ~~is the naive cache sufficient or do we need to track multiple requests to the same peer at a time?~~ Consolidated with existing cache.

I'd recommend reviewing commit by commit as the first commit is a file rename.

Fixes #2894.

Drafted as it has only been tested on a local network.


niklaslong:
Rebased.

howardwu (approved):

PR #2903, opened 2023-12-08
fix: block sync issue

joske:
If your node is syncing many blocks, it can stop syncing. The cause is the following:
* node sends out block request to 3 nodes
* only 2 respond, now sync can't continue as the response is not 'complete'
* sync proceeds to send all block requests it can so no more are sent out (50 outstanding)
* concurrently another thread removes the peer that was not responding using `remove_block_requests_to_peer`
* now this request IS considered complete, so it's not eligible for timeout handling
* incoming responses are **only** considered if there are BlockResponses coming in, but none will ever come in anymore (as we've already sent out and received them all) - completely stuck

The fix is to check if we have any responses that can proceed in `try_block_sync` (this is called every 5 seconds) if we don't have any requests to send out. This seems to solve the problem.

Included is also a smaller bugfix that will reduce memory somewhat (we were not cleaning up the `request_timestamps` when handling `remove_block_response`, keeping these longer in the map than necessary - these would be cleaned up 15s later anyway, but better to correct this).

@howardwu Matthew Davids confirms this fix works.

### DIFF
--- a/node/sync/src/block_sync.rs
+++ b/node/sync/src/block_sync.rs
@@ -218,6 +218,13 @@ impl<N: Network> BlockSync<N> {
         let block_requests = self.prepare_block_requests();
         trace!("Prepared {} block requests", block_requests.len());
 
+        // If there are no block requests, then try if we can advance with responses we already
+        // have in the sync pool.
+        if block_requests.is_empty() && !self.responses.read().is_empty() {
+            trace!("No block requests to send - try handling responses we already have");
+            self.handle_responses(self.canon.latest_block_height());
+        }
+
         // Process the block requests.
         'outer: for (height, (hash, previous_hash, sync_ips)) in block_requests {
             // Insert the block request into the sync pool.
@@ -281,8 +288,14 @@ impl<N: Network> BlockSync<N> {
         };
 
         // Retrieve the latest block height.
-        let mut current_height = self.canon.latest_block_height();
+        let current_height = self.canon.latest_block_height();
         // Try to advance the ledger with the sync pool.
+        self.handle_responses(current_height);
+        Ok(())
+    }
+
+    /// Handles the block responses from the sync pool.
+    fn handle_responses(&self, mut current_height: u32) {
         while let Some(block) = self.remove_block_response(current_height + 1) {
             // Ensure the block height matches.
             if block.height() != current_height + 1 {
@@ -302,7 +315,6 @@ impl<N: Network> BlockSync<N> {
             // Update the latest height.
             current_height = self.canon.latest_block_height();
         }
-        Ok(())
     }
 }
 
@@ -548,6 +560,8 @@ impl<N: Network> BlockSync<N> {
         }
         // Remove the request entry for the given height.
         requests.remove(&height);
+        // Remove the request timestamp entry for the given height.
+        self.request_timestamps.write().remove(&height);
         // Remove the response entry for the given height.
         self.responses.write().remove(&height)
     }
@@ -572,6 +586,7 @@ impl<N: Network> BlockSync<N> {
 
     /// Removes all block requests for the given peer IP.
     fn remove_block_requests_to_peer(&self, peer_ip: &SocketAddr) {
+        trace!("Removing all block requests to peer {peer_ip}");
         // Acquire the write lock on the requests map.
         let mut requests = self.requests.write();
         // Acquire the read lock on the responses map.
--- a/node/sync/src/block_sync.rs
+++ b/node/sync/src/block_sync.rs
@@ -218,11 +218,15 @@ impl<N: Network> BlockSync<N> {
         let block_requests = self.prepare_block_requests();
         trace!("Prepared {} block requests", block_requests.len());
 
-        // If there are no block requests, then try if we can advance with responses we already
-        // have in the sync pool.
+        // If there are no block requests, but there are pending block responses in the sync pool,
+        // then try to advance the ledger using these pending block responses.
         if block_requests.is_empty() && !self.responses.read().is_empty() {
-            trace!("No block requests to send - try handling responses we already have");
-            self.handle_responses(self.canon.latest_block_height());
+            // Retrieve the latest block height.
+            let current_height = self.canon.latest_block_height();
+            // Try to advance the ledger with the sync pool.
+            trace!("No block requests to send - try advancing with block responses (at block {current_height})");
+            self.try_advancing_with_block_responses(current_height);
+            // Return early.
             return;
         }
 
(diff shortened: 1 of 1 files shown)


niklaslong (approved):
Seems reasonable 👍 I haven't run this locally yet but I believe a devnet is being run with this change?

node/sync/src/block_sync.rs
```diff
@@ -215,6 +215,13 @@ impl<N: Network> BlockSync<N> {
         let block_requests = self.prepare_block_requests();
         trace!("Prepared {} block requests", block_requests.len());
 
+        // If there are no block requests, then try if we can advance with responses we already
+        // have in the sync pool.
+        if block_requests.is_empty() && !self.responses.read().is_empty() {
+            trace!("No block requests to send - try handling responses we already have");
+            self.handle_responses(self.canon.latest_block_height());
```
kpp: This is not a fix of the issue. There is `construct_request()` that calculates `num_sync_ips` which is sometimes `REDUNDANCY_FACTOR=3`.

The fix should be done in other way:

```diff
/// A struct that keeps track of the current block sync state.
///
/// # State
/// - When a request is inserted, the `requests` map and `request_timestamps` map insert an entry for the request height.
/// - When a response is inserted, the `requests` map inserts the entry for the request height.
-/// - When a request is completed, the `requests` map still has the entry, but its `sync_ips` is empty;
+/// - When a request is completed, the `requests` map still has the entry, but its `sync_ips` is empty AND num_sync_ips = 0;
///   the `request_timestamps` map remains unchanged.
/// - When a response is removed/completed, the `requests` map and `request_timestamps` map also remove the entry for the request height.
/// - When a request is timed out, the `requests`, `request_timestamps`, and `responses` map remove the entry for the request height;
#[derive(Clone, Debug)]
pub struct BlockSync<N: Network> {
...

/// Inserts the given block response, after checking that the request exists and the response is well-formed.
-/// On success, this function removes the peer IP from the requests map.
+/// On success, this function removes the peer IP from the requests map AND decrements num_sync_ips.
/// On failure, this function removes all block requests from the given peer IP.
fn insert_block_response(&self, peer_ip: SocketAddr, block: Block<N>) -> Result<()> { ... }


fn remove_timed_out_block_requests(&self) -> usize {
...
    // Determine if the request is incomplete.
    let is_request_incomplete =
-        !requests.get(height).map(|(_, _, peer_ips)| peer_ips.is_empty()).unwrap_or(false);
+        !requests.get(height).map(|(_, _, _, num_sync_ips)| num_sync_ips > 0).unwrap_or(false);
```

ljedrz: Shouldn't we `return Ok(());` after handling the responses here?

ljedrz (approved):
LGTM with one question.

node/sync/src/block_sync.rs
joske: The method `try_block_sync` doesn't return anything, so not sure why you say that?

joske: We could return early, but it's not really necessary as we know that the `block_requests` is empty.

joske: This doesn't even compile.

ljedrz: > The method try_block_sync doesn't return anything

ah yeah, then just `return;`

ljedrz: > it's not really necessary as we know that the block_requests is empty.

indeed, I only mean it as a readability improvement

kpp: Good to know.

@ljedrz there is a malicious attempt from the author of this PR to reduce the number of sync nodes to fetch blocks from.

ljedrz: how so? this new logic is triggered only if no `block_requests` are available

kpp: Because the completeness of block request must be defined by the number of (correct) responses not by peer_ips.empty().

Imagine you got 1 block response from Mallory while Alice and Bob are timed out (or removed by remove_block_requests_to_peer. You would want to mark this block request as timed out, generate a new one (with other random peers) and get a complete block response from them. You don’t want to proceed with the one block response from Mallory.

niklaslong: @kpp, we've taken another look at the logic, I'll attempt to summarise here. The checks implemented on snarkOS (and down in VM) when progressing to the next block should be sufficient to allay the concerns you've mentioned above about exploiting timeouts in sync to misguide a validator. Specifically, we check the authority of the block (amongst other things)—this makes receiving even a single block safe for sync purposes. In short, we don't rely on communication redundancy for safety. Don't hesitate to follow up if you're unsure about this and still have concerns. 

Please note that sync logic is in flux at the moment, this PR is part of a larger effort to improve this aspect of the network.

kpp: Thank you!

HarukaMa:
It seems I started to have some issues after trying to apply this patch:

1. Sometimes the following panic happens and the node stops syncing:

```
thread 'tokio-runtime-worker' panicked at 'assertion failed: self.database.atomic_batch.lock().is_empty()', /home/mrx/snarkVM/ledger/store/src/helpers/rocksdb/internal/map.rs:111:9
2023-12-15T09:19:18.114699Z  WARN snarkos_node::client::router: Disconnecting from '...' - task 70531 panicked  
```

then:

```
2023-12-15T09:19:20.478289Z  WARN snarkos_node_sync::block_sync: The next block (969045) is invalid - Cannot start an atomic batch write operation while another one is already in progress.
```

I've never seen this before so it's possible that this is introduced by this pr.

2. Sometimes the node would still stop to sync at a random height and will stuck at loops like:

```
2023-12-15T00:04:22.274442Z TRACE snarkos_node_sync::block_sync: Updating is_block_synced: greatest_peer_height = 962574, canon_height = 958768
2023-12-15T00:04:22.274893Z TRACE snarkos_node_sync::block_sync: Prepared 0 block requests
2023-12-15T00:04:22.274907Z TRACE snarkos_node_sync::block_sync: No block requests to send - try handling responses we already have
```

this seems different from #2916 as the node is not even trying to send block requests at this point.

joske:
Hi @HarukaMa I don't think 1. is caused by this PR. For 2, can you try to run with `--verbosity 4`? This would give me helpful info.

HarukaMa:
If 1. is not caused by this PR then I know where to report to.

I'm now running with verbosity 4. Let's see if the node could replicate that itself.

HarukaMa:
I don't really know where to cut the log as there are too many messages, but here's the compressed log file. You can search for the last "Advanced to" to see where it stopped syncing.

Original file name is `snarkos.log.zst`.

snarkos log zst

node/sync/src/block_sync.rs
```diff
@@ -572,6 +587,7 @@ impl<N: Network> BlockSync<N> {
 
     /// Removes all block requests for the given peer IP.
     fn remove_block_requests_to_peer(&self, peer_ip: &SocketAddr) {
+        trace!("Removing all block requests to peer {peer_ip}");
```
```suggestion
        trace!("Block sync is removing all block requests to peer {peer_ip}");
```

howardwu (approved):

PR #2909, opened 2023-12-09
feat(cli): Add sign and verify sub commands to the account command

Meshiest:
## Motivation

Signatures can be created/verified programmatically both from [AleoHQ/sdk](https://github.com/AleoHQ/sdk) (web-ui via [aleo.tools](https://aleo.tools/account)) and within [leo](https://developer.aleo.org/leo/language/#signatures) but not from the official snarkOS CLI.

This PR creates two straightforward commands for working with Aleo signatures:

- `snarkos account sign <--private-key-file <privateKeyFile>|--private-key <privateKey>> -m <message> -s [seed] [--raw]`
    outputs an Aleo signature of the message, signed with the private key.
- `snarkos account verify -s <signature> -m <message> -a <address> [--raw]`
    outputs `verified` or `invalid` if the signature is verified or invalid respectfully.

By default, all messages are snarkVM values so signatures are compatible with the [`signature::verify`](https://developer.aleo.org/leo/operators#signatureverify) leo operator. If the `--raw` flag is passed, the message will be parsed as bytes, which is compatible with the [AleoHQ/sdk].

This PR also includes "lite" documentation in the form of an FAQ entry ([rendered here](https://github.com/AleoHQ/snarkOS/blob/ddd901ba256d40b315f16f79d09a5a1f2d7117c1/README.md#4-how-do-i-use-the-cli-to-sign-and-verify-a-message)) for how to sign/verify messages.

## Test Plan

No code outside of the new features were changed and test coverage is as follows:

- Signature creation (unseeded)
- Seeded signature creation
- Valid signature verification
- Invalid signature verification (message is incorrect)
- Invalid signature verification (address is incorrect)
- All of the above using plaintext messages (rather than aleo values)

## Related PRs

No related work


howardwu:
@d0cd do we want to support this signature call? It does not match with the signature scheme used in Leo and Aleo instructions. If we're fine with non-compatibility, then we can merge this in.

Meshiest:
@howardwu I've added support for signing any aleo literal value, with regular plaintext as bytes fallback.

image
image
```leo
program helloworld_nefsp0h.aleo {
  transition main(public s: signature, a: address, m: field) -> bool {
      return s.verify(a, m);
  }
}
```

`leo run main sign1w0gwak002kqxyqxknkft7zxj45vvthk3nd3jqy29xlj2qtllvuqdqjhn5y3jvj56pw44vqppxynkv3wywh7kk02vauvcgymrhdk5sqawfz3e8r7u2tm5h7jun68vsmc9 aleo18pfumcshatzm2r6te5erxeaxkrynsk2qkdpfy3wd90fhq573cgysj9kpx4 5field`

HarukaMa:
I don't think this kind of ambiguity of input message format is good though - what if we want to sign a string which happens to be able to interpreted as `Literal`? Plus in the future we might have string type as well.

This is mainly because we probably will want to have the capability to sign arbitrary content.

Meshiest:
@HarukaMa makes a very good point. I've added a flag `-l`/ `--literal` to gate the literal parsing feature.

This ensures the `sign` and `verify` commands have compatibility with the wasm sdk (which does not have compatibility with signing aleo literals) when the flag is absent and compatibility with leo when present.

image
image

d0cd:
@Meshiest thank you for your contribution! 
I think the default option should be to support signing snarkVM datatypes (`Value`).
We can then add a `--raw, -r` flag to support signing raw messages.

Meshiest:
@d0cd I've inverted `--literal` into `--raw` as requested

Meshiest:
I can reword the language in the CLI and README to be "snarkVM values" if desired

Meshiest:
zeroize doesn't support PathBuf so I had to move that parsing from `clap` derive macro into the enum handler

README.md
```diff
@@ -152,6 +152,15 @@ APrivateKey1xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
 - Before running the command above (`snarkos account new`) try `source ~/.bashrc`
 - Also double-check the spelling of `snarkos`. Note the directory is `/snarkOS`, and the command is `snarkos`
 
+### 4. How do I use the CLI to sign and verify a message?
+
+1. Generate an account with `snarkos account new` if you haven't already
+2. Sign a message with your private key using `snarkos account sign --raw -m "Message" --private-key-file=<PRIVATE_KEY_FILE>`
+3. Verify your signature with `snarkos account verify --raw -m "Message" -s sign1SignatureHere -a aleo1YourAccountAddress`
+
+Note, using the `--raw` flag with the command will sign plaintext messages as bytes rather than [leo](https://leo-lang.org/) values such as `1u8` or `100field`.
```
d0cd:
```suggestion
Note, using the `--raw` flag with the command will sign plaintext messages as bytes rather than [Aleo](https://developer.aleo.org/aleo/language#data-types-and-values) values such as `1u8` or `100field`.
```

cli/src/commands/account.rs
```diff
@@ -46,6 +54,45 @@ pub enum Account {
...
+        /// Specify the path to a file containing the account private key of the node
+        #[clap(long = "private-key-file")]
+        private_key_file: Option<String>,
+        /// Message (Aleo value) to sign
+        #[clap(short = 'm', long)]
+        message: String,
+        /// Seed the RNG with a numeric value
+        #[clap(short = 's', long)]
+        seed: Option<String>,
+        /// When enabled, parses the message as bytes instead of aleo literals
```
d0cd:
```suggestion
        /// When enabled, parses the message as bytes instead of Aleo literals
```

cli/src/commands/account.rs
```diff
@@ -46,6 +54,45 @@ pub enum Account {
...
+        /// Address to use for verification
+        #[clap(short = 'a', long)]
+        address: String,
+        /// Signature to verify
+        #[clap(short = 's', long)]
+        signature: String,
+        /// Message (Aleo value) to verify the signature against
+        #[clap(short = 'm', long)]
+        message: String,
+        /// When enabled, parses the message as bytes instead of aleo literals
```
d0cd:
```suggestion
        /// When enabled, parses the message as bytes instead of Aleo literals
```

cli/src/commands/account.rs
```diff
@@ -46,6 +54,45 @@ pub enum Account {
...
+        #[clap(short = 'r', long)]
+        raw: bool,
+    },
+}
+
+/// Parse a raw leo input into fields
```
d0cd:
```suggestion
/// Parse a raw Aleo input into fields
```

cli/src/commands/account.rs
```diff
@@ -176,6 +238,60 @@ impl Account {
...
+
+        // Parse the private key
+        let private_key =
+            PrivateKey::<Network>::from_str(&key).map_err(|_| anyhow!("Failed to parse a valid private key"))?;
+        // Sign the message
+        let signature = if raw {
+            private_key.sign_bytes(message.as_bytes(), &mut rng)
+        } else {
+            let fields =
+                aleo_literal_to_fields(&message).map_err(|_| anyhow!("Failed to parse a valid aleo literal"))?;
```
d0cd:
```suggestion
                aleo_literal_to_fields(&message).map_err(|_| anyhow!("Failed to parse a valid Aleo literal"))?;
```

cli/src/commands/account.rs
```diff
@@ -176,6 +238,60 @@ impl Account {
...
+        let address = Address::<Network>::from_str(&address).map_err(|_| anyhow!("Failed to parse a valid address"))?;
+        // Parse the signature
+        let signature =
+            Signature::<Network>::from_str(&signature).map_err(|_| anyhow!("Failed to parse a valid signature"))?;
+        // Verify the signature
+        let verified = if raw {
+            signature.verify_bytes(&address, message.as_bytes())
+        } else {
+            let fields =
+                aleo_literal_to_fields(&message).map_err(|_| anyhow!("Failed to parse a valid aleo literal"))?;
```
d0cd:
```suggestion
                aleo_literal_to_fields(&message).map_err(|_| anyhow!("Failed to parse a valid Aleo literal"))?;
```

cli/src/commands/account.rs
```diff
@@ -176,6 +238,60 @@ impl Account {
...
+            signature.verify(&address, &fields)
+        };
+
+        // Return the verification result
+        Ok(if verified { "verified" } else { "invalid" }.to_string())
```
d0cd:
```suggestion
        match verified {
            true => Ok("✅ The signature is valid"),
            false =>  bail!("❌ The signature is invalid"),
        }
```

cli/src/commands/account.rs
```diff
@@ -46,6 +54,45 @@ pub enum Account {
...
+        raw: bool,
+    },
+}
+
+/// Parse a raw leo input into fields
+fn aleo_literal_to_fields(input: &str) -> Result<Vec<Field<Network>>> {
+    let literal = snarkvm::circuit::Literal::<A>::from_str(input)?;
+    let plaintext = snarkvm::circuit::Plaintext::<A>::from(literal);
+    let value = snarkvm::circuit::Value::<A>::Plaintext(plaintext);
+    Ok(value.to_fields().into_iter().map(|f| f.eject_value()).collect())
```
d0cd:
```suggestion
    let value = snarkvm::console::Value::<Network>::from_str(input)?;
    Ok(value.to_field().collect())
```

d0cd: ^ The idea is to use the console values directly.

cli/src/commands/account.rs
```diff
@@ -26,10 +30,14 @@ use crossterm::ExecutableCommand;
...
 use rayon::prelude::*;
-use std::io::{Read, Write};
+use std::{
+    io::{Read, Write},
+    path::PathBuf,
+};
 use zeroize::Zeroize;
 
 type Network = snarkvm::prelude::Testnet3;
+type A = snarkvm::circuit::AleoV0;
```
d0cd: I believe this is not needed if you are using the console values.

d0cd (changes requested):
@Meshiest awesome work on this PR! 
Left some suggestions and some nits.
I would also double check that the signature produced by the CLI tool actually works with the `sign.verify` instruction

Meshiest:
I've implemented all of your feedback (in one commit rather than a bunch of individual ones).

I've verified that the latest changes w/o the --raw flag work with the leo CLI (https://play.leo-lang.org/ is down)

image
image

```rs
program foo.aleo {
    transition main(public s: signature, a: address, m: field) -> bool {
      return s.verify(a, m);
    }
}
```

Meshiest:
As Leo Playground is back, here's a screenshot of it working there too 

image

d0cd (approved):
LGTM!

howardwu:
Apologies for the delay, just seeing this LGTM, proceeding to merge.

PR #2914, opened 2023-12-11
Don't stop the node abruptly when syncing with CDN

ljedrz:
This PR ensures that the CDN sync process can be interrupted cleanly. I've tested it several times and haven't encountered any issues.

note: normally I'd prefer to signal the sync to stop and then conclude the shutdown from within the signal handler, but due to the sync being done via an async stream, it is much simpler for the signal handler to notify the stream so that it calls the shutdown instead.

### DIFF
--- a/.integration/src/lib.rs
+++ b/.integration/src/lib.rs
@@ -42,7 +42,8 @@ mod tests {
         // Perform the sync.
         let rt = tokio::runtime::Runtime::new().unwrap();
         rt.block_on(async {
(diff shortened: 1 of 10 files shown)


ljedrz:
Update: the PR now also reduces the probability of interrupting block insertion beyond the CDN sync, and it reuses the `shutdown` flag instead of introducing a new one.

howardwu (approved):

PR #2927, opened 2023-12-18
Fetch certificate from ledger as backup and increase timeouts

raychu86:
## Motivation

This PR attempts to fetch a certificate from the ledger if the certificate ID (from a CertificateRequest) does not exist in the Certificate Storage. In addition the request timeouts are increased for the following:

Block requests: 15s -> 25s
Batch Certificate: 2.5s -> 5s
Transmission: 2.5s -> 5s

Note: These values are experimental and should be tuned.


raychu86:
Closing in favor of #3079

PR #2948, opened 2023-12-26
feat: add support for displaying private key discreetly for `snarkos account new`

Halimao:
## Motivation

Currently, running `snarkos account new` will show private key dicrectly in current terminal, it would be better to output it in an alternate screen. See following pics

image
image

### DIFF
--- a/account/src/lib.rs
+++ b/account/src/lib.rs
@@ -21,8 +21,6 @@ use snarkvm::{
 
 use colored::*;
 use core::fmt;
-use crossterm::ExecutableCommand;
-use std::io::{Read, Write};
 
 /// A helper struct for an Aleo account.
 #[derive(Clone, Debug)]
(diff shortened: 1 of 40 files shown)


account/src/lib.rs
```diff
@@ -146,11 +148,14 @@ impl<N: Network> FromStr for Account<N> {
 impl<N: Network> Display for Account<N> {
     /// Renders the account as a string.
     fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        display_string_discreetly(
```
howardwu: Let's add a CLI flag, such as `--discreet` to make this approach optional.

Halimao: Okay, I will add it later😊

Halimao: @howardwu Addressed

Halimao:
@howardwu Hi sir, suggestions has addressed, looking forward for your review again

Halimao:
@howardwu Hi sir, is there anything else need to add? Or can we get this merged then😁

cli/src/commands/account.rs
```diff
@@ -39,32 +41,35 @@ pub enum Account {
         /// Try until an address with the vanity string is found
         #[clap(short = 'v', long)]
         vanity: Option<String>,
+        /// Print sensitive information(such as private key) discreetly to an alternate screen
```
howardwu:
```suggestion
        /// Print sensitive information (such as the private key) discreetly in an alternate screen
```

howardwu (approved):

PR #2953, opened 2023-12-27
feat: add rest api for broadcast solution

ghostant-1017:
## Motivation

Add rest api for broadcast prover solution


ghostant-1017:
Could anyone merge this pr in please?
I think it wouldn't cause any side effect.

howardwu (approved):

PR #2956, opened 2023-12-27
fix: limit outstanding certificate requests

### LINKED ISSUE #2935: [Bug] A malicious validator can broadcast invalid BatchCertificate or Propose that cause other validators stack overflow
# https://hackerone.com/reports/2279584

## Summary:
The `Primary` recursively fetch certificates from other peer upon receiving new `Propose` and `BatchCertificate `. Although it will check the certificate before storing into db, the check is at end of the recursion. 

A malicious validator can create fake certificate that contains other fake certificate at the same round. Then the recursion will never ends and lead to stack overflow. The call stack will be like this:

```
sync_with_certificate_from_peer
    sync_with_batch_header_from_peer
        sync_with_certificate_from_peer
             ....
```

The attacker can crash all other validator in this way.

## Steps To Reproduce:

See the commit: https://github.com/Gooong/snarkOS/commit/3e7c6866203c02dd9db225f81d794ad9d2aae146

1. Clone the repo above
2. Run ./devnet.sh
3. Wait round 5
4. Check the log and we will find some node crashed: `fatal runtime error: stack overflow`

## Proof-of-Concept (PoC)

The attacker just broadcast fake certificate. This will cause all other validator continously crash and shutdown the network.

See the attachment 

## Impact

Cause validators crash and shutdown the network.

## Fix suggestions:

Strictly check certificate before the recursion.

joske:
This attempts to fix https://github.com/AleoHQ/snarkOS/issues/2935. 

I was able to reproduce using the following malicious code (credit @niklaslong): https://github.com/niklaslong/snarkOS/tree/tmp/2935-replication

Run one node with this code, and the rest with normal `testnet3`. After 20 rounds, the malicious node will send fake certificate requests.

With this fix, the nodes reject the bad certs after the max outstanding requests of 20 is reached, and then carry on, before all nodes would crash with stack overflow.

The `MAX_OUTSTANDING_CERTIFICATE_REQUESTS` is chosen arbitrarily at 20, please change as you see fit.

### DIFF
--- a/node/bft/src/primary.rs
+++ b/node/bft/src/primary.rs
@@ -60,7 +60,7 @@ use futures::stream::{FuturesUnordered, StreamExt};
 use indexmap::{IndexMap, IndexSet};
 use parking_lot::{Mutex, RwLock};
 use std::{
-    collections::{hash_map::Entry, HashMap, HashSet},
+    collections::{HashMap, HashSet},
     future::Future,
     net::SocketAddr,
     sync::Arc,
@@ -1492,25 +1492,16 @@ impl<N: Network> Primary<N> {
             if !self.storage.contains_certificate(*certificate_id) {
                 trace!("Primary - Found a new certificate ID for round {round} from '{peer_ip}'");
                 let mut outstanding_requests = self.outstanding_certificate_requests.lock();
-                let mut value = 1;
                 let entry = outstanding_requests.entry((peer_ip, round));
-                match entry {
-                    Entry::Occupied(mut entry) => {
-                        *entry.get_mut() += 1;
-                        value = *entry.get();
-                        if value > MAX_OUTSTANDING_CERTIFICATE_REQUESTS {
-                            entry.remove();
-                            return Err(anyhow!(
-                                "Too many outstanding certificate requests ({}) for peer '{peer_ip}'",
-                                value
-                            ));
-                        }
-                    }
-                    Entry::Vacant(entry) => {
-                        entry.insert(1);
-                    }
+                // Get the number of outstanding requests for the peer. Note: the map is cleared on each round.
+                let num_outstanding_requests = entry.and_modify(|value| *value += 1).or_insert(1);
+                if *num_outstanding_requests > MAX_OUTSTANDING_CERTIFICATE_REQUESTS {
+                    return Err(anyhow!(
+                        "Too many outstanding certificate requests ({}) for peer '{peer_ip}'",
+                        num_outstanding_requests
+                    ));
                 }
-                debug!("Sending a certificate request to '{peer_ip}' ({value} outstanding)");
+                debug!("Sending a certificate request to '{peer_ip}' ({num_outstanding_requests} outstanding)");
                 // Send an certificate request to the peer.
                 fetch_certificates.push(self.sync.send_certificate_request(peer_ip, *certificate_id));
             }
--- a/node/bft/src/primary.rs
+++ b/node/bft/src/primary.rs
@@ -60,7 +60,7 @@ use futures::stream::{FuturesUnordered, StreamExt};
 use indexmap::{IndexMap, IndexSet};
 use parking_lot::{Mutex, RwLock};
 use std::{
-    collections::{HashMap, HashSet},
+    collections::{hash_map::Entry, HashMap, HashSet},
     future::Future,
     net::SocketAddr,
     sync::Arc,
@@ -70,6 +70,8 @@ use tokio::{
     sync::{Mutex as TMutex, OnceCell},
     task::JoinHandle,
 };
+
+const MAX_OUTSTANDING_CERTIFICATE_REQUESTS: u32 = 20;
 
 /// A helper type for an optional proposed batch.
 pub type ProposedBatch<N> = RwLock<Option<Proposal<N>>>;
@@ -96,6 +98,8 @@ pub struct Primary<N: Network> {
     handles: Arc<Mutex<Vec<JoinHandle<()>>>>,
     /// The lock for propose_batch.
     propose_lock: Arc<TMutex<u64>>,
+    /// The outstanding requests to a peer.
+    outstanding_certificate_requests: Arc<RwLock<HashMap<(SocketAddr, u64), u32>>>,
 }
 
 impl<N: Network> Primary<N> {
@@ -124,6 +128,7 @@ impl<N: Network> Primary<N> {
             signed_proposals: Default::default(),
             handles: Default::default(),
             propose_lock: Default::default(),
+            outstanding_certificate_requests: Default::default(),
         })
     }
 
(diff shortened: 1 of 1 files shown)


node/bft/src/primary.rs
```diff
@@ -96,6 +98,8 @@ pub struct Primary<N: Network> {
     handles: Arc<Mutex<Vec<JoinHandle<()>>>>,
     /// The lock for propose_batch.
     propose_lock: Arc<TMutex<u64>>,
+    /// The outstanding requests to a peer.
+    outstanding_certificate_requests: Arc<RwLock<HashMap<(SocketAddr, u64), u32>>>,
```
```suggestion
    outstanding_certificate_requests: Arc<Mutex<HashMap<(SocketAddr, u64), u32>>>,
```
since we never take advantage of concurrent `read`s

randomsleep:
I am the reporter of #2935. The root cause of the issue is that the validator didn't verify the certificate before entering recursion. I think we can resolve the issue more directly:

1. Upon receiving a `BatchPropose`, verify the direct parent certificates are all valid.
2. Upon receiving a `Certificate` through broadcast, immediately verify if it is valid.

Also, we may support `Batch Certificate Request` and avoid using recursion.

ghostant-1017:
> I am the reporter of #2935. The root cause of the issue is that the validator didn't verify the certificate before entering recursion. I think we can resolve the issue more directly:
> 
> 1. Upon receiving a `BatchPropose`, verify the direct parent certificates are all valid.
> 2. Upon receiving a `Certificate` through broadcast, immediately verify if it is valid.
> 
> Also, we may support `Batch Certificate Request` and avoid using recursion.

Verify the `BatchPropose` or `BatchCertificate` rely on the previous certificates. And we don't know the committee has changed or not when receive a `BatchPropose` or `BatchCertificate`.

randomsleep:
> > I am the reporter of #2935. The root cause of the issue is that the validator didn't verify the certificate before entering recursion. I think we can resolve the issue more directly:
> > 
> > 1. Upon receiving a `BatchPropose`, verify the direct parent certificates are all valid.
> > 2. Upon receiving a `Certificate` through broadcast, immediately verify if it is valid.
> > 
> > Also, we may support `Batch Certificate Request` and avoid using recursion.
> 
> Verify the `BatchPropose` or `BatchCertificate` rely on the previous certificates. And we don't know the committee has changed or not when receive a `BatchPropose` or `BatchCertificate`.

Yes, this is a problem for the current implementation. 

I submitted another bug report before, which is about the committee update. The report shows that updating the committee at every block will cause some consensus issues. Maybe we can fix that issue first to make it easier to validate the certificate.

howardwu:
@randomsleep @ghostant-1017 Is the issue still present in latest commit of `mainnet-staging`? If there is a better design, we can close this PR and work on a new issue.

randomsleep:
@howardwu I believe this issue still exists in `mainnet-staging`. Now the committee is fixed for every round. We can address this issue by:

1. Upon receiving a Certificate, immediately check if it has reached quorum before recursion. If the Certificate is the parent of an already checked Certificate, we can skip the check.
2. Upon receiving a BatchPropose, immediately perform some checks before recursion. Make sure the direct parent Certificate has reached quorum and `parent_round == batch_propose_round + 1`

node/bft/src/primary.rs
```diff
@@ -1127,6 +1132,7 @@ impl<N: Network> Primary<N> {
                 fast_forward_round = self.storage.increment_to_next_round(fast_forward_round)?;
                 // Clear the proposed batch.
                 *self.proposed_batch.write() = None;
+                self.outstanding_certificate_requests.lock().clear();
```
kaimast, line 1135: Could we document here why it is safe to clear all outstanding requests, not just the ones for the given round?

It looks like all requests are always from the most recent round. If so, do we need to track the round number in the HashMap?

node/bft/src/primary.rs
```diff
@@ -71,6 +71,8 @@ use tokio::{
     task::JoinHandle,
 };
 
+const MAX_OUTSTANDING_CERTIFICATE_REQUESTS: u32 = 20;
```
kaimast, line 74: Might be good to add a comment that indicates why we picked 20? Also maybe write that this is per peer and not the total number of outstanding requests (if I understand correctly).

node/bft/src/primary.rs
```diff
@@ -96,6 +98,8 @@ pub struct Primary<N: Network> {
     handles: Arc<Mutex<Vec<JoinHandle<()>>>>,
     /// The lock for propose_batch.
     propose_lock: Arc<TMutex<u64>>,
+    /// The outstanding requests to a peer.
+    outstanding_certificate_requests: Arc<Mutex<HashMap<(SocketAddr, u64), u32>>>,
```
kaimast, line 102: Similarly, here it would be good to write what this maps. Something like "(peer_ip x round_number) -> num_outstanding_requests" would be sufficient.

Maybe in the future we can have a round number type. u64 is always a little less descriptive than I would like.

node/bft/src/primary.rs
```diff
@@ -1485,7 +1491,17 @@ impl<N: Network> Primary<N> {
             // If we do not have the certificate, request it.
             if !self.storage.contains_certificate(*certificate_id) {
                 trace!("Primary - Found a new certificate ID for round {round} from '{peer_ip}'");
-                // TODO (howardwu): Limit the number of open requests we send to a peer.
+                let mut outstanding_requests = self.outstanding_certificate_requests.lock();
+                let entry = outstanding_requests.entry((peer_ip, round));
+                // Get the number of outstanding requests for the peer. Note: the map is cleared on each round.
+                let num_outstanding_requests = entry.and_modify(|value| *value += 1).or_insert(1);
```
kaimast, line 1497: Is this correct? We increment before checking if `num_oustanding_requests` exceeds the limit.

node/bft/src/primary.rs
```diff
@@ -1485,7 +1491,17 @@ impl<N: Network> Primary<N> {
...
+                if *num_outstanding_requests > MAX_OUTSTANDING_CERTIFICATE_REQUESTS {
+                    return Err(anyhow!(
```
kaimast, line 1499: If we return here and there are already futures in `fetch_certificates`, those will cancelled. In other words, if `fetch_missing_certificates` gets a list of more than 20 certificates, none of them will be fetched and we will not be able to fetch any certificates from this peer for this round.

Is that the desired behavior? If so, we should expand the function documentation to state that, especially if there are cases where a benign/honest validator sends more than 20 requests.

kaimast (changes requested):
I added a few comments to the code. Feel free to reject some of the suggestions, I am still new to the codebase and might misunderstand some things :)

vicsn:
@kaimast my bad for requesting a review, the author of this PR (Hi Jos!) is not contracting anymore, can you re-create the PR and invite @ljedrz me and Ray?

PR #2961, opened 2023-12-29
Minor refactor to fit update to cost function updates

vicsn:
## Motivation

Fit to updated API in: https://github.com/AleoHQ/snarkVM/pull/2271


vicsn:
Closing this as people are having different opinions on the right API. The minor change will be easy to fix once a release is made.

PR #2970, opened 2024-01-03
Limit amount of transactions which clients verify in parallel

vicsn:
## Motivation

Our clients can and do run out of memory, probably because they have no limit on how many transactions they verify in parallel. This PR proposes to queue them (just like the validator does in `Consensus`) and limit how much parallel verification we do.

We can do a lot of clever things to increase the processing speed, like check how many constraints the incoming transactions have, await on a channel to rapidly start verifying, but the focus for now is simplicity and safety.

## Test Plan

- Verified that the client still runs fine with https://github.com/AleoHQ/snarkOS/pull/2972

## Related PRs

Implicitly assumes we have https://github.com/AleoHQ/snarkVM/pull/2271 to limit size of transactions coming through.
And this would help to minimize the overhead of transaction verification when syncing blocks with already seen transactions: https://github.com/AleoHQ/snarkVM/pull/2270


node/src/client/mod.rs
```diff
@@ -181,6 +201,72 @@ impl<N: Network, C: ConsensusStorage<N>> Client<N, C> {
...
+                let num_transactions = match previous_verification_counter {
+                    // Determine how many transactions we can verify.
+                    Ok(previous_value) => num_transactions.saturating_sub(previous_value),
+                    // If we are already verifying sufficient transactions, don't verify any more for now.
+                    Err(_) => continue,
+                };
+
+                // For each transaction, spawn a task to verify it.
+                for _ in 0..num_transactions {
+                    if let Some((_, (peer_ip, serialized, transaction))) = node.transaction_queue.lock().pop_lru() {
```
ljedrz: I think the cache should be unlocked just once for the entirety of the loop, i.e. we should bind `node.transaction_queue.lock()` before it

node/src/client/mod.rs
```diff
@@ -181,6 +201,72 @@ impl<N: Network, C: ConsensusStorage<N>> Client<N, C> {
...
+                        let _node = node.clone();
+                        tokio::spawn(async move {
```
ljedrz, line 260: I wonder if these task handles shouldn't be accounted for in case of a shutdown :thinking:

node/src/client/mod.rs
```diff
@@ -39,17 +39,27 @@ use snarkvm::{
...
+    sync::{
+        atomic::{AtomicBool, AtomicUsize, Ordering::Relaxed},
+        Arc,
+    },
 };
 use tokio::task::JoinHandle;
 
+const VERIFICATION_CONCURRENCY_LIMIT: usize = 6; // 8 deployments of MAX_NUM_CONSTRAINTS will run out of memory.
+
+type TransactionContents<N> = (SocketAddr, UnconfirmedTransaction<N>, Transaction<N>);
```
ljedrz, line 65: we might want to comment on why we're preserving the serialized transaction too (i.e. for faster propagation)

node/src/client/mod.rs
vicsn: Hmm, interesting idea, but we don't currently pass a handler into the transaction verifier, so adding an extra shutdown check would not help much I think without a large refactor. We also don't do this for the [solution verifier](https://github.com/AleoHQ/snarkOS/blob/max_concurrent_deploy_verification/node/src/client/router.rs#L275).

node/src/client/mod.rs
vicsn: 989e741ee

node/src/client/mod.rs
vicsn: Sure: 321cd023e
Just so our mental models are aligned: given that we want it to take the "oldest seen" transactions, this won't change the order of transactions which are validated right? But just be more robust and perhaps speedy.

Added this to ensure we will really process by oldest tx first: 0d383e525

node/src/client/mod.rs
```diff
@@ -181,6 +203,73 @@ impl<N: Network, C: ConsensusStorage<N>> Client<N, C> {
...
+        self.handles.lock().push(tokio::spawn(async move {
+            loop {
+                // If the Ctrl-C handler registered the signal, stop the node.
+                if node.shutdown.load(Relaxed) {
+                    info!("Shutting down transaction verification");
+                    break;
+                }
+
+                // Sleep briefly to allow transactions to be validated.
+                tokio::time::sleep(std::time::Duration::from_millis(100)).await;
```
ljedrz: perhaps this value could be influenced by `queue_len`?

node/src/client/mod.rs
ljedrz: as long as we are still able to shut down normally (i.e. any ongoing tx verification isn't slowing things down too much), this shouldn't be an issue :ok_hand:

node/src/client/router.rs
```diff
@@ -303,11 +303,13 @@ impl<N: Network, C: ConsensusStorage<N>> Inbound<N> for Client<N, C> {
...
             return true; // Maintain the connection.
         }
-        // Check that the transaction is well-formed and unique.
-        if self.ledger.check_transaction_basic(&transaction, None, &mut rand::thread_rng()).is_ok() {
-            // Propagate the `UnconfirmedTransaction`.
-            self.propagate(Message::UnconfirmedTransaction(serialized), &[peer_ip]);
+
+        // Try to add the transaction to the verification queue, without changing LRU status of known txs.
+        let mut tx_queue = self.transaction_queue.lock();
+        if tx_queue.peek(&transaction.id()).is_none() {
```
ljedrz:
```suggestion
        if !tx_queue.contains(&transaction.id()) {
```

node/src/client/mod.rs
ljedrz: this change to locking shouldn't affect any order; it should just make task spawning a bit faster

node/src/client/mod.rs
```diff
@@ -181,6 +203,73 @@ impl<N: Network, C: ConsensusStorage<N>> Client<N, C> {
...
+                        let _node = node.clone();
+                        tokio::spawn(async move {
+                            // Check the transaction.
+                            if _node.ledger.check_transaction_basic(&transaction, None, &mut rand::thread_rng()).is_ok()
+                            {
+                                // Propagate the `UnconfirmedTransaction`.
+                                _node.propagate(Message::UnconfirmedTransaction(serialized), &[peer_ip]);
+                            }
+                            // Reduce the verification counter.
+                            _node.verification_counter.fetch_sub(1, Relaxed);
```
ljedrz, line 268: this could be moved after the loop so that we update the atomic only once (subtracting `num_transactions`)

node/src/client/mod.rs
vicsn: What about d2ba98384
Keeping in mind we should keep it simple to review

node/src/client/router.rs
vicsn: 7e12d015d

node/src/client/mod.rs
vicsn: I think that would no longer serve our goal, we only want the verification counter to be released when verification of a transaction is finished, not when we spawned all the verifications. Clarified the meaning a bit: 61e2a1d18

ljedrz: ah, good point

node/src/client/mod.rs
```diff
@@ -181,6 +204,80 @@ impl<N: Network, C: ConsensusStorage<N>> Client<N, C> {
...
+
+                // Determine the current length of the queue and counter.
+                let queue_len = node.transaction_queue.lock().len();
+                let counter = node.verification_counter.load(Relaxed);
+
+                // Sleep to allow the queue to be filled and transactions to be validated.
+                match (queue_len, counter) {
+                    // If the queue is empty, sleep longer and retry.
+                    (0, _) => {
+                        sleep(Duration::from_millis(500)).await;
```
ljedrz: I'm wondering if this isn't a bit much; we probably want to eagerly verify transactions as they come

vicsn: f6debd6f4

ljedrz (approved):
LGTM :+1:

node/src/client/mod.rs
```diff
@@ -39,17 +39,30 @@ use snarkvm::{
...
+    sync::{
+        atomic::{AtomicBool, AtomicUsize, Ordering::Relaxed},
+        Arc,
+    },
+    time::Duration,
 };
-use tokio::task::JoinHandle;
+use tokio::{task::JoinHandle, time::sleep};
+
+const VERIFICATION_CONCURRENCY_LIMIT: usize = 6; // 8 deployments of MAX_NUM_CONSTRAINTS will run out of memory.
```
raychu86, line 61: Should this be dynamic based on your machine? 

How much memory was this estimation based on?

vicsn: This was based on the `Clients: 16GB of memory` requirement in the snarkOS README. Please confirm if you want me to make it a dynamic value from a cli flag. 

Sidenote: given our syncing issues I understood the CPU/memory requirements need to go up anyway.

vicsn:
Closing in favour of: https://github.com/AleoNet/snarkOS/pull/3358

PR #2975, opened 2024-01-07
chore: fix the tmux window management in devnet script

bishopcheckmate:
## Motivation

Currently `.devnet.sh` script fails to start due to the creation of the second `window0` window, which then leads to the error in the script and the tmux session with no commands running in windows.

```
❯ ./devnet.sh
Enter the total number of validators (default: 4): 
Do you want to run 'cargo install --path .' to build the binary? (y/n, default: y): n
Do you want to clear the existing ledger logs? (y/n, default: n): n
can't find window: window0
```

The fix is to simply skip creating another window for the first validator.

## Test Plan

manual check

### DIFF
--- a/devnet.sh
+++ b/devnet.sh
@@ -51,13 +51,12 @@ for validator_index in "${validator_indices[@]}"; do
   # Generate a unique and incrementing log file name based on the validator index
   log_file="$log_dir/validator-$validator_index.log"
 
-  # Create a new window with a unique name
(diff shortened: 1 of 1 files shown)


howardwu:
Can you verify whether this PR is compatible with this merged in PR/change?
https://github.com/AleoHQ/snarkOS/pull/2972

thanks!

bishopcheckmate:
Hi @howardwu, yes, it is compatible and still valid

bishopcheckmate:
closing as it was also part of #2976

PR #2976, opened 2024-01-07
chore: respect tmux's base-index config in devnet.sh

bishopcheckmate:
## Motivation

I have a following line in my `.tmux.conf`
```
set -g base-index 1
```

which configures tmux to create windows with indexes starting from 1 instead of the default 0. I use it so that when I want to switch the windows using their indexes, the order of windows matches the order of num keys on keyboard.

This however doesn't play nice with the `devnet.sh` script, which assumes the indexing of windows from 0, thus failing on creating the second window `window1` as the index 1 got taken by the `window0`.

It would be cool if this tmux setting was taken into account.

I've opened this PR on top of the #2975 as it is kinda required, however I didn't want to include this directly in the previous PR as I  am not sure if you'd like to support this config option.

## Test Plan

manual check

### DIFF
--- a/devnet.sh
+++ b/devnet.sh
@@ -51,13 +51,12 @@ for validator_index in "${validator_indices[@]}"; do
   # Generate a unique and incrementing log file name based on the validator index
   log_file="$log_dir/validator-$validator_index.log"
 
-  # Create a new window with a unique name
-  tmux new-window -t "devnet:$validator_index" -n "window$validator_index"
-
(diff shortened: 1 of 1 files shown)


howardwu:
Can you verify whether this PR is compatible with this merged in PR/change?
https://github.com/AleoHQ/snarkOS/pull/2972

thanks!

bishopcheckmate:
This is also compatible now, I applied index offsetting to client windows too

howardwu (approved):

PR #2977, opened 2024-01-08
Process unconfirmed_transactions by priority_fee and oldest timestamp

vicsn:
## Motivation

Following EIP-1559 and common decency, this PR ensures we process transactions by priority_fee. If the priority_fee is the same, we order by timestamp (in nanoseconds), and otherwise randomly by transaction_id.

It may be narrowly individually rational for validators to only sort by priority_fee and ignore the timestamp/id, but "It is recommended that transactions with the same priority fee be sorted by time the transaction was received" - https://eips.ethereum.org/EIPS/eip-1559 . This improves the performance and security of the whole network, which also matters for validators.

## Test Plan

Separated the queue insertion and popping functions so they can be tested.


node/consensus/Cargo.toml
```diff
@@ -57,7 +57,7 @@ version = "=2.2.7"
 path = "../bft/ledger-service"
 version = "=2.2.7"
 default-features = false
-features = [ "ledger", "ledger-write" ]
+features = [ "ledger", "ledger-write", "test" ]
```
ljedrz: it'd probably be better if we didn't use `test` features in non`-dev` dependencies; we should probably make the required items available outside of test scenarios instead

node/consensus/src/lib.rs
```diff
@@ -234,10 +240,6 @@ impl<N: Network> Consensus<N> {
         {
             let transaction_id = transaction.id();
 
-            // Check that the transaction is not a fee transaction.
-            if transaction.is_fee() {
-                bail!("Transaction '{}' is a fee transaction {}", fmt_id(transaction_id), "(skipping)".dimmed());
-            }
```
ljedrz: we may want to keep this for performance (this avoids one lock access and 2 lookups); also, we currently don't touch `seen_transactions` in case of this early return

node/consensus/src/lib.rs
```diff
@@ -247,9 +249,18 @@ impl<N: Network> Consensus<N> {
             if self.ledger.contains_transmission(&TransmissionID::from(&transaction_id))? {
                 bail!("Transaction '{}' exists in the ledger {}", fmt_id(transaction_id), "(skipping)".dimmed());
             }
-            // Add the transaction to the memory pool.
+            // Get the priority_fee amount from the Deploy or Execute Transaction. Fee Transactions are not expected here.
+            let priority_fee = match &transaction {
```
ljedrz, line 280: since this is only used for queuing, I wonder if it wouldn't be cleaner if we obtained the fee within `add_transaction_to_queue`, which would only need the `Transaction` as a parameter

node/consensus/Cargo.toml
vicsn: Good catch, I only need it in the test scope: 38e973002

node/consensus/src/lib.rs
vicsn: 4a92b3a22

node/consensus/src/lib.rs
vicsn: It would indeed be cleaner, but because it will then obtain priority_fee from the Transaction, we'd need to make new `sample_deployment`/`sample_fee` functions, passing in a custom `priority_fee`'s to test. And I don't want to clutter the sampling code too much.

node/consensus/Cargo.toml
```diff
@@ -86,3 +86,13 @@ version = "1.19"
...
+
+[dev-dependencies.snarkvm]
+workspace = true
+features = [ "test-helpers" ]
+
+[dev-dependencies.snarkos-node-bft-ledger-service]
+path = "../bft/ledger-service"
+version = "=2.2.7"
+default-features = false
+features = [ "test" ]
```
raychu86:
```suggestion
features = [ "test" ]

```

nit

node/consensus/src/lib.rs
```diff
@@ -65,7 +71,7 @@ pub struct Consensus<N: Network> {
     /// The unconfirmed solutions queue.
     solutions_queue: Arc<Mutex<IndexMap<PuzzleCommitment<N>, ProverSolution<N>>>>,
     /// The unconfirmed transactions queue.
-    transactions_queue: Arc<Mutex<IndexMap<N::TransactionID, Transaction<N>>>>,
+    transactions_queue: Arc<Mutex<BTreeMap<TransactionKey<N>, Transaction<N>>>>,
```
raychu86, line 77: This may not catch duplicate transactions that come in at different times. We may now have multiple instances of the same transaction in the `transactions_queue`.

vicsn: I see, good catch.

Just for shared context: Before adding to the queue, we already check if the tx was seen recently in an LRU cache: `self.seen_transactions.lock().put(transaction_id, ()).is_some()`. But `seen_transactions` can evict txs before they are popped from the queue, causing duplicate transactions to be added to the queue, passed to workers to verify, wasting compute.

Best I could come up with for now is to add yet another `IndexSet` which tracks `transaction_id`'s in the queue: ac3a95be7. Independently, the LRU cache is still useful as a fast check.

Somewhat related:
- we didn't have any limit on the size of `transactions_queue`. Someone could spam e.g. 64_000 unique 1MB deployments and make the validators go OOM. Solutions don't have this issue because they are constant size. I added a limit of ~1 GB: 1704d3573
- I think we also need to evict from the cache to let retransmission succeed: ef61a9dce

raychu86:
Likely superseded by https://github.com/ProvableHQ/snarkOS/pull/3697

PR #2993, opened 2024-01-11
fix: bump timestamps of all requests when a BlockResponse is handled

joske:
Partial fix for https://github.com/AleoHQ/snarkOS/issues/2978

Rationale:

If there's only 1 peer connected, and it's slow to send `BlockRespons`es, or we are slow processing them, the `BlockRequest`s can expire in the `self.requests` map before they can be handled. If that happens, the only peer is disconnected, and syncing comes to a halt.

This PR bumps all the timestamps for all outstanding `BlockRequest`s when we've successfully handled a `BlockResponse`.

@HarukaMa confirms this fix works.

### DIFF
--- a/node/sync/src/block_sync.rs
+++ b/node/sync/src/block_sync.rs
@@ -319,6 +319,11 @@ impl<N: Network> BlockSync<N> {
             }
             // Update the latest height.
             current_height = self.canon.latest_block_height();
+            // Bump the timestamps of all outstanding requests if we successfully handled a BlockResponse.
+            let mut timestamps = self.request_timestamps.write();
+            for (_height, instant) in timestamps.iter_mut() {
+                *instant = Instant::now();
+            }
         }
     }
 }
--- a/node/sync/src/block_sync.rs
+++ b/node/sync/src/block_sync.rs
@@ -320,9 +320,10 @@ impl<N: Network> BlockSync<N> {
             // Update the latest height.
             current_height = self.canon.latest_block_height();
             // Bump the timestamps of all outstanding requests if we successfully handled a BlockResponse.
+            let now = Instant::now();
             let mut timestamps = self.request_timestamps.write();
-            for (_height, instant) in timestamps.iter_mut() {
-                *instant = Instant::now();
+            for instant in timestamps.values_mut() {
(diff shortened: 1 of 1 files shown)


node/sync/src/block_sync.rs
```diff
@@ -319,6 +319,11 @@ impl<N: Network> BlockSync<N> {
             }
             // Update the latest height.
             current_height = self.canon.latest_block_height();
+            // Bump the timestamps of all outstanding requests if we successfully handled a BlockResponse.
+            let mut timestamps = self.request_timestamps.write();
+            for (_height, instant) in timestamps.iter_mut() {
+                *instant = Instant::now();
```
```suggestion
            let new_timestamp = Instant::now();
            let mut timestamps = self.request_timestamps.write();
            for (_height, instant) in timestamps.iter_mut() {
                *instant = new_timestamp;
```
for a small perf boost

node/sync/src/block_sync.rs
```diff
@@ -319,6 +319,11 @@ impl<N: Network> BlockSync<N> {
             }
             // Update the latest height.
             current_height = self.canon.latest_block_height();
+            // Bump the timestamps of all outstanding requests if we successfully handled a BlockResponse.
+            let mut timestamps = self.request_timestamps.write();
+            for (_height, instant) in timestamps.iter_mut() {
```
```suggestion
            for instant in timestamps.values_mut() {
```

node/sync/src/block_sync.rs
```diff
@@ -319,6 +319,12 @@ impl<N: Network> BlockSync<N> {
             }
             // Update the latest height.
             current_height = self.canon.latest_block_height();
+            // Bump the timestamps of all outstanding requests if we successfully handled a BlockResponse.
+            let mut timestamps = self.request_timestamps.write();
+            let now = Instant::now();
```
ljedrz, line 323: it'd be best to obtain the time before accessing the lock (as it involves a system call and would prolong the access)

joske: Indeed! Note to self: don't work on PRs on Sunday evening :sweat_smile:

raychu86:
Closing. Superseded by #3079

PR #2999, opened 2024-01-13
Limit the number of transactions and limit deployment priority in the mempool

vicsn:
## Motivation

It is unfortunate that we can't take speculative execution fees from deployments at the moment, so as a **temporary fix** this should prevent DoS attacks.

Setting the priority_fee to 0 means that in theory, there could be a situation where deployments will never come to the front of the queue. But then we're in a scenario where we are bombarded with fee-paying executions, which means the community has enough fun with the existing deployments. Alternatively, we could create a separate `deployments_queue` [where we can always pop some off](https://github.com/AleoHQ/snarkOS/blob/priority_fee/node/consensus/src/lib.rs#L338).


node/consensus/src/lib.rs
```diff
@@ -263,14 +263,23 @@ impl<N: Network> Consensus<N> {
             if self.ledger.contains_transmission(&TransmissionID::from(&transaction_id))? {
                 bail!("Transaction '{}' exists in the ledger {}", fmt_id(transaction_id), "(skipping)".dimmed());
             }
-            // Check if the queue is full.
-            if std::mem::size_of_val(&self.transactions_queue.lock()) >= MAX_TX_QUEUE_SIZE {
-                // If the queue is full, return early.
+            // Check the size of the transctions queue.
+            let transactions_queue_size = std::mem::size_of_val(&self.transactions_queue.lock());
```
```suggestion
            let transactions_queue_size = self.transactions_queue.lock().len() * (std::mem::size_of::<TransactionKey<N>>() + std::mem::size_of::<Transaction<N>>());
```
this would be more accurate (the current value won't change regardless of length), but still not perfect; I'll provide a precise calculation tomorrow

vicsn: Took a stab: 6cac104c0

node/consensus/src/lib.rs
```diff
@@ -74,6 +75,8 @@ pub struct Consensus<N: Network> {
     solutions_queue: Arc<Mutex<IndexMap<PuzzleCommitment<N>, ProverSolution<N>>>>,
     /// The unconfirmed transactions queue.
     transactions_queue: Arc<Mutex<BTreeMap<TransactionKey<N>, Transaction<N>>>>,
+    /// Size of transactions queue.
+    transactions_queue_size: Arc<Mutex<TransactionSize>>,
```
ljedrz: this should be an `Arc<AtomicI64>`; it can be another type alias (e.g. `AtomicTransactionSize`) if practical

vicsn: 753d6d0c5

node/consensus/src/lib.rs
```diff
@@ -316,32 +340,38 @@ impl<N: Network> Consensus<N> {
         priority_fee: PriorityFee<N>,
         id: N::TransactionID,
         transaction: Transaction<N>,
+        size: TransactionSize,
```
ljedrz, line 352:
```suggestion
        new_queue_size: TransactionSize,
```

node/consensus/src/lib.rs
```diff
@@ -316,32 +340,38 @@ impl<N: Network> Consensus<N> {
...
     ) -> Result<()> {
         // Get the current timestamp.
         let timestamp = now_nanos();
         // Add the Transaction to the transactions_in_queue.
         let inserted = self.transactions_in_queue.lock().insert(id);
         ensure!(inserted, "Transaction '{}' exists in the queue", fmt_id(id));
+        // Add the size to the tracking transactions_queue_size.
+        self.transactions_queue_size.fetch_update(Relaxed, Relaxed, |x| Some(x.saturating_add(size))).map_err(error)?;
```
ljedrz, line 360:
```suggestion
        self.transactions_queue_size.store(new_queue_size, Relaxed);
```

node/consensus/src/lib.rs
ljedrz: either that, or https://github.com/AleoHQ/snarkOS/pull/2999/files#diff-8b822371aae41d2275688c666d4c5d5d304e67aa422628ee42c405b5e8d7f572R306 needs adjustment

node/consensus/src/lib.rs
```diff
@@ -316,32 +340,38 @@ impl<N: Network> Consensus<N> {
...
         // Acquire the lock on the queue.
         let mut queue = self.transactions_queue.lock();
         // Determine the number of transactions to send.
         let num_transactions = queue.len().min(capacity);
         // Pop the transactions from the queue.
         let transactions = (0..num_transactions).map(|_| queue.pop_last().unwrap()).collect::<Vec<_>>();
-        // Remove the transactions from the queue index and mark them again freshly as seen.
-        for (_, tx) in transactions.iter() {
+        // Remove the transactions from the queue index, size tracker and mark them again freshly as seen.
+        for ((_, _, negative_size, _), tx) in transactions.iter() {
```
ljedrz: why use negative size here instead of subtracting?

vicsn: Because we're storing a negative size, because I want to prioritize ordering smaller transactions as extra DoS protection. It is very ugly, but it makes sense, and is the same ugliness as we already have by storing negative timestamps... Ideas are welcome.

ljedrz: if we just want to reverse the sorting order we can use [`cmp::Reverse`](https://doc.rust-lang.org/std/cmp/struct.Reverse.html) instead; it conveys the rationale better IMO

vicsn: TIL, brilliant fee6b4666

node/consensus/src/lib.rs
vicsn: My bad I was passing the wrong value: 7cd6dce96
We need to pass the individual tx size so we can order the transaction by it

node/consensus/src/lib.rs
vicsn: See comment above

vicsn:
Closing due to merging of priority fees

PR #3000, opened 2024-01-15
Store a bitset of transmissions in compact subdag.

vicsn:
## Motivation

Subdags can include duplicate transmission_ids which scales very badly at the moment (30MB w/ 50 validators). In this PRs, Subdags can track transmission_ids using a succinct bitset.

New types:
- CompactCertificate
- CompactHeader
- Subdag enum can hold either
- NarwhalCertificate trait provides an interface
- We use BitSet, which used to be part of standard library

## Test Plan

- [x] Unit tests pass.
- [x] Local network passes.
- [ ] Integration with snarkOS on existing testnet passes. Or we just go for mainnet integration.

## Related PRs

https://github.com/AleoHQ/snarkVM/pull/2304

### DIFF
--- a/node/bft/src/bft.rs
+++ b/node/bft/src/bft.rs
@@ -506,6 +506,8 @@ impl<N: Network> BFT<N> {
         };
         // Initialize a map for the deduped transmissions.
         let mut transmissions = IndexMap::new();
+        // Initialize a map for the deduped prior transmission ids.
+        let mut prior_included_transmissions = IndexSet::new();
         // Start from the oldest leader certificate.
         for certificate in commit_subdag.values().flatten() {
             // Update the DAG.
@@ -518,11 +520,6 @@ impl<N: Network> BFT<N> {
                 if transmissions.contains_key(transmission_id) {
                     continue;
                 }
-                // If the transmission already exists in the ledger, skip it.
-                // Note: On failure to read from the ledger, we skip including this transmission, out of safety.
-                if self.ledger().contains_transmission(transmission_id).unwrap_or(true) {
-                    continue;
-                }
                 // Retrieve the transmission.
                 let Some(transmission) = self.storage().get_transmission(*transmission_id) else {
                     bail!(
@@ -531,8 +528,14 @@ impl<N: Network> BFT<N> {
                         certificate.round()
                     );
                 };
-                // Add the transmission to the set.
-                transmissions.insert(*transmission_id, transmission);
+                // TODO: this also means that they won't necessarily be prior_included_transactions, and it might fail other checks... What do? Repeated this check until we succeed, or make the verify block check way more expensive.
+                // Add the transmission to the set of prior or new transmissions.
+                // Note: On failure to read from the ledger, we skip including this transmission, out of safety.
+                if self.ledger().contains_transmission(transmission_id).unwrap_or(true) {
+                    prior_included_transmissions.insert(*transmission_id);
+                } else {
+                    transmissions.insert(*transmission_id, transmission);
+                }
             }
         }
         // If the node is not syncing, trigger consensus, as this will build a new block for the ledger.
@@ -557,7 +560,10 @@ impl<N: Network> BFT<N> {
                 // Initialize a callback sender and receiver.
                 let (callback_sender, callback_receiver) = oneshot::channel();
                 // Send the subdag and transmissions to consensus.
-                consensus_sender.tx_consensus_subdag.send((subdag, transmissions, callback_sender)).await?;
+                consensus_sender
+                    .tx_consensus_subdag
+                    .send((subdag, transmissions, prior_included_transmissions, callback_sender))
+                    .await?;
                 // Await the callback to continue.
                 match callback_receiver.await {
                     Ok(Ok(())) => (), // continue
@@ -573,7 +579,7 @@ impl<N: Network> BFT<N> {
             }
 
             info!(
-                "\n\nCommitting a subdag from round {anchor_round} with {num_transmissions} transmissions: {subdag_metadata:?}\n"
+                "\n\nCommitting a subdag from round {anchor_round} with {num_transmissions} new transmissions: {subdag_metadata:?}\n"
             );
             // Update the DAG, as the subdag was successfully included into a block.
             let mut dag_write = self.dag.write();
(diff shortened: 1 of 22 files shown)


node/bft/src/bft.rs
```diff
@@ -518,33 +522,45 @@ impl<N: Network> BFT<N> {
...
                 }
-                // If the transmission already exists in the ledger, skip it.
-                // Note: On failure to read from the ledger, we skip including this transmission, out of safety.
-                if self.ledger().contains_transmission(transmission_id).unwrap_or(true) {
-                    continue;
+                // Check if the ledger contains the transmission already.
+                match self.ledger().contains_transmission(transmission_id) {
+                    // On failure to read from the ledger, we skip including this transmission, out of safety.
+                    Err(err) => {
+                        warn!("{}", err.to_string());
```
ljedrz:
```suggestion
                        warn!("{}", err);
```
the `Display` impl (giving us `to_string`) should suffice

node/bft/src/bft.rs
```diff
@@ -518,33 +522,45 @@ impl<N: Network> BFT<N> {
...
+                        aborted_transmissions.insert(*transmission_id);
+                        continue;
```
ljedrz:
```suggestion
```
(doesn't affect the logic)

node/bft/src/bft.rs
```diff
@@ -518,33 +522,45 @@ impl<N: Network> BFT<N> {
...
+                    }
+                    // If the transmission already exists in the ledger, save just the transmission ID.
+                    Ok(true) => {
+                        prior_included_transmissions.insert(*transmission_id);
+                        continue;
```
ljedrz:
```suggestion
```
ditto

node/bft/src/bft.rs
```diff
@@ -557,7 +573,10 @@ impl<N: Network> BFT<N> {
                 // Initialize a callback sender and receiver.
                 let (callback_sender, callback_receiver) = oneshot::channel();
                 // Send the subdag and transmissions to consensus.
-                consensus_sender.tx_consensus_subdag.send((subdag, transmissions, callback_sender)).await?;
+                consensus_sender
+                    .tx_consensus_subdag
+                    .send((subdag, transmissions, prior_included_transmissions, aborted_transmissions, callback_sender))
```
ljedrz: I wonder if a helper `struct` to hold all these would make sense yet

node/bft/src/sync/mod.rs
```diff
@@ -203,48 +207,40 @@ impl<N: Network> Sync<N> {
...
-                    // If the block authority is a subdag, then retrieve the certificates.
-                    Authority::Quorum(subdag) => {
-                        Some((subdag.leader_certificate().clone(), subdag.election_certificate_ids().clone()))
-                    }
-                }
-            })
-            .collect::<Vec<_>>();
-        if leader_certificates.is_empty() {
+        // Check if the blocks contain a quorum authority, and by extension a leader certificate.
+        let contains_leader_certificate = blocks.iter().any(|block| block.authority().is_quorum());
```
ljedrz: can't we register this while we're iterating over the blocks in L208?

node/bft/src/sync/mod.rs
```diff
@@ -203,48 +207,40 @@ impl<N: Network> Sync<N> {
...
+        if !contains_leader_certificate {
             return Ok(());
         }
 
         // Construct a list of the certificates.
-        let certificates = blocks
+        let mut leader_certificates = Vec::with_capacity(blocks.len());
+        let num_certificates = blocks
```
ljedrz: ditto

node/bft/src/sync/mod.rs
```diff
@@ -203,48 +207,40 @@ impl<N: Network> Sync<N> {
...
-                }
-            })
-            .flatten()
-            .collect::<Vec<_>>();
+            .map(|block| if let Authority::Quorum(subdag) = block.authority() { subdag.num_certificates() } else { 0 })
+            .sum::<usize>();
+        let mut certificates = Vec::with_capacity(num_certificates);
+        for block in blocks {
+            // If the block authority is a quorum subdag, then retrieve the certificates.
+            if block.authority().is_quorum() {
```
ljedrz: ditto (especially since we're checking for quorum again)

ljedrz (commented):
Left a few comments; seems like we need a rebase, and there are some failing CI jobs.

node/bft/src/bft.rs
vicsn: 2368bad08

node/bft/src/bft.rs
vicsn: 6f5144ec1

node/bft/src/bft.rs
vicsn: 6f5144ec1

node/bft/src/bft.rs
vicsn: Definitely: 687d56779

There is still a lot of duplication inside of snarkVM's `fn prepare_advance_to_next_quorum_block`. https://github.com/AleoHQ/snarkVM/pull/2304/commits/c349c71d51ac0800f6f355b79d2f36a977c622f4

@raychu86 just a heads up, if and when either of us adds `aborted_solution_ids` we can think of a proper abstraction.

node/bft/src/sync/mod.rs
vicsn: Yes you're right: de9cc18c0

Note that I'm still iterating first to determine the total number of certificates for vector preallocation. In theory we could `collect` all the needed components through a single iterator but that becomes quite messy to program and read.

node/bft/src/sync/mod.rs
vicsn: See comment above.

node/bft/src/sync/mod.rs
vicsn: See comment above.

node/bft/src/sync/mod.rs
ljedrz: I agree, making iteration fully functional can make things less readable :+1:

howardwu:
Closing, was not in scope.

PR #3002, opened 2024-01-15
Filter peer response for requester's IP before sending it

kpandl:
## Motivation

- Fix for the following issue: https://github.com/AleoHQ/snarkOS/issues/2837
- Previously, before sending a peer response after a peer request, snarkOS doesn't filter the peers for the requester's IP
- By not sending the requester's IP in the response, the requester node will not attempt to connect to itself, which frees up resources

## Test Plan

- Ran devnets with validators and clients locally and on AWS. Logged peerresponses through the kp/fix/stress_test_6 branch

## Related PRs

(Link any related PRs here)


ljedrz (approved):
LGTM :+1:

howardwu (approved):

PR #3012, opened 2024-01-17
[fix] Add condition to avoid invalid sync logic for validators

raychu86:
## Motivation

This PR is a quick attempt at addressing a possible condition for causing forks in the network due to syncing issues.

The case observed is as follows:

1. A validator is slightly behind it's peers and needs to sync.
2. The validator sends block requests and receives responses (which are stored in the `BlockSync` module).
3. The validator calls `try_block_sync` on a heartbeat and occasionally performs the sync process of the router before the standard validator sync process of handling the certificates.
4. This updates the validator's ledger, but not the BFT state. (It skips a subdag commit for that block)
5. In the next block, it will commit additional certificates that should have been committed in the previous round, thus leading to a fork.


howardwu (approved):

node/sync/src/block_sync.rs
```diff
@@ -220,7 +220,8 @@ impl<N: Network> BlockSync<N> {
 
         // If there are no block requests, but there are pending block responses in the sync pool,
         // then try to advance the ledger using these pending block responses.
-        if block_requests.is_empty() && !self.responses.read().is_empty() {
+        // This is only necessary in router mode because the gateway must sync via the certificate approach.
```
```suggestion
        // Note: This condition is guarded by `mode.is_router()` because validators sync blocks
        // using another code path that updates both `storage` and `ledger` when advancing blocks.
```

PR #3015, opened 2024-01-17
[HackerOne-2279770] Prevent handshake based signature forgery

### LINKED ISSUE #2983: [Bug] Validator sign arbitrary nonce can lead to downgraded length of secure bit.
# https://hackerone.com/reports/2279770

## Summary:

During handshake, validator directly sign nonce (u64) sent from counterparty:

```
let Ok(our_signature) = self.account.sign_bytes(&peer_request.nonce.to_le_bytes(), rng)
```

The attacker can exploit this to let validator to sign the message they want.

For example, the attacker can brute force and try to find some transaction `hash` that satify `hash < u64.max`. Then let the validator sign this hash as `peer_request.nonce`. In this case the length of secure bit is downgraded from `252/2 = 126` bits to `(252-64)/2 = 94` bits. 

## Proof-of-Concept (PoC)

As described above.

## Fix Suggestions:

Instead of directly sign the nonce, the validator can sign something like `hash('validator_handshake_nonce', nonce)`. Also, consider adding prefix string when generating `batch_id` and `block_hash` to avoid future schema conflict.

niklaslong:
Fixes #2983.

Intuition: 
1. an attacker could locally craft a preimage that matches the `ChallengeRequest::nonce` value range. 
2. the target then signs the nonce during the handshake.
3. the attacker is able to reuse the signed nonce in any other non-handshake interactions as an authenticity proof.

In order to mitigate this issue, this PR removes the possibility of brute forcing the preimage locally by generating a `response_nonce` that is signed alongside the initial `nonce` and sent in the `ChallengeResponse` so the signature can be verified.

### DIFF
--- a/node/bft/events/src/challenge_response.rs
+++ b/node/bft/events/src/challenge_response.rs
@@ -17,6 +17,7 @@ use super::*;
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct ChallengeResponse<N: Network> {
     pub signature: Data<Signature<N>>,
+    pub nonce: u64,
 }
 
 impl<N: Network> EventTrait for ChallengeResponse<N> {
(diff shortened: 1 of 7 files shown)


raychu86:
Note: This will require a version bump for the `Event` and `Message` types, because this will break the current handshake protocol

ljedrz (commented):
LGTM with a version bump as suggested by @raychu86.

ljedrz (approved):

howardwu:
Targeting this to `mainnet` branch given it contains breaking changes

howardwu (approved):

PR #3017, opened 2024-01-18
More fine-grained CDN sync

ljedrz:
The current future-stream-based approach is pretty concise, but there are some issues with it:
- adjusting its logic is tricky
- error-handling is inconvenient
- it's hard to keep track of the number of pending blocks and predict the buffering behavior in general
- the backoff logic is more complicated than it needs to be
- despite best efforts, a shutdown may still cause storage issues
- some downloads can be very slow due to competition for shared objects

In this PR I propose that we instead switch to a simpler approach where we spawn a background task responsible for making concurrent requests to the CDN, and we can control both the number of desired concurrent requests and the number of pending blocks that we keep in memory.

In my local runs, it performs at least as fast as the current code, while providing us with a much greater degree of control over and insight into the progress. In addition, despite many restarts I haven't run into any storage failures.


node/cdn/src/blocks.rs
```diff
@@ -138,143 +146,192 @@ pub async fn load_blocks<N: Network>(
...
+                                debug!(
+                                    "Received {ctx} in {:.2?} ({} queued for insertion)",
+                                    request_time.elapsed(),
+                                    pending_blocks.len()
+                                );
+                                break;
+                            }
+                            Err(error) => {
+                                // Increment the attempt counter, and wait with a linear backoff.
+                                attempts += 1;
```
joske: should we have some max attempts?

ljedrz: happy to impose some limit; @howardwu would you like to suggest one?

ljedrz:
Rebased and applied a bit of refactoring to make the changes easier to digest.

node/cdn/src/blocks.rs
ljedrz: I introduced a limit of 10 attempts for now; if it is breached, the sync gets aborted.

ljedrz:
With https://github.com/AleoHQ/snarkVM/pull/2322 applied I've almost managed to sync to the tip; tomorrow I'll be able to confirm that I've tested all possible sync scenarios using this PR.

ljedrz:
I was able to sync to the tip :+1:, and at no time during my syncing have I encountered any storage issues despite many restarts.

howardwu (approved):

howardwu:
@ljedrz can you confirm that the interfaces do not have any changes for the block height?

I see that the tests needed to be updated to reflect new range changes. This will cause breakages downstream for the Explorer, faucet, and all developers using the CDN to date.

Issue #3030, opened 2024-01-23
[Bug]


nomadnft69:
## Your Environment

- 
- 
- 
issue when 

trying to isstall node.

For more information about this error, try `rustc --explain E0308`.
error: could not compile `snarkvm-ledger-store` (lib) due to previous error
error: failed to compile `snarkos v2.2.7 (/root/snarkOS)`, intermediate artifacts can be found at `/root/snarkOS/target`.
To reuse those artifacts with a future compilation, set the environment variable `CARGO_TARGET_DIR` to that path.
root@vmi1617775:~/snarkOS# snarkos account new

joske:
another dup of #3023

Issue #3023, opened 2024-01-22
[Bug]  compiling snarkvm-ledger-store fails with error snarkvm-ledger-store


ranjithrajaram:
## 🐛 Bug Report

Executing ./build_ubuntu.sh fails with the following error. I did not find a similar report, so filing a new one

```
error[E0308]: mismatched types
   --> /root/.cargo/registry/src/index.crates.io-6f17d22bba15001f/snarkvm-ledger-store-0.16.11/src/helpers/rocksdb/internal/mod.rs:113:69
    |
113 |                 let primary = aleo_std::aleo_ledger_dir(network_id, dev);
    |                               -------------------------             ^^^ expected `StorageMode`, found `Option<u16>`
    |                               |
    |                               arguments to this function are incorrect
    |
    = note: expected enum `StorageMode`
               found enum `std::option::Option<u16>`
note: function defined here
   --> /root/.cargo/registry/src/index.crates.io-6f17d22bba15001f/aleo-std-storage-0.1.7/src/lib.rs:89:8
    |
89  | pub fn aleo_ledger_dir(network: u16, mode: StorageMode) -> PathBuf {
    |        ^^^^^^^^^^^^^^^
help: call `Into::into` on this expression to convert `std::option::Option<u16>` into `StorageMode`
    |
113 |                 let primary = aleo_std::aleo_ledger_dir(network_id, dev.into());
    |                                                                        +++++++
```

## Steps to Reproduce

git clone https://github.com/AleoHQ/snarkOS.git --depth 1
cd snarkOS
./build_ubuntu.sh

## Expected Behavior

Compilation should work successfully

## Your Environment

-  v2.2.5 
-    rustc 1.75.0
-  Ubuntu 22.04.3 LTS

ljedrz:
It's due to `aleo-std` just having received some breaking changes and a version bump. I'm pretty sure snarkOS will be adjusted to it shortly, but in the meantime you can fix it locally by tweaking the contents of `node/Cargo.toml` and `cli/Cargo.toml` as follows: just find
```
[dependencies.aleo-std]
version = "0.1.18"
```
and change it to
```
[dependencies.aleo-std]
version = "=0.1.18"
```

nomadnft69:
[...]

PR #3031, opened 2024-01-23
Fix/compile

joske:
Recent API changes in aleo-std break snarkOS compilation: https://github.com/AleoHQ/snarkOS/issues/3028, #3023 #3030 

This fixes compilation, but may not be the correct way to fix this.
- bumped snarkVM to a commit that fixes the API change: 06bd41f
- bumped aleo-std to 0.1.24
- use `StorageMode` based on `dev` parameter

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -75,9 +75,9 @@ dependencies = [
 
 [[package]]
 name = "aleo-std"
-version = "0.1.18"
+version = "0.1.24"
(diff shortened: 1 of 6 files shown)


howardwu:
@ljedrz can you please fix this PR to support custom paths? You have a PR for this already but it needs to be updated.

https://github.com/AleoHQ/snarkOS/pull/2872

howardwu (approved):
Merging to resolve compilation issues. We still need to properly update support.

PR #3035, opened 2024-01-25
chore: Corrected a spelling error

keienWang:
The reason for the spelling error "offseted" is that in English, the past participle form of the verb "offset" is "offset," not "offseted." The past participle form is typically used to indicate an action or state that has already occurred. In this context, we want to convey that the index has already been offset, so using "offset" is more in line with standard English grammar. "Offseted" is not a common or correct spelling, hence the correction to "offset."


howardwu (approved):

keienWang:
@all-contributors please add @keienWang  for design

PR #3036, opened 2024-01-26
Stop processing block sync when ctrlc

elderhammer:
## Motivation

This PR improves the synchronization logic of CDN: when the program receives the sigint signal (ctrl-c), the CDN synchronization process of the block is stopped in time to avoid data inconsistency and damage.

## Related PRs

https://github.com/AleoHQ/snarkOS/issues/2951


ljedrz:
This is already done in https://github.com/AleoHQ/snarkOS/pull/3017.

PR #3041, opened 2024-01-28
Created Install Script for Debian

fabrinasr:
## Motivation
I looked at the Ubuntu script created by Howard and I made one main modification to make it work for Debian. In the Debian repos the LLVM and Clang packages are old. To solve this issue I grabbed the LLVM and Clang packages directly from the LLVM repo. For the reviewer checking this... the original source for the LLVM part of the script is below. 

Source: https://apt.llvm.org/llvm.sh

I made some minor modifications to the above script and Howard's Ubuntu Script to create a working Install Script for Debian.

I also added Debian Install instructions to the Docs and I added a Beta tag next to the Debian instructions in the Docs for clarity. 
## Test Plan
I have tested the script multiple times in Debian with no errors. I purged all installed packages and re-ran the script multiple times and got almost exactly the same output that you get when you run the Ubuntu Script in an Ubuntu OS. 

## Related PRs

(Link any related PRs here)


joske:
Hi, thx for taking the time to make this. However, this is based on an older version of the ubuntu script. It's better form to only require root for the actions that need it (installing packages, firewall rules), and not for the build. This was updated a few days ago for the ubuntu script. Note that the `ufw` commands were removed, and it's left up to the user to make sure these ports are open.

As a general comment, I feel this script tries to do too much. I understand you need to fetch the LLVM packages as debian repos are always hopelessly outdated, but does this script really need to support all kinds of ubuntu and debian variants?

So please rebase your branch to bring it in line with the current script. I'll also add some comments on specific lines.

build_debian.sh
```diff
@@ -0,0 +1,216 @@
...
+
+
+# install everything
+
+if [[ ! -f /etc/apt/trusted.gpg.d/apt.llvm.org.asc ]]; then
+    # download GPG key once
+    wget -qO- https://apt.llvm.org/llvm-snapshot.gpg.key | tee /etc/apt/trusted.gpg.d/apt.llvm.org.asc
+fi
+
+if [[ -z "`apt-key list 2> /dev/null | grep -i llvm`" ]]; then
```
joske: backticks are discouraged these days, please replace with the more portable `$(...)` notation

build_debian.sh
```diff
@@ -0,0 +1,216 @@
...
+      if [ "$2" == "all" ]; then
+          # Install all packages
+          ALL=1
+          OPTIND=3
+      fi
+    fi
+fi
+
+while getopts ":hm:n:" arg; do
+    case $arg in
```
joske: should add a `*)` case to handle invalid flags.

build_debian.sh
```diff
@@ -0,0 +1,216 @@
...
+            LINKNAME=
+        else
+            LINKNAME=-${CODENAME}
+        fi
+        CODENAME_FROM_ARGUMENTS="true"
+        ;;
+    esac
+done
+
+if [[ $EUID -ne 0 ]]; then
```
joske: use sudo only on the commands that need it, don't require the entire script to be ran as root.

build_debian.sh
```diff
@@ -0,0 +1,216 @@
...
+    ufw
+
+# Install Rust
+
+curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
+source $HOME/.cargo/env
+
+# Install snarkOS
+# cargo clean
+cargo install --path .
```
joske: should use `cargo install --locked ...` as current build_ubuntu.sh. This is because `cargo install` by default *doesn't* use the lock file, and many users are reporting problems with breaking deps, so please use shipped `Cargo.lock`.

build_debian.sh
```diff
@@ -0,0 +1,216 @@
...
+echo "             are enabled on your local network."
+echo ""
+echo " Cloud Providers - Enable ports 4133 and 3033"
+echo "                   in your network firewall"
+echo ""
+echo " Home Users - Enable port forwarding or NAT rules"
+echo "              for 4133 and 3033 on your router."
+echo "=================================================="
+
+# Open ports on system
```
joske: please remove the `ufw` commands.

build_debian.sh
```diff
@@ -0,0 +1,216 @@
...
+done
+if [[ ${#missing_binaries[@]} -gt 0 ]] ; then
+    echo "You are missing some tools this script requires: ${missing_binaries[@]}"
+    echo "(hint: apt install lsb-release wget software-properties-common gnupg)"
+    exit 4
+fi
+
+# Set default values for commandline arguments
+# We default to the current stable branch of LLVM
+LLVM_VERSION=$CURRENT_LLVM_STABLE
```
ljedrz: do we need all this setup just to download `LLVM` and `clang`? I don't use Debian, but wouldn't it be possible just with `apt-get install`?

joske: the problem is these packages are (of course) outdated in debian.

fabrinasr:
I found a better solution for the best way to integrate Debian.

I think it is a mistake to create one script for Debian like we have for Ubuntu.

We should instead have a two step install process

Step 1) run the official script provided by the LLVM organisation to install updated version of Clang and LLVM for Debian

Step 2) run the SnarkOS Install script for Debian

The SnarkOS install script for Debian will be almost identical to the updated install script for Ubuntu,
The only difference will be that "apt-get clang" and "apt-get llvm" will be removed to avoid the installation of the outdated
versions of these packages in Debian. 


This is in my opinion the best solution for three reasons:

1) we use the official install script provided by the LLVM organisation for Debian, this has already been
 tested by them, I believe there is no need to make changes to this. It is already working and stable.

2) we can avoid running the SnarkOS install script as root which is ideal. 

3) In future, if we want to edit the Debian SnarkOS install script we will have a simple and clean file to edit,
it makes life easier for devs. The Ubuntu and Debian install scripts will look almost identical. 

I have made the following changes 

1) Grabbed the official install script from LLVM for Debian and added to SnarkOS repo (Source: https://apt.llvm.org/llvm.sh)
2) Modified Ubuntu Install script for use in Debian
3) Edited readme to include install instructions for Debian

joske:
Yes this seems like a better approach. I'm not sure we should bundle `llvm.sh` with snarkOS though.

fabrinasr:
Yes, including a link to the llvm script in the readme is a better approach than bundling it. I removed the llvm script file.

README.md
```diff
@@ -94,6 +94,27 @@ cargo install --locked --path .
...
+A helper script to install dependencies is available. From the `snarkOS` directory, run:
+```
+./build_debian.sh
+```
+
+Lastly, install `snarkOS`:
+```
+cargo install --locked --path .
+```
+
```
ljedrz:
```suggestion
```
This is already contained in the script, so no need for it; can you also remove it in the Ubuntu docs right above?

build_debian.sh
```diff
@@ -0,0 +1,41 @@
...
+	build-essential \
+	curl \
+	gcc \
+	libssl-dev \
+	make \
+	pkg-config \
+	tmux \
+	xz-utils \
+	ufw
+
```
ljedrz:
```suggestion
```

README.md
```diff
@@ -94,6 +94,27 @@ cargo install --locked --path .
 
 Please ensure ports `4133/tcp` and `3033/tcp` are open on your router and OS firewall.
 
+**[*Beta* - For Debian users]**
```
ljedrz:
```suggestion
**[For Debian users]** 
```

README.md
```diff
@@ -94,6 +94,27 @@ cargo install --locked --path .
 
 Please ensure ports `4133/tcp` and `3033/tcp` are open on your router and OS firewall.
 
+**[*Beta* - For Debian users]** 
+
+Firstly, install the updated version of LLVM and Clang.
```
ljedrz:
```suggestion
Firstly, install the most recent versions of LLVM and Clang.
```

README.md
fabrinasr: I removed those lines from ubuntu and debian instructions.

README.md
```diff
@@ -87,13 +87,24 @@ cd snarkOS
 ./build_ubuntu.sh
 ```
 
-Lastly, install `snarkOS`:
+Please ensure ports `4133/tcp` and `3033/tcp` are open on your router and OS firewall.
```
ljedrz: why don't we move this right after the `cd snarkOS` step, so we don't duplicate it for the scripts?

fabrinasr: done

ljedrz (dismissed):
LGTM, with the caveat that I don't currently have a Debian environment to test it in.

joske:
The file `build_debian.sh` should be committed with 755 permissions.

build_debian.sh
```diff
@@ -0,0 +1,40 @@
...
+sudo apt-get update
+sudo apt-get install -y \
+	build-essential \
+	curl \
+	gcc \
+	libssl-dev \
+	make \
+	pkg-config \
+	tmux \
+	xz-utils \
```
joske, line 19: add `sudo`, it's not installed in the default debian docker images. It's also missing from ubuntu docker images, so please add it in `build_ubuntu.sh` too.

joske (approved):
LGTM now

joske:
The decision to merge is not mine. @Howardwu will merge when appropriate.
This is nice to have but not necessary for mainnet launch, we are focussing
our efforts on mandatory changes ATM.

fabrinasr:
I understand. I was thinking about it from the point of view of the people running Aleo Provers. I think Debian would be a better option for them. In my experience it is more stable than Ubuntu. This PR also made some minor improvements to the Ubuntu script. But I understand there might be bigger priorities right now for @howardwu

fabrinasr:
@ljedrz 

I re-added the instructions "cargo install --locked --path...." to the readme. It was originally removed [here](https://github.com/AleoHQ/snarkOS/pull/3041#discussion_r1497682708).

It is still needed, see [here](https://github.com/AleoHQ/snarkOS/pull/3131#issuecomment-1974912139) from @howardwu

build_debian.sh
```diff
@@ -0,0 +1,41 @@
...
+	build-essential \
+	curl \
+	gcc \
+	libssl-dev \
+	make \
+	pkg-config \
+        sudo \
+	tmux \
+	xz-utils \
+	ufw
```
HarukaMa: As a debian user and ubuntu hater I really don't like ufw. As the script is not using it as well, probably we should leave that to users to decide how to do port forwarding.

build_debian.sh
```diff
@@ -0,0 +1,41 @@
...
+
+curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
+source $HOME/.cargo/env
+
+# Install snarkOS
+# cargo clean
+cargo install --locked --path .
+
+echo "=================================================="
+echo " Attention - Please ensure ports 4133 and 3033"
```
HarukaMa: We should update the ports as well as we are using 4130 and 3030 for all networks. (Should we consider using different ports for different networks?)

build_debian.sh
fabrinasr: removed ufw

build_debian.sh
fabrinasr: I changed ports.

fabrinasr:
@zosorock 

I think this is good to go. It does not block mainnet - see label. It is not urgent but still very useful for new Aleo users. We've had a few users [asking for debian](https://github.com/AleoNet/snarkOS/issues/3255)

zosorock:
@fabrinasr can you re-target to the `staging` branch please?

README.md
```diff
@@ -82,17 +82,31 @@ Next, move into the `snarkOS` directory:
 cd snarkOS
 ```
 
+Lastly, install `snarkOS`:
+```
+cargo install --locked --path .
+```
+
+Please ensure ports `4133/tcp` and `3033/tcp` are open on your router and OS firewall.
```
vicsn, line 90: Let's get rid of port recommendations. Also these are not even the defaults? 🤔

PR #3044, opened 2024-01-29
Customizable ledger location

ljedrz:
Supersedes https://github.com/AleoHQ/snarkOS/pull/2872.

Filing as a draft, as it depends on https://github.com/AleoHQ/snarkVM/pull/2326 and will fail to compile without it in place.

### DIFF
--- a/cli/src/commands/start.rs
+++ b/cli/src/commands/start.rs
@@ -12,7 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use aleo_std::StorageMode;
 use snarkos_account::Account;
 use snarkos_display::Display;
 use snarkos_node::{bft::MEMORY_POOL_PORT, router::messages::NodeType, Node};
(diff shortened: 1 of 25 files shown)


ljedrz:
I updated the snarkVM dep to the tip of `testnet3`; the PR may now be merged as-is, or it can wait for the next release of snarkVM.

joske (approved):

cli/src/commands/start.rs
```diff
@@ -566,7 +583,7 @@ fn load_or_compute_genesis<N: Network>(
     /* Otherwise, compute the genesis block and store it. */
 
     // Initialize a new VM.
-    let vm = VM::from(ConsensusStore::<N, ConsensusMemory<N>>::open(None)?)?;
+    let vm = VM::from(ConsensusStore::<N, ConsensusMemory<N>>::open(storage_mode)?)?;
```
raychu86: Is passing `storage_mode` through necessary here? We are using ConsensusMemory, so any StorageMode will have the same result.

howardwu (approved):

cli/src/commands/start.rs
howardwu: Resolved [2b1b881](https://github.com/AleoHQ/snarkOS/pull/3044/commits/2b1b881e9ec9bd5839cdab33f45ce744f70818a6)

PR #3045, opened 2024-01-29
Ensure backwards compatibility of load_blocks

### LINKED ISSUE #3043: CDN range check
@ljedrz can you confirm that the interfaces do not have any changes for the block height?

I see that the tests needed to be updated to reflect new range changes. This will cause breakages downstream for the Explorer, faucet, and all developers using the CDN to date.

_Originally posted by @howardwu in https://github.com/AleoHQ/snarkOS/issues/3017#issuecomment-1913922168_

ljedrz:
This PR special-cases calls to `load_blocks` made outside of the context of a node, so that the function maintains backwards compatibility with the logic from before https://github.com/AleoHQ/snarkOS/pull/3017 for such calls.

Fixes https://github.com/AleoHQ/snarkOS/issues/3043.

### DIFF
--- a/node/cdn/src/blocks.rs
+++ b/node/cdn/src/blocks.rs
@@ -213,9 +213,14 @@ pub async fn load_blocks<N: Network>(
                     std::process::exit(0);
                 }
 
+                // Register the next block's height, as the block gets consumed next.
+                let block_height = block.height();
+
                 // Insert the block into the ledger.
                 process_clone(block)?;
-                current_height += 1;
+
+                // Update the current height.
+                current_height = block_height;
 
                 // Log the progress.
                 log_progress::<BLOCKS_PER_FILE>(timer, current_height, cdn_start, cdn_end, "block");
@@ -465,10 +470,10 @@ mod tests {
     }
 
     #[test]
-    fn test_load_blocks_1_to_50() {
-        let start_height = 1;
+    fn test_load_blocks_0_to_50() {
+        let start_height = 0;
         let end_height = Some(50);
-        check_load_blocks(start_height, end_height, 49);
+        check_load_blocks(start_height, end_height, 50);
     }
 
     #[test]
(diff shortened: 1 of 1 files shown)


node/cdn/src/blocks.rs
```diff
@@ -465,10 +475,10 @@ mod tests {
     }
 
     #[test]
-    fn test_load_blocks_1_to_50() {
-        let start_height = 1;
+    fn test_load_blocks_0_to_50() {
+        let start_height = 0;
         let end_height = Some(50);
-        check_load_blocks(start_height, end_height, 49);
+        check_load_blocks(start_height, end_height, 50);
```
howardwu, line 476: Can you check the `check_load_blocks` calls in the tests below?

In the previous PR, they were also changed.

Note the one below is also erroneous (it should say 100)

ljedrz: [these](https://github.com/AleoHQ/snarkOS/pull/3017/files#diff-90406784933ef232b46d15981d5179f0b7917b8e006d9d31f1f856c6066e5c4cL447-L464) were the only 2 adjustments to tests, with the 1st one being reverted in this PR, and I just checked that reverting the other one (from `1_to_123` to `0_to_123` with an extra expected block) also passes; if you'd like for the second test's expected value to be `100` (which would also make sense, as 100 blocks are actually downloaded), _that_ would be a deviation from the state from before #3017

howardwu: Screen Shot 2024-01-30 at 2 01 46 PM

I'm still seeing changes here that break the original interface

ljedrz: I restored the original version of the test

node/cdn/src/blocks.rs
```diff
@@ -228,6 +231,13 @@ pub async fn load_blocks<N: Network>(
         .map_err(|e| (current_height, e))?;
     }
 
+    // While the node will never request the genesis block, non-node callers may do so; in that case 1
+    // needs to be subtracted from the returned value, as the number of processed blocks is lower than
+    // the height of the ledger.
+    if start_height == 0 {
+        current_height = current_height.saturating_sub(1);
+    }
```
howardwu: Why would this impact the current_height? Sorry, I'm not following the change.

ljedrz: I can see how this is unclear; I updated the PR and now, instead of any special-casing, the `current_height` just gets updated with the last block's height, which is simpler to digest

howardwu (approved):

PR #3046, opened 2024-01-30
check transaction before broadcasting / accepting

HarukaMa:
## Motivation

Check if the transaction is valid before broadcasting. Prevents users accidentally generating invalid transactions then wonder where did they go after it got rejected by other nodes.

This check was there before Sep 2022. Not sure why it was removed.

## Test Plan

My invalid tx could be rejected.

```
⚠️  ❌ Failed to broadcast execution 'credits.aleo/transfer_private' to https://explorer.hamp.app/testnet3/transaction/broadcast: (status code 500: "Something went wrong: Invalid transaction: The input ID '7798950585313772830287778856136311499267216010577198944948157594845486864875field' already exists in the ledger")
```

## Related PRs

(Link any related PRs here)


HarukaMa:
Oops, edited the code but forgot to git add, will change later

howardwu:
There is the concern that it is possible the user broadcasts 2 or more transactions, whereby the first transaction makes the second (and later) transactions valid. But the verifier will not know this without speculative execution in advance.

HarukaMa:
This check is also present in the inbound method for `UnconfirmedTransaction`, so unless the user is broadcasting directly to a validator, they won't succeed as their later transactions won't be accepted by other nodes anyway.

howardwu:
Great point, we did add that check in as a preemptive safeguard. I agree it wouldn't be a significant feature change to include this one too.

Do you know if there are any major performance implications we should be aware of with this PR?

HarukaMa:
Actually I did think about if one can just spam invalid transactions, but didn't actually try to profile the check function. It does seems a little expensive as the "basic" check did try to thoroughly verify many things included in the transaction.

Without the check though, the spammer will be spamming all connected nodes instead. But admittedly losing a public API endpoint would be worse.

Probably I need to generate some fake data to test if this could have severe performance implications.

HarukaMa:
I just realized we have https://github.com/AleoHQ/snarkOS/pull/2942 so we don't need to specifically worry about attacks as you can still DDoS other endpoints, especially `/blocks`. 

I did test the check function and it took 50~100ms on my server to verify a [valid transaction](https://explorer.hamp.app/transaction?id=at1k372fktp0ke8m3jny4l3wtc8helhaa0zhqj0xkyfyrqpjj5tg5qq953vd5) and similar transactions with a bad execution proofs. The timing variation seems to come from storage accesses to verify duplicated IDs. I currently don't have resources to actually test if flooding would take down a node, but from the time above, heuristically it's indeed possible even with legitimate traffic (think a large amount of valid transactions during a event of some project).

That said, that amount of traffic could disable other nodes processing `UnconfirmedTransaction` messages as well if we don't add this check here.

kaimast:
This change would be useful to quickly see if the network will reject a transaction.

Did we already decide whether this is worth the (potential) performance hit?

HarukaMa:
I didn't check and I actually forgot I opened this PR

vicsn (commented):
I see the benefit of getting a fast user response. However, transaction confirmation latency has been a primary concern for users, so this small performance hit is unfortunately a concern for wallets and apps submitting large amounts of transactions. Moreover, it opens a (mostly theoretical) OOM issue, when someone submits 100 large deployments in parallel.

Instead, transaction verification can happen by a client on an as-need basis, e.g.:
- in the leo CLI: https://github.com/ProvableHQ/leo/issues/28528#issuecomment-2906648938
- from the respective SDK an explorer or wallet may be using

@HarukaMa I suggest we close, let me know if you feel strongly otherwise

HarukaMa:
So I've been trying to write a code to locally generate transactions in batch, and now I have exactly this issue again: my code generated transactions that is rejected somewhere and there is 0 way for me right now to know what went wrong.

As I need to batch generate the transactions locally, it's not possible to use wallets to do that, so I'm generating the transaction by directly calling snarkVM functions. The `VM::execute` returned me a transaction, but when I broadcast it, my node accepted the transaction (because it's not checking anything), and it went nowhere.

I would accept it if there is at least a way to verify the transaction without having to use the tool to also generate it, but now I'm completely out of ideas why the transaction is not going anywhere.

<hr>
Probably not useful, but my code to generate transaction:

```py
def main():
    program = Program.load(BytesIO(aleo_explorer_rust.parse_program(open("../../src/aleo-explorer/node/credits_v1.aleo").read())))

    for d in data:
        transaction_bytes = aleo_transaction.execute(
            "testnet",
            os.environ.get("PRIVATE_KEY", ""),
            program.dump(),
            "transfer_public",
            [
                PlaintextValue(
                    plaintext=LiteralPlaintext(
                        literal=Literal(
                            type_=Literal.Type.Address,
                            primitive=Address.loads(d["address"])
                        )
                    )
                ).dump(),
                PlaintextValue(
                    plaintext=LiteralPlaintext(
                        literal=Literal(
                            type_=Literal.Type.U64,
                            primitive=u64(d["amount"])
                        )
                    )
                ).dump()
            ],
            "https://testnet.aleoscan.io"
        )
        transaction = Transaction.load(BytesIO(transaction_bytes))
        r = requests.post("https://testnet.aleoscan.io/testnet/transaction/broadcast", json=transaction.json_compatible())
        print(r.status_code)
        print(r.json())
```

```rust

type M = snarkvm_console_network::MainnetV0;
type T = snarkvm_console_network::TestnetV0;
type C = snarkvm_console_network::CanaryV0;

fn execute_network<'a, N: Network>(
    private_key: &str,
    program: &[u8],
    function_name: &str,
    inputs: Vec<Vec<u8>>,
    query: &str,
) -> PyResult<Cow<'a, [u8]>> {
    let private_key = PrivateKey::<N>::from_str(private_key).unwrap();
    let program = Program::<N>::from_bytes_le(program).unwrap();
    let function_name = Identifier::<N>::from_str(function_name).unwrap();
    let inputs: Vec<Value<N>> = inputs
        .iter()
        .map(|input| Value::<N>::from_bytes_le(input).unwrap())
        .collect();
    let query = Query::<N, BlockMemory<N>>::REST(query.to_string());

    let store = ConsensusStore::<N, ConsensusMemory<N>>::open(StorageMode::Production).unwrap();
    let vm = VM::from(store).unwrap();

    vm.process().write().add_program(&program).unwrap();

    let transaction = vm
        .execute(
            &private_key,
            (program.id(), function_name),
            inputs.iter(),
            None,
            0,
            Some(&query),
            &mut rand::thread_rng(),
        )
        .unwrap()
        .to_bytes_le()
        .unwrap();

    Ok(Cow::Owned(transaction))
}

#[pyfunction]
fn execute<'a>(
    network: &str,
    private_key: &str,
    program: &[u8],
    function_name: &str,
    inputs: Vec<Vec<u8>>,
    query: &str,
) -> PyResult<Cow<'a, [u8]>> {
    if network == "mainnet" {
        execute_network::<M>(private_key, program, function_name, inputs, query)
    } else if network == "testnet" {
        execute_network::<T>(private_key, program, function_name, inputs, query)
    } else {
        execute_network::<C>(private_key, program, function_name, inputs, query)
    }
}
```

vicsn:
Yeah that is annoying. I'll follow up with you in Telegram. New draft PR: https://github.com/ProvableHQ/snarkOS/pull/3753

PR #3048, opened 2024-01-31
Avoid deprecation warnings from indexmap

ljedrz:
This PR anticipates deprecation warnings related to the use of IndexMap::remove coming in `indexmap 2.2.1`.

The changes to `::swap_remove` leave the existing logic intact; there were a few spots where I suspect that the insertion order matters, though, and I used `::shift_remove` there instead.

The one larger diff comes from an optimization of lookups I applied as a drive-by.


ljedrz:
Note: we might want to revise our use of `IndexMap`/`IndexSet` and change it to `HashMap`/`HashSet` for collections where the insertion order doesn't matter, for both performance and clarity.

joske (approved):

howardwu (approved):

PR #3049, opened 2024-01-31
Corrected minimum Ubuntu version in README.md

dembelekaro:
## Motivation

SnarkOS Readme states minimum Ubuntu version is 22.04.  
Dev docs state that minimum Ubuntu requirement is 20.04.  See here: https://developer.aleo.org/testnet/getting_started/installation/#21-requirements

## Test Plan

I tested on Ubuntu 20.04 and it works fine. The minimum should be Ubuntu 20.04.


pnoken:
I can confirm it works well on Ubuntu 20.04 as I also tested it too
@all-contributors please add @pnoken for README

pnoken (approved):
Go to go

PR #3050, opened 2024-01-31
fix: build should not run as root

joske:
## Motivation

Running the entire `build-ubuntu.sh` as root will also create root-owned folder/files which will create problems later. It's better to put `sudo` in front of the commands that need it only (apt, ufw).

### DIFF
--- a/build_ubuntu.sh
+++ b/build_ubuntu.sh
@@ -33,7 +33,7 @@ source $HOME/.cargo/env
 
 # Install snarkOS
 # cargo clean
-cargo install --path .
+cargo install --locked --path .
 
 echo "=================================================="
 echo " Attention - Please ensure ports 4133 and 3033"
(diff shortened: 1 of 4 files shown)


ljedrz (approved):
I'm in favor of this change; we could also adjust the `cargo install` command to be executed with the `--locked` param.

There may also be some other spots (e.g. the README) where this should be applied/documented.

build_ubuntu.sh
```diff
@@ -1,28 +1,30 @@
...
+	pkg-config \
+	tmux \
+	xz-utils \
+	ufw
+
+# Open ports on system
+
+sudo ufw allow 5000/tcp
+sudo ufw allow 4133/tcp
+sudo ufw allow 3033/tcp
```
howardwu:
```suggestion
```

howardwu:
Not my preferred change, merging anyways

PR #3051, opened 2024-02-01
Clarification needed on minimum Ubuntu Version

dembelekaro:
## Motivation
Snark OS wiki states minimum Ubuntu version is 22.04 (LTS)
Dev docs linked below say that Minimum Ubuntu version is 20.04 (LTS) which worked fine for me. I mistakenly closed a previous PR with the same edit because I thought the issue was fixed but error is still there in the Snark OS wiki. 

Link: https://developer.aleo.org/testnet/getting_started/installation/#21-requirements


howardwu:
22.04 is correct.

PR #3064, opened 2024-02-07
Add a histogram for tcp events based on Event type

miazn:
## Motivation
To better understand what is going on in SnarkOS TCP land, it may be beneficial to create a histogram (1 metric in Prometheus) that shows count and sum in bytes of each Event Message [type](https://github.com/AleoHQ/snarkOS/blob/testnet3/node/bft/events/src/lib.rs#L94-L110), so this PR adds one called `snarkos_tcp_gateway_messages_received`, with dynamic labeling based on the event name, in the `Decoder` impl of the `EventCodec`. I am using the basic `metrics` rust crate to call the [histogram macro](https://docs.rs/metrics/latest/metrics/macro.histogram.html), however it appears that in SnarkVM this functionality is overridden by the implementation of the snarkVM metrics crate, so I figured the easiest thing to do was to just add an alias for the default crate.

Note, the `histogram!` macro used currently only needs to be called once, so we don't need to register the metrics in `/node/metrics/src/names.rs`, which might be worth revisiting at a later date.

## Test Plan
Run on local devnets

Screenshot 2024-02-07 at 10 44 48 PM

### DIFF
--- a/node/bft/events/Cargo.toml
+++ b/node/bft/events/Cargo.toml
@@ -36,6 +36,10 @@ path = "../../metrics"
 version = "=2.2.7"
 optional = true
 
+[dependencies.metrics_external]
+version = "0.22"
+package = "metrics" 
+
 [dependencies.rayon]
 version = "1"
 
--- a/node/bft/events/src/helpers/codec.rs
+++ b/node/bft/events/src/helpers/codec.rs
@@ -89,6 +89,7 @@ impl<N: Network> Decoder for EventCodec<N> {
         let reader = bytes.reader();
         match Event::read_le(reader) {
             Ok(event) => {
+                #[cfg(feature = "metrics")]
                 histogram!(metrics::tcp::TCP_GATEWAY, "event" => event.name()).record(bytes_len);
                 Ok(Some(event))
             },
--- a/node/bft/events/src/helpers/codec.rs
+++ b/node/bft/events/src/helpers/codec.rs
@@ -21,6 +21,7 @@ use rayon::{
     iter::{IndexedParallelIterator, ParallelIterator},
     prelude::ParallelSlice,
 };
+use metrics_external::histogram;
 use snow::{HandshakeState, StatelessTransportState};
 use std::{io, sync::Arc};
 use tokio_util::codec::{Decoder, Encoder, LengthDelimitedCodec};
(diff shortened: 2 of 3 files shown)


cli/src/commands/start.rs
```diff
@@ -125,6 +126,13 @@ pub struct Start {
     pub dev_num_validators: Option<u16>,
 }
 
+// Create a struct to hold the account data
+#[derive(Serialize)]
+struct AccountData {
+    address: String,
+    private_key: String,
+}
+
```
howardwu: Please revert

cli/src/commands/start.rs
```diff
@@ -442,6 +450,15 @@ impl Start {
...
 
+        let account_data = AccountData {
+            address: account.address().to_string(),
+            private_key: account.private_key().to_string(), // Ensure you have a way to access the private key
+        };
+
+        let json = serde_json::to_string(&account_data).expect("Error serializing account data");
+        let mut file = File::create("/home/ubuntu/snarkOS/account_data.json").expect("Unable to create file");
+        file.write_all(json.as_bytes()).expect("Unable to write data");
+
```
howardwu: Please revert

node/bft/events/Cargo.toml
```diff
@@ -36,6 +36,10 @@ path = "../../metrics"
 version = "=2.2.7"
 optional = true
 
+[dependencies.metrics_external]
+version = "0.22"
+package = "metrics" 
+
```
howardwu, line 42: Please revert

node/bft/events/src/helpers/codec.rs
```diff
@@ -83,11 +84,15 @@ impl<N: Network> Decoder for EventCodec<N> {
...
         };
-
+        let bytes_len = bytes.len() as f64;
         // Convert the bytes to an event, or fail if it is not valid.
         let reader = bytes.reader();
         match Event::read_le(reader) {
-            Ok(event) => Ok(Some(event)),
+            Ok(event) => {
+                #[cfg(feature = "metrics")]
+                histogram!(metrics::tcp::TCP_GATEWAY, "event" => event.name()).record(bytes_len);
```
howardwu, line 93: Use the histogram from `snarkvm_metrics`

node/bft/events/src/helpers/codec.rs
```diff
@@ -21,6 +21,7 @@ use rayon::{
     iter::{IndexedParallelIterator, ParallelIterator},
     prelude::ParallelSlice,
 };
+use metrics_external::histogram;
```
howardwu, line 24: Please revert

node/bft/events/src/helpers/codec.rs
miazn: In snarkvm_metrics, you can only update a histogram with a static name and value, which would require creating a separate metric name for every event type-- using just the regular histogram macro allows you to add labels which keeps the number of separate metrics lower and allows for easier analysis in Prometheus
I tried to call the original metrics crate histogram macro from here but it looked like I would need to add it to expose it in the snarkvm_metrics crate 😅

howardwu: Got it, thanks for the clarification! I'll take a look at this when things are more stable.

miazn:
closing this in favor of https://github.com/AleoHQ/snarkOS/pull/3097

PR #3075, opened 2024-02-09
Make run_prover.sh respect the lockfile

### LINKED ISSUE #3071: Compilation Failure Due to Type Mismatch and Deprecated remove Method
My last bug was marked as closed, but I'm still unable to run a prover node on a fresh Ubuntu install, as of an hour ago. So here goes again:

🐛 Bug Report

I encountered a compilation failure while working with snarkOS. There are two primary issues noted:

    Deprecated usage of the remove method, which suggests using swap_remove instead to maintain map order.
    A type mismatch error (E0308), where a String was found when a Body was expected in an error handling function.

This bug prevents successful compilation of the snarkOS node and seems quite serious as it affects the fundamental operation of the software.
Steps to Reproduce

    Clone the snarkOS repository.
    Attempt to compile snarkOS using Rust.
    Observe the warnings about the deprecated remove method.
    Encounter the E0308 type mismatch error related to the .error_handler function expecting a Body type instead of a String.

Expected Behavior

The expected behavior is for snarkOS to compile without any errors or deprecation warnings. Instead of a String, the .error_handler should either be supplied with or be able to convert the provided string into a Body type, and the deprecated remove method should be updated to swap_remove.
Your Environment

    snarkOS Version: v2.2.7
    Rust Version: rustc 1.74.0 (79e9716c9 2023-11-13)
    Computer OS: Windows 10 Version 22H2 (OS Build 10945.3930)

ljedrz:
I'm not sure what the rationale was for removing it in the update loop, but it has been causing trouble due to it being recreated with updates afterwards.

This should fix https://github.com/AleoHQ/snarkOS/issues/3071, filing as a draft first just to double-check.

### DIFF
--- a/run-prover.sh
+++ b/run-prover.sh
@@ -39,7 +39,6 @@ while :
 do
   echo "Checking for updates..."
   git stash
-  rm Cargo.lock
   STATUS=$(git pull)
 
   if [ "$STATUS" != "Already up to date." ]; then


joske (approved):

joske:
This block:
```bash
for word in $*;
do
  COMMAND="${COMMAND} ${word}"
done
```
can be replaced by just 
```bash
COMMAND="${COMMAND}" "$@"
```

xE33:
> ```shell
> COMMAND="${COMMAND}" "$@"
> ```

it works now, how do i check log to be sure that prover is running

ljedrz:
@xE33 there is likely some `bash`-foo that would bring the logs back to `stdout`, but a solution I know is to run `tail -f /tmp/snarkos.log`.

xE33:
> @xE33 there is likely some `bash`-foo that would bring the logs back to `stdout`, but a solution I know is to run `tail -f /tmp/snarkos.log`.

thanks, everything works now

PR #3081, opened 2024-02-10
[ZKS-00] Revert election certificates

howardwu:
## Motivation

This change was made as part of a potential fix for the BFT, however it became clear we actually do not need it. So we are removing it to bring this back to vanilla Bullshark.

The sister PR can be found here: https://github.com/AleoHQ/snarkVM/pull/2343

This was originally a fix intended for Audit Finding: **[zksecurity 00] Commit Flow Can Lead to Safety Violation**, however it is now unnecessary with the new fixes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3526,7 +3526,7 @@ dependencies = [
 [[package]]
 name = "snarkvm"
 version = "0.16.18"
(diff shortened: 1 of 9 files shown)


raychu86 (approved):
LGTM

PR #3083, opened 2024-02-10
Verify incoming `TransmissionResponse`s

raychu86:
## Motivation

This PR performs transmission verification when processing transmissions from `TransmissionResponse`s. Previously we were blindly accepting them into the batch, which could lead to possible halting cases, where malicious nodes inject faulty transactions.

This should fix:
https://github.com/AleoHQ/snarkOS/issues/2990
https://github.com/AleoHQ/snarkOS/issues/2991

### DIFF
--- a/node/bft/ledger-service/src/ledger.rs
+++ b/node/bft/ledger-service/src/ledger.rs
@@ -171,9 +171,8 @@ impl<N: Network, C: ConsensusStorage<N>> LedgerService<N> for CoreLedgerService<
         }
     }
 
-    // TODO (raychu86): Rename to `check_transmission_basic`.
-    /// Ensures the given transmission ID matches the given transmission.
-    async fn ensure_transmission_id_matches(
+    /// Checks the given transmission is well-formed and unique.
+    async fn check_transmission_basic(
         &self,
         transmission_id: TransmissionID<N>,
         transmission: &mut Transmission<N>,
--- a/node/bft/ledger-service/src/ledger.rs
+++ b/node/bft/ledger-service/src/ledger.rs
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use crate::{fmt_id, spawn_blocking, LedgerService};
+use crate::{spawn_blocking, LedgerService};
 use snarkvm::{
     ledger::{
         block::{Block, Transaction},
@@ -183,18 +183,13 @@ impl<N: Network, C: ConsensusStorage<N>> LedgerService<N> for CoreLedgerService<
             (TransmissionID::Transaction(expected_transaction_id), Transmission::Transaction(transaction_data)) => {
                 match transaction_data.clone().deserialize_blocking() {
                     Ok(transaction) => {
-                        let deserialized_transaction_data = Data::Object(transaction);
+                        let deserialized_transaction = Data::Object(transaction);
 
                         // Check that the transaction is valid.
-                        if let Err(err) = self
-                            .check_transaction_basic(expected_transaction_id, deserialized_transaction_data.clone())
-                            .await
-                        {
-                            bail!("Failed to check transmission - {err}")
-                        }
+                        self.check_transaction_basic(expected_transaction_id, deserialized_transaction.clone()).await?;
 
                         // Update the transmission with the deserialized transaction.
-                        *transaction_data = deserialized_transaction_data;
+                        *transaction_data = deserialized_transaction;
                     }
                     Err(err) => {
                         bail!("Failed to deserialize transaction: {err}");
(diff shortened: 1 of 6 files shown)


vicsn:
I believe this risks OOMing the validator. I know of two paths where we request transmissions:
- from pings
- from `batch_proposal`s

We could receive many pings in parallel, and a single `batch_proposal`s transmissions are retrieved in parallel as well. The maximum RAM to verify a 1M constraint program is 2GB. So 32-64 transmissions in the worst case and we're out.

For pings I'd imagine it would be better to follow the path of `UnconfirmedTransaction` messages, which will add them to the mempool queue. The mempool queue currently serially processes transactions by a single worker so should not OOM.

For `batch_proposal`s, pushing them to the back of the mempool queue might cause significant delays, so I was thinking we should add a second separate simple mempool queue for `batch_proposal`s.

node/bft/src/worker.rs
```diff
@@ -363,7 +363,7 @@ impl<N: Network> Worker<N> {
         self.spawn(async move {
             while let Some((peer_ip, transmission_response)) = rx_transmission_response.recv().await {
                 // Process the transmission response.
-                self_.finish_transmission_request(peer_ip, transmission_response);
+                self_.finish_transmission_request(peer_ip, transmission_response).await;
```
ljedrz, line 366: do we need to process the transmission response entries in the order they arrive? if not, we could spawn tasks that performs this action, in order to be able to perform multiple of them concurrently

node/bft/src/worker.rs
```diff
@@ -393,19 +393,19 @@ impl<N: Network> Worker<N> {
...
-            // Ensure the transmission ID matches the transmission.
-            match self.ledger.ensure_transmission_id_matches(transmission_id, &mut transmission) {
+            // Check that the given transmission is well-formed and unique.
+            match self.ledger.check_transmission_basic(transmission_id, &mut transmission).await {
                 Ok(()) => {
                     // Remove the transmission ID from the pending queue.
                     self.pending.remove(transmission_id, Some(transmission));
                 }
-                Err(err) => warn!("Failed to finish transmission response from peer '{peer_ip}': {err}"),
+                Err(err) => warn!("Malicious peer ('{peer_ip}') send an invalid transmission: {err}"),
```
ljedrz:
```suggestion
                Err(err) => warn!("Malicious peer ('{peer_ip}') sent an invalid transmission: {err}"),
```
nit

node/bft/ledger-service/src/ledger.rs
```diff
@@ -182,16 +182,13 @@ impl<N: Network, C: ConsensusStorage<N>> LedgerService<N> for CoreLedgerService<
             (TransmissionID::Transaction(expected_transaction_id), Transmission::Transaction(transaction_data)) => {
                 match transaction_data.clone().deserialize_blocking() {
```
ljedrz, line 183: it's not ideal that we do blocking deserialization in an `async` context; we should either do it before calling `check_transaction_basic` (with a deserialized transaction), or use a blocking task here

node/bft/ledger-service/src/ledger.rs
```diff
@@ -201,16 +198,13 @@ impl<N: Network, C: ConsensusStorage<N>> LedgerService<N> for CoreLedgerService<
             (TransmissionID::Solution(expected_commitment), Transmission::Solution(solution_data)) => {
                 match solution_data.clone().deserialize_blocking() {
```
ljedrz, line 199: ditto

ljedrz (commented):
Left a few comments; LGTM in general, but there are some performance considerations.

randomsleep (commented):
`check_transmission_basic` is too strict for all `TransmissionResponse`. It will cause liveness issues.  For example, it's common that a valid `Certificate` contains a transaction that tries to consume a serial number that is already consumed. With this patch, the validator will fail to sync the transaction and thus will fail to fetch the `Certificate`.

The validator should perform some more basic checks (e.g. exclude Fee transaction, invalid format) before signing `BatchPropose`. It should not reject the transaction if already included in a `Certificate`.

vicsn:
@randomsleep can you elaborate? 

> For example, it's common that a valid Certificate contains a transaction that tries to consume a serial number that is already consumed.

This sounds like something which should be invalid.

> The validator should perform some more basic checks (e.g. exclude Fee transaction, invalid format) before signing BatchPropose. It should not reject the transaction if already included in a Certificate.

This PR introduces calling `check_transaction_basic` on every transaction *before* signing batch proposals, so if all goes well we will only include valid transactions as Certificate and they won't be at risk of further rejection perse.

raychu86:
I believe @randomsleep might be correct that we may not need to verify the transmissions in the BatchCertificate, because it may prevent us from syncing certificates that are included in a subsequent subdag. Perhaps these transactions/double-spends were going to be aborted in the block; instead, we forcefully ignore ignore transmissions which prevents us from syncing up with the certificate. The validator would then get stuck until is starts syncing via blocks. Definitely needs more thought.

However, we definitely do need to verify the transmissions from the proposals.

randomsleep:
Yes, @raychu86 have explained.

However, while checking 'BatchPropose', we should also allow duplicated transactions. This is because at the time some validarors may have committed some transactions but others haven't.

randomsleep:
My thoughts on transaction checking rules:


1. Before proposing my own 'BatchPropose': do all kinds of necessary check. Prevent attack from users.
2. While checking incoming 'BatchPropose': only do basic check. The check should be independent from latest ledger state because validators may have different state. Prevent attack from malicious validators.
3. While adding 'Certificate': no check.
4. While commiting 'Subdag': Make sure bad transactions won't block executing.

raychu86:
This should be replaced by https://github.com/AleoHQ/snarkVM/pull/2376 and https://github.com/AleoHQ/snarkOS/pull/3104.

PR #3089, opened 2024-02-12
[Fix] Limit number of deployments in mempool.

vicsn:
A new approach to https://github.com/AleoHQ/snarkOS/pull/2999 with a focus on simplicity and readability.

At 10 deployments at most in the deployments_queue, we won't run out of memory (the network message size is at most 300 MB).

And we will never verify more than 1 deployment sequentially if there are any executions around, making sure we won't take too much compute time.

We always interweave the verification of some executions and deployments.

### DIFF
--- a/node/consensus/src/lib.rs
+++ b/node/consensus/src/lib.rs
@@ -58,6 +58,27 @@ const CAPACITY_FOR_DEPLOYMENTS: usize = 20;
 /// Percentage of mempool transactions capacity reserved for executions.
 const CAPACITY_FOR_EXECUTIONS: usize = 80;
 
+/// Helper struct to track incoming transactions.
+struct TransactionsQueue<N: Network> {
+    pub deployments: LruCache<N::TransactionID, Transaction<N>>,
+    pub executions: LruCache<N::TransactionID, Transaction<N>>,
+}
+
+impl<N: Network> Default for TransactionsQueue<N> {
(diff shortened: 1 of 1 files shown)


node/consensus/src/lib.rs
```diff
@@ -274,14 +290,20 @@ impl<N: Network> Consensus<N> {
...
-            // Determine the number of transactions to send.
-            let num_transactions = queue.len().min(capacity);
-            // Drain the solutions from the queue.
-            (0..num_transactions)
-                .filter_map(|_| queue.pop_lru().map(|(_, transaction)| transaction))
-                .collect::<Vec<_>>()
+            // Acquire the lock on the deployments queue.
+            let mut deployments_queue = self.deployments_queue.lock();
+            // Acquire the lock on the executions queue.
+            let mut executions_queue = self.executions_queue.lock();
```
ljedrz: future consideration: ideally, collections that are always accessed together could be under a single lock (for added performance), though that may require a little bit more type boilerplate

vicsn: Made it anyway so Howard can decide: f93060b9d
Did not implement any further methods for TransactionsQueue to avoid having to add input checking for those methods and to keep it simple.

ljedrz (approved):
LGTM :+1:

howardwu (approved):

PR #3099, opened 2024-02-16
feat(ubuntu) : checks of installation

cryptoryda:
### Check if Ubuntu dependencies installation was successful.

### DIFF
--- a/build_ubuntu.sh
+++ b/build_ubuntu.sh
@@ -5,30 +5,44 @@ echo " This will request root permissions with sudo."
 echo "================================================"
 
 # Install Ubuntu dependencies
-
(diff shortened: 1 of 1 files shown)


joske:
Although nothing wrong with this change, could you maybe try adding `set -euo` in the beginning of the script, then all this won't even be needed, and it will stop the script on other errors too.

vicsn:
I'm gonna close this - sadly clearly airdrop farming spam which we should not incentivise

cryptoryda:
No reward or incentive required thanks. I code because It will boost my reputation to have PRs merged with some of the best web3 projects.

PR #3100, opened 2024-02-16
[ZKS-04] Clean the DAG state of committed certificates on bootup

raychu86:
## Motivation

This PR adds logic during bootup, to clear the DAG of certificates that should have already been committed in blocks on the ledger.

### DIFF
--- a/node/bft/src/bft.rs
+++ b/node/bft/src/bft.rs
@@ -809,11 +809,11 @@ impl<N: Network> BFT<N> {
             // Acquire the BFT write lock.
             let mut dag = self.dag.write();
             // Iterate over the certificates.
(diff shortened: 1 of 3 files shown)


node/bft/src/bft.rs
```diff
@@ -809,7 +809,7 @@ impl<N: Network> BFT<N> {
             // Acquire the BFT write lock.
             let mut dag = self.dag.write();
             // Iterate over the certificates.
-            for certificate in certificates {
+            for certificate in certificates.clone() {
```
ljedrz: we should probably iterate by reference and only clone in case of insertion

howardwu: resolved in 073dab41f

PR #3102, opened 2024-02-19
[Fix] Handle more POSIX signals on UNIX systems

ljedrz:
This PR extends the signal handler on UNIX; in addition to `SIGINT`, it now also supports `SIGTERM`, `SIGQUIT`, and `SIGHUP`.

### DIFF
--- a/node/src/traits.rs
+++ b/node/src/traits.rs
@@ -17,6 +17,8 @@ use snarkvm::prelude::{Address, Network, PrivateKey, ViewKey};
 
 use once_cell::sync::OnceCell;
 use std::{
+    future::Future,
+    io,
     sync::{
         atomic::{AtomicBool, Ordering},
         Arc,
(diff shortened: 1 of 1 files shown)


joske (approved):

iamalwaysuncomfortable (commented):
Is it possible to bump up the number of seconds waiting for the shutdown? Even if there are SIGINT signals, a 3 second wait time might not be enough to clean everything up.

Maybe something like
```rust
tokio::time::sleep(Duration::from_secs(10)).await;
```

ljedrz:
@iamalwaysuncomfortable what cleanups do you have in mind? Anything that's not (persistent) storage-related doesn't really require any special treatment.

howardwu:
Adding a note for context here:
```
SIGTERM terminates the process gracefully
SIGQUIT terminates the process with a core dump
SIGKILL immediately and forcefully terminates the process; it can't be caught
SIGSTP suspends execution
SIGSTOP suspends execution and can't be caught
SIGHUP can be received when the owning terminal is shut down; it can be worked around on UNIXes with nohup or disown
SIGABRT terminates the process as if it was done from within the process
```

howardwu (approved):

PR #3103, opened 2024-02-19
[Stress Test] Check if the number of peers received is less than the amount we expect

vicsn:
## Motivation

It makes sense to drop messages which are clearly violating our software and may be a DoS vector. This is optional and defense in depth.

## Test Plan

Found while defining a stress test for malicious PeerResponse attack vectors (this will not be public).


ljedrz (approved):
LGTM :+1:

howardwu (approved):

Issue #2991, opened 2024-01-11
[Bug] BatchPropose with invalid transaction can halt the network


feezybabee:
# https://hackerone.com/reports/2311934

## Summary:

 BatchPropose with invalid transaction halt the network

## Steps To Reproduce:

    1. Use this branch: https://github.com/ghostant-1017/snarkOS/tree/demo/invalid_tx
    2. Run the devnet script: `.devnet` 
    3. Check the logs: ` Unable to advance to the next block - Invalid transaction found in the transactions list: Execution verification failed: Failed to verify a transition input`

## Proof-of-Concept (PoC)

The `BatchPropose` with invalid transactions will be certified and the `check_next_block` will fail as the transaction is invalid

## Supporting Material/References:

Logs: https://github.com/ghostant-1017/logs/blob/master/logs-20240111094535.tar.gz

## Summary:
BatchPropose with invalid transaction halt the network

ghostant-1017:
Same as the invalid `Deployment transaction`

raychu86:
Please try to reproduce this bug with the changes here - https://github.com/AleoHQ/snarkOS/pull/3083

raychu86:
@ghostant-1017 I've created a new proposed fix for this to just do verification for the proposals. Take a look here - https://github.com/AleoHQ/snarkOS/pull/3098. Would be helpful to think about the implications of this fix.

PR #3104, opened 2024-02-19
[HackerOne-2310746] Add fee check to transmissions fetched in batch proposal.

### LINKED ISSUE #2990: [Bug] A malicious validator can broadcast certificate with `Transaction::Fee` and block the network
# https://hackerone.com/reports/2310746

## Summary:

Currently, Aleo does not speculate on `Transaction::Fee`:

```
                    // There are no finalize operations here.
                    // Note: This will abort the entire atomic batch.
                    Transaction::Fee(..) => Err("Cannot speculate on a fee transaction".to_string()),
```

However, malicious can propose a `BatchPropose` with `Fee` transaction. Other validators won't check if a `BatchPropose` contains a `Fee` transaction and will sign the proposal. Once the `Certificate` with `Fee` transaction is committed, it will block the execution of the block. All the validators will be blocked.

## Steps To Reproduce:

1. Clone snarkOS
2. Checkout testnet3: `git checkout 052457bddbf736f88227c543646890f8588a0efb`
3. Apply the patch unchecked_fee_transaction.patch in the attachment: `git apply unchecked_fee_transaction.patch`
4. Run ./devnet.sh (To make it easy to reproduce, run all 4 validators with the patch)
5. Wait some round and find the network stuck:
image

Also see the attached sample log `sample_log.zip`
[sample_log.zip](https://github.com/AleoHQ/snarkOS/files/13897085/sample_log.zip)

## Proof-of-Concept (PoC)

See above

## Fix Suggestions:

Check if a `BatchPropose` contains `Transaction::Fee` before signing.

## Impact

A malicious validator can broadcast certificate with `Transaction::Fee` and block the whole network

raychu86:
## Motivation

This PR performs basic checks on transmissions fetched from batch proposals. We now check that the transmission id's match and that the transmission is not a `Fee` transaction. 

This fix does not address all the issues, but is an added mitigation for the malicious Fee transmission case. This is an incremental fix for #3098.

## Related PRs

This PR is related to #3098, and is a simpler approach that addresses one case of the problem.

This PR fixes #2990, but is insufficient for #2991.

### DIFF
--- a/node/bft/src/primary.rs
+++ b/node/bft/src/primary.rs
@@ -549,7 +549,16 @@ impl<N: Network> Primary<N> {
         }
 
         // If the peer is ahead, use the batch header to sync up to the peer.
-        let transmissions = self.sync_with_batch_header_from_peer(peer_ip, &batch_header).await?;
+        let mut transmissions = self.sync_with_batch_header_from_peer(peer_ip, &batch_header).await?;
+
+        // Check that the transmission ids match and are not fee transactions.
+        for (transmission_id, transmission) in transmissions.iter_mut() {
+            // If the transmission is invalid, then return early.
+            if let Err(err) = self.ledger.ensure_transmission_id_matches(*transmission_id, transmission) {
+                debug!("Batch propose from '{peer_ip}' contains an invalid transmission - {err}",);
+                return Ok(());
+            }
+        }
 
         // Ensure the batch is for the current round.
         // This method must be called after fetching previous certificates (above),
--- a/node/bft/src/primary.rs
+++ b/node/bft/src/primary.rs
@@ -573,8 +573,8 @@ impl<N: Network> Primary<N> {
 
         // Check that the transmission ids match and are not fee transactions.
         for (transmission_id, transmission) in transmissions.iter_mut() {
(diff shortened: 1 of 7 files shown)


howardwu (approved):

node/bft/src/primary.rs
```diff
@@ -569,7 +569,16 @@ impl<N: Network> Primary<N> {
         }
 
         // If the peer is ahead, use the batch header to sync up to the peer.
-        let transmissions = self.sync_with_batch_header_from_peer(peer_ip, &batch_header).await?;
+        let mut transmissions = self.sync_with_batch_header_from_peer(peer_ip, &batch_header).await?;
+
+        // Check that the transmission ids match and are not fee transactions.
+        for (transmission_id, transmission) in transmissions.iter_mut() {
```
howardwu, line 575: @ljedrz can you profile this and determine if an optimization is possible with rayon or tokio?

howardwu: Feel free to open a new PR if so

raychu86: Note that the number of transmissions here can be up to 50, and each one will be deserializing the Transmission Data object.

PR #3105, opened 2024-02-20
[Fix] Update `is_proposing` check 

raychu86:
## Motivation

This PR attempts to solve the issue where validators are stuck on an old proposal and unable to advance/participate in future rounds of consensus. The current check was preventing the round from advancing if we had a pending proposal, because we were too aggressively assuming the other validators would still sign it. The new logic will now advance the round even if we are proposing if the peer's certificate is already past our proposal round.

If the peer already has a certificate that hit availability threshold, then we can assume they won't be signing for old proposals.

Related to #3074

### DIFF
--- a/node/bft/src/primary.rs
+++ b/node/bft/src/primary.rs
@@ -722,6 +722,8 @@ impl<N: Network> Primary<N> {
 
         // Retrieve the batch certificate author.
         let author = certificate.author();
+        // Retrieve the batch certificate round.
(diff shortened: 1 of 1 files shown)


joske:
I tested this branch, and this seems to work. On `mainnet` branch I could still reproduce the issue.

howardwu (approved):
LGTM, reviewed it by hand as well, this change is desirable

PR #3106, opened 2024-02-20
[Optimize] Cache the recently queried committees

ljedrz:
Committee queries are very prominent in CPU profiles, even with small committee sizes - they are performed quite often, and the deserialization involved is relatively expensive.

This PR adds a simple, small LRU cache that speeds them up; I've double-checked its results by obtaining a new CPU profile (which - as expected - is very different), and ensuring that the cache gets hit (which happens quite a lot).

### DIFF
--- a/node/bft/ledger-service/src/ledger.rs
+++ b/node/bft/ledger-service/src/ledger.rs
@@ -26,6 +26,8 @@ use snarkvm::{
 };
 
 use indexmap::IndexMap;
+use lru::LruCache;
+use parking_lot::Mutex;
 use std::{
     fmt,
     ops::Range,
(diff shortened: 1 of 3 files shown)


joske (approved):

node/bft/ledger-service/src/ledger.rs
```diff
@@ -127,20 +134,30 @@ impl<N: Network, C: ConsensusStorage<N>> LedgerService<N> for CoreLedgerService<
...
-                    true => Ok(current_committee),
+                    true => current_committee,
                     false => bail!("No committee found for round {round} in the ledger"),
                 }
             }
-        }
+        };
+
+        // Insert the committee into the cache.
+        self.committee_cache.lock().push(round, committee.clone());
```
howardwu: @ljedrz this only works for the `match` case on `Some`. It does not work for the committee from `None`.

Can you scope the cache for only committees from `Some` and profile the performance again?

_(If you want to go the extra mile, you can also try adding a `cache` for the underlying impl of `self.current_committee()`, but this extra overhead may not be worth the complexity. The majority of calls will route to `Some` I believe)_

ljedrz: Fixed; I ran the branch locally and all I can see are still cache hits. As for the current committee, I agree that its overhead is low; we can always go back to that idea in the future if need be.

node/bft/ledger-service/src/ledger.rs
```diff
@@ -35,18 +37,23 @@ use std::{
     },
 };
 
+/// The capacity of the LRU holiding the recently queried committees.
+const COMMITTEE_CACHE_SIZE: usize = 16;
```
raychu86, line 41: Is there a reason for using `16`?

This committee cache should just need the rounds near the ones being processed, but not sure what a good number would be.

ljedrz: Not really; the idea was just that it would be small (for low overhead), but still large enough to cover all possible hits. It could probably be a bit smaller, but any resulting difference should be negligible.

howardwu (approved):

PR #3107, opened 2024-02-20
fix: transaction per block metrics

miazn:
## Motivation

~To go along with adding TCP Events [received](https://github.com/AleoHQ/snarkOS/pull/3097), this PR adds TCP Events sent, as well as adds the `node/router` plaintext Messages sent and received.~ edit: removed these since part of the histogram labeling was taking up unnecessary space in prometheus, and the metrics arent really needed
It also changes the transactions per block to be a gauge rather than a counter (which we already registered it as [here](https://github.com/AleoHQ/snarkOS/blob/mainnet/node/metrics/src/names.rs#L24), since a counter cannot decrease, thus fixing the transaction metrics in Grafana

### DIFF
--- a/node/consensus/src/lib.rs
+++ b/node/consensus/src/lib.rs
@@ -407,7 +407,7 @@ impl<N: Network> Consensus<N> {
             let block_latency = next_block_timestamp - current_block_timestamp;
 
             metrics::gauge(metrics::blocks::HEIGHT, next_block.height() as f64);
-            metrics::counter(metrics::blocks::TRANSACTIONS, next_block.transactions().len() as u64);
+            metrics::gauge(metrics::blocks::TRANSACTIONS, next_block.transactions().len() as f64);
             metrics::gauge(metrics::consensus::LAST_COMMITTED_ROUND, next_block.round() as f64);
             metrics::gauge(metrics::consensus::COMMITTED_CERTIFICATES, num_committed_certificates as f64);
             metrics::histogram(metrics::consensus::CERTIFICATE_COMMIT_LATENCY, elapsed.as_secs_f64());
--- a/node/consensus/src/lib.rs
+++ b/node/consensus/src/lib.rs
@@ -407,7 +407,7 @@ impl<N: Network> Consensus<N> {
             let block_latency = next_block_timestamp - current_block_timestamp;
 
             metrics::gauge(metrics::blocks::HEIGHT, next_block.height() as f64);
-            metrics::gauge(metrics::blocks::TRANSACTIONS, next_block.transactions().len() as f64);
+            metrics::increment_gauge(metrics::blocks::TRANSACTIONS, next_block.transactions().len() as f64);
             metrics::gauge(metrics::consensus::LAST_COMMITTED_ROUND, next_block.round() as f64);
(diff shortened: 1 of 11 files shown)


miazn:
@joske I believe we will also have to update the TPS stat in the Grafana dash, should it just be the `avg()` rather than the `sum(rate())`?

node/router/messages/src/helpers/codec.rs
```diff
@@ -59,8 +59,17 @@ impl<N: Network> Encoder<Message<N>> for MessageCodec<N> {
...
 
         let serialized_message = dst.split_to(dst.len()).freeze();
-
-        self.codec.encode(serialized_message, dst)
+        #[cfg(feature = "metrics")]
+        let num_bytes = serialized_message.len() as f64;
+        self.codec.encode(serialized_message, dst)?;
+        #[cfg(feature = "metrics")]
+        metrics::histogram_label(
+            metrics::tcp::TCP_ROUTER_MESSAGES_OUTBOUND,
```
miazn: @ljedrz I noticed in testing with an only validator devnet that this and the router messages inbound metric dont record any values-- is this to be expected?

node/consensus/src/lib.rs
```diff
@@ -407,7 +407,7 @@ impl<N: Network> Consensus<N> {
             let block_latency = next_block_timestamp - current_block_timestamp;
 
             metrics::gauge(metrics::blocks::HEIGHT, next_block.height() as f64);
-            metrics::counter(metrics::blocks::TRANSACTIONS, next_block.transactions().len() as u64);
+            metrics::increment_gauge(metrics::blocks::TRANSACTIONS, next_block.transactions().len() as f64);
```
howardwu, line 410: @joske the intention is to calculate the number of transactions per block here. Will increment be the right operation?

joske: No this will add `next_block.transactions().len()` to the existing value of the gauge. I think you should use the `metrics::gauge` method, which will set the current value of transactions per block.

miazn: Since prometheus is a time series database, even if we set this gauge to be the existing number of transactions, it would not be per block. Since we scrape the endpoint every [15s ](https://github.com/AleoHQ/snarkOS/blob/testnet3/node/metrics/prometheus.yml#L2), I'm pretty confident that we would only get the reading of the current state, i.e. we would be "missing" data from one block if blocktime is under 15 seconds.

I decided to make the metric increment because the metric name is `snarkos_blocks_transactions_total` and I figured this way we would capture all the data. Now, it is correct that this metric _should_ be a counter, since it is always increasing. However, our snarkVM implementation only exposes an `increment_counter` function that increases by the [constant 1](https://github.com/AleoHQ/snarkVM/blob/mainnet/metrics/src/lib.rs#L50-L52) .

I think that this transaction metric should be a total, to reflect the metric name as well as not miss any block data. The best way to do this would be to change the increment counter function in snarkvm to take a value, rather than make it a constant, but switching the metric to a gauge was the lowest-lift method. Let me know what is preferred here.

howardwu (approved):

PR #3109, opened 2024-02-20
[ZKS-01] Add round check in signed proposals cache

raychu86:
## Motivation

This PR adds a check for malicious validators that try to have peers double-sign proposals by circumventing the `signed_proposals` cache. The simple check is to now disconnect from peers that request signatures on proposals that are behind the `signed_round` (of other proposals they've sent).

Audit Finding: **[zksecurity 01] Lack of Dag Containment Could Lead To Safety Violation**


mimoo (approved):
beautiful!

howardwu (approved):

PR #3112, opened 2024-02-21
[Optimize] Skip the scheduled batch proposal if already proposing

ljedrz:
If I'm not missing anything, a batch proposal can be triggered by a timed loop or round progression, and if we're already proposing a batch for a new round, it seems reasonable to me that we may skip the timed one. This would save some network traffic and CPU use for the validators.

I tested it locally and can see the skip happening from time to time, which means those attempts can currently overlap.

### DIFF
--- a/node/bft/src/primary.rs
+++ b/node/bft/src/primary.rs
@@ -900,6 +900,11 @@ impl<N: Network> Primary<N> {
                     debug!("Skipping batch proposal {}", "(node is syncing)".dimmed());
                     continue;
                 }
+                // A best-effort attempt to skip the scheduled batch proposal if
+                // round progression already triggered one.
+                if self_.propose_lock.try_lock().is_err() {
+                    continue;
+                };
                 // If there is no proposed batch, attempt to propose a batch.
                 // Note: Do NOT spawn a task around this function call. Proposing a batch is a critical path,
                 // and only one batch needs be proposed at a time.
--- a/node/bft/src/primary.rs
+++ b/node/bft/src/primary.rs
@@ -903,6 +903,7 @@ impl<N: Network> Primary<N> {
                 // A best-effort attempt to skip the scheduled batch proposal if
                 // round progression already triggered one.
                 if self_.propose_lock.try_lock().is_err() {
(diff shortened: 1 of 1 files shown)


joske (approved):

node/bft/src/primary.rs
```diff
@@ -900,6 +900,11 @@ impl<N: Network> Primary<N> {
                     debug!("Skipping batch proposal {}", "(node is syncing)".dimmed());
                     continue;
                 }
+                // A best-effort attempt to skip the scheduled batch proposal if
+                // round progression already triggered one.
+                if self_.propose_lock.try_lock().is_err() {
+                    continue;
```
vvp, line 907: Perhaps add some `debug!` message here for observing that there's an ongoing batch proposal attempt?

vvp (approved):

node/bft/src/primary.rs
ljedrz: It could be useful, but if my intuition is correct, such an occurrence could just be ignored (so a log would only increase verbosity); either way works for me.

raychu86:
The `propose_batch` call a few lines after this new skip already has a `propose_lock`. So having this check will only ensure that we can continue with this loop (and perform extra `self_.sync.is_synced()` checks) instead of awaiting on the `propose_batch` call. 

I believe this change is safe, but don't see that much of a difference in terms of resource optimizations. Let me know if I'm missing anything.

ljedrz:
@raychu86 you're right, it's nothing extreme, but it can avoid a duplicate proposal to be produced right after we produced one, which reduces CPU use and network traffic.

node/bft/src/primary.rs
howardwu, line 907:
```suggestion
                    trace!("Skipping batch proposal {}", "(node is already proposing)".dimmed());
                    continue;
```

howardwu: Resolved in [44f498b](https://github.com/AleoHQ/snarkOS/pull/3112/commits/44f498bc29cde5301178936115fe2835beaeebcc)

howardwu (approved):

PR #3115, opened 2024-02-22
[Optimize] Cache the latest leader

ljedrz:
Leader computation is a relatively costly operation (it can take more than 1ms with a 4-member committee), and it is currently called several times for the same round. With a little bit of extra code and with negligible overhead we can cache the latest computed leader and in case of a hit, the leader can be procured 3 orders of magnitude (under 1 µs) faster.

note: this PR has a little bit of overlap (mostly related to dependencies) with https://github.com/AleoHQ/snarkOS/pull/3106


ljedrz:
~Ah, forgot about the mock ledger; will update shortly.~

Ready for review.

joske (approved):

SplittyDev (approved):

PR #3117, opened 2024-02-23
[ZKS-04] Perform garbage collection on subdag commit

raychu86:
## Motivation

This PR performs garbage collection on subdag commits only. Previously we were performing GC on every round advancement, which was relying on the improbability of going 100 rounds without committing a subdag. Now we guarantee that we don't throw away older certificates in the case we do go more than MAX_GC_ROUNDS without a commit.

Note: We may also consider adding a manual skip of older certificates in the commit phase.

Audit Finding: **[zksecurity 04] Garbage Collection Can Block Commits From Happening**

### DIFF
--- a/node/bft/src/helpers/storage.rs
+++ b/node/bft/src/helpers/storage.rs
@@ -204,6 +204,12 @@ impl<N: Network> Storage<N> {
         // Update the current round.
         self.current_round.store(next_round, Ordering::SeqCst);
 
+        // Perform garbage collection.
+        self.garbage_collect_certificates(next_round);
+    }
+
+    /// Update the storage by performing garbage collection based on the next round.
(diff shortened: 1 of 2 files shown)


node/bft/src/helpers/storage.rs
```diff
@@ -203,7 +203,10 @@ impl<N: Network> Storage<N> {
     fn update_current_round(&self, next_round: u64) {
         // Update the current round.
         self.current_round.store(next_round, Ordering::SeqCst);
+    }
 
+    /// Update the storage by performing garbage collection based on the next round.
+    pub(crate) fn garbage_collect_certificates(&self, next_round: u64) {
```
mimoo, line 209: what a clean cut!

mimoo (approved):
LGTM!

howardwu (approved):

PR #3118, opened 2024-02-23
[ZKS-02] Update the set of authorized validators 

mdelle1:
## Motivation

This PR updates the set of authorized validators to include current committee members along with committee members from rounds up until the sync range tolerance, which is currently set at one block before nodes are considered out-of-sync. 

Previously, the code would fetch the committee lookback at the latest ledger round, but this is not always the same as the committee lookback from the storage’s current round. For example, if the latest ledger round is ‘n’ but the current round is ‘n+2’, the committees may disagree. In some cases, this may lead to validators being disconnected who should still be considered as authorized to participate in consensus. 

One of the implications of the previous logic is that if these committees differ substantially and affect (2f+1), then a validator will be unable to connect to a quorum of nodes in the current committee and halt. 

Audit Finding: **[zksecurity 02] Dynamic Committee Feature is Not Safe**

### DIFF
--- a/node/bft/ledger-service/src/mock.rs
+++ b/node/bft/ledger-service/src/mock.rs
@@ -32,7 +32,7 @@ use tracing::*;
 #[derive(Debug)]
 pub struct MockLedgerService<N: Network> {
     committee: Committee<N>,
-    height_to_hash: Mutex<BTreeMap<u32, N::BlockHash>>,
+    height_to_hash: Mutex<BTreeMap<u32, (u64, N::BlockHash)>>,
 }
 
 impl<N: Network> MockLedgerService<N> {
@@ -45,7 +45,7 @@ impl<N: Network> MockLedgerService<N> {
     pub fn new_at_height(committee: Committee<N>, height: u32) -> Self {
         let mut height_to_hash = BTreeMap::new();
         for i in 0..=height {
-            height_to_hash.insert(i, (Field::<N>::from_u32(i)).into());
+            height_to_hash.insert(i, (i as u64 * 2, Field::<N>::from_u32(i).into()));
         }
         Self { committee, height_to_hash: Mutex::new(height_to_hash) }
     }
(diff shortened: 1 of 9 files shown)


node/bft/ledger-service/src/mock.rs
```diff
@@ -32,35 +32,35 @@ use tracing::*;
 #[derive(Debug)]
 pub struct MockLedgerService<N: Network> {
     committee: Committee<N>,
-    height_to_hash: Mutex<BTreeMap<u32, N::BlockHash>>,
+    height_to_round_and_hash: Mutex<BTreeMap<u32, (u64, N::BlockHash)>>,
```
ljedrz, line 35: it would probably be better if this was an `RwLock`, as I can see several read-only operations available

node/bft/src/gateway.rs
```diff
@@ -330,18 +334,35 @@ impl<N: Network> Gateway<N> {
...
         //  2. Existing validators must remain connected until they are no longer bonded as a committee member.
         //     (i.e. meaning they must stay online until the next block has been produced)
-        self.ledger
-            .get_committee_lookback_for_round(self.ledger.latest_round())
-            .map_or(false, |committee| committee.is_committee_member(validator_address))
-            || self
-                .ledger
-                .current_committee()
-                .map_or(false, |committee| committee.is_committee_member(validator_address))
+        exists_in_previous_committee_lookback || exists_in_current_committee_lookback || exists_in_latest_committee
```
ljedrz: there's some potential for faster early returns here; once we know that `exists_in_previous_committee_lookback` is `true`, we can immediately `return true;`, same thing with `exists_in_current_committee_lookback`; we only need to proceed with further checks if the previous ones are `false`

ljedrz: in addition, I'd guess that the first of the checks is the slowest and the last one is the fastest, so it would probably be a good idea to reverse their order for the purpose for potential early returns

ljedrz (commented):
I've left some perf-relevant comments; also, 2 CI jobs seem to be failing.

raychu86:
Note: This PR was tested on a network where validators continuously bonded/unbonded and did not encounter block production issues. 

The `mainnet-staging` branch as of https://github.com/AleoHQ/snarkOS/commit/3b4737598690881e4b438ac5e19403724a2e9fa0 (without this change), still ran into halting issues with that same test scenario.

howardwu (approved):

PR #3119, opened 2024-02-23
[Fix] Updates conditions to advance from odd and even rounds 

mdelle1:
## Motivation

This PR updates the conditions to advance from odd and even rounds by adding a quorum check. Previously, in odd rounds the code would return ‘true’ if the previous even round leader was unseen [1]. In some scenarios, this would cause nodes to pre-maturely increment rounds before seeing ‘2f+1’ certificates, the quorum threshold to increment rounds in Bullshark. For example, if the leader in even round ‘r’ was unseen, in odd round ‘r+1’ a malicious node could send a batch header with a round number beyond the GC limit. When nodes try to sync with the batch header [2], they will increment their BFT to the next round since the previous leader was unseen. 

[1] https://github.com/AleoHQ/snarkOS/blob/mainnet/node/bft/src/bft.rs#L370 
[2] https://github.com/AleoHQ/snarkOS/blob/mainnet/node/bft/src/primary.rs#L1249

### DIFF
--- a/node/bft/src/bft.rs
+++ b/node/bft/src/bft.rs
@@ -323,10 +323,9 @@ impl<N: Network> BFT<N> {
         self.is_even_round_ready_for_next_round(current_certificates, committee_lookback, current_round)
     }
 
-    /// Returns 'true' under one of the following conditions:
-    ///  - If the leader certificate is set for the current even round,
-    ///  - The timer for the leader certificate has expired, and we can
-    ///    achieve quorum threshold (2f + 1) without the leader.
+    /// Returns 'true' if the quorum threshold is reached for this round under one of the following conditions:
+    ///  - If the leader certificate is set for the current even round.
+    ///  - The timer for the leader certificate has expired.
     fn is_even_round_ready_for_next_round(
         &self,
         certificates: IndexSet<BatchCertificate<N>>,
@@ -360,9 +359,8 @@ impl<N: Network> BFT<N> {
         self.leader_certificate_timer.load(Ordering::SeqCst) + MAX_LEADER_CERTIFICATE_DELAY_IN_SECS <= now()
     }
 
-    /// Returns 'true' if any of the following conditions hold:
-    ///  - The leader certificate is 'None'.
-    ///  - The leader certificate reached quorum threshold `(2f + 1)` (in the previous certificates in the current round).
+    /// Returns 'true' if the quorum threshold is reached for this round under one of the following conditions:
+    ///  - The leader certificate is `None`.
     ///  - The leader certificate is not included up to availability threshold `(f + 1)` (in the previous certificates of the current round).
     ///  - The leader certificate timer has expired.
     fn is_leader_quorum_or_nonleaders_available(&self, odd_round: u64) -> bool {
@@ -853,6 +851,7 @@ mod tests {
     use snarkos_node_bft_ledger_service::MockLedgerService;
     use snarkos_node_bft_storage_service::BFTMemoryService;
     use snarkvm::{
+        console::account::{Address, PrivateKey},
         ledger::{
             committee::Committee,
             narwhal::batch_certificate::test_helpers::{sample_batch_certificate, sample_batch_certificate_for_round},
@@ -861,7 +860,7 @@ mod tests {
     };
 
     use anyhow::Result;
-    use indexmap::IndexSet;
+    use indexmap::{IndexMap, IndexSet};
     use std::sync::Arc;
 
     type CurrentNetwork = snarkvm::console::network::MainnetV0;
--- a/node/bft/src/bft.rs
+++ b/node/bft/src/bft.rs
@@ -398,11 +398,9 @@ impl<N: Network> BFT<N> {
             // If there is no leader certificate for the previous round, return 'true'.
             return true;
         };
-        // Retrieve the leader certificate ID.
-        let leader_certificate_id = leader_certificate.id();
         // Compute the stake for the leader certificate.
         let (stake_with_leader, stake_without_leader) =
-            self.compute_stake_for_leader_certificate(leader_certificate_id, current_certificates, &committee_lookback);
+            self.compute_stake_for_leader_certificate(leader_certificate.id(), current_certificates, &committee_lookback);
         // Return 'true' if any of the following conditions hold:
         stake_with_leader >= committee_lookback.availability_threshold()
             || stake_without_leader >= committee_lookback.quorum_threshold()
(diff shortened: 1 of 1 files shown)


howardwu:
@mdelle1 can you address this PR's issues?

mimoo (approved):
good catch

node/bft/src/bft.rs
```diff
@@ -322,6 +322,13 @@ impl<N: Network> BFT<N> {
         committee: Committee<N>,
         current_round: u64,
     ) -> bool {
+        // Retrieve the authors for the current round.
```
vicsn, line 335: Should the function comment `Returns 'true' under one of the following conditions:` also be corrected, or was it already correct?

node/bft/src/bft.rs
```diff
@@ -381,7 +377,20 @@ impl<N: Network> BFT<N> {
                 return false;
             }
         };
-
+        // Retrieve the authors of the current certificates.
```
vicsn, line 389: Should the function comment `Returns 'true' if any of the following conditions hold:` also be corrected, or was it already correct?

node/bft/src/bft.rs
raychu86: I've updated the documentation to be more clear.

node/bft/src/bft.rs
raychu86: I've updated the documentation to be more clear.

raychu86:
Note: This PR was tested on a network with a malicious validator that was spoofing rounds and did not run into block production issues.

vicsn:
I ran this PR (c9e9d01) with 5 nodes on my laptop and every node periodically gets an error, after this PR it should be a debug statement because we won't be ready until we see quorum.
```
2024-03-12T10:52:50.035313Z ERROR BFT - A leader certificate was found, but 'is_ready' is false is_ready=false
```

I ran mainnet-latest (3b4737598) and it indeed didn't have the same issue.

node/bft/src/bft.rs
```diff
@@ -392,7 +386,20 @@ impl<N: Network> BFT<N> {
...
+        // Retrieve the leader certificate.
+        let Some(leader_certificate) = self.leader_certificate.read().clone() else {
+            // If there is no leader certificate for the previous round, return 'true'.
+            return true;
+        };
+        // Retrieve the leader certificate ID.
+        let leader_certificate_id = leader_certificate.id();
         // Compute the stake for the leader certificate.
         let (stake_with_leader, stake_without_leader) =
             self.compute_stake_for_leader_certificate(leader_certificate_id, current_certificates, &committee_lookback);
```
howardwu:
```suggestion
        // Compute the stake for the leader certificate.
        let (stake_with_leader, stake_without_leader) =
            self.compute_stake_for_leader_certificate(leader_certificate.id(), current_certificates, &committee_lookback);
```

node/bft/src/bft.rs
```diff
@@ -392,7 +386,20 @@ impl<N: Network> BFT<N> {
...
-
+        // Retrieve the authors of the current certificates.
+        let authors = current_certificates.clone().into_iter().map(|c| c.author()).collect();
+        // Check if quorum threshold is reached.
+        if !committee_lookback.is_quorum_threshold_reached(&authors) {
+            debug!("BFT failed reach quorum threshold in odd round {current_round}. ");
+            return false;
+        }
+        // Retrieve the leader certificate.
+        let Some(leader_certificate) = self.leader_certificate.read().clone() else {
```
howardwu, line 397: Why was this moved down? We can return early if the leader certificate is not found, and not incur the quorum threshold check.

node/bft/src/bft.rs
```diff
@@ -356,9 +359,8 @@ impl<N: Network> BFT<N> {
         self.leader_certificate_timer.load(Ordering::SeqCst) + MAX_LEADER_CERTIFICATE_DELAY_IN_SECS <= now()
     }
 
-    /// Returns 'true' if any of the following conditions hold:
-    ///  - The leader certificate is 'None'.
-    ///  - The leader certificate reached quorum threshold `(2f + 1)` (in the previous certificates in the current round).
+    /// Returns 'true' if the quorum threshold is reached for this round under one of the following conditions:
```
howardwu:
```suggestion
    /// Returns 'true' if the quorum threshold `(2f + 1)` is reached for this round under one of the following conditions:
```

node/bft/src/bft.rs
```diff
@@ -323,16 +323,22 @@ impl<N: Network> BFT<N> {
         self.is_even_round_ready_for_next_round(current_certificates, committee_lookback, current_round)
     }
 
-    /// Returns 'true' under one of the following conditions:
-    ///  - If the leader certificate is set for the current even round,
-    ///  - The timer for the leader certificate has expired, and we can
-    ///    achieve quorum threshold (2f + 1) without the leader.
+    /// Returns 'true' if the quorum threshold is reached for this round under one of the following conditions:
```
howardwu:
```suggestion
    /// Returns 'true' if the quorum threshold `(2f + 1)` is reached for this round under one of the following conditions:
```

node/bft/src/bft.rs
```diff
@@ -392,10 +386,21 @@ impl<N: Network> BFT<N> {
                 return false;
             }
         };
-
+        // Retrieve the authors of the current certificates.
+        let authors = current_certificates.clone().into_iter().map(|c| c.author()).collect();
+        // Check if quorum threshold is reached.
+        if !committee_lookback.is_quorum_threshold_reached(&authors) {
```
raychu86, line 392: Why is this particular `quroum_threshold_reached` check needed? Does this not invalidate the subsequent checks below?

```
 `stake_with_leader >= committee_lookback.availability_threshold()
            || stake_without_leader >= committee_lookback.quorum_threshold()`
```

node/bft/src/bft.rs
mdelle1: If the leader was not found, you should advance to the next round only if you've seen a quorum of certificates. Otherwise, nodes may prematurely jump to the next round on a faulty leader before seeing quorum.

node/bft/src/bft.rs
mdelle1: It wouldn't invalidate the previous check because you can still see 'f+1' votes for the leader but not see '2f+1' certificates in the round. Here you should commit the leader, but wait to see '2f+1' certificates before advancing to the next round. Previously, the code was also returning 'true' if the timer had expired. But again, with an expired timer we should still include a quorum check before advancing.

howardwu (approved):

PR #3121, opened 2024-02-23
[Optimize] Adjust blocking tasks in the primary

ljedrz:
This PR adjusts the placement of blocking tasks in the `primary` module based on how long the blocking operations can take, so that we avoid blocking the async executor.

### DIFF
--- a/node/bft/src/primary.rs
+++ b/node/bft/src/primary.rs
@@ -630,7 +630,7 @@ impl<N: Network> Primary<N> {
         let BatchSignature { batch_id, signature } = batch_signature;
 
         // Retrieve the signer.
-        let signer = spawn_blocking!(Ok(signature.to_address()))?;
+        let signer = signature.to_address();
 
         // Ensure the batch signature is signed by the validator.
         if self.gateway.resolver().get_address(peer_ip).map_or(true, |address| address != signer) {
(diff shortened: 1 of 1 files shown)


node/bft/src/primary.rs
```diff
@@ -630,7 +630,7 @@ impl<N: Network> Primary<N> {
         let BatchSignature { batch_id, signature } = batch_signature;
 
         // Retrieve the signer.
-        let signer = spawn_blocking!(Ok(signature.to_address()))?;
+        let signer = signature.to_address();
```
ljedrz, line 633: note: this operation was found to be fast enough to not require blocking

node/bft/src/primary.rs
```diff
@@ -1113,14 +1118,16 @@ impl<N: Network> Primary<N> {
...
     async fn store_and_broadcast_certificate(&self, proposal: &Proposal<N>, committee: &Committee<N>) -> Result<()> {
-        // Create the batch certificate and transmissions.
-        let (certificate, transmissions) = proposal.to_certificate(committee)?;
-        // Convert the transmissions into a HashMap.
-        // Note: Do not change the `Proposal` to use a HashMap. The ordering there is necessary for safety.
-        let transmissions = transmissions.into_iter().collect::<HashMap<_, _>>();
-        // Store the certified batch.
-        let (storage, certificate_) = (self.storage.clone(), certificate.clone());
-        spawn_blocking!(storage.insert_certificate(certificate_, transmissions))?;
+        let certificate = tokio::task::block_in_place(|| {
```
ljedrz, line 1121: note: this is done via `block_in_place` instead of `spawn_blocking` in order to not have to clone the `Proposal` (which could get expensive enough to defeat the purpose)

node/bft/src/primary.rs
```diff
@@ -1114,7 +1119,7 @@ impl<N: Network> Primary<N> {
     /// Stores the certified batch and broadcasts it to all validators, returning the certificate.
     async fn store_and_broadcast_certificate(&self, proposal: &Proposal<N>, committee: &Committee<N>) -> Result<()> {
         // Create the batch certificate and transmissions.
-        let (certificate, transmissions) = proposal.to_certificate(committee)?;
+        let (certificate, transmissions) = tokio::task::block_in_place(|| proposal.to_certificate(committee))?;
```
ljedrz, line 1122: note: this is done via `block_in_place` instead of `spawn_blocking` in order to not have to clone the `Proposal`, which could get expensive enough to defeat the purpose

node/bft/src/primary.rs
```diff
@@ -1840,7 +1845,7 @@ mod tests {
         );
     }
 
-    #[tokio::test]
+    #[tokio::test(flavor = "multi_thread")]
```
ljedrz, line 1848: note: this is a requirement for using `block_in_place`

ljedrz:
These changes were tested using `tokio-console` and 4 validators in `--dev` mode; the code of the non-blocking `Primary`-related tasks that were showing high `busy` values was investigated for the presence of blocking operations, and once those were put inside blocking tasks, the `busy` values have dropped, meaning the async worker threads and the executor were no longer being blocked (i.e. slowed down).

howardwu (approved):

PR #3123, opened 2024-02-23
[Optimize] Use a blocking task to execute test transactions

ljedrz:
This is likely (depending on the resources and the number of nodes) going to make a noticeable difference in local `--dev` runs, and make those runs more predictable.

### DIFF
--- a/node/src/validator/mod.rs
+++ b/node/src/validator/mod.rs
@@ -16,7 +16,7 @@ mod router;
 
 use crate::traits::NodeInterface;
 use snarkos_account::Account;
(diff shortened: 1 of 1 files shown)


ljedrz:
These changes were tested using `tokio-console` and 4 validators in `--dev` mode; the code of the non-blocking tasks of the `dev 0` node that were showing high `busy` values was investigated for the presence of blocking operations, and the transaction loop was a clear outlier. Once it was wrapped in a blocking task, the corresponding `busy` value dropped, meaning the async worker threads and the executor were no longer being blocked (i.e. slowed down).

howardwu (approved):

PR #3129, opened 2024-02-26
[ZKS-02] Add committee ID checks when processing proposals and certificates

raychu86:
## Motivation

This PR adds an additional committee ID check when processing batch proposals and batch certificates from peers. This check guarantees that the proposal/certificate from a validator must have the same committee id that your validator node expects for the given round, which means both validators have consistent committee state.

If this committee ID from a peer's batch proposal is not what was expected, that means the peer is either being malicious or has deviated state. The new logic will now ignore the proposal and disconnect from the peer in an attempt to not perpetuate a subdag based on deviated committee state.

The sister PR can be found here: https://github.com/AleoHQ/snarkVM/pull/2374

Audit Finding: **[zksecurity 02] Dynamic Committee Feature is Not Safe**

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -46,7 +46,7 @@ default-features = false
 
 [workspace.dependencies.snarkvm]
 git = "https://github.com/AleoHQ/snarkVM.git"
-rev = "569cf5a"
+rev = "ee10658"
 #version = "=0.16.18"
 features = [ "circuit", "console", "rocks" ]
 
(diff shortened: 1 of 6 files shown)


acoglio (approved):

Cargo.toml
```diff
@@ -46,7 +46,7 @@ default-features = false
 
 [workspace.dependencies.snarkvm]
 git = "https://github.com/AleoHQ/snarkVM.git"
-rev = "569cf5a"
+rev = "ee10658"
```
howardwu:
```suggestion
rev = "46f2625"
```

howardwu (approved):

PR #3130, opened 2024-02-26
Development experience improvements to devnet.sh

vvp:
This PR introduces two improvements to development experience when using devnet.sh to start dev nodes locally. 

**Support for custom executables**. This PR enables devnet.sh to start local devnet nodes using a custom executable instead of currently installed `snarkos` executable, by setting the path to an environment variable. For example:
```
export SNARKOS_BIN=target/release/snarkos
./devnet.sh
```
When `SNARKOS_BIN` is set, the script skips the question of whether to rebuild+install the `snarkos` executable.

**Auto-restart option**. This PR also implements a simple auto-restart for `snarkOS`-processes using a infinite while-loop in bash. With this, you can test dev nodes going off and back online simply by sending `SIGTERM` signals to processes. The length of the auto-restart delay can be configured during script startup.

### DIFF
--- a/devnet.sh
+++ b/devnet.sh
@@ -1,5 +1,12 @@
 #!/bin/bash
 
+# Check if using a non-installed binary via environment variable (e.g. SNARKOS_BIN=target/release/snarkos)
+executable=${SNARKOS_BIN:-"snarkos"}
+
+if [[ $executable != "snarkos" ]]; then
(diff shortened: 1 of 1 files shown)


kaimast:
I will close this for now due to being inactive for over a year and conflicts with upstream changes. If you want to us to still consider this change, please resolve the conflicts and reply to this PR.

PR #3132, opened 2024-02-27
[Feature] Add --no-dev-txs flag

vicsn:
## Motivation

Adds a `--no-dev-txs` flag which determines if node 0 generates transactions if the `--dev` flag is enabled.

## Test Plan

Manually ran the `devnet.sh` script with and without the `--no-dev-txs` flag to observe correct behaviour.

### DIFF
--- a/cli/src/commands/start.rs
+++ b/cli/src/commands/start.rs
@@ -138,7 +138,7 @@ pub struct Start {
     #[clap(long)]
     pub dev_num_validators: Option<u16>,
     /// If developtment mode is enabled, specify whether node 0 should generate traffic to drive the network
-    #[clap(long = "dev-traffic")]
+    #[clap(default_value = "false", long = "dev-traffic")]
     pub dev_traffic: bool,
     /// Specify the path to a directory containing the ledger
     #[clap(long = "storage_path")]
--- a/cli/src/commands/start.rs
+++ b/cli/src/commands/start.rs
@@ -137,6 +137,9 @@ pub struct Start {
     /// If development mode is enabled, specify the number of genesis validators (default: 4)
     #[clap(long)]
     pub dev_num_validators: Option<u16>,
+    /// If developtment mode is enabled, specify whether node 0 should generate traffic to drive the network
+    #[clap(long = "dev-traffic")]
+    pub dev_traffic: bool,
     /// Specify the path to a directory containing the ledger
     #[clap(long = "storage_path")]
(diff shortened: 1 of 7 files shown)


joske:
There's also https://github.com/AleoHQ/snarkOS/pull/2822, which I think is more useful (but is outdated).

joske:
@vicsn I've updated the linked PR.

cli/src/commands/start.rs
```diff
@@ -137,6 +137,9 @@ pub struct Start {
     /// If development mode is enabled, specify the number of genesis validators (default: 4)
     #[clap(long)]
     pub dev_num_validators: Option<u16>,
+    /// If developtment mode is enabled, specify whether node 0 should generate traffic to drive the network
+    #[clap(long = "dev-traffic")]
+    pub dev_traffic: bool,
```
howardwu: Is this `true` by default?

node/src/validator/mod.rs
```diff
@@ -82,6 +82,7 @@ impl<N: Network, C: ConsensusStorage<N>> Validator<N, C> {
         trusted_validators: &[SocketAddr],
         genesis: Block<N>,
         cdn: Option<String>,
+        dev_traffic: bool,
         storage_mode: StorageMode,
```
howardwu: nit: can you swap the order between `storage_mode` and `dev_traffic`?

cli/src/commands/start.rs
vicsn: It is `false` by default, made it explicit: d3cc72bbd

I realise this *may* lead to issues in the community because nodes require this flag to auto-advance blocks. But I think it is the right default for testing and the updated README 281acb5d3, `devnet.sh` should be sufficient guidance.

Let me know if you prefer an alternative `dev-no-traffic` command.

node/src/validator/mod.rs
vicsn: 71230f240 Sure, though note `cargo fmt` then makes the comment indentation different:
Screenshot 2024-03-03 at 19 37 36

cli/src/commands/start.rs
howardwu: It should be true by default.

Can use `no-dev-txs` as the flag name. Traffic implies network traffic among clients/validators.

vicsn: Done

de2bd055c

8a0145c35

howardwu (approved):

PR #3135, opened 2024-02-28
Unify pending object and clear expired items from pending queue

raychu86:
## Motivation

This PR unifies the `pending` and `callbacks` objects in the `Pending` struct for better tracking/access.

This PR also adds additional logic to properly clear the items from the `Pending` queue. Previously, we would clear the expired callbacks, but not clear the items from the `pending` queue; this leave an artifact in the queue for every unique element that was not `removed` explicitly. 

In addition, we were only clearing items when an `insert`, `remove`, or `num_callbacks` function was called for that particular item. So if for any reason, the nodes are unable to complete the pending request, the item/callbacks will stay in the queue until the node is restarted. 

The new changes will properly clear the expired items and periodically clear the pending queue of the expired callbacks and items. This should mitigate the incremental memory growth and also the redundant requests issue.

The `CALLBACK_EXPIRATION_IN_SECS` was also updated to prevent truncation, which should address the case where the callback expired but the fetch timeout hasn't. 

## Related PRs

https://github.com/AleoHQ/snarkOS/pull/3088


node/bft/src/helpers/pending.rs
```diff
@@ -128,12 +131,29 @@ impl<T: Copy + Clone + PartialEq + Eq + Hash, V: Clone> Pending<T, V> {
...
             // Fetch the current timestamp.
             let now = OffsetDateTime::now_utc().unix_timestamp();
             // Remove the callbacks that have expired.
-            callbacks.retain(|(_, timestamp)| now - *timestamp <= CALLBACK_TIMEOUT_IN_SECS);
+            callbacks.retain(|(_, timestamp)| now - *timestamp <= CALLBACK_EXPIRATION_IN_SECS);
+        }
+
+        // If there are no more remaining callbacks for the item, remove the item from the pending queue.
+        // If the item was added to the pending queue without a callback, do not remove the item.
+        let mut callbacks = self.callbacks.lock();
```
ljedrz: we can obtain this guard just once (and bind it), before clearing the expired callbacks, and it will be applicable for both blocks

ljedrz: also, since we're querying the same item twice, we could do it with a single query with the `Entry` API to avoid borrowing issues (a matched `Entry::Occupied` can be removed from the collection)

raychu86: Good catch, I've made it a singular lock. I left the `get_mut` and `remove` approach rather than changing to the `Entry` API, but can easily be amended if `Entry` is a better approach.

ljedrz: as long as the borrow checker is happy, `get_mut` is just fine :+1:

node/bft/src/helpers/pending.rs
```diff
@@ -128,12 +131,29 @@ impl<T: Copy + Clone + PartialEq + Eq + Hash, V: Clone> Pending<T, V> {
...
+                self.pending.write().remove(&item);
+            }
+        }
+    }
+
+    /// Removes the callbacks for all items have that expired.
+    pub fn clear_expired_callbacks(&self) {
+        let items = self.pending.read().keys().copied().collect::<Vec<T>>();
+        for item in items.into_iter() {
+            self.clear_expired_callbacks_for_item(item);
```
ljedrz: I missed this earlier, but if we can expect many items here, it might be a good idea to obtain the current timestamp and the lock over `callbacks` before this loop

ljedrz (approved):
LGTM, left 1 last comment with an additional, optional perf improvement

joske (approved):

node/bft/src/helpers/pending.rs
```diff
@@ -105,10 +105,10 @@ impl<T: Copy + Clone + PartialEq + Eq + Hash, V: Clone> Pending<T, V> {
         // Clear the callbacks that have expired.
         self.clear_expired_callbacks_for_item(item);
         // Return the number of live callbacks.
-        self.callbacks
-            .lock()
+        self.pending
+            .write()
```
vvp:
```suggestion
            .read()
```

node/bft/src/helpers/pending.rs
```diff
@@ -105,10 +105,10 @@ impl<T: Copy + Clone + PartialEq + Eq + Hash, V: Clone> Pending<T, V> {
...
             .get(&item)
-            .map_or(0, |callbacks| callbacks.iter().filter(|(_, _, request_sent)| *request_sent).count())
+            .map_or(0, |peers| peers.values().flatten().filter(|(_, _, request_sent)| *request_sent).count())
```
vvp, line 113: Since `num_sent_requests()` will be called very frequently by the `Worker::send_transmission_request()` and iterating nested hashmaps isn't constant-time, it might make sense to consider memoizing these counts in `Pending`-struct.

ljedrz: In addition, if we do need a linear lookup here, we might want to call `.clone()` after the `.get` in order to release the lock sooner

raychu86: I believe since we are now clearing the callbacks after expiration, the size of this vec shouldn't grow too large.

I'd like to keep the change-set small enough s.t. it is easy to review, but would be open to further optimization PRs down the line.

node/bft/src/helpers/pending.rs
```diff
@@ -82,33 +82,35 @@ impl<T: Copy + Clone + PartialEq + Eq + Hash, V: Clone> Pending<T, V> {
...
     /// Returns the number of pending callbacks for the specified `item`.
     pub fn num_callbacks(&self, item: impl Into<T>) -> usize {
         let item = item.into();
+        let now = OffsetDateTime::now_utc().unix_timestamp();
         // Clear the callbacks that have expired.
-        self.clear_expired_callbacks_for_item(item);
+        self.clear_expired_callbacks_for_item(now, item);
         // Return the number of live callbacks.
-        self.callbacks.lock().get(&item).map_or(0, |callbacks| callbacks.len())
+        self.pending.read().get(&item).map_or(0, |peers| peers.values().flatten().count())
```
ljedrz:
```suggestion
        self.pending.read().get(&item).map_or(0, |peers| peers.values().map(|v| v.len()).sum::<usize>())
```
I think the flattening would cause the counting to be done in a linear fashion instead of using the readily available lengths

node/bft/src/helpers/pending.rs
```diff
@@ -146,29 +160,55 @@ impl<T: Copy + Clone + PartialEq + Eq + Hash, V: Clone> Pending<T, V> {
...
-        if let Some(callbacks) = self.callbacks.lock().remove(&item) {
-            if let Some(callback_value) = callback_value {
-                // Send a notification to the callback.
-                for (callback, _, _) in callbacks {
-                    callback.send(callback_value.clone()).ok();
+        // Remove the item from the pending queue and process any remaining callbacks.
+        match self.pending.write().remove(&item) {
+            Some(callbacks) => {
+                // Get the peer IPs.
+                let peer_ips = callbacks.keys().cloned().collect();
```
ljedrz:
```suggestion
                let peer_ips = callbacks.keys().copied().collect();
```
since `SocketAddr` is `Copy`

node/bft/src/helpers/pending.rs
```diff
@@ -146,29 +160,55 @@ impl<T: Copy + Clone + PartialEq + Eq + Hash, V: Clone> Pending<T, V> {
...
+                pending.remove(&item);
+            }
+        }
+    }
+
+    /// Removes the callbacks for all items have that expired.
+    pub fn clear_expired_callbacks(&self) {
+        let items = self.pending.read().keys().copied().collect::<Vec<T>>();
+        let now = OffsetDateTime::now_utc().unix_timestamp();
+        for item in items.into_iter() {
```
ljedrz: once this loop is done, it might make sense to call `.shrink_to_fit` on `self.pending`, as long as we don't call this method too often

node/bft/src/helpers/pending.rs
```diff
@@ -146,29 +160,55 @@ impl<T: Copy + Clone + PartialEq + Eq + Hash, V: Clone> Pending<T, V> {
...
-        result
     }
 
     /// Removes the callbacks for the specified `item` that have expired.
-    pub fn clear_expired_callbacks_for_item(&self, item: impl Into<T>) {
+    pub fn clear_expired_callbacks_for_item(&self, now: i64, item: impl Into<T>) {
+        let item = item.into();
+
+        // Acquire the pending lock.
+        let mut pending = self.pending.write();
```
ljedrz, line 187: this lock is acquired in a loop; it might make sense to tweak the API so that it's done just once for the duration of the `clear_expired_callbacks` loop

node/bft/src/sync/mod.rs
```diff
@@ -118,6 +118,18 @@ impl<N: Network> Sync<N> {
...
 
+        // Start the pending queue expiration loop.
+        let self_ = self.clone();
+        self.spawn(async move {
+            loop {
+                // Sleep briefly.
+                tokio::time::sleep(Duration::from_millis(MAX_FETCH_TIMEOUT_IN_MS)).await;
+
+                // Remove the expired pending transmission requests.
+                self_.pending.clear_expired_callbacks();
```
ljedrz: we might want this to be called within a blocking task

node/bft/src/worker.rs
```diff
@@ -354,6 +354,18 @@ impl<N: Network> Worker<N> {
...
 
+        // Start the pending queue expiration loop.
+        let self_ = self.clone();
+        self.spawn(async move {
+            loop {
+                // Sleep briefly.
+                tokio::time::sleep(Duration::from_millis(MAX_FETCH_TIMEOUT_IN_MS)).await;
+
+                // Remove the expired pending certificate requests.
+                self_.pending.clear_expired_callbacks();
```
ljedrz: ditto

ljedrz (commented):
Left a few new comments and suggestions.

node/bft/src/helpers/pending.rs
```diff
@@ -324,6 +369,53 @@ mod tests {
...
+    #[test]
+    fn test_expired_items() {
+        let rng = &mut TestRng::default();
+
+        // Initialize the ready queue.
+        let pending = Pending::<TransmissionID<CurrentNetwork>, ()>::new();
+
+        // Check initially empty.
+        assert!(pending.is_empty());
+        assert_eq!(pending.len(), 0);
```
niklaslong, line 394: No need to check both the length and `is_empty()`, the latter checks `len == 0` in the `hashbrown` implementation (which is used internally by `std`).

node/bft/src/helpers/pending.rs
```diff
@@ -324,6 +369,53 @@ mod tests {
...
+        // Wait for ` CALLBACK_EXPIRATION_IN_SECS + 1` seconds.
+        thread::sleep(Duration::from_secs(CALLBACK_EXPIRATION_IN_SECS as u64 + 1));
+
+        // Expire the pending callbacks.
+        pending.clear_expired_callbacks();
+
+        // Ensure that the items have been expired.
+        assert_eq!(pending.num_callbacks(solution_id_1), 0);
+        assert_eq!(pending.num_callbacks(solution_id_2), 0);
+        assert_eq!(pending.len(), 0);
```
niklaslong:
```suggestion
        assert!(pending.is_empty());
```

niklaslong (approved):
LGTM pending Lukasz's comments. Left a few small suggestions.

node/bft/src/helpers/pending.rs
niklaslong: (ditto in the other tests)

node/bft/src/worker.rs
raychu86: Resolved with https://github.com/AleoHQ/snarkOS/pull/3135/commits/a39996d213f0a3e9c2556bb305d1fce2e00bc1d6

node/bft/src/sync/mod.rs
raychu86: Resolved with https://github.com/AleoHQ/snarkOS/pull/3135/commits/a39996d213f0a3e9c2556bb305d1fce2e00bc1d6

node/bft/src/helpers/pending.rs
raychu86: Resolved with https://github.com/AleoHQ/snarkOS/pull/3135/commits/a39996d213f0a3e9c2556bb305d1fce2e00bc1d6

node/bft/src/helpers/pending.rs
raychu86: Since this is testing code only and it doesn't necessarily hurt, I'll elect to leave it in for changing less LOC.

node/bft/src/helpers/pending.rs
raychu86: @ljedrz We call `clear_expired_callbacks` quite often - on an interval, so I'd imagine we'd need some way to determine when it'd be appropriate to call `shrink_to_fit`.

node/bft/src/helpers/pending.rs
raychu86: Resolved in https://github.com/AleoHQ/snarkOS/pull/3135/commits/964599be38d5dfe56f5d4701ad4cccf98819c8d1. 

This also removes a use of `collect`. 

@ljedrz Can you review the new change?

node/bft/src/helpers/pending.rs
ljedrz: I don't recall having seen this collection in the past heap profiles, but I'll keep it in mind, then 👍.

node/bft/src/helpers/pending.rs
ljedrz: Looks good :+1:; we can now consider if we still need the `clear_expired_callbacks_for_item` method.

raychu86: We still use the `clear_expired_callbacks_for_item` on individual calls to `num_sent_requests`, `num_callbacks`, and `insert`. Primarily to clear the expired callbacks in order to fetch the correct output number.

ghostant-1017:
LGTM

PR #3139, opened 2024-02-29
Update fetch timeouts as a factor of committee size

raychu86:
## Motivation

This PR updates the `MAX_FETCH_TIMEOUT_IN_MS` to a dynamic timeout based on the number of validators in the committee set. 

The proposed formula for the fetch timeout is `BASE_FETCH_TIMEOUT_IN_MS` + `X`ms per validator in the committee set with a maximum of 30 seconds. This `X` is currently set to 200ms, but can be adjusted based on need. 

 The `BASE_FETCH_TIMEOUT_IN_MS` is the original `MAX_FETCH_TIMEOUT_IN_MS` from before the change, so some example values for timeouts would be:

10 validators => 200 * 10 + 7500 = 9500ms
50 validators => 200 * 50 + 7500 = 17500ms
100 validators => 200 * 100 + 7500 = 27500ms
113+ validators => 30000ms

## Related PRs

Based on https://github.com/AleoHQ/snarkOS/pull/3135

### DIFF
--- a/node/bft/src/lib.rs
+++ b/node/bft/src/lib.rs
@@ -48,14 +48,14 @@ pub const MEMORY_POOL_PORT: u16 = 5000; // port
 
 /// The maximum number of milliseconds to wait before proposing a batch.
 pub const MAX_BATCH_DELAY_IN_MS: u64 = 2500; // ms
-/// The maximum number of milliseconds to wait before timing out on a fetch.
-pub const MAX_FETCH_TIMEOUT_IN_MS: u64 = 3 * MAX_BATCH_DELAY_IN_MS; // ms
 /// The maximum number of seconds allowed for the leader to send their certificate.
 pub const MAX_LEADER_CERTIFICATE_DELAY_IN_SECS: i64 = 2 * MAX_BATCH_DELAY_IN_MS as i64 / 1000; // seconds
 /// The maximum number of seconds before the timestamp is considered expired.
 pub const MAX_TIMESTAMP_DELTA_IN_SECS: i64 = 10; // seconds
 /// The maximum number of workers that can be spawned.
 pub const MAX_WORKERS: u8 = 1; // worker(s)
+/// The base number of milliseconds to wait before timing out on a fetch.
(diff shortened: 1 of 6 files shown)


node/bft/src/lib.rs
```diff
@@ -73,3 +73,19 @@ macro_rules! spawn_blocking {
...
+    const MAX_FETCH_TIMEOUT_IN_MS: u64 = 30000; // 30 seconds
+
+    // Calculate the timeout.
+    let timeout = BASE_FETCH_TIMEOUT_IN_MS + 200 * num_validators;
+
+    // Bound the timeout to the maximum allowed fetch timeout.
+    match timeout > MAX_FETCH_TIMEOUT_IN_MS {
+        true => MAX_FETCH_TIMEOUT_IN_MS,
+        false => timeout,
+    }
```
ljedrz, line 90:
```suggestion
    std::cmp::min(timeout, MAX_FETCH_TIMEOUT_IN_MS)
```
nit

node/bft/src/sync/mod.rs
```diff
@@ -436,8 +443,14 @@ impl<N: Network> Sync<N> {
                 trace!("Skipped sending redundant request for certificate {} to '{peer_ip}'", fmt_id(certificate_id));
             }
         }
+        // Calculate the dynamic fetch timeout in milliseconds.
+        let num_validators = self
+            .ledger
+            .get_committee_lookback_for_round(self.storage.current_round())
+            .map_or(Committee::<N>::MAX_COMMITTEE_SIZE as u64, |committee| committee.num_members() as u64);
+        let timeout_in_ms = max_fetch_timeout_in_ms(num_validators);
```
ljedrz, line 451: since this is the next time this is calculated, maybe a dedicated method would be a good idea?

ljedrz: (I can also see this done in the worker module)

node/bft/src/lib.rs
raychu86: This won't compile because it's a `const` function

ljedrz: ah, I forgot `min` is not `const` yet :+1:

PR #3140, opened 2024-03-01
[Fix] Update fetch redundancy as a factor of committee size

raychu86:
## Motivation

This PR updates the `NUM_REDUNDANT_REQUESTS` to a dynamic number based on the number of validators in the committee set.

The current formula for maximum allowed request redundancy is  `1 + (num_validators_in_committee_lookback / 3)`. This can be adjusted if it is too lenient or too aggressive.

Additionally, the `num_pending_requests` that was used to measure redundancy is now adjusted to the number of sent `Event::TransmissionRequest`/`Event::CertificateRequest`. Previously we were using the number of callbacks, which was not a good gauge of how many request messages were actually being sent out to peers.

## Related PRs

Related to #3135 and #3139

### DIFF
--- a/node/bft/src/helpers/pending.rs
+++ b/node/bft/src/helpers/pending.rs
@@ -27,7 +27,10 @@ const CALLBACK_TIMEOUT_IN_SECS: i64 = MAX_FETCH_TIMEOUT_IN_MS as i64 / 1000;
 
 /// Returns the maximum number of redundant requests for the specified number of validators.
 pub const fn max_redundant_requests(num_validators: usize) -> usize {
-    num_validators.saturating_div(2)
+    // Note: It is adequate to set this value to the availability threshold,
+    // as with high probability one will respond honestly (in the best and worst case
+    // with stake spread across the validators evenly and unevenly, respectively).
+    1 + num_validators.saturating_div(3)
 }
 
 #[derive(Debug)]
--- a/node/bft/src/helpers/pending.rs
+++ b/node/bft/src/helpers/pending.rs
@@ -29,7 +29,7 @@ use tokio::sync::oneshot;
 const CALLBACK_TIMEOUT_IN_SECS: i64 = MAX_FETCH_TIMEOUT_IN_MS as i64 / 1000;
 
 /// Returns the maximum number of redundant requests for the number of validators in the specified round.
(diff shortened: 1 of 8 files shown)


node/bft/src/helpers/pending.rs
```diff
@@ -23,19 +23,20 @@ use std::{
...
-#[cfg(not(test))]
-pub const NUM_REDUNDANT_REQUESTS: usize = 2;
-#[cfg(test)]
-pub const NUM_REDUNDANT_REQUESTS: usize = 10;
-
 const CALLBACK_TIMEOUT_IN_SECS: i64 = MAX_FETCH_TIMEOUT_IN_MS as i64 / 1000;
 
+/// Returns the maximum number of redundant requests for the specified number of validators.
+pub const fn max_redundant_requests(num_validators: usize) -> usize {
+    num_validators.saturating_div(2)
```
howardwu:
```suggestion
    // Note: It is adequate to set this value to the availability threshold,
    // as with high probability one will respond honestly (in the best and worst case
    // with stake spread across the validators evenly and unevenly, respectively).
    1 + num_validators.saturating_div(3)
```

It is adequate to set this value to the **availability threshold**, as with high probability one will respond honestly (in the best case and worst case with stake spread across the validators evenly and unevenly, respectively).

Thus, the proposed change is to set it to `1 + num_validators.saturating_div(3)`.

node/bft/src/worker.rs
```diff
@@ -384,18 +390,30 @@ impl<N: Network> Worker<N> {
...
-        if num_pending_requests <= NUM_REDUNDANT_REQUESTS {
+        if should_send_request {
             // Send the transmission request to the peer.
             if self.gateway.send(peer_ip, Event::TransmissionRequest(transmission_id.into())).await.is_none() {
                 bail!("Unable to fetch transmission - failed to send request")
             }
         } else {
-            trace!("Skipped sending redundant request for transmission {} to '{peer_ip}'", fmt_id(transmission_id));
+            debug!(
+                "Skipped sending redundant request for transmission {} to '{peer_ip}' (Already pending {num_sent_requests} requests)",
```
howardwu:
```suggestion
                "Skipped sending request for transmission {} to '{peer_ip}' ({num_sent_requests} redundant requests)",
```

howardwu (approved):

PR #3144, opened 2024-03-04
[ZKS-04] Do not perform GC on bootup

raychu86:
## Motivation

This PR just updates the `commit_leader_certificate` function to not perform GC during the bootup sync, as it should only be performing garbage collection during actual BFT consensus.

Audit Finding: **[zksecurity 04] Garbage Collection Can Block Commits From Happening**


howardwu (approved):

howardwu:
Please note that this one-liner is an addendum to PR #3117.

PR #3145, opened 2024-03-04
[ZKS-04] Properly set the `gc_round` on `Storage` initialization

raychu86:
## Motivation

This PR calls GC on `Storage` initialization to properly set the `gc_round` variable set in `Storage`. By default this value was set to 0, which is incorrect and would cause bootup to look too far back for certificates.

## Test Plan

Tests hav been added to ensure that the storage has the proper `gc_round` during initialization and that the BFT calls GC when committing a leader certificate.

## Related PRs

Related to https://github.com/AleoHQ/snarkOS/pull/3144

### DIFF
--- a/node/bft/src/helpers/storage.rs
+++ b/node/bft/src/helpers/storage.rs
@@ -114,6 +114,8 @@ impl<N: Network> Storage<N> {
         }));
         // Update the storage to the current round.
         storage.update_current_round(current_round);
+        // Perform GC on the current round.
+        storage.garbage_collect_certificates(current_round);
         // Return the storage.
         storage
     }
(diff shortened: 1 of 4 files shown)


howardwu (approved):

node/bft/src/helpers/storage.rs
```diff
@@ -114,6 +114,8 @@ impl<N: Network> Storage<N> {
         }));
         // Update the storage to the current round.
         storage.update_current_round(current_round);
+        // Perform GC on the current round.
+        storage.garbage_collect_certificates(current_round);
```
mimoo, line 118: I'm not sure what is happening here. Shouldn't you garbage collect based on the last committed round instead?

howardwu: This case here is called in the constructor.

On bootup, we load the `current_round` which is technically the last committed round from the `ledger`'s perspective.

I believe this is equivalent to what your intention is as well.

PR #3147, opened 2024-03-04
[Feature] feat: more TPS metrics

joske:
## Motivation

https://github.com/AleoHQ/snarkOS/issues/3146

### DIFF
--- a/node/consensus/src/lib.rs
+++ b/node/consensus/src/lib.rs
@@ -212,6 +212,11 @@ impl<N: Network> Consensus<N> {
 impl<N: Network> Consensus<N> {
     /// Adds the given unconfirmed solution to the memory pool.
     pub async fn add_unconfirmed_solution(&self, solution: ProverSolution<N>) -> Result<()> {
+        #[cfg(feature = "metrics")]
+        {
+            metrics::increment_gauge(metrics::consensus::SOLUTIONS, 1f64);
+            metrics::increment_gauge(metrics::consensus::TRANSMISSIONS, 1f64);
+        }
         // Process the unconfirmed solution.
         {
             let solution_id = solution.commitment();
@@ -265,6 +270,11 @@ impl<N: Network> Consensus<N> {
 
     /// Adds the given unconfirmed transaction to the memory pool.
     pub async fn add_unconfirmed_transaction(&self, transaction: Transaction<N>) -> Result<()> {
+        #[cfg(feature = "metrics")]
+        {
+            metrics::increment_gauge(metrics::consensus::TRANSACTIONS, 1f64);
+            metrics::increment_gauge(metrics::consensus::TRANSMISSIONS, 1f64);
+        }
         // Process the unconfirmed transaction.
         {
             let transaction_id = transaction.id();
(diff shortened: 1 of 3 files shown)


node/consensus/src/lib.rs
```diff
@@ -237,6 +237,11 @@ impl<N: Network> Consensus<N> {
         if num_unconfirmed > N::MAX_SOLUTIONS || num_unconfirmed > BatchHeader::<N>::MAX_TRANSMISSIONS_PER_BATCH {
             return Ok(());
         }
+        #[cfg(feature = "metrics")]
+        {
+            metrics::increment_gauge(metrics::consensus::SOLUTIONS, 1f64);
```
vicsn: Please add it to the top of the function, same for transactions and transmissions, we want to track the amount going into the node and compare it to what goes into the block

node/consensus/src/lib.rs
```diff
@@ -212,6 +212,11 @@ impl<N: Network> Consensus<N> {
 impl<N: Network> Consensus<N> {
     /// Adds the given unconfirmed solution to the memory pool.
     pub async fn add_unconfirmed_solution(&self, solution: ProverSolution<N>) -> Result<()> {
+        #[cfg(feature = "metrics")]
+        {
+            metrics::increment_gauge(metrics::consensus::SOLUTIONS, 1f64);
```
miazn: I know we had this conversation on the prior tps pr, but I'm wondering now if we are adding so many tps metrics, whether we should change the snarkVM metrics crate `increment_counter` function to take a param to increase the number by rather than just a constant, since this should technically be a counter

node/consensus/src/lib.rs
```diff
@@ -212,6 +212,11 @@ impl<N: Network> Consensus<N> {
 impl<N: Network> Consensus<N> {
     /// Adds the given unconfirmed solution to the memory pool.
     pub async fn add_unconfirmed_solution(&self, solution: ProverSolution<N>) -> Result<()> {
+        #[cfg(feature = "metrics")]
```
miazn, line 215: I think we should probably add this to after the unconfirmed solutions and transmissions are added, since there are a few points where we fail or return early

joske: Initially I did that, but @vicsn requested to move to top.

node/metrics/src/names.rs
```diff
@@ -53,13 +58,18 @@ pub mod bft {
...
+    pub const TRANSMISSIONS: &str = "snarkos_blocks_transmissions_total";
+    pub const SOLUTIONS: &str = "snarkos_blocks_solutions_total";
 }
 
 pub mod consensus {
     pub const CERTIFICATE_COMMIT_LATENCY: &str = "snarkos_consensus_certificate_commit_latency_secs";
     pub const COMMITTED_CERTIFICATES: &str = "snarkos_consensus_committed_certificates_total";
     pub const LAST_COMMITTED_ROUND: &str = "snarkos_consensus_last_committed_round";
     pub const BLOCK_LATENCY: &str = "snarkos_consensus_block_latency_secs";
+    pub const TRANSACTIONS: &str = "snarkos_consensus_transactions_total";
```
miazn: should we maybe add unconfirmed to the consensus txn metric names just so its clear?

node/consensus/src/lib.rs
joske: That's also possible, but then we have 2 linked PRs (snarkVM + snarkOS) which makes it more annoying to merge.

node/consensus/src/lib.rs
miazn: yep, I saw that. I think if we want it at the top we'd have to add metrics flagged code to decrement the gauge on failure or early return, which probably isn't optimal

node/consensus/src/lib.rs
miazn: yeah, that's why I avoided it on the first one-- however if we do end up adding more and more metrics it makes sense to eventually change it-- could be an issue for a later date in snarkvm? @vicsn

node/metrics/src/names.rs
joske: Seems useful to me :+1:

node/consensus/src/lib.rs
joske: I think the point is to see the difference with how many come in and how many actually make it into a block.

miazn: Fair, although in that case it would probably be better to track it from the delivering side, i.e. whatever we are sending. I think maybe that in this case, we are omitting metrics maybe one layer too deep in the code since what we really want is to omit metrics whenever these post endpoints [are called](https://github.com/AleoHQ/snarkOS/blob/mainnet/node/rest/src/routes.rs#L312). If we do want to just track how many times those endpoints are hit it might be cleaner to move it there- @vicsn what do you think?

node/consensus/src/lib.rs
vicsn: Sure, I personally don't see the need but make an issue for the future if you think the added function will make us more robust!

node/consensus/src/lib.rs
vicsn: Rationale for putting it in `add_unconfirmed_{transaction, solution}` is so we also count transactions received via the router.

I think measuring it on the delivering side (i.e. tx-cannon) would be even better, but that requires more coordination and time to figure out.

node/metrics/src/names.rs
vicsn: Sure!

vicsn (approved):
I don't know how to validate the grafana template changes, otherwise LGTM

howardwu (approved):

PR #3149, opened 2024-03-04
[Optimize] Process batch proposal transmissions using rayon

ljedrz:
This PR uses `rayon` to process transmissions in `process_batch_propose_from_peer`; I've double-checked that this doesn't slow things down when the number of transmissions is small.

### DIFF
--- a/node/bft/src/primary.rs
+++ b/node/bft/src/primary.rs
@@ -573,7 +573,7 @@ impl<N: Network> Primary<N> {
         let mut transmissions = self.sync_with_batch_header_from_peer(peer_ip, &batch_header).await?;
 
         // Check that the transmission ids match and are not fee transactions.
-        if let Err(err) = transmissions.par_iter_mut().try_for_each(|(transmission_id, transmission)| {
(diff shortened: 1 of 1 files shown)


node/bft/src/primary.rs
```diff
@@ -572,12 +573,12 @@ impl<N: Network> Primary<N> {
         let mut transmissions = self.sync_with_batch_header_from_peer(peer_ip, &batch_header).await?;
 
         // Check that the transmission ids match and are not fee transactions.
-        for (transmission_id, transmission) in transmissions.iter_mut() {
+        if let Err(err) = transmissions.par_iter_mut().try_for_each(|(transmission_id, transmission)| {
```
raychu86: Why not use `cfg_iter_mut`?

ljedrz: we could, but do we support snarkOS without `rayon` :thinking:?

howardwu: I think it is better to use it as a form of convention

ljedrz: done

howardwu (approved):

PR #3154, opened 2024-03-05
[Feature] feat: jemalloc by default on linux

joske:
Enable jemalloc by default on linux.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -119,7 +119,7 @@ version = "=2.2.7"
 path = "./node/tcp"
 version = "=2.2.7"
 
-[target.'cfg(target_os = "linux")'.dependencies]
+[target.'cfg(all(target_os = "linux", target_arch = "x86_64"))'.dependencies]
 tikv-jemallocator = "0.5"
 
 [dev-dependencies.rusty-hook]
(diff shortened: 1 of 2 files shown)


ljedrz (approved):
LGTM as long as we're ok with enabling it only for Linux; ideally we'd use a target-specific feature, but it is not doable yet: https://github.com/rust-lang/cargo/issues/1197.

howardwu:
Is this also targetable for macOS?

joske:
Yes, it works as far as we've tested, but the jemalloc test suite doesn't completely succeed on macOS, so there may be issues. See platform support on https://crates.io/crates/jemallocator.

Maybe we should even restrict to linux on x86 (though unlikely people will be running snarkOS on aarch64 linux at this point).

howardwu:
We can't merge this if this PR introduces issues.

howardwu:
Moving back to draft until this is resolved, please prioritize as P0

ljedrz (approved):
LGTM, this makes `jemalloc` default on the platforms that are fully supported: https://crates.io/crates/tikv-jemallocator#platform-support.

howardwu (approved):

PR #3162, opened 2024-03-06
[Fix] Initialize the BFT DAG with the correct certificates at bootup

raychu86:
## Motivation

This PR initializes the BFT `DAG` with the correct committed certificate state by updating `sync_bft_dag_at_bootup` to simply commit all the `certificates` that were fetched from ledger. Previously, we were calling `DAG::insert` and `BFT::commit_leader_certificate` to simulate the commit process, but because we can guarantee these certificates have already been committed (exist in the ledger), this logic was incorrect. 

In addition, we've updated `order_dag_with_dfs` to skip certificates that already exist in the ledger for the same reason. There is no need to reorder certificates that have been ordered before in previous subdags.

With the removal of the extraneous logic, we could also simplify the `sync_bft_dag_at_bootup` channel and remove the `IS_SYNCING` const generic.

## Test Plan

Unit tests have been added to check that the DAG state of the BFT after bootup has the correct commit state.

### DIFF
--- a/node/bft/src/bft.rs
+++ b/node/bft/src/bft.rs
@@ -1335,79 +1335,91 @@ mod tests {
     #[tokio::test]
     #[tracing_test::traced_test]
     async fn test_sync_bft_dag_at_bootup_shutdown() -> Result<()> {
-
-        /* 
-        BFT bootup unit test - 
+        /*
+        BFT bootup unit test -
         1. Run one uninterrupted BFT on a set of certificates for 2 leader commits.
-        2. Run a separate bootup BFT that syncs with a set of pre shutdown certificates, and then commits a second leader normally over a set of post shutdown certificates. 
-        3. Observe that the uninterrupted BFT and the bootup BFT end in the same state. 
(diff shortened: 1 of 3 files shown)


node/bft/src/bft.rs
```diff
@@ -824,51 +804,20 @@ impl<N: Network> BFT<N> {
...
-                dag.commit(leader_certificate, self.storage().max_gc_rounds());
-            }
-        }
-        // Commit the latest leader certificate.
-        if let Err(e) = self.commit_leader_certificate::<true, true>(leader_certificate).await {
-            error!("BFT failed to update the DAG with the latest leader certificate - {e}");
+    /// This method commits all the certificates into the dag.
+    /// Note that there is no need to insert the certificates into the DAG, because these certificates
+    /// already exist in the ledger and therefor do not need to be re-ordered into future
+    /// committed subdags.
```
howardwu:
```suggestion
    /// This method commits all the certificates into the DAG.
    /// Note that there is no need to insert the certificates into the DAG, because these certificates
    /// already exist in the ledger and therefore do not need to be re-ordered into future committed subdags.
```

howardwu (approved):

PR #3163, opened 2024-03-06
[Feature] feat: allow-external-peers

joske:
Feature https://github.com/AleoHQ/snarkOS/issues/3155

### DIFF
--- a/.devnet/start.sh
+++ b/.devnet/start.sh
@@ -37,7 +37,7 @@ start_snarkos_in_tmux() {
     tmux new-session -d -s snarkos-session
 
     # Send the snarkOS start command to the tmux session with the NODE_ID
-    tmux send-keys -t "snarkos-session" "snarkos start --nodisplay --bft 0.0.0.0:5000 --rest 0.0.0.0:3030 --allow-outside-peers --peers $NODE_IP:4130 --validators $NODE_IP:5000 --rest-rps 1000 --verbosity $VERBOSITY --dev $NODE_ID --dev-num-validators $NUM_INSTANCES --validator --metrics" C-m
+    tmux send-keys -t "snarkos-session" "snarkos start --nodisplay --bft 0.0.0.0:5000 --rest 0.0.0.0:3030 --allow-external-peers --peers $NODE_IP:4130 --validators $NODE_IP:5000 --rest-rps 1000 --verbosity $VERBOSITY --dev $NODE_ID --dev-num-validators $NUM_INSTANCES --validator --metrics" C-m
 
     exit  # Exit root user
 EOF
--- a/.devnet/start.sh
+++ b/.devnet/start.sh
@@ -37,7 +37,7 @@ start_snarkos_in_tmux() {
     tmux new-session -d -s snarkos-session
 
     # Send the snarkOS start command to the tmux session with the NODE_ID
-    tmux send-keys -t "snarkos-session" "snarkos start --nodisplay --bft 0.0.0.0:5000 --rest 0.0.0.0:3030 --peers $NODE_IP:4130 --validators $NODE_IP:5000 --rest-rps 1000 --verbosity $VERBOSITY --dev $NODE_ID --dev-num-validators $NUM_INSTANCES --validator --metrics" C-m
+    tmux send-keys -t "snarkos-session" "snarkos start --nodisplay --bft 0.0.0.0:5000 --rest 0.0.0.0:3030 --allow-outside-peers --peers $NODE_IP:4130 --validators $NODE_IP:5000 --rest-rps 1000 --verbosity $VERBOSITY --dev $NODE_ID --dev-num-validators $NUM_INSTANCES --validator --metrics" C-m
 
     exit  # Exit root user
 EOF
--- a/cli/src/commands/start.rs
+++ b/cli/src/commands/start.rs
@@ -144,6 +144,10 @@ pub struct Start {
     #[clap(long)]
     /// If development mode is enabled, specify the custom bonded balances as a json object. (default: None)
     dev_bonded_balances: Option<BondedBalances>,
+
+    #[clap(long = "allow-outside-peers")]
+    /// If the flag is set, the validator will allow untrusted peers to connect
+    allow_outside_peers: bool,
 }
 
 impl Start {
@@ -527,7 +531,7 @@ impl Start {
         // Initialize the node.
         let bft_ip = if self.dev.is_some() { self.bft } else { None };
         match node_type {
-            NodeType::Validator => Node::new_validator(self.node, bft_ip, rest_ip, self.rest_rps, account, &trusted_peers, &trusted_validators, genesis, cdn, storage_mode).await,
+            NodeType::Validator => Node::new_validator(self.node, bft_ip, rest_ip, self.rest_rps, account, &trusted_peers, &trusted_validators, genesis, cdn, storage_mode, self.allow_outside_peers).await,
             NodeType::Prover => Node::new_prover(self.node, account, &trusted_peers, genesis, storage_mode).await,
             NodeType::Client => Node::new_client(self.node, rest_ip, self.rest_rps, account, &trusted_peers, genesis, cdn, storage_mode).await,
         }
--- a/cli/src/commands/start.rs
+++ b/cli/src/commands/start.rs
@@ -145,9 +145,9 @@ pub struct Start {
     /// If development mode is enabled, specify the custom bonded balances as a json object. (default: None)
     dev_bonded_balances: Option<BondedBalances>,
 
-    #[clap(long = "allow-outside-peers")]
+    #[clap(long = "allow-external-peers")]
     /// If the flag is set, the validator will allow untrusted peers to connect
-    allow_outside_peers: bool,
+    allow_external_peers: bool,
 }
 
 impl Start {
@@ -531,7 +531,7 @@ impl Start {
         // Initialize the node.
         let bft_ip = if self.dev.is_some() { self.bft } else { None };
         match node_type {
-            NodeType::Validator => Node::new_validator(self.node, bft_ip, rest_ip, self.rest_rps, account, &trusted_peers, &trusted_validators, genesis, cdn, storage_mode, self.allow_outside_peers).await,
+            NodeType::Validator => Node::new_validator(self.node, bft_ip, rest_ip, self.rest_rps, account, &trusted_peers, &trusted_validators, genesis, cdn, storage_mode, self.allow_external_peers).await,
             NodeType::Prover => Node::new_prover(self.node, account, &trusted_peers, genesis, storage_mode).await,
             NodeType::Client => Node::new_client(self.node, rest_ip, self.rest_rps, account, &trusted_peers, genesis, cdn, storage_mode).await,
         }
--- a/cli/src/commands/start.rs
+++ b/cli/src/commands/start.rs
@@ -144,12 +144,12 @@ pub struct Start {
     #[clap(long = "storage_path")]
     pub storage_path: Option<PathBuf>,
 
-    #[clap(long)]
     /// If development mode is enabled, specify the custom bonded balances as a json object. (default: None)
+    #[clap(long)]
     dev_bonded_balances: Option<BondedBalances>,
 
-    #[clap(long = "allow-external-peers")]
     /// If the flag is set, the validator will allow untrusted peers to connect
+    #[clap(long = "allow-external-peers")]
     allow_external_peers: bool,
 }
 
@@ -546,7 +546,7 @@ impl Start {
         // Initialize the node.
         let bft_ip = if self.dev.is_some() { self.bft } else { None };
         match node_type {
-            NodeType::Validator => Node::new_validator(self.node, bft_ip, rest_ip, self.rest_rps, account, &trusted_peers, &trusted_validators, genesis, cdn, storage_mode, dev_txs, self.allow_external_peers).await,
+            NodeType::Validator => Node::new_validator(self.node, bft_ip, rest_ip, self.rest_rps, account, &trusted_peers, &trusted_validators, genesis, cdn, storage_mode, self.allow_external_peers, dev_txs).await,
             NodeType::Prover => Node::new_prover(self.node, account, &trusted_peers, genesis, storage_mode).await,
             NodeType::Client => Node::new_client(self.node, rest_ip, self.rest_rps, account, &trusted_peers, genesis, cdn, storage_mode).await,
         }
--- a/devnet.sh
+++ b/devnet.sh
@@ -64,12 +64,12 @@ for validator_index in "${validator_indices[@]}"; do
 
   # Send the command to start the validator to the new window and capture output to the log file
   if [ "$validator_index" -eq 0 ]; then
-    tmux send-keys -t "devnet:window$validator_index" "snarkos start --nodisplay --dev $validator_index --allow-outside-peers --dev-num-validators $total_validators --validator --logfile $log_file --metrics" C-m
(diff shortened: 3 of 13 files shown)


ljedrz:
Ok, so if `nogossip` is set, the node will not be requesting more peers; shouldn't this also mean that the node doesn't "shuffle" its peers, so that it doesn't lose them?

joske:
@vicsn ?

vicsn:
> shouldn't this also mean that the node doesn't "shuffle" its peers, so that it doesn't lose them?

Yes, I did not have this logic in mind. @joske can you analyse and update the PR? Depending on the scope of required changes, we may also need to rename the flag to make it sound appropriate.

Can you also test the behavior manually? One suggested approach: start a network of e.g. 10 validators each having the new flag and the same 10 hardcoded client peers, and after half an hour you can check if all validators still have all client peers connected. Feel free to suggest another approach. Soon™️ this will be an automated integration test in CI, but that will take a few weeks still. :)

howardwu:
Closing, this was NOT an assigned task.

cli/src/commands/start.rs
```diff
@@ -144,6 +144,10 @@ pub struct Start {
     #[clap(long)]
     /// If development mode is enabled, specify the custom bonded balances as a json object. (default: None)
     dev_bonded_balances: Option<BondedBalances>,
+
+    #[clap(long = "allow-outside-peers")]
+    /// If the flag is set, the node will engage in P2P gossip to request more peers. (default: false)
```
vicsn:
```suggestion
    /// If the flag is set, the validator will engage in P2P gossip to request more peers. (default: false)
```

vicsn (commented):
For clarity, can you share with how many validators and clients you tested?

I just tested using the `devnet.sh` script, without the `--allow-outside-peers` flag, and the clients still connect, is that expected behaviour? Or is there extra code we need so we can easily test in `devnet.sh` if this feature works?

Can you also confirm this is not an issue? https://github.com/AleoHQ/snarkOS/pull/3163#issuecomment-1980505760

cli/src/commands/start.rs
joske: reworded to reflect the new requirements: only allow trusted peers by default

vicsn (changes requested):
Logic mostly LGTM.

Test `duplicate_disconnect_attempts` is failing, I assume because we set the new flag to `false` when the test expects it to be true.

Can you add the new flag to the validators in the `devnet.sh` and `.devnet/start.sh` scripts?

howardwu:
I would advocate for `allow-external-peers` or `no-external-peers` (preferred) as the default flag

vicsn:
> I would advocate for allow-external-peers or no-external-peers (preferred) as the default flag

This PR ensures default behaviour WITHOUT flag is to NOT connect with peers. Therefore calling the flag `--allow-external-peers` sounds good to me, `--no-external-peers` is confusing.

howardwu:
`--allow-external-peers` works for me. This means the default behavior is to not allow external peers.

@joske can you update this PR to reflect this change?

joske:
Yes, was waiting for confirmation on the final name :)

howardwu:
> Yes, was waiting for confirmation on the final name :)

Please see above: --allow-external-peers is confirmed

.devnet/start.sh
```diff
@@ -37,7 +37,7 @@ start_snarkos_in_tmux() {
     tmux new-session -d -s snarkos-session
 
     # Send the snarkOS start command to the tmux session with the NODE_ID
-    tmux send-keys -t "snarkos-session" "snarkos start --nodisplay --bft 0.0.0.0:5000 --rest 0.0.0.0:3030 --peers $NODE_IP:4130 --validators $NODE_IP:5000 --rest-rps 1000 --verbosity $VERBOSITY --dev $NODE_ID --dev-num-validators $NUM_INSTANCES --validator --metrics" C-m
+    tmux send-keys -t "snarkos-session" "snarkos start --nodisplay --bft 0.0.0.0:5000 --rest 0.0.0.0:3030 --allow-outside-peers --peers $NODE_IP:4130 --validators $NODE_IP:5000 --rest-rps 1000 --verbosity $VERBOSITY --dev $NODE_ID --dev-num-validators $NUM_INSTANCES --validator --metrics" C-m
```
howardwu: The flag needs to be updated here

devnet.sh
```diff
@@ -64,12 +64,12 @@ for validator_index in "${validator_indices[@]}"; do
 
   # Send the command to start the validator to the new window and capture output to the log file
   if [ "$validator_index" -eq 0 ]; then
-    tmux send-keys -t "devnet:window$validator_index" "snarkos start --nodisplay --dev $validator_index --dev-num-validators $total_validators --validator --logfile $log_file --metrics" C-m
+    tmux send-keys -t "devnet:window$validator_index" "snarkos start --nodisplay --dev $validator_index --allow-outside-peers --dev-num-validators $total_validators --validator --logfile $log_file --metrics" C-m
```
howardwu: The flag needs to be updated here

devnet.sh
```diff
@@ -64,12 +64,12 @@ for validator_index in "${validator_indices[@]}"; do
...
   else
     # Create a new window with a unique name
     window_index=$((validator_index + index_offset))
     tmux new-window -t "devnet:$window_index" -n "window$validator_index"
-    tmux send-keys -t "devnet:window$validator_index" "snarkos start --nodisplay --dev $validator_index --dev-num-validators $total_validators --validator --logfile $log_file" C-m
+    tmux send-keys -t "devnet:window$validator_index" "snarkos start --nodisplay --dev $validator_index --allow-outside-peers --dev-num-validators $total_validators --validator --logfile $log_file" C-m
```
howardwu: The flag needs to be updated here

.devnet/start.sh
joske: doh, sorry!

node/src/client/mod.rs
```diff
@@ -117,6 +117,7 @@ impl<N: Network, C: ConsensusStorage<N>> Client<N, C> {
             trusted_peers,
             Self::MAXIMUM_NUMBER_OF_PEERS as u16,
             matches!(storage_mode, StorageMode::Development(_)),
+            false,
```
vicsn: I should have spotted this sooner, but this is a bug. I think this should be true, and we should set it true for all the other invocations of `Router::new` as well, except for the validator where its set by a parameter. Because clients/provers are expected to always do peer discovery.

node/src/prover/mod.rs
```diff
@@ -110,6 +110,7 @@ impl<N: Network, C: ConsensusStorage<N>> Prover<N, C> {
             trusted_peers,
             Self::MAXIMUM_NUMBER_OF_PEERS as u16,
             matches!(storage_mode, StorageMode::Development(_)),
+            false,
```
raychu86: > https://github.com/AleoHQ/snarkOS/pull/3163/files#r1525287483

@vicsn Should this be `true` as well according to your comment above?

vicsn: Yes

howardwu:
@joske can you update all instances of `false` to `true`? This includes the tests.

node/router/tests/common/mod.rs
```diff
@@ -110,6 +112,7 @@ pub async fn validator(listening_port: u16, max_peers: u16) -> TestRouter<Curren
         &[],
         max_peers,
         true,
+        false,
```
joske, line 114: @vicsn should this be `true`? I'm inclined to leave this one as false as this is the default for validators.

vicsn: Indeed, this should remain false, thank you for confirming.

vicsn (approved):

node/router/src/handshake.rs
```diff
@@ -264,6 +264,12 @@ impl<N: Network> Router<N> {
         if self.is_connected(&peer_ip) {
             bail!("Dropping connection request from '{peer_ip}' (already connected)")
         }
+        // Only allow trusted peers to connect if we are a validator
+        // (unless allow_external_peers is set)
+        let is_validator = self.node_type().is_validator();
```
howardwu: nit: is_validator is no longer necessary as part of the check (because allow_external_peers takes precedence)

node/router/src/heartbeat.rs
```diff
@@ -107,6 +107,12 @@ pub trait Heartbeat<N: Network>: Outbound<N> {
             return;
         }
 
+        let is_validator = self.router().node_type().is_validator();
+        // Skip if the node is not requesting peers.
+        if is_validator && !self.router().allow_external_peers() {
```
howardwu: nit: is_validator is no longer necessary as part of the check (because allow_external_peers takes precedence)

node/router/src/heartbeat.rs
```diff
@@ -216,9 +222,12 @@ pub trait Heartbeat<N: Network>: Outbound<N> {
             for peer_ip in self.router().candidate_peers().into_iter().choose_multiple(rng, num_deficient) {
                 self.router().connect(peer_ip);
             }
-            // Request more peers from the connected peers.
-            for peer_ip in self.router().connected_peers().into_iter().choose_multiple(rng, 3) {
-                self.send(peer_ip, Message::PeerRequest(PeerRequest));
+            let is_validator = self.router().node_type().is_validator();
+            if !is_validator || self.router().allow_external_peers() {
```
howardwu: nit: is_validator is no longer necessary as part of the check (because allow_external_peers takes precedence)

node/router/src/inbound.rs
```diff
@@ -125,6 +125,9 @@ pub trait Inbound<N: Network>: Reading + Outbound<N> {
                 if !self.router().cache.contains_outbound_peer_request(peer_ip) {
                     bail!("Peer '{peer_ip}' is not following the protocol (unexpected peer response)")
                 }
+                if self.router().node_type().is_validator() && !self.router().allow_external_peers() {
```
howardwu: nit: is_validator is no longer necessary as part of the check (because allow_external_peers takes precedence)

node/router/src/lib.rs
```diff
@@ -95,6 +95,8 @@ pub struct InnerRouter<N: Network> {
     handles: Mutex<Vec<JoinHandle<()>>>,
     /// The boolean flag for the development mode.
     is_dev: bool,
+    /// If the flag is set, the node will not engage in P2P gossip to request more peers.
+    allow_external_peers: bool,
```
howardwu, line 99: nit: can you reverse the order of declaration here?

node/router/src/lib.rs
```diff
@@ -116,6 +118,7 @@ impl<N: Network> Router<N> {
         trusted_peers: &[SocketAddr],
         max_peers: u16,
         is_dev: bool,
+        allow_external_peers: bool,
```
howardwu, line 121: nit: can you reverse the order of declaration here?

node/router/src/lib.rs
```diff
@@ -133,6 +136,7 @@ impl<N: Network> Router<N> {
             restricted_peers: Default::default(),
             handles: Default::default(),
             is_dev,
+            allow_external_peers,
```
howardwu, line 139: nit: can you reverse the order of declaration here?

node/router/tests/common/mod.rs
howardwu, line 115: nit: can you reverse the order of declaration here?

node/src/node.rs
```diff
@@ -51,6 +51,7 @@ impl<N: Network> Node<N> {
         cdn: Option<String>,
         storage_mode: StorageMode,
         dev_txs: bool,
+        allow_external_peers: bool,
```
howardwu, line 54: nit: can you reverse the order of declaration here?

node/src/node.rs
```diff
@@ -65,6 +66,7 @@ impl<N: Network> Node<N> {
                 cdn,
                 storage_mode,
                 dev_txs,
+                allow_external_peers,
```
howardwu, line 69: nit: can you reverse the order of declaration here?

node/src/prover/mod.rs
```diff
@@ -110,6 +110,7 @@ impl<N: Network, C: ConsensusStorage<N>> Prover<N, C> {
             trusted_peers,
             Self::MAXIMUM_NUMBER_OF_PEERS as u16,
             matches!(storage_mode, StorageMode::Development(_)),
+            true,
```
howardwu, line 113: nit: can you reverse the order of declaration here?

node/src/validator/mod.rs
```diff
@@ -84,6 +84,7 @@ impl<N: Network, C: ConsensusStorage<N>> Validator<N, C> {
         cdn: Option<String>,
         storage_mode: StorageMode,
         dev_txs: bool,
+        allow_external_peers: bool,
```
howardwu, line 87: nit: can you reverse the order of declaration here?

node/src/validator/mod.rs
```diff
@@ -126,6 +127,7 @@ impl<N: Network, C: ConsensusStorage<N>> Validator<N, C> {
             trusted_peers,
             Self::MAXIMUM_NUMBER_OF_PEERS as u16,
             matches!(storage_mode, StorageMode::Development(_)),
+            allow_external_peers,
```
howardwu, line 130: nit: can you reverse the order of declaration here?

node/tests/common/node.rs
```diff
@@ -60,6 +60,7 @@ pub async fn validator() -> Validator<CurrentNetwork, ConsensusMemory<CurrentNet
         None,                   // No CDN.
         StorageMode::Production,
         false, // No dev traffic in production mode.
+        true,
```
howardwu, line 63: nit: can you reverse the order of declaration here?

node/src/client/mod.rs
```diff
@@ -117,6 +117,7 @@ impl<N: Network, C: ConsensusStorage<N>> Client<N, C> {
             trusted_peers,
             Self::MAXIMUM_NUMBER_OF_PEERS as u16,
             matches!(storage_mode, StorageMode::Development(_)),
+            true,
```
howardwu, line 120: nit: can you reverse the order of declaration here?

cli/src/commands/start.rs
```diff
@@ -147,6 +147,10 @@ pub struct Start {
     #[clap(long)]
     /// If development mode is enabled, specify the custom bonded balances as a json object. (default: None)
     dev_bonded_balances: Option<BondedBalances>,
+
+    #[clap(long = "allow-external-peers")]
+    /// If the flag is set, the validator will allow untrusted peers to connect
```
howardwu:
```suggestion
    /// If the flag is set, the validator will allow untrusted peers to connect
    #[clap(long = "allow-external-peers")]
```

cli/src/commands/start.rs
```diff
@@ -542,7 +546,7 @@ impl Start {
         // Initialize the node.
         let bft_ip = if self.dev.is_some() { self.bft } else { None };
         match node_type {
-            NodeType::Validator => Node::new_validator(self.node, bft_ip, rest_ip, self.rest_rps, account, &trusted_peers, &trusted_validators, genesis, cdn, storage_mode, dev_txs).await,
+            NodeType::Validator => Node::new_validator(self.node, bft_ip, rest_ip, self.rest_rps, account, &trusted_peers, &trusted_validators, genesis, cdn, storage_mode, dev_txs, self.allow_external_peers).await,
```
howardwu, line 549:
```suggestion
            NodeType::Validator => Node::new_validator(self.node, bft_ip, rest_ip, self.rest_rps, account, &trusted_peers, &trusted_validators, genesis, cdn, storage_mode, self.allow_external_peers, dev_txs).await,
```

raychu86 (approved):

PR #3169, opened 2024-03-14
feat: unconfirmed transmission latency metric & stale transmissions

miazn:
## Motivation

It would be beneficial for validators to have the ability to track the time delta between when each node first sees a transmission ( in `consensus.add_unconfirmed_solution` and `consensus.add_unconfirmed_transaction`) and when it gets added to the block ( in `consensus.try_advance_to_next_block`-- where the other block metrics are currently being updated). I created one histogram metric, `snarkos_consensus_transmission_latency` in prometheus, that is split with a label distinguishing transactions and solutions.

In addition, I added a counter (always increasing) metric for transmissions that have not been seen in a block for a certain threshold (currently set to 30 mins, but could be changed)

## Test Plan

I ran this on a 10 node devnet:

Transmission latency average across all 10 nodes(so far only seeing transactions)
Screenshot 2024-03-14 at 2 15 28 PM

Stale transactions:
Screenshot 2024-03-14 at 2 16 00 PM


miazn:
for some reason when i compile the code locally the check-clippy test is just a warning rather than an error 🤔

node/consensus/src/lib.rs
```diff
@@ -131,6 +135,8 @@ impl<N: Network> Consensus<N> {
             seen_solutions: Arc::new(Mutex::new(LruCache::new(NonZeroUsize::new(1 << 16).unwrap()))),
             seen_transactions: Arc::new(Mutex::new(LruCache::new(NonZeroUsize::new(1 << 16).unwrap()))),
             handles: Default::default(),
+            #[cfg(feature = "metrics")]
+            transmissions_queue_timestamps: Arc::new(Mutex::new(HashMap::new())),
```
```suggestion
            transmissions_queue_timestamps: Default::default(),
```

node/consensus/src/lib.rs
```diff
@@ -216,6 +222,9 @@ impl<N: Network> Consensus<N> {
         {
             metrics::increment_gauge(metrics::consensus::UNCONFIRMED_SOLUTIONS, 1f64);
             metrics::increment_gauge(metrics::consensus::UNCONFIRMED_TRANSMISSIONS, 1f64);
+            self.transmissions_queue_timestamps
+                .lock()
+                .insert(TransmissionID::Solution(solution.commitment()), snarkos_node_bft::helpers::now());
```
ljedrz: it might be a good idea to obtain the timestamp before inserting into the collection; that would reduce the duration of the lock guard

node/consensus/src/lib.rs
```diff
@@ -274,6 +283,9 @@ impl<N: Network> Consensus<N> {
         {
             metrics::increment_gauge(metrics::consensus::UNCONFIRMED_TRANSACTIONS, 1f64);
             metrics::increment_gauge(metrics::consensus::UNCONFIRMED_TRANSMISSIONS, 1f64);
+            self.transmissions_queue_timestamps
+                .lock()
+                .insert(TransmissionID::Transaction(transaction.id()), snarkos_node_bft::helpers::now());
```
ljedrz: ditto

node/consensus/src/lib.rs
```diff
@@ -431,6 +445,50 @@ impl<N: Network> Consensus<N> {
         Ok(())
     }
 
+    #[cfg(feature = "metrics")]
+    fn add_transmission_latency_metric(&self, next_block: &Block<N>) {
+        const AGE_THRESHOLD_SECONDS: i32 = 30 * 60; // 30 minutes set as stale transmission threshold
+
+        let mut keys_to_remove = Vec::new();
+        let mut transmission_queue_timestamps = self.transmissions_queue_timestamps.lock();
```
ljedrz: I'd move this just before the loop, to minimize its lifetime

node/consensus/src/lib.rs
```diff
@@ -431,6 +445,50 @@ impl<N: Network> Consensus<N> {
...
+
+        let solution_ids: std::collections::HashSet<_> = next_block.solutions().solution_ids().collect();
+        let transaction_ids: std::collections::HashSet<_> = next_block.transaction_ids().collect();
+
+        for (key, timestamp) in transmission_queue_timestamps.iter() {
+            let elapsed_time = std::time::Duration::from_secs((snarkos_node_bft::helpers::now() - *timestamp) as u64);
```
ljedrz: wouldn't it suffice for us to obtain the current timestamp just once before the loop?

node/consensus/src/lib.rs
```diff
@@ -431,6 +445,50 @@ impl<N: Network> Consensus<N> {
...
+                        Some("transaction")
+                    }
+                    _ => None,
+                };
+
+                if let Some(transmission_type_string) = transmission_type {
+                    metrics::histogram_label(
+                        metrics::consensus::TRANSMISSION_LATENCY,
+                        "transmission_type",
+                        transmission_type_string.to_string(),
```
```suggestion
                        transmission_type_string.to_owned(),
```
just to not suggest there is any conversion involved

node/consensus/src/lib.rs
```diff
@@ -431,6 +445,50 @@ impl<N: Network> Consensus<N> {
...
+        let mut transmission_queue_timestamps = self.transmissions_queue_timestamps.lock();
+
+        let solution_ids: std::collections::HashSet<_> = next_block.solutions().solution_ids().collect();
+        let transaction_ids: std::collections::HashSet<_> = next_block.transaction_ids().collect();
+
+        for (key, timestamp) in transmission_queue_timestamps.iter() {
+            let elapsed_time = std::time::Duration::from_secs((snarkos_node_bft::helpers::now() - *timestamp) as u64);
+
+            if elapsed_time.as_secs() > AGE_THRESHOLD_SECONDS as u64 {
+                // This entry is stale-- remove it from transmission queue and record it as a stale transmission
```
```suggestion
                // This entry is stale - remove it from transmission queue and record it as a stale transmission.
```

node/consensus/src/lib.rs
```diff
@@ -431,6 +445,50 @@ impl<N: Network> Consensus<N> {
...
+                };
+
+                if let Some(transmission_type_string) = transmission_type {
+                    metrics::histogram_label(
+                        metrics::consensus::TRANSMISSION_LATENCY,
+                        "transmission_type",
+                        transmission_type_string.to_string(),
+                        elapsed_time.as_secs_f64(),
+                    );
+                    keys_to_remove.push(key.clone());
```
ljedrz: should we remove this key even if it's not stale yet?

ljedrz (commented):
Left a few comments and suggestions; also, some CI jobs seem to be failing.

node/consensus/src/lib.rs
miazn: The original request here was for a transition latency metric, from when the block first sees the transition to when it lands in the block-- the keys are removed from the `transitionid, timestamp` queue when either 1) the id has not been included in a block in 30 mins, it goes to the stale transmission metric (can change this threshold), or 2) the id is included in the block in under 30 mins, it goes to the transmission latency metric. This way we periodically clear the queue so it is not infinitely increasing.

node/consensus/src/lib.rs
miazn: The M/double dash is correct here, but I can add a period :D

joske (approved):

howardwu (approved):

miazn:
tested and everything works, removed registering the histogram without labels bc it creates extra unused/empty data timeseries in grafana

PR #3170, opened 2024-03-14
Smooth out sync-to-tip

raychu86:
## Motivation

This PR allows syncing nodes to start processing peer certificates slightly earlier. Previously, the syncing node needed to be synced up in order to be processing incoming batch certificates. Which puts the node behind all the peers, who should have already constructed new rounds past the last commit. This causes a cycle of lagging just out of reach of the latest round and preventing the node from quickly rejoining consensus. Breaking the cycle relies on the node and message timing such that it can fetch the certificates before new ones can be formed on the network.

The change here allows the node to start processing certificates 1 block earlier, in preparation of reaching tip.


raychu86:
Closing, as this is a bandaid optimization.

PR #3178, opened 2024-03-21
[Optimize] Skip proposing duplicated transmissions

raychu86:
## Motivation

This PR skips adding transmissions to proposals if the transmission was included in storage. This is because if the node's proposal/certificate gets included in a subdag, it must also include the previous certificates, meaning we'd have duplicated transmission ids. 

This optimization reduces state bloat in blocks and also opens more space in proposals for unique transmissions, which would allow more TPS. The second change in this PR processes more transmissions from the worker in case there were some in the initial `num_transmissions_per_worker` that were skipped.

This PR also increase the number of unconfirmed transmissions we pass into the memory pool. Previously it was bounded by `MAX_TRANSMISSIONS_PER_BATCH`, however if any of these transmissions are invalid or duplicated, then we effectively limit the theoretical throughput. We still only include `MAX_TRANSMISSIONS_PER_BATCH` in each proposal, however we now have a larger pool of transmissions to choose from.

TLDR - The changes are:

1. Skip transmissions that are already included in storage.
2. Increase number of transmissions sent into the mempool.

This PR needs to be validated with internal devnet testing, particularly with the mempool change.

## Test Plan

A test has been added that we properly skip the transmissions that already exist in storage.

### DIFF
--- a/node/bft/src/primary.rs
+++ b/node/bft/src/primary.rs
@@ -1787,6 +1787,72 @@ mod tests {
     }
 
     #[tokio::test]
+    async fn test_propose_batch_skip_transmissions_from_previous_certificates() {
+        let round = 3;
+        let prev_round = round - 1;
+        let mut rng = TestRng::default();
+        let (primary, accounts) = primary_without_handlers(&mut rng).await;
+        let peer_account = &accounts[1];
+        let peer_ip = peer_account.0;
+
+        // Fill primary storage.
+        store_certificate_chain(&primary, &accounts, round, &mut rng);
+
+        // Get transmissions from previous certificates.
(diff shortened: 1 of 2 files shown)


vicsn (approved):
Logic LGTM, makes a lot of sense

kpandl:
Did an initial run and was able to reach around 38 TPS on an ongoing basis with credits.aleo transfer_public transactions

howardwu (approved):

howardwu:
@onetrickwolf's testing has also confirmed that block times have fallen significantly with this PR, **when there are `tx-cannons` driving the network forward**

PR #3180, opened 2024-03-21
Switch tests to release-profile from dev-profile

vvp:
This PR configures CI to execute tests with `cargo test --release` and fixes the tests that break when running with `release` profile.

All failing tests fail due to `snarkVM` test-helpers generating a sample genesis block, which panics only when running in `release` profile (`The genesis block must contain exactly 1 ratification`, originated from https://github.com/AleoHQ/snarkVM/blob/mainnet-staging/ledger/block/src/header/genesis.rs#L24-L27). It might make sense to refactor the `VM::genesis_beacon()`-based fix over to `snarkVM` test-helpers side.

In general, tests should be run on as production-like build as possible, so if `snarkVM` runs its tests with `dev` profile, at least `snarkOS` should do `release` tests for appropriate coverage.


howardwu:
@vvp can you revert the Cargo.toml spacing formatter changes? These are auto-gen'ed by our release engine.

kaimast:
This would be nice to have and could speed up CI a little.

I rebased your commit on recent staging and opened a new PR, because I wasn't able to push to your branch. Closing this in favor of #3673.

PR #3187, opened 2024-03-24
[Cleanup] Timestamp GC 

mdelle1:
## Motivation

In Bullshark, timestamps on certificates are used during garbage collection to ensure that slow validators are still able to contribute to the DAG, rather than having their certificates dropped due to a slow network. To accomplish this, Bullshark garbage collects entire rounds rather than individual certificates. In particular, during GC, Bullshark will check if the difference between the median timestamp of a round and the timestamp of the leader being committed exceeds a specified upper bound. If it does, the entire round and those preceding it are garbage collected. Bullshark uses this algorithm to reach a balance between fairness and bounded memory.


howardwu (approved):
LGTM, this TODO is deprecated.

PR #3189, opened 2024-03-25
[Fix] Make --node optional to ensure it has an influence even when using --dev; renames `--storage_path` to `--storage`

vicsn:
## Motivation

Previously, when using `--dev`, the node ports would automatically be incremented alongside the the dev index. This makes it hard to indicate `--peers` on other machines, as you'll need to know the dev index / port number before it works. This design is similar to the `--bft` flag.

This PR also renames `--storage_path` to `--storage` in the CLI `start` command.

## Test Plan

This only changes a node's behaviour when both `--node` and `--dev` are used.

Tested manually locally and on aws.

## Related PRs

Needed because we're testing: https://github.com/AleoHQ/snarkOS/pull/3163

### DIFF
--- a/cli/src/commands/start.rs
+++ b/cli/src/commands/start.rs
@@ -90,8 +90,8 @@ pub struct Start {
     pub private_key_file: Option<PathBuf>,
 
     /// Specify the IP address and port for the node server
-    #[clap(default_value = "0.0.0.0:4130", long = "node")]
-    pub node: SocketAddr,
+    #[clap(long = "node")]
+    pub node: Option<SocketAddr>,
     /// Specify the IP address and port for the BFT
     #[clap(long = "bft")]
     pub bft: Option<SocketAddr>,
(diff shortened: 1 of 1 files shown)


christianwwwwwwww (approved):
LGTM 🤓

vicsn:
Made an issue for the flaky test, which was already flaky before this PR: https://github.com/AleoHQ/snarkOS/issues/3190

vicsn:
Made an issue for the flaky test, which was already flaky before this PR: https://github.com/AleoHQ/snarkOS/issues/3190

howardwu:
- Does this change the production `snarkos start` command for any node operators?
- Does the `devnet.sh` script or `.devnet/start.sh` script need updating?
- Do the `run-client.sh` or `run-prover.sh` scripts need updating?
- Have the above items been tested by hand and confirmed to work as before?

vicsn:
Note behavior only changes if you use both `--node` and `--dev`.

> Does this change the production snarkos start command for any node operators?

Both devnet and canary do not use `--dev`, so not impacted.
 
> Does the devnet.sh script or .devnet/start.sh script need updating?

They do not use `--node`, so not impacted.

> Do the run-client.sh or run-prover.sh scripts need updating?

They do not use `--dev` or `--node`, so not impacted.

> Have the above items been tested by hand and confirmed to work as before?

I ran `devnet.sh` and `run-client.sh` and they still work. I did not run `.devnet/start.sh` but it has the same `--dev` as `devnet.sh`.

raychu86:
Tested on `.devnet/start.sh` and it seems to work.

howardwu:
^ This commit renames `--storage_path` to `--storage` in the CLI `start` command.

howardwu (approved):

PR #3191, opened 2024-03-26
Feature-gated committee size configuration

niklaslong:
This PR exposes a `test-committee` feature on OS that compiles in VM test helpers including the appropriate constants for running larger networks. Running snarkOS with `--features test-committee` will build with a `MAX_COMMITTEE_SIZE` of 100 rather than the standard 10. 

@raychu86 please let me know if we need more granular control over the constants vs all `test-helpers` in VM. For now, I've opted for the simplest possible change.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -56,6 +56,7 @@ path = "snarkos/main.rs"
 
 [features]
 metrics = [ "snarkos-node-metrics", "snarkos-node/metrics" ]
+test-committee = [ "snarkvm/test-helpers" ]
 
 [dependencies.anyhow]
 version = "1.0.79"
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -56,6 +56,7 @@ path = "snarkos/main.rs"
 
 [features]
 metrics = [ "snarkos-node-metrics", "snarkos-node/metrics" ]
+test-committee = [ "snarkvm/test-helpers" ]
```
raychu86, line 59: We may want to make a dedicated feature flag in snarkVM just for this. We can also lump a few other dev-mode only parameter changes, such as:
- Lower genesis coinbase and puzzle targets @d0cd

raychu86:
Closing in favor of the approach in #3199

PR #3192, opened 2024-03-26
[Fix] Fix flaky connection cleanup case

### LINKED ISSUE #3190: [Bug] test_connection_cleanups is flaky
## 🐛 Bug Report

It appears that sometimes, when a validator is picked in the test, it fails:
```
thread 'test_connection_cleanups' panicked at node/router/tests/cleanups.rs:74:9:
the deadline has elapsed for condition: tcp0.num_connected() == 1 && tcp1.num_connected() == 1
```

A quick print statement indicated: `Unable to connect to '127.0.0.1:63168' - '127.0.0.1:63168' disconnected before sending "Message::ChallengeResponse"`

## Steps to Reproduce

`cd node/router && cargo test test_connection_cleanups`

Due to random sampling of the node type, the test only sometimes fails.

## Expected Behavior

The test should pass.

## Your Environment

Ran it on `mainnet-staging` `491f3078a`

niklaslong:
Stricter peering rules were introduced in https://github.com/AleoHQ/snarkOS/pull/3163 which caused connections to fail if peers were not in the validator's trusted peer set (which we are not setting in these cases). Relaxing the requirements in this particular case should fix the flakiness. 

Closes #3190.

### DIFF
--- a/node/router/tests/common/mod.rs
+++ b/node/router/tests/common/mod.rs
@@ -111,7 +111,7 @@ pub async fn validator(listening_port: u16, max_peers: u16) -> TestRouter<Curren
         sample_account(),
         &[],
         max_peers,
-        false,
+        true,
         true,
     )
     .await
--- a/node/router/tests/common/mod.rs
+++ b/node/router/tests/common/mod.rs
@@ -111,7 +111,7 @@ pub async fn validator(listening_port: u16, max_peers: u16) -> TestRouter<Curren
         sample_account(),
         &[],
         max_peers,
-        true,
+        true, // Router tests require validators to connect to peers.
         true,
     )
     .await


howardwu (approved):

node/router/tests/common/mod.rs
```diff
@@ -111,7 +111,7 @@ pub async fn validator(listening_port: u16, max_peers: u16) -> TestRouter<Curren
         sample_account(),
         &[],
         max_peers,
-        false,
+        true,
```
vicsn: Thank you!!

The default for validators is false, so I suggest to add a comment, same as https://github.com/AleoHQ/snarkOS/pull/3163/files#diff-0490634f09b53b897acaf1041eb100043b7bd7627128eb7e77f745df58b64377R62

```suggestion
        true, // This test requires validators to connect to peers.
```

vicsn (approved):
One nit

joske (approved):

PR #3193, opened 2024-03-26
[Fix] Lower bound the batch proposal production time

raychu86:
## Motivation

This PR lower bounds the production and signing of batch proposals. Previously this was unguarded, so blocks could be produced as fast as validators could find subdags to commit. This could lead to much shorter block times, especially if malicious validators decide to keep proposals empty to bring down block times.

This is important, because lowering the block time increases the number of blocks produced, which in turn advances the amount of block rewards paid out. The fix is to enforce that the difference in time between an author's proposals must be at least `MAX_BATCH_DELAY_IN_SECS`.

## Test Plan

Existing tests have been updated to reflect the rule change for a minimum elapsed time between batch proposals.
New tests have also been added to enforce that batch proposals that are created too quickly are not signed.

### DIFF
--- a/node/bft/src/lib.rs
+++ b/node/bft/src/lib.rs
@@ -49,6 +49,8 @@ pub const MEMORY_POOL_PORT: u16 = 5000; // port
 
 /// The maximum number of milliseconds to wait before proposing a batch.
 pub const MAX_BATCH_DELAY_IN_MS: u64 = 2500; // ms
+/// The maximum number of seconds to wait before proposing a batch (rounded down to the nearest second).
+pub const MAX_BATCH_DELAY_IN_SECS: u64 = MAX_BATCH_DELAY_IN_MS / 1000; // seconds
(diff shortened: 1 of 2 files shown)


node/bft/src/lib.rs
```diff
@@ -49,6 +49,8 @@ pub const MEMORY_POOL_PORT: u16 = 5000; // port
 
 /// The maximum number of milliseconds to wait before proposing a batch.
 pub const MAX_BATCH_DELAY_IN_MS: u64 = 2500; // ms
+/// The maximum number of seconds to wait before proposing a batch (rounded down to the nearest second).
+pub const MAX_BATCH_DELAY_IN_SECS: u64 = MAX_BATCH_DELAY_IN_MS / 1000; // seconds
```
howardwu: Shouldn't this just be the `MIN_BATCH_DELAY` now?

howardwu (approved):

PR #3198, opened 2024-04-01
[Test] Add unit test for validator Router connections

### LINKED ISSUE #3196: [Bug] Peers are not sufficiently connecting to each other
## 🐛 Bug Report

It seems that our nodes are not connected as optimally as we'd like. Perhaps existing PRs ([link](https://github.com/AleoHQ/snarkOS/pull/3164)) will improve the situation

## Steps to Reproduce

Set up a network with 10 validators and 100 clients. The 10 validators should connect to each other, and each validator should connect to 10 trusted peers using `--peers`. We would prefer `snarkos_router_connected_total` to approach 19, but it is only 8.

In contrast, a local default `devnet.sh` run does reveal `N-1` validators are connected via `snarkos_router_connected_total`.

## Your Environment

SnarkOS commit `491f3078a`.

vicsn:
## Motivation

Due to the recently introduced `--allow-external-peers` flag, validator Routers do not by default accept connections from other validator Routers. We noticed on tests with only validators that the `snarkos_router_connected_total` metric returned 0.

To make this behavior clear, added a unit test `test_validator_connection`. In order for validator's Routers to connect to one another, their address has to be explicitly passed via `--peers`.

Note that if we want to change that behaviour, we need to either:
- add Gateway features to Router (e.g. a `Resolver` to track whether we are not connected to the same address twice)
- do a bigger refactor to merge Gateway and Router so the validator has just one TCP stack

## Related PRs

Closes https://github.com/AleoHQ/snarkOS/issues/3196
Corrects yet another regression from https://github.com/AleoHQ/snarkOS/pull/3163
Improves on https://github.com/AleoHQ/snarkOS/pull/3192

### DIFF
--- a/node/router/src/handshake.rs
+++ b/node/router/src/handshake.rs
@@ -17,7 +17,6 @@ use crate::{
     Peer,
     Router,
 };
-use snarkos_node_router_messages::NodeType;
 use snarkos_node_tcp::{ConnectionSide, Tcp, P2P};
 use snarkvm::{
     ledger::narwhal::Data,
@@ -203,7 +202,7 @@ impl<N: Network> Router<N> {
         let peer_ip = peer_ip.unwrap();
 
         // Knowing the peer's listening address, ensure it is allowed to connect.
-        if let Err(forbidden_message) = self.ensure_peer_is_allowed(peer_ip, peer_request.node_type) {
+        if let Err(forbidden_message) = self.ensure_peer_is_allowed(peer_ip) {
             return Err(error(format!("{forbidden_message}")));
         }
         // Verify the challenge request. If a disconnect reason was returned, send the disconnect message and abort.
@@ -252,7 +251,7 @@ impl<N: Network> Router<N> {
     }
 
     /// Ensure the peer is allowed to connect.
-    fn ensure_peer_is_allowed(&self, peer_ip: SocketAddr, node_type: NodeType) -> Result<()> {
+    fn ensure_peer_is_allowed(&self, peer_ip: SocketAddr) -> Result<()> {
         // Ensure the peer IP is not this node.
         if self.is_local_ip(&peer_ip) {
             bail!("Dropping connection request from '{peer_ip}' (attempted to self-connect)")
@@ -266,7 +265,7 @@ impl<N: Network> Router<N> {
             bail!("Dropping connection request from '{peer_ip}' (already connected)")
         }
         // Only allow trusted peers to connect if allow_external_peers is set
-        if !node_type.is_validator() && !self.allow_external_peers() && !self.is_trusted(&peer_ip) {
+        if !self.allow_external_peers() && !self.is_trusted(&peer_ip) {
             bail!("Dropping connection request from '{peer_ip}' (untrusted)")
         }
         // Ensure the peer is not restricted.
--- a/node/router/src/handshake.rs
+++ b/node/router/src/handshake.rs
@@ -17,6 +17,7 @@ use crate::{
     Peer,
     Router,
 };
+use snarkos_node_router_messages::NodeType;
 use snarkos_node_tcp::{ConnectionSide, Tcp, P2P};
 use snarkvm::{
     ledger::narwhal::Data,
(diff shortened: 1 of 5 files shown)


node/router/src/handshake.rs
```diff
@@ -251,7 +252,7 @@ impl<N: Network> Router<N> {
     }
 
     /// Ensure the peer is allowed to connect.
-    fn ensure_peer_is_allowed(&self, peer_ip: SocketAddr) -> Result<()> {
+    fn ensure_peer_is_allowed(&self, peer_ip: SocketAddr, node_type: NodeType) -> Result<()> {
```
niklaslong: Note that we have to trust the node type as it's given to us by the peer, I don't think we previously used it for anything so we're introducing a new assumption.

node/router/tests/connect.rs
```diff
@@ -150,11 +150,48 @@ async fn test_connect_with_handshake() {
...
+
+    // Start listening.
+    node0.tcp().enable_listener().await.unwrap();
+    node1.tcp().enable_listener().await.unwrap();
+
+    {
+        // Connect node0 to node1.
+        node0.connect(node1.local_ip());
+        // Sleep briefly.
+        tokio::time::sleep(Duration::from_millis(200)).await;
```
niklaslong, line 174: `deadline` could be used here to avoid the `sleep`.

node/router/src/handshake.rs
vicsn: You are right. The easiest immediate resolution is to pass in validator routers explicitly via `--peers`. I added a unit test for it and updated the PR description. 297461215

node/router/tests/connect.rs
vicsn: Will keep it in mind for future tests, for now I'll keep the logic similar to the rest of the tests in the file.

ljedrz: IIRC `connect` returns a `JoinHandle` that can be `.await`ed (and once it's complete, it means the connection is either ready or that it has failed), so there is no need for a `sleep` here

node/router/tests/connect.rs
```diff
@@ -150,11 +150,64 @@ async fn test_connect_with_handshake() {
...
+        assert_eq!(node1.number_of_connected_peers(), 1);
+
+        // Disconnect the nodes.
+        node0.disconnect(node1.local_ip());
+        node1.disconnect(node0.local_ip());
+
+        // Connect node1 to node0.
+        node1.connect(node0.local_ip());
+        // Sleep briefly.
+        tokio::time::sleep(Duration::from_millis(200)).await;
```
ljedrz, line 194: ditto

node/router/tests/connect.rs
```diff
@@ -150,11 +150,64 @@ async fn test_connect_with_handshake() {
...
+
+    // Create second router, trusting the first router, and start listening.
+    let node1 = validator(0, 2, &[addr0], false).await;
+    assert_eq!(node1.number_of_connected_peers(), 0);
+    node1.enable_handshake().await;
+    node1.tcp().enable_listener().await.unwrap();
+
+    {
+        // Connect node0 to node1.
+        node0.connect(node1.local_ip());
```
ljedrz, line 172: aren't we racing against the `Heartbeat` (that might cause `node1` to attempt to connect to `node0`) here? then again, that direction should fail, so it's probably fine

ljedrz (commented):
LGTM with some nits addressed.

node/router/tests/connect.rs
vicsn: Just tried it. Just awaiting node0's connect is not enough, because node1 might not be connected yet. I suggest we leave refactoring this file for later.

node/router/tests/connect.rs
vicsn: heartbeat is not turned on, only if we call `initialize_routing`.

node/router/tests/connect.rs
vicsn: ditto comment above

howardwu (approved):

PR #3200, opened 2024-04-02
Store batch proposal to disk on shutdown

### LINKED ISSUE #3171: [Feature] Store batch proposal to disk on shutdown
## 🚀 Feature

Store the current proposal to DB on shutdown. 

Currently when a node shutdowns/reboots while a batch proposal is still in progress, the proposal is dropped. Once the node restarts, it will need to generate a new proposal for the same round. Other nodes who see this new proposal will think the node is being malicious because each validator can't have more than one proposal per round.

This is really only meaningful if the other nodes have not progressed passed this round between the shutdown and reboot.

The solution is to store the pending batch proposal to disk on shutdown and reload it on bootup.

ljedrz:
This PR persists the current batch proposal to the disk on shutdown, and loads it on startup when available, deleting it afterwards.

Filing as a draft until it's sufficiently tested.

Closes https://github.com/AleoHQ/snarkOS/issues/3171.

### DIFF
--- a/cli/src/commands/clean.rs
+++ b/cli/src/commands/clean.rs
@@ -16,6 +16,7 @@ use aleo_std::StorageMode;
 use anyhow::{bail, Result};
 use clap::Parser;
 use colored::Colorize;
+use snarkos_node::bft::batch_proposal_path;
 use std::path::PathBuf;
 
 /// Cleans the snarkOS node storage.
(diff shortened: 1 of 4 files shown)


ljedrz:
Some open questions:
- should we distinguish between production and dev modes?
- should we consider the network IDs?
- should the creation of the file path be dedicated to a function to deduplicate code a little bit?

raychu86:
The file path should definitely be differentiated by dev_mode and network ID. Especially in the case of local devnets being run. 

In addition, `snarkos clean` should also remove these files.

ljedrz:
Addressed the points in the previous comment.

raychu86:
This logic may need to be more comprehensive:

Given a scenario where the latest block is at round 50:

1. A node shuts down and stores its proposal at round 52 (before committing a new block).

2. On bootup the node will sync up with the ledger blocks and advance storage to round 50. 

3. Then the node will try to load the proposal for round 52, but won't have the certificates in `Storage` to properly advance to round 52. Nor will it have the certificates used to create the proposal in round 52.
    - The node will need to request the certificates for round 51 from its peers and properly advance to round 52.

ljedrz:
Rebased.

raychu86:
Closing in favor of https://github.com/AleoHQ/snarkOS/pull/3239

PR #3202, opened 2024-04-02
Introduce proposal expiration

raychu86:
## Motivation

This PR introduces an expiration time for proposals, where nodes expire their own proposal and no longer track other validator proposals past the expiration time. The current expiration time is 1 minute.

This is needed to mitigate the concerns where validator nodes that reboot lose their in-memory proposal and other validator nodes will no longer accept new proposals for that validator. 

There are two cases:
1. If a validator's  own proposal is not certified after the expiration time, it is thrown away and a new proposal is created.
2. Validators can sign new proposals on the same round and author if the original proposal has expired.

This is one of the solutions to address a portion of https://github.com/AleoHQ/snarkOS/issues/3171.

### DIFF
--- a/node/bft/src/primary.rs
+++ b/node/bft/src/primary.rs
@@ -576,43 +576,32 @@ impl<N: Network> Primary<N> {
                 bail!("Malicious peer - proposed a batch for a previous round ({})", batch_header.round());
             }
 
-            // If the signed round matches the peer's batch round then check the contents.
-            if signed_round == batch_header.round() {
-                // Determine if the proposal has expired.
-                let is_proposal_expired = now().saturating_sub(timestamp) >= PROPOSAL_EXPIRATION_IN_SECS;
-                match is_proposal_expired {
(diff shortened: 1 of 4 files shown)


node/bft/src/primary.rs
```diff
@@ -562,36 +567,52 @@ impl<N: Network> Primary<N> {
...
+                            return Ok(());
+                        }
+                        // If the batch ID differs, then the validator is malicious.
+                        false => {
+                            // Proceed to disconnect the validator.
+                            self.gateway.disconnect(peer_ip);
+                            bail!("Malicious peer - proposed another batch for the same round ({signed_round})");
+                        }
+                    },
+                }
```
howardwu: Can you please revert and minimize the code diff here? This nesting behavior reduces readability and complicates the review process. It also impacts much more LOCs than needed.

howardwu (approved):

PR #3203, opened 2024-04-03
[Log] Reduces log sensitivity for failed solutions near epoch changes.

d0cd:
This PR reduces  log sensitivity for failed solutions near epoch changes by bumping down the the level from `warn` to `trace`.

### DIFF
--- a/node/consensus/src/lib.rs
+++ b/node/consensus/src/lib.rs
@@ -266,7 +266,7 @@ impl<N: Network> Consensus<N> {
                 // If the BFT is synced, then log the warning.
                 if self.bft.is_synced() {
                     // If error occurs after the first 10 blocks of the epoch, log it as a warning, otherwise ignore.
-                    if self.ledger().latest_block_height() % N::NUM_BLOCKS_PER_EPOCH <= 10 {
+                    if self.ledger().latest_block_height() % N::NUM_BLOCKS_PER_EPOCH > 10 {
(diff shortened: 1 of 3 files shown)


node/consensus/src/lib.rs
```diff
@@ -265,7 +265,17 @@ impl<N: Network> Consensus<N> {
             if let Err(e) = self.primary_sender().send_unconfirmed_solution(solution_id, Data::Object(solution)).await {
                 // If the BFT is synced, then log the warning.
                 if self.bft.is_synced() {
-                    warn!("Failed to add unconfirmed solution '{}' to the memory pool - {e}", fmt_id(solution_id));
+                    // If error occurs in the first 10 blocks of the epoch, then log it as a trace, otherwise log it as a warning.
+                    match self.ledger().latest_block_height() % N::NUM_BLOCKS_PER_EPOCH <= 10 {
```
raychu86: Why check against the first 10 blocks in the epoch? 

The more comprehensive check is to see if the solution is using the `epoch_hash` of the previous epoch. But that might incur a lookup of some sort or a cache

d0cd: Great point!
The intention of the PR was to reduce warning logs that would occur around epoch changes, hence the first 10 blocks.

I attempted an alternate solution (in commit 9b3f8ca) where we downgrade the log if the solution's epoch hash does not match the existing one. It does not exactly look at the previous epoch hash, but I feel wary about adding caching just for the sake of reduced logs.

howardwu (approved):

PR #3207, opened 2024-04-08
use tcmalloc allocator

joske:
## Motivation

switch to tcmalloc allocator.

Note that libtcmalloc must be installed on the target (build) machine. On ubuntu systems this means installing `libgoogle-perftools-dev`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -120,7 +120,7 @@ path = "./node/tcp"
 version = "=2.2.7"
 
 [target.'cfg(all(target_os = "linux", target_arch = "x86_64"))'.dependencies]
-tcmalloc = { git = "https://github.com/bhansconnect/tcmalloc-rs", branch = "master" }
+tcmalloc = { git = "https://github.com/joske/tcmalloc-rs", tag = "v2.9.1" }
 
 [dev-dependencies.rusty-hook]
 version = "0.11.2"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -120,7 +120,7 @@ path = "./node/tcp"
 version = "=2.2.7"
 
 [target.'cfg(all(target_os = "linux", target_arch = "x86_64"))'.dependencies]
-tikv-jemallocator = "0.5"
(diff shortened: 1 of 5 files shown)


Cargo.toml
```diff
@@ -120,7 +120,7 @@ path = "./node/tcp"
 version = "=2.2.7"
 
 [target.'cfg(all(target_os = "linux", target_arch = "x86_64"))'.dependencies]
-tikv-jemallocator = "0.5"
+tcmalloc = { git = "https://github.com/bhansconnect/tcmalloc-rs", branch = "master" }
```
ljedrz: we might want to pin to a specific SHA; we are publishing the lockfile, but just to be on the safe side; also, that particular repo uses `gperftools = 2.9.1`, while the current version is `2.15`; perhaps it's worth forking?

joske: As discussed before, it's not really that important (as it uses the system libtcmalloc.so), but it might be in the future, or the fork I was using goes away. So, yeah, I'll fork it.

joske: In fact, ubuntu has 2.9.1 for `libgoogle-perftools-dev` up until the next LTS (24.04) which is not out yet, so I guess best fix on 2.9.1 for now.

howardwu: do they not have a release on crates.io?

also, i think it's a non-starter if we have to install a new binary

joske: There is a release, but it's 5 years old, that's why I was using a fork of the main repo.

howardwu: What are the main changes in the last 5 years? If there were material security fixes, then we need to use the remote instead of the release. But regardless, if one needs to install something extra, this may be a nonstarter.

joske: The only difference is it builds the rust bindings against a more recent native tcmalloc.

vicsn:
I suggest we close this for now, as our tests have shown even though its a big improvement, it doesn't solve our core memory issue.

joske:
@vicsn I agree, even more so now we found the jemalloc options that give similar results.

PR #3211, opened 2024-04-11
[Optimize] Primary and Worker configuration tweaks

raychu86:
## Motivation

This PR updates the configurations with the following changes:

1. Randomly select transmissions for the worker ping
    - Previously, we would always take the first 5
    - This changes gives a better distribution of transmissions.

2. Relax the proposal requirement to allow proposals with no transmissions.
    - Blocks with no transactions is already allowed, this relaxes the requirements on the snarkOS side

3. Reinsert transmissions to the worker if a proposal fails
   - Previously we were erroneously discarding transmissions if the proposal fails, now we keep them around.


vicsn:
If we don't need transactions anymore in batch proposals, is this also a good moment to remove `initialize_transaction_pool` (and the commented out code above it, and the `--no-dev-txs` flag)?

howardwu (approved):

PR #3214, opened 2024-04-12
[Optimize] Removes `request_timeouts` from `BlockSync`

howardwu:
## Motivation

[Optimize] Removes `request_timeouts` from `BlockSync`


raychu86 (commented):
LGTM. `request_timeouts` was unnecessary and could be held in memory forever.

raychu86 (approved):

PR #3216, opened 2024-04-13
Adding transaction types and aborted transactions/solutions to metrics

miazn:
## Motivation
It's useful to have more granular metrics as to what types of transactions are included in blocks, as well as aborted transactions and solutions.

This pr adds 6 new metrics:
```
1. Accepted Deploy
2. Accepted Execute
3. Rejected Deploy
4. Rejected Execute
5. Aborted Transactions
6. Aborted Solutions
```

I did not yet remove the existing confirmed transactions (which we can get by summing the accepted deploys and executes), but that would probably be best to avoid duplicate data (can sum in post processing for analysis).
## Test Plan
Screenshot 2024-04-11 at 3 42 45 PM
Still testing to get all types recorded


miazn:
@ljedrz is there a more optimal way to do the enum matching?

ljedrz:
Was this found to be slow? Enum matching on its own shouldn't be a big deal, it's the mutable borrows that e.g. disallow parallel processing; if the iteration needs to be faster, the counters can be made atomic, then the `_mut`s can be dropped from the `get`s, and the iteration can be parallelized with `rayon`.

node/bft/ledger-service/src/traits.rs
```diff
@@ -118,4 +118,8 @@ pub trait LedgerService<N: Network>: Debug + Send + Sync {
     /// Adds the given block as the next block in the ledger.
     #[cfg(feature = "ledger-write")]
     fn advance_to_next_block(&self, block: &Block<N>) -> Result<()>;
+
+    // Updates the block metrics.
+    #[cfg(feature = "metrics")]
+    fn update_block_metrics(&self, _block: &Block<N>) {}
```
iamalwaysuncomfortable: Just as a general nit, traits are interfaces and when other Rust code needs to accept a trait parameter elsewhere, it expects the concrete type that implements it to have it implemented as well. 

Putting a feature flag in a trait basically makes other implementors of that trait need to also implement a feature flag in their implementation further down the line which can be lead to a lot of problems. Thus in general it's good NOT to put feature flags in traits for this reason unless it's somehow absolutely necessary. (There is another feature flag in the trait above, but this is now adding a separate one) 

In the concrete implementation of this above, the &self parameter isn't used and only block data from the parameter `block` input itself is ever used (i.e. you don't need to access &self at all). Therefore you don't need it as part of the trait and can make the `update_block_metrics` call from either a standalone function or a method on another object.

Luckily though this should be a simple change!

node/bft/ledger-service/src/ledger.rs
```diff
@@ -343,9 +345,49 @@ impl<N: Network, C: ConsensusStorage<N>> LedgerService<N> for CoreLedgerService<
...
             metrics::increment_gauge(metrics::blocks::TRANSACTIONS, num_tx as f64);
+            self.update_block_metrics(block);
         }
 
         tracing::info!("\n\nAdvanced to block {} at round {} - {}\n", block.height(), block.round(), block.hash());
         Ok(())
     }
+
+    #[cfg(feature = "metrics")]
+    fn update_block_metrics(&self, block: &Block<N>) {
```
iamalwaysuncomfortable: To track Aborted Transactions & Aborted Solutions. Block has two methods:

`aborted_solution_ids` and `aborted_transaction_ids` - you can use these two methods to count the aborted solutions & transactoins.

iamalwaysuncomfortable (dismissed):
Left a comment about how to track aborted solutions & transactions.

The one major nit is to remove the metrics method from the ledger trait. Reasons why are explained int he comments.

Thanks for working to contribute this to the node software. This will make the life of all node operators easier once it's merged!

node/bft/ledger-service/src/traits.rs
miazn: Yeah I was initially trying to make it a standalone function but for some reason it would not compile. Trying this again now

node/bft/ledger-service/src/ledger.rs
miazn: I used those in the function already, they are the last two lines

node/bft/ledger-service/src/traits.rs
miazn: Changed :D

node/bft/ledger-service/src/ledger.rs
iamalwaysuncomfortable: Nice! Let me take a second look.

node/bft/ledger-service/src/lib.rs
```diff
@@ -60,3 +67,38 @@ macro_rules! spawn_blocking {
         }
     };
 }
+
+#[cfg(feature = "metrics")]
+fn update_block_metrics<N: Network>(block: &Block<N>) {
```
iamalwaysuncomfortable: My personal opinion on where this should go is in the `metrics` crate. This way if there's any question about the metrics collection logic, we have a singular place to look. My fear is that if we end up having a lot of metrics collection logic beyond this, it may get hard to keep track of it - and keeping as much as is possible in the metrics crate will help us keep track of our metrics logic.

Does anyone have any other opinions on where this function should live?

node/metrics/snarkOS-grafana.json
```diff
@@ -203,7 +203,7 @@
             "uid": "${DS_PROMETHEUS}"
           },
           "editorMode": "code",
-          "expr": "avg(\n  1 / rate(snarkos_blocks_height_total{}[1m])\n) < +inf",
+          "expr": "avg(\n  1 / rate(snarkos_bft_height_total{}[1m])\n) < +inf",
```
miazn, line 206: Changed these names just to reflect changes in https://github.com/AleoHQ/snarkOS/pull/3210

node/bft/ledger-service/src/lib.rs
```diff
@@ -60,3 +67,38 @@ macro_rules! spawn_blocking {
...
+    use snarkvm::ledger::ConfirmedTransaction;
+
+    let accepted_deploy = AtomicUsize::new(0);
+    let accepted_execute = AtomicUsize::new(0);
+    let rejected_deploy = AtomicUsize::new(0);
+    let rejected_execute = AtomicUsize::new(0);
+
+    // Add transaction to atomic counter based on enum type match.
+    block.transactions().par_iter().for_each(|tx| match tx {
```
howardwu: `cfg_iter!`?

iamalwaysuncomfortable: @miazn for context `cfg_iter!` is a Rust macro within SnarkVM that will iterate through a collection either in parallel or serial depending on the feature flag. 

A usage example is can be found here [here](https://github.com/AleoHQ/snarkVM/blob/d48f6fb7d7ca36d134407baebcc59952fe58c546/ledger/block/src/verify.rs#L245-L255). 

The actual source code behind this macro is [here](https://github.com/AleoHQ/snarkVM/blob/d48f6fb7d7ca36d134407baebcc59952fe58c546/utilities/src/parallel.rs#L88-L98).

Since there exists a `serial` flag that will disable parallel iteration. It's important to use this macro to ensure that this code will still run if the `serial` flag is enabled.

miazn: ah gotcha, wasn't privy to that -- changed :D

node/bft/ledger-service/src/lib.rs
iamalwaysuncomfortable: Pending other reviewer's comments on where this function should live, this PR has my approval.

node/bft/ledger-service/Cargo.toml
```diff
@@ -51,6 +51,9 @@ optional = true
 version = "0.8"
 optional = true
 
+[dependencies.rayon]
+version = "1"
```
howardwu: `optional = true` and add `serial`?

miazn: done!

miazn:
after discussing with @iamalwaysuncomfortable , we decided it might make the most sense to move the metrics related functions to the actual metrics crate. This makes it possible to import `rayon` optionally when `metrics` is enabled and `serial` is disabled , as well as the added bonus of cleaning up protocol logic code

howardwu:
thanks for making the change!

node/bft/ledger-service/Cargo.toml
```diff
@@ -51,6 +52,10 @@ optional = true
 version = "0.8"
 optional = true
 
+[dependencies.rayon]
```
howardwu: This should be removed.

node/bft/ledger-service/Cargo.toml
```diff
@@ -23,6 +23,7 @@ ledger-write = [ ]
 metrics = ["dep:metrics", "snarkvm/metrics"]
 mock = [ "parking_lot", "tracing" ]
 prover = [ ]
+serial = [ ]
```
howardwu: This should be removed.

node/metrics/Cargo.toml
```diff
@@ -17,15 +17,27 @@ license = "Apache-2.0"
...
+
 [dependencies.snarkvm]
 workspace = true
 features = [ "metrics" ]
 
 [dependencies.tokio]
 version = "1.28"
 features = [ "rt" ]
+
+[dependencies.time]
```
howardwu: nit: this should be reordered for alphabetical ordering

node/metrics/Cargo.toml
```diff
@@ -17,15 +17,27 @@ license = "Apache-2.0"
...
+version = "0.3"
```
raychu86:
```suggestion
version = "0.3"

```

nit: new line at end of file

node/metrics/src/lib.rs
```diff
@@ -16,8 +16,23 @@ mod names;
 
 // Expose the names at the crate level for easy access.
 pub use names::*;
+use parking_lot::Mutex;
 // Re-export the snarkVM metrics.
+#[cfg(not(feature = "serial"))]
+use rayon::prelude::*;
 pub use snarkvm::metrics::*;
```
howardwu: nit: please fix formatting and import ordering

node/metrics/src/lib.rs
```diff
@@ -16,8 +16,24 @@ mod names;
 
 // Expose the names at the crate level for easy access.
 pub use names::*;
+use parking_lot::Mutex;
+
 // Re-export the snarkVM metrics.
+#[cfg(not(feature = "serial"))]
+use rayon::prelude::*;
 pub use snarkvm::metrics::*;
```
howardwu:
```suggestion
pub use names::*;

// Re-export the snarkVM metrics.
pub use snarkvm::metrics::*;

#[cfg(not(feature = "serial"))]
use rayon::prelude::*;

use parking_lot::Mutex;
```

howardwu (approved):

PR #3217, opened 2024-04-14
[Fix] Add aborted transmission IDs to storage service.

raychu86:
## Motivation

This PR adds support for tracking aborted transmission IDs in storage. This solves an edge case in syncing via BFT when there are aborted transmissions in certificates:

1. We can’t find a transmission  in block `X` because it was aborted in block `X-1`
2. BFT block sync requires certificates from block `X` in order to commit block `X-1`
3. We have not added Block `X-1` into our ledger and started tracking that the transmission was indeed aborted
4. When we process certificates in block `X`, we don’t know where the transmission is and that it was supposed to be aborted in block `X-1`, which causes an error.

The sister PR is here - https://github.com/AleoHQ/snarkVM/pull/2433

## Test Plan

The tests have been updated to check that we properly track aborted transmissions in storage.

### DIFF
--- a/node/bft/src/primary.rs
+++ b/node/bft/src/primary.rs
@@ -2552,7 +2552,7 @@ mod tests {
 
         // Ensure the certificate exists in storage.
         assert!(primary.storage.contains_certificate(certificate_id));
-        // Ensure that the aborted transmissions are exist in the storage.
+        // Ensure that the aborted transmission IDs exist in storage.
         for aborted_transmission_id in aborted_transmissions {
             assert!(primary.storage.contains_transmission(aborted_transmission_id));
             assert!(primary.storage.get_transmission(aborted_transmission_id).is_none());
--- a/node/bft/src/primary.rs
+++ b/node/bft/src/primary.rs
@@ -2545,9 +2545,17 @@ mod tests {
         );
 
         // Insert the certificate to storage.
-        primary.storage.insert_certificate(certificate, transmissions_without_aborted, aborted_transmissions).unwrap();
+        primary
+            .storage
+            .insert_certificate(certificate, transmissions_without_aborted, aborted_transmissions.clone())
+            .unwrap();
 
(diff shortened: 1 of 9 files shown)


node/bft/src/primary.rs
```diff
@@ -2545,9 +2545,17 @@ mod tests {
...
         // Insert the certificate to storage.
-        primary.storage.insert_certificate(certificate, transmissions_without_aborted, aborted_transmissions).unwrap();
+        primary
+            .storage
+            .insert_certificate(certificate, transmissions_without_aborted, aborted_transmissions.clone())
+            .unwrap();
 
         // Ensure the certificate exists in storage.
         assert!(primary.storage.contains_certificate(certificate_id));
+        // Ensure that the aborted transmissions are exist in the storage.
```
iamalwaysuncomfortable: Comment nit.

```rust
// Ensure that the aborted transmissions exist in storage.
```

iamalwaysuncomfortable (commented):
Small comment nit

iamalwaysuncomfortable (commented):
The 1 hour long-running e2e load tests against this passed.

node/bft/storage-service/src/memory.rs
```diff
@@ -124,6 +130,23 @@ impl<N: Network> StorageService<N> for BFTMemoryService<N> {
                 }
             }
         }
+        // Inserts the aborted transmission IDs.
+        for aborted_transmission_id in aborted_transmission_ids {
+            // Retrieve the transmission entry.
+            match aborted_transmission_ids_lock.entry(aborted_transmission_id) {
```
vicsn, line 136: If Pratyush would see this he would tell you to use [or_insert](https://doc.rust-lang.org/std/collections/hash_map/enum.Entry.html#method.or_insert)

vicsn (approved):
LGTM

Issue #3220, opened 2024-04-16
[Feature] Need a way to NOT change REST port number when --dev parameter is specified


robwafle:
## 🚀 Feature

Need a way to NOT change REST port number when --dev parameter is specified

## Motivation

This will allow clients to be in a load balancer similar to a production environment where the load balancer points to all 30 clients (e.g. 3 clients for each of the 10 validators). We need development testing environments to match production as closely as possible, and without this we can't have a dev environment that closely resembles production.

## Implementation

Hopefully, this is only logic to parse the parameter and one if statement wrapping a small amount of logic that actually doesn't happen in production networks.   (when --dev parameter is NOT specified )

I believe the +dev on this line is part of what would have to be made conditional:
https://github.com/AleoHQ/snarkOS/blob/mainnet-staging/cli/src/commands/start.rs#L328

vicsn:
See for a similar or the same issue: https://github.com/AleoHQ/snarkOS/pull/3189

raychu86:
Should be resolved with https://github.com/AleoNet/snarkOS/pull/3271

PR #3221, opened 2024-04-18
[Fix] Make --rest optional to prevent auto-change of REST port with --dev

christianwwwwwwww:
## Motivation

Solves Issue #3220 wherein specified rest port was overwritten by the default when the `--dev` flag was being used. 

## Test Plan

This will only change a node's behavior when both `--dev` and `--rest` are used. 

Tested locally. 

## Related PRs

(Link any related PRs here)

### DIFF
--- a/cli/src/commands/start.rs
+++ b/cli/src/commands/start.rs
@@ -506,8 +506,8 @@ impl Start {
         };
         // Parse the REST IP.
             let rest_ip = match self.norest {
-                true => None,
-                false => self.rest,
+            true => None,
+            false => self.rest,
         };
 
         // If the display is not enabled, render the welcome message.
@@ -856,7 +856,7 @@ mod tests {
         config.parse_development(&mut trusted_peers, &mut trusted_validators).unwrap();
         let expected_genesis = config.parse_genesis::<CurrentNetwork>().unwrap();
         assert_eq!(config.node, Some(SocketAddr::from_str("0.0.0.0:4130").unwrap()));
-        assert_eq!(config.rest, SocketAddr::from_str("0.0.0.0:3030").unwrap());
+        assert_eq!(config.rest, Some(SocketAddr::from_str("0.0.0.0:3030").unwrap()));
         assert_eq!(trusted_peers.len(), 0);
         assert_eq!(trusted_validators.len(), 1);
         assert!(!config.validator);
@@ -871,7 +871,7 @@ mod tests {
         config.parse_development(&mut trusted_peers, &mut trusted_validators).unwrap();
         let genesis = config.parse_genesis::<CurrentNetwork>().unwrap();
         assert_eq!(config.node, Some(SocketAddr::from_str("0.0.0.0:4131").unwrap()));
-        assert_eq!(config.rest, SocketAddr::from_str("0.0.0.0:3031").unwrap());
+        assert_eq!(config.rest, Some(SocketAddr::from_str("0.0.0.0:3031").unwrap()));
         assert_eq!(trusted_peers.len(), 1);
         assert_eq!(trusted_validators.len(), 1);
         assert!(config.validator);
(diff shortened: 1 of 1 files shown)


cli/src/commands/start.rs
```diff
@@ -903,7 +901,7 @@ mod tests {
         config.parse_development(&mut trusted_peers, &mut trusted_validators).unwrap();
         let genesis = config.parse_genesis::<CurrentNetwork>().unwrap();
         assert_eq!(config.node, Some(SocketAddr::from_str("0.0.0.0:4133").unwrap()));
-        assert_eq!(config.rest, SocketAddr::from_str("0.0.0.0:3033").unwrap());
+        assert_eq!(config.rest, Some(SocketAddr::from_str("0.0.0.0:3033").unwrap()));
```
iamalwaysuncomfortable, line 943: The second ask would be to add some further tests here.

iamalwaysuncomfortable (changes requested):

cli/src/commands/start.rs
```diff
@@ -509,7 +507,7 @@ impl Start {
         // Parse the REST IP.
         let rest_ip = match self.norest {
             true => None,
-            false => Some(self.rest),
+            false => self.rest,
```
iamalwaysuncomfortable: Now that self.rest is an `Option` enum, this match statement needs to return a `Some` value with the default `0.0.0.0::3030` in the `false` case if the user does not specify as REST IP - otherwise both true or false could return a `None` which would NOT be the desired behavior.

iamalwaysuncomfortable (changes requested):

cli/src/commands/start.rs
christianwwwwwwww: fixed ty

cli/src/commands/start.rs
christianwwwwwwww: I can add more tests tomorrow 🙏

cli/src/commands/start.rs
iamalwaysuncomfortable: Thanks!

cli/src/commands/start.rs
iamalwaysuncomfortable: Appreciate it!

cli/src/commands/start.rs
```diff
@@ -837,6 +835,25 @@ mod tests {
...
+        assert_eq!(config.rest, Some(SocketAddr::from_str("127.0.0.1:8080").unwrap()));
+
+        // Test default REST IP when REST flag is not passed in prod mode
+        let config = Start::try_parse_from(["snarkos"].iter()).unwrap();
+        assert_eq!(config.rest, Some(SocketAddr::from_str("0.0.0.0:3030").unwrap()));
+
+        // Test specified REST IP when passed in prod mode
+        let config = Start::try_parse_from(["snarkos", "--rest", "192.168.1.1:8080"].iter()).unwrap();
+        assert_eq!(config.rest, Some(SocketAddr::from_str("192.168.1.1:8080").unwrap()));
+    }
```
iamalwaysuncomfortable: This is a great start, can you write a few more test cases covering more potential cases?

iamalwaysuncomfortable (commented):
This needs a few more tests, but looking good

cli/src/commands/start.rs
christianwwwwwwww: 🫡

damons:
I'm not sure of the scope of this change, but no matter what, I want to make sure that anytime we require a port that we can set both host:port to a specific values.  There's a tendency for us to simply bind to 0.0.0.0:port and this breaks multi-homed deployments in production environments.  Dev environments, I understand, are special, but for everything else, these values, at a minimum, should be configurable either via command line or in a config file.

PR #3223, opened 2024-04-18
[Optimize] Do not `PeerRefresh` syncing peers.

raychu86:
## Motivation

This PR adds 2 changes:

1. A cache for `inbound_block_requests` to limit the number of block requests a node can send every minute.
2. Prevent `PeerRefresh` on peers that are syncing from you or if you are syncing from that peer.

Note: The current rate limit is quite generous at 256 block requests every 60 seconds.

### DIFF
--- a/node/router/src/heartbeat.rs
+++ b/node/router/src/heartbeat.rs
@@ -123,6 +123,8 @@ pub trait Heartbeat<N: Network>: Outbound<N> {
             .get_connected_peers()
             .iter()
             .filter(|peer| !trusted.contains(&peer.ip()) && !bootstrap.contains(&peer.ip()))
+            .filter(|peer| !self.router().cache.contains_inbound_block_request(&peer.ip())) // Skip if the peer is syncing.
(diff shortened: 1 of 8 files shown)


node/router/src/heartbeat.rs
```diff
@@ -123,6 +123,8 @@ pub trait Heartbeat<N: Network>: Outbound<N> {
             .get_connected_peers()
             .iter()
             .filter(|peer| !trusted.contains(&peer.ip()) && !bootstrap.contains(&peer.ip()))
+            .filter(|peer| !self.router().cache.contains_inbound_block_request(&peer.ip())) // Skip if the peer is syncing.
+            .filter(|peer| self.is_block_synced() || self.router().cache.num_outbound_block_requests(&peer.ip()) == 0) // Skip if you are syncing from this peer.
```
vicsn, line 127: Using a single call to filter() doing all the exclusion checks is likely more efficient. It may not be material given the small number of peers, but still a nice improvement.

vicsn (approved):
One nit but otherwise LGTM.

Don't forget to make a reminder/issue to lower `MAXIMUM_BLOCK_REQUESTS_PER_INTERVAL` after we tested the ideal number of blocks to sync and the request rate it leads to. :)

howardwu (approved):

PR #3225, opened 2024-04-18
[Optimize] Add guards for processing unconfirmed solutions

raychu86:
## Motivation

This PR adds a few changes to improve syncing and reduce network spam:

1. Clients perform solution verification in `POST /broadcast/solution` prior to solution propagation.
    - This reduces the propagation of invalid solutions on the network.
2. Nodes do not process incoming unconfirmed solutions if they are syncing.
    - This reduces the resource load of nodes and lets them focus on syncing.
3. Nodes do not process incoming unconfirmed transactions if they are syncing
    - This reduces the resource load of nodes and lets them focus on syncing.

### DIFF
--- a/node/rest/src/routes.rs
+++ b/node/rest/src/routes.rs
@@ -315,7 +315,7 @@ impl<N: Network, C: ConsensusStorage<N>, R: Routing<N>> Rest<N, C, R> {
     ) -> Result<ErasedJson, RestError> {
         // Do not process the transaction if the node is syncing.
         if !rest.routing.is_block_synced() {
-            return Err(RestError(format!("Unable to broadcast transaction {} - node is syncing.", fmt_id(tx.id()))));
+            return Err(RestError(format!("Unable to broadcast transaction '{}' (node is syncing)", fmt_id(tx.id()))));
         }
 
         // If the consensus module is enabled, add the unconfirmed transaction to the memory pool.
@@ -345,7 +345,7 @@ impl<N: Network, C: ConsensusStorage<N>, R: Routing<N>> Rest<N, C, R> {
         // Do not process the solution if the node is syncing.
         if !rest.routing.is_block_synced() {
             return Err(RestError(format!(
-                "Unable to broadcast solution {} - node is syncing.",
+                "Unable to broadcast solution '{}' (node is syncing)",
                 fmt_id(solution.id())
             )));
         }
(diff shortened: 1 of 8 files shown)


node/rest/src/routes.rs
```diff
@@ -338,9 +338,27 @@ impl<N: Network, C: ConsensusStorage<N>, R: Routing<N>> Rest<N, C, R> {
         Json(solution): Json<Solution<N>>,
     ) -> Result<ErasedJson, RestError> {
         // If the consensus module is enabled, add the unconfirmed solution to the memory pool.
-        if let Some(consensus) = rest.consensus {
+        // Otherwise, verify it prior to broadcasting.
```
vicsn, line 354: I can imagine spamming a publicly accessible REST endpoint with solutions could halt a client due to the amount of verifications required. But we should test (cc @iamalwaysuncomfortable).

If it turns out to be an issue, we can:
- accept the risk and tell everyone not to expose their REST endpoint...
- add a separate [transmission rate limit](https://docs.rs/tower_governor/0.3.2/tower_governor/governor/struct.GovernorConfigBuilder.html#method.methods) using a second `.layer(GovernorLayer { }`.
- queue transmissions to verify, takes more work but seems more robust. See my comment below for an example implementation

node/rest/src/routes.rs
```diff
@@ -338,9 +338,27 @@ impl<N: Network, C: ConsensusStorage<N>, R: Routing<N>> Rest<N, C, R> {
...
+        match rest.consensus {
             // Add the unconfirmed solution to the memory pool.
-            consensus.add_unconfirmed_solution(solution).await?;
+            Some(consensus) => consensus.add_unconfirmed_solution(solution).await?,
+            // Verify the solution.
+            None => {
```
vicsn, line 359: If clients receive transactions/solutions via P2P, they are already verified **before** propagation in `fn unconfirmed_{transaction, solution}`. So I suggest we just call that function from here. This is what the validator also does.

For consistency and clarity, we should also do this for transactions. A solution takes longer to verify than an execution, but less than a deployment, so there's no difference justifying a different verification approach.

Recall this PR was triggered in part because recently we did a large invalid solution spam, but @iamalwaysuncomfortable and I just realized we never tried spamming very large amounts of invalid transactions.

vicsn: This PR also remains relevant in the face of solution or deployment spam: https://github.com/AleoHQ/snarkOS/pull/2970

iamalwaysuncomfortable:
An automated integration test has been kicked off by Robert

iamalwaysuncomfortable:
The following measures are being done to test this

- [x] Run a 30 client, 10 validator devnet with a high load of solution spam
- [ ] Analyze the results

The devnet run has completed, and analysis of it is forthcoming.

howardwu (approved):

PR #3227, opened 2024-04-19
Disconnect from requests decreasing in height

vicsn:
## Motivation

Prevents a DoS attack by a peer who keeps requesting large amounts of the same blocks. We disconnect as soon as we encounter a repetitive request, and it is allowed again after `INBOUND_BLOCK_REQUEST_INTERVAL = 60 seconds`.

I tried to re-use the Cache abstraction without imposing too many additional assumptions.

## Test Plan

Added some unit tests, but this should definitely be tested on a running network.

## Related PRs

Extends: https://github.com/AleoHQ/snarkOS/pull/3223

### DIFF
--- a/node/router/messages/src/block_request.rs
+++ b/node/router/messages/src/block_request.rs
@@ -18,7 +18,7 @@ use snarkvm::prelude::{FromBytes, ToBytes};
 
 use std::borrow::Cow;
 
-#[derive(Copy, Clone, Debug, PartialEq, Eq, Hash)]
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Hash, Default)]
 pub struct BlockRequest {
     /// The starting block height (inclusive).
     pub start_height: u32,
--- a/node/router/messages/src/block_request.rs
+++ b/node/router/messages/src/block_request.rs
@@ -18,7 +18,7 @@ use snarkvm::prelude::{FromBytes, ToBytes};
 
 use std::borrow::Cow;
 
-#[derive(Copy, Clone, Debug, PartialEq, Eq, Hash, Default)]
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Hash)]
 pub struct BlockRequest {
     /// The starting block height (inclusive).
     pub start_height: u32,
--- a/node/router/src/helpers/cache.rs
+++ b/node/router/src/helpers/cache.rs
@@ -244,7 +244,7 @@ impl<N: Network> Cache<N> {
             timestamps.pop_front();
         }
         // Check whether the request is higher than the last.
-        let incrementing = timestamps.back().map(|(_, last)| request.start_height > last.end_height).unwrap_or(true);
+        let incrementing = timestamps.back().map(|(_, last)| request.start_height >= last.end_height).unwrap_or(true);
         // Insert the new timestamp and request if its higher than the last.
         if incrementing {
             timestamps.push_back((now, request));
@@ -340,7 +340,7 @@ mod tests {
         assert_eq!(cache.seen_inbound_block_requests.read().len(), 0);
 
         // Insert a block request.
-        let request = BlockRequest { start_height: 1, end_height: 1 };
+        let request = BlockRequest { start_height: 1, end_height: 2 };
         assert_eq!(cache.insert_inbound_block_request(peer_ip, request), (1, true));
         // Check that the cache contains the block request.
         assert!(cache.contains_inbound_block_request(&peer_ip));
@@ -352,13 +352,13 @@ mod tests {
         assert!(cache.contains_inbound_block_request(&peer_ip));
 
         // Insert another block request for the same peer with a lower start height.
-        let request = BlockRequest { start_height: 1, end_height: 3 };
+        let request = BlockRequest { start_height: 2, end_height: 3 };
         assert_eq!(cache.insert_inbound_block_request(peer_ip, request), (2, false));
         // Check that the cache does contains the block requests.
         assert!(cache.contains_inbound_block_request(&peer_ip));
 
         // Insert another block request for the same peer with a low start height.
-        let request = BlockRequest { start_height: 1, end_height: 1 };
+        let request = BlockRequest { start_height: 1, end_height: 2 };
         assert_eq!(cache.insert_inbound_block_request(peer_ip, request), (2, false));
         // Check that the cache does contains the block requests.
         assert!(cache.contains_inbound_block_request(&peer_ip));
@@ -369,7 +369,7 @@ mod tests {
         ));
 
         // Insert another block request for the same peer with a low start height.
-        let request = BlockRequest { start_height: 1, end_height: 1 };
+        let request = BlockRequest { start_height: 1, end_height: 2 };
         assert_eq!(cache.insert_inbound_block_request(peer_ip, request), (1, true));
         // Check that the cache does contains the block requests.
         assert!(cache.contains_inbound_block_request(&peer_ip));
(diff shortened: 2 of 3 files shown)


node/router/messages/src/block_request.rs
```diff
@@ -18,7 +18,7 @@ use snarkvm::prelude::{FromBytes, ToBytes};
 
 use std::borrow::Cow;
 
-#[derive(Copy, Clone, Debug, PartialEq, Eq, Hash)]
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Hash, Default)]
```
howardwu: I would suggest not using `Default` here because this would technically be a malformed `BlockRequest` (which goes against the spirit of the purpose for `Default` in Rust)

node/router/src/helpers/cache.rs
```diff
@@ -221,9 +226,36 @@ impl<N: Network> Cache<N> {
...
+
+        let mut map_write = map.write();
+        // Load the entry for the key.
+        let timestamps = map_write.entry(key).or_default();
+        // Retain only the timestamps that are within the recent interval.
+        while timestamps.front().map_or(false, |(t, _)| now - *t > Duration::seconds(interval_in_secs)) {
+            timestamps.pop_front();
+        }
+        // Check whether the request is higher than the last.
+        let incrementing = timestamps.back().map(|(_, last)| request.start_height > last.end_height).unwrap_or(true);
```
raychu86:
```suggestion
        let incrementing = timestamps.back().map(|(_, last)| request.start_height >= last.end_height).unwrap_or(true);
```

Should it be a `>=` because the `end_height` is exclusive whereas the `start_height` is inclusive?

node/router/messages/src/block_request.rs
vicsn: 7e16a7fab

node/router/src/helpers/cache.rs
vicsn: 255a459a0

node/router/src/inbound.rs
```diff
@@ -78,11 +78,15 @@ pub trait Inbound<N: Network>: Reading + Outbound<N> {
...
-                let frequency = self.router().cache.insert_inbound_block_request(peer_ip);
+                let (frequency, increasing_height) = self.router().cache.insert_inbound_block_request(peer_ip, message);
                 // Check if the number of block requests is within the limit.
                 if frequency > Self::MAXIMUM_BLOCK_REQUESTS_PER_INTERVAL {
                     bail!("Peer '{peer_ip}' is not following the protocol (excessive block requests)")
                 }
+                // Check if the requested height is higher than the last request.
+                if !increasing_height {
+                    bail!("Peer '{peer_ip}' is not following the protocol (block request heights not increasing)")
+                }
```
iamalwaysuncomfortable, line 89: The one safety nit here is what if the sending node perhaps fails to send OR the block response fails and a legitimate retry is needed? 

Potentially we should allow more than 1 try introduce a growing backoff?

howardwu:
What's the status on this PR? I see one unresolved conversation.

node/router/src/inbound.rs
vicsn: I had the same concern. My assumption is that failing responses are unlikely. And if they do happen, the "penalty" is only a 60 second disconnect. So I think there's no need for more complexity.

To validate the assumption, @kpandl can you run [the new client sync test](https://github.com/AleoHQ/stress-observability/pull/136 ) on this PR to see if we observe a regression in client sync speed?

@iamalwaysuncomfortable let us know if you think we should run other tests against this PR, or if we can do further testing on mainnet-staging :)

vicsn:
> What's the status on this PR? I see one unresolved conversation.

Still unresolved, thx for the ping

node/router/src/inbound.rs
kpandl: Yes I'll test that!

kpandl: Here are the benchmark results across 10 clients syncing for 30 mins:
* `mainnet-staging`: client sync block time 1.281-1.491s, validator block time 4.671s
* `this PR`: client sync block time 0.874-4.615s, validator block time 3.244s

The mainnet-staging hash was `6b54869c32803f703b559d7880cd1f05031ddc4f`, for this PR I cherry-picked the 3 commits on top of it.
One thing to note is that the TPS using `this PR` was repeatedly lower (6-8) than in `mainnet-staging` with 17, despite using the same test setup. Validators were `m5.16xlarge`, clients `m5.4xlarge`.

vicsn: Thank you! Very interesting data.

Some follow up questions:
1. Are you measuring the sync speed of the 10 "last" clients which you add at the end? Can you also report the sync speed range of the 20 initial clients for either setup?
2. I'm assuming the lower TPS is because `this PR`'s clients are out of sync more frequently. 
a. Do you see `(node is syncing)` logs more frequently in `this PR`'s clients compared to `mainnet-staging`? 
b. What about `Unable to broadcast transaction`?
3. In my understanding, the only reason this PR would lower sync speed, is if clients are disconnected for 60+ seconds for asking the same block more than once. Can you validate this:
a. Do you see any ~60 second stalls when you're looking at `this PR`'s clients? And in comparison, is `mainnet-staging` a smoother straight line going up?
b. Do you see any WARN level `(block request heights not increasing)` in the logs of `this PR`'s clients?

kpandl: I did two further runs based of commit `fc340c679960e63612c536d69e71405b77e113f4`, here's the data:

1. Yes, it was the 10 last clients previously. Here is the data for all clients, note that the initial clients sync a larger share of empty blocks, as the tx cannons boot up later than the snarkos nodes:
a) `this PR`: initial client sync block time 0.493-0.705s, later client sync block time 1.608-2.917s, validator block time 3.545s
b) `mainnet-staging`: initial client sync block time 0.727-0.895s, later client sync block time 2.187-2.34s, validator block time 5.252s (interestingly, the latter is higher this time)
2. The TPS this time was 10 (`this PR`) vs. 17.
a) `node is syncing` is logged by the tx cannons (but not by the clients), and it is only logged in the case of `this PR`. Of two network driver cannon logs that I fetched, it was logged by one cannon 20 times, another cannon 28 times. On an absolute basis, this number appears low, as the two cannons were running for 10 hours and each had >5000 logs of `broadcast_successful`.
b) `Unable to broadcast transaction` was logged at the same time as `node is syncing`, same tx cannons and number of occurrences. The entire log looks like this and contains both of these logs twice (so effectively we could divide the number of occurrences by 2): `[2m2024-05-07T12:51:04.726911Z[0m [31mERROR[0m [2mtx_cannon::commands::batch_execute::helpers::step[0m[2m:[0m [2m216:[0m {"event": "step_execution_failed", "account": "APrivateKey1zkp8CZNn3yeCseEtxuVPbDCwSyhGW6yZKUYKfgXmcpoGPWH", "order": 1, "transaction_error": "Err(❌ Failed to broadcast execution to http://snarkos-client-lb-1292457601.us-west-2.elb.amazonaws.com:3030/mainnet/transaction/broadcast: (status code 500: "Something went wrong: Unable to broadcast transaction 'at1rq9tndlylcag9..' (node is syncing)"))", "function": "credits.aleo/transfer_public", "retries": 3, "error": "Err(❌ Failed to broadcast execution to http://snarkos-client-lb-1292457601.us-west-2.elb.amazonaws.com:3030/mainnet/transaction/broadcast: (status code 500: "Something went wrong: Unable to broadcast transaction 'at1rq9tndlylcag9..' (node is syncing)"))" }`
4. Here's the data:
a. Yes exactly, see the two figures attached below. The one with higher block times is `this PR`.
b. Yes, plenty. 3689 occurrences across 25 log files, no occurrence of it in `mainnet-staging`.

this_pr
mainnet-staging

vicsn:
Closing this because:
1. [honest nodes request the same block twice sometimes under very normal conditions](https://github.com/AleoHQ/snarkOS/pull/3227#discussion_r1593739411). Great find @kpandl !
2. It's not that great of a DoS protection, an attacker could use multiple clients which intermittendly keep requesting increasing block ranges

PR #3229, opened 2024-04-21
[Feature] Check transaction size limit in codec

niklaslong:
Enforces the `MAX_TRANSACTION_SIZE` on the decode half of the router codec for unconfirmed transactions. See https://github.com/AleoHQ/snarkVM/pull/2437#discussion_r1570512045 and the corresponding PR in VM: https://github.com/AleoHQ/snarkVM/pull/2438 for context.

I chose to avoid modifying the `FromBytes` and `ToBytes` trait implementation on `Message`  for now and instead read the ID bytes twice to keep special casing simple but I'd be happy to implement a `decode_le` that takes in the ID as well.

Can be merged once the corresponding PRs in VM are in.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -46,7 +46,7 @@ default-features = false
 
 [workspace.dependencies.snarkvm]
 git = "https://github.com/AleoHQ/snarkVM.git"
-rev = "afacda3"
+rev = "57641ca"
 #version = "=0.16.18"
 features = [ "circuit", "console", "rocks" ]
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -46,7 +46,7 @@ default-features = false
 
 [workspace.dependencies.snarkvm]
 git = "https://github.com/AleoHQ/snarkVM.git"
-rev = "aef07da"
+rev = "afacda3"
 #version = "=0.16.18"
 features = [ "circuit", "console", "rocks" ]
 
--- a/node/bft/ledger-service/src/ledger.rs
+++ b/node/bft/ledger-service/src/ledger.rs
@@ -216,29 +216,26 @@ impl<N: Network, C: ConsensusStorage<N>> LedgerService<N> for CoreLedgerService<
         match (transmission_id, transmission) {
             (TransmissionID::Ratification, Transmission::Ratification) => {}
             (TransmissionID::Transaction(expected_transaction_id), Transmission::Transaction(transaction_data)) => {
-                match transaction_data.clone().deserialize_blocking() {
-                    Ok(transaction) => {
-                        // Ensure the transaction ID matches the expected transaction ID.
-                        if transaction.id() != expected_transaction_id {
-                            bail!(
-                                "Received mismatching transaction ID  - expected {}, found {}",
-                                fmt_id(expected_transaction_id),
-                                fmt_id(transaction.id()),
-                            );
-                        }
-
-                        // Ensure the transaction is not a fee transaction.
-                        if transaction.is_fee() {
-                            bail!("Received a fee transaction in a transmission");
-                        }
-
-                        // Update the transmission with the deserialized transaction.
-                        *transaction_data = Data::Object(transaction);
-                    }
-                    Err(err) => {
-                        bail!("Failed to deserialize transaction: {err}");
-                    }
+                // Deserialize the transaction. If the transaction exceeds the maximum size, then return an error.
+                let transaction = match transaction_data.clone() {
+                    Data::Object(transaction) => transaction,
+                    Data::Buffer(bytes) => Transaction::<N>::read_le(&mut bytes.take(N::MAX_TRANSACTION_SIZE as u64))?,
+                };
+                // Ensure the transaction ID matches the expected transaction ID.
+                if transaction.id() != expected_transaction_id {
+                    bail!(
+                        "Received mismatching transaction ID  - expected {}, found {}",
+                        fmt_id(expected_transaction_id),
+                        fmt_id(transaction.id()),
+                    );
                 }
+                // Ensure the transaction is not a fee transaction.
+                if transaction.is_fee() {
+                    bail!("Received a fee transaction in a transmission");
+                }
+
+                // Update the transmission with the deserialized transaction.
+                *transaction_data = Data::Object(transaction);
             }
             (TransmissionID::Solution(expected_solution_id), Transmission::Solution(solution_data)) => {
                 match solution_data.clone().deserialize_blocking() {
--- a/node/bft/ledger-service/src/ledger.rs
+++ b/node/bft/ledger-service/src/ledger.rs
@@ -22,7 +22,7 @@ use snarkvm::{
         store::ConsensusStorage,
         Ledger,
     },
-    prelude::{bail, Address, Field, Network, Result},
+    prelude::{bail, Address, Field, FromBytes, Network, Result},
 };
 
 use indexmap::IndexMap;
@@ -30,6 +30,7 @@ use lru::LruCache;
 use parking_lot::{Mutex, RwLock};
 use std::{
     fmt,
+    io::Read,
     ops::Range,
     sync::{
         atomic::{AtomicBool, Ordering},
@@ -294,8 +295,13 @@ impl<N: Network, C: ConsensusStorage<N>> LedgerService<N> for CoreLedgerService<
         transaction_id: N::TransactionID,
         transaction: Data<Transaction<N>>,
     ) -> Result<()> {
-        // Deserialize the transaction.
-        let transaction = spawn_blocking!(transaction.deserialize_blocking())?;
+        // Deserialize the transaction. If the transaction exceeds the maximum size, then return an error.
+        let transaction = spawn_blocking!({
+            match transaction {
+                Data::Object(transaction) => Ok(transaction),
+                Data::Buffer(bytes) => Ok(Transaction::<N>::read_le(&mut bytes.take(N::MAX_TRANSACTION_SIZE as u64))?),
+            }
+        })?;
         // Ensure the transaction ID matches in the transaction.
         if transaction_id != transaction.id() {
             bail!("Invalid transaction - expected {transaction_id}, found {}", transaction.id());
--- a/node/rest/src/routes.rs
+++ b/node/rest/src/routes.rs
@@ -335,8 +335,10 @@ impl<N: Network, C: ConsensusStorage<N>, R: Routing<N>> Rest<N, C, R> {
         }
 
         // If the transaction exceeds the transaction size limit, return an error.
+        // The buffer is initially roughly sized to hold a `transfer_public`,
+        // most transactions will be smaller and this reduces unnecessary allocations.
         // TODO: Should this be a blocking task?
-        let buffer = Vec::with_capacity(N::MAX_TRANSACTION_SIZE);
+        let buffer = Vec::with_capacity(3000);
         if tx.write_le(LimitedWriter::new(buffer, N::MAX_TRANSACTION_SIZE)).is_err() {
             return Err(RestError("Transaction size exceeds the byte limit".to_string()));
         }
--- a/node/rest/src/routes.rs
+++ b/node/rest/src/routes.rs
@@ -16,14 +16,13 @@ use super::*;
 use snarkos_node_router::{messages::UnconfirmedSolution, SYNC_LENIENCY};
 use snarkvm::{
     ledger::puzzle::Solution,
-    prelude::{block::Transaction, Identifier, Plaintext, ToBytes},
+    prelude::{block::Transaction, Identifier, LimitedWriter, Plaintext, ToBytes},
 };
 
 use indexmap::IndexMap;
 use rayon::prelude::*;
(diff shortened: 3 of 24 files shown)


node/router/messages/src/helpers/codec.rs
```diff
@@ -75,6 +75,23 @@ impl<N: Network> Decoder for MessageCodec<N> {
...
+        }
+
+        // Check the first two bytes for the message ID.
+        let id_bytes: [u8; 2] = (&bytes[..2])
+            .try_into()
+            .map_err(|_| std::io::Error::new(std::io::ErrorKind::InvalidData, "id couldn't be deserialized"))?;
+        let id = u16::from_le_bytes(id_bytes);
+
+        // SPECIAL CASE: check the transaction message isn't too large.
+        if id == 12 && len > N::MAX_TRANSACTION_SIZE {
```
howardwu: Could we move this check to be inside the specific message type itself so it is scoped appropriately?

niklaslong: Done. Let me know if I misinterpreted what you meant.

niklaslong:
Note, I've split the lock file changes out of the first commit for easier rebasing (originally by @d0cd, lmk if you have an email I can credit you with in the new commit).

node/rest/src/routes.rs
```diff
@@ -334,6 +335,13 @@ impl<N: Network, C: ConsensusStorage<N>, R: Routing<N>> Rest<N, C, R> {
             return Err(RestError(format!("Unable to broadcast transaction '{}' (node is syncing)", fmt_id(tx.id()))));
         }
 
+        // If the transaction exceeds the transaction size limit, return an error.
+        // TODO: Should this be a blocking task?
```
d0cd, line 340: Could we spawn a blocking task here?

niklaslong: I think `write_le` should be fast even for large txs. The only way to be sure is to hit the endpoint with a very large tx and measure the call, not sure if we could do that as a burn-in? Otherwise I suggest we leave this for a follow-up.

node/router/messages/src/lib.rs
```diff
@@ -152,6 +152,28 @@ impl<N: Network> Message<N> {
...
+        }
+
+        // Check the first two bytes for the message ID.
+        let id_bytes: [u8; 2] = (&bytes[..2])
+            .try_into()
+            .map_err(|_| io::Error::new(io::ErrorKind::InvalidData, "id couldn't be deserialized"))?;
+        let id = u16::from_le_bytes(id_bytes);
+
+        // SPECIAL CASE: check the transaction message isn't too large.
+        if id == 12 && len > N::MAX_TRANSACTION_SIZE {
```
howardwu, line 171: @d0cd can this `check_size` method be moved in to the respective file for message type 12?

niklaslong: I've moved it down into `UnconfirmedTransaction`, I think that's what you meant? 
One thing to bear in mind is this function takes in _all_ message bytes, not just `UnconfirmedTransaction`s, even though it only does the length check for the latter, it could be expanded in future to check other message types.

niklaslong:
(Rebased)

node/bft/ledger-service/src/ledger.rs
```diff
@@ -294,8 +295,13 @@ impl<N: Network, C: ConsensusStorage<N>> LedgerService<N> for CoreLedgerService<
...
     ) -> Result<()> {
-        // Deserialize the transaction.
-        let transaction = spawn_blocking!(transaction.deserialize_blocking())?;
+        // Deserialize the transaction. If the transaction exceeds the maximum size, then return an error.
+        let transaction = spawn_blocking!({
+            match transaction {
+                Data::Object(transaction) => Ok(transaction),
+                Data::Buffer(bytes) => Ok(Transaction::<N>::read_le(&mut bytes.take(N::MAX_TRANSACTION_SIZE as u64))?),
+            }
+        })?;
```
howardwu, line 301: You will need to do the same check here:
https://github.com/AleoHQ/snarkOS/blob/mainnet-staging/node/bft/ledger-service/src/ledger.rs#L218

node/rest/src/routes.rs
```diff
@@ -334,6 +335,13 @@ impl<N: Network, C: ConsensusStorage<N>, R: Routing<N>> Rest<N, C, R> {
             return Err(RestError(format!("Unable to broadcast transaction '{}' (node is syncing)", fmt_id(tx.id()))));
         }
 
+        // If the transaction exceeds the transaction size limit, return an error.
```
howardwu, line 337: Technically, this function's signature takes in: `Json(tx): Json<Transaction<N>>` as argument, which can also bypass the size check there. Do we have any fundamental risk of an OOM from that attack vector?

node/router/messages/src/helpers/codec.rs
```diff
@@ -75,6 +75,8 @@ impl<N: Network> Decoder for MessageCodec<N> {
             None => return Ok(None),
         };
 
+        UnconfirmedTransaction::<N>::check_size(&bytes)?;
```
howardwu: I'm confused. Why is the UnconfirmedTransaction logic being used by MessageCodec for all message types?

Cargo.toml
```diff
@@ -46,7 +46,7 @@ default-features = false
 
 [workspace.dependencies.snarkvm]
 git = "https://github.com/AleoHQ/snarkVM.git"
-rev = "aef07da"
+rev = "afacda3"
```
howardwu: Can you update the rev to the latest snarkVM `mainnet-staging`?

node/rest/src/routes.rs
```diff
@@ -16,13 +16,14 @@ use super::*;
...
     ledger::puzzle::Solution,
-    prelude::{block::Transaction, Identifier, Plaintext},
+    prelude::{block::Transaction, Identifier, Plaintext, ToBytes},
 };
 
 use indexmap::IndexMap;
 use rayon::prelude::*;
 use serde::{Deserialize, Serialize};
 use serde_json::json;
+use snarkvm::prelude::LimitedWriter;
```
howardwu: nit: ordering

node/router/messages/src/unconfirmed_transaction.rs
```diff
@@ -56,11 +56,38 @@ impl<N: Network> FromBytes for UnconfirmedTransaction<N> {
...
+        let id = u16::from_le_bytes(id_bytes);
+
+        // SPECIAL CASE: check the transaction message isn't too large.
+        if id == 12 && len > N::MAX_TRANSACTION_SIZE {
+            return Err(io::Error::new(io::ErrorKind::InvalidData, "transaction is too large"))?;
+        }
+
+        Ok(())
+    }
+}
```
howardwu: I don't think we need this one. There is a duplicate copy in the file above.

niklaslong: This wasn't deliberate, forgot to delete the dupe 🤦 sorry! I'll remove this one again (reverting the last commit and moving the check back up to `Message` as discussed in the previous comments).

node/router/messages/src/helpers/codec.rs
niklaslong: Oh, I thought this is what you meant with https://github.com/AleoHQ/snarkOS/pull/3229#discussion_r1583886044, my bad. 
This check needs to be done from the codec unless we choose to not use `FromBytes` and `ToBytes` in the codec. This is because we can't know the message type before decoding the message ID and currently that is coupled with the rest of message deserialisation.

node/rest/src/routes.rs
niklaslong: We limit request sizes here so we should be ok: https://github.com/AleoHQ/snarkOS/blob/mainnet-staging/node/rest/src/lib.rs#L204.

iamalwaysuncomfortable: @niklaslong should we consider lowering that limit? 10MB could add up quickly. 

Transactions were the largest thing we accepted and since the size is now capped, does it make sense to allow body sizes greater than 1MB?

node/rest/src/routes.rs
iamalwaysuncomfortable: Basically I think a benchmark should be run on `LimitedWriter::new`. That should quickly tell us if running too many of these in the tokio executor would block it.

node/rest/src/routes.rs
```diff
@@ -334,6 +334,13 @@ impl<N: Network, C: ConsensusStorage<N>, R: Routing<N>> Rest<N, C, R> {
             return Err(RestError(format!("Unable to broadcast transaction '{}' (node is syncing)", fmt_id(tx.id()))));
         }
 
+        // If the transaction exceeds the transaction size limit, return an error.
+        // TODO: Should this be a blocking task?
+        let buffer = Vec::with_capacity(N::MAX_TRANSACTION_SIZE);
```
iamalwaysuncomfortable: Normally `with_capacity` is a good idea. However, here the consequence of this is 128KB being allocated for each transaction call. At `30 TPS` this is roughly `~225 MB` worth of memory allocation per minute. 

Most transactions won't be that large so it's a lot of allocation.

Given credits.aleo functions are likely to be the largest we see. Maybe here's better to allocate the size of something like `transfer_public` to start?

node/rest/src/routes.rs
niklaslong: I don't see why not, I can file another PR for this.

howardwu: Is the 10MB on the request or response size limit?

Remember blocks do not have a maximum size, just like in Ethereum.

node/rest/src/routes.rs
howardwu: +1 to this. The max size should not be allocated each time

node/rest/src/routes.rs
niklaslong: It _should_ apply only to requests (as it's implemented for the request extractor and not the response).

node/rest/src/routes.rs
iamalwaysuncomfortable: Here's the benchmarks for the `LimitedWriter.write_le` and `check_transaction`.

This is for a program called [too_big.aleo](https://github.com/AleoHQ/snarkVM/blob/cc28c8da25a82924da2d267c6421b43030a6d5a8/ledger/benches/transaction.rs#L188) which results in a Transaction size of exactly `164613 bytes``

```
LimitedWriter::new - transfer_public
                        time:   [11.136 µs 12.007 µs 13.021 µs]

Transaction::Execute(transfer_public) - verify
                        time:   [15.048 ms 19.123 ms 22.293 ms]

LimitedWriter::new - transfer_private
                        time:   [11.076 µs 12.442 µs 13.927 µs]

Transaction::Execute(transfer_private) - verify
                        time:   [14.654 ms 19.755 ms 25.494 ms]

LimitedWriter::new - too_big.aleo (164613 bytes)
                        time:   [2.1626 ms 2.4037 ms 2.6260 ms]

Transaction::Execute(too_big.aleo) - verify (164613 bytes)
                        time:   [2.0601 ms 2.2421 ms 2.4375 ms]
```

It shows that for deployment and execution transactions of small functions the cost is negligible.

`spawn_blocking` is probably the safer route here, but we should be mindful of the effect of that on the BFT logic as those processes make extensive use of the blocking threadpool.

node/rest/src/routes.rs
iamalwaysuncomfortable: `transfer_public` is `2880 bytes` so we could bound this by `3000`.

node/rest/src/routes.rs
iamalwaysuncomfortable: @ljedrz @pauan 

Could you comment on what the better call would be here?

ljedrz: I'm not sure what the cost is here at first glance, but the rule of thumb would be that anything that can go into the millisecond scale should be done within a blocking task; several microseconds can be ignored.

`tokio-console` is also quite good at detecting which tasks spend too much time blocking, in case we ever miss/suspect something.

node/rest/src/routes.rs
niklaslong: Set to `3000` in [2b75824](https://github.com/AleoHQ/snarkOS/pull/3229/commits/2b75824d45a8d985507dfceba72077796366738a).

node/rest/src/routes.rs
niklaslong: I'd avoid using the blocking thread pool for anything that isn't truly expensive, a couple ms isn't super worrisome (if its >100ms then there's no question), especially as this is a larger tx than our current limit of 128kb?

niklaslong: In any case I think we should confirm by testing on devnet (without making this a blocker for merging). It'll go well with our current testing of the REST server.

howardwu (approved):

node/rest/src/routes.rs
Pauan: 15 ms is a *lot*... in general it's a bad idea for async code to block for that long, because it blocks the executor and can prevent *other* async tasks from running concurrently.

PR #3231, opened 2024-04-22
[Metrics] Track synced blocks.

d0cd:
This PR introduces a metric to track the number of blocks added through syncing.

### DIFF
--- a/node/sync/src/block_sync.rs
+++ b/node/sync/src/block_sync.rs
@@ -17,6 +17,7 @@ use crate::{
     locators::BlockLocators,
 };
 use snarkos_node_bft_ledger_service::LedgerService;
+use snarkos_node_metrics as metrics;
 use snarkos_node_router::messages::DataBlocks;
 use snarkos_node_sync_communication_service::CommunicationService;
 use snarkos_node_sync_locators::{CHECKPOINT_INTERVAL, NUM_RECENT_BLOCKS};
@@ -345,6 +346,8 @@ impl<N: Network> BlockSync<N> {
             }
             // Update the latest height.
             current_height = self.canon.latest_block_height();
+            // Update the `SYNCED` metric.
+            metrics::increment_gauge(metrics::bft::SYNCED, 1);
         }
     }
 }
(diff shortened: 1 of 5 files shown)


node/sync/src/block_sync.rs
```diff
@@ -461,6 +461,9 @@ impl<N: Network> BlockSync<N> {
         let is_synced = num_blocks_behind <= max_blocks_behind;
         // Update the sync status.
         self.is_block_synced.store(is_synced, Ordering::SeqCst);
+        // Update the `IS_SYNCED` metric.
+        #[cfg(feature = "metrics")]
+        metrics::gauge(metrics::bft::IS_SYNCED, is_synced);
```
miazn, line 466: This looks good! One thing that might make a difference in recording is that metrics are "registered" before they are seen, which basically initializes all metrics in the names folder at 0. Its probably a trivial difference, as for this metric we initialize it and then update sync status pretty quickly, however just wanted to point out that it will be initialized at 0. If we only want to start recording the metric the first time this line of code is hit, the metric name should be removed from the `names.rs` file

d0cd: I think it should be fine since `is_block_synced` is initialized with `Default::default() == false`. So we'll have the property that `is_block_synced` matches `IS_SYNCED`.

iamalwaysuncomfortable: That's true it'll match the value in the `BlockSync` struct but 

If a node starts off synced it's possible for the value stored in `is_block_synced` to be `false` while the node IS indeed synced. So it doesn't match the picture of what's currently synced. This might cause confusions on the Dashboards.

I'd advocate to initialize this guage as `true`/`1` because as soon as a node begins the syncing process it'll immediately set the metric to 0.

miazn: Doesnt it update when its initializing the sync module? I think it should reflect the current state 🤔

iamalwaysuncomfortable: @miazn - reading further, this does get triggered with the sync is initialized. You're right.

howardwu (approved):

Issue #3226, opened 2024-04-19
[Bug] Due to changes from PR #3217, BFT sync logic is not safe, 2 malicious validators can send invalid block_locators and blocks to other nodes and halt the network


feezybabee:
# https://hackerone.com/reports/2469178

## Summary:

BFT sync logic is not safe after PR #3217 ; 2 malicious validators can send invalid block_locators and blocks to other nodes, and the other nodes will accept them and get halted.

## Steps To Reproduce:

1. Use this [commit](https://github.com/ghostant-1017/snarkOS/commit/540fe2731b1f898e4824135f707e6379f837d2bc)
2. Run script `./devnet.sh` with 8 validators.
3. Check the state of all nodes

## Proof-of-Concept (PoC)

This attack only requires **two validator** nodes and **is independent of the amount staked** by the validators, so it doesn't violate the Byzantine assumption.

1. Validator0 and Validator1 are two malicious nodes. Validator0 is the main attacker, while for Validator1, we only need to use its private key, referred to as `partner_pk`. The purpose of using this private key is to generate a `BatchCertificate` that can pass validation checks.
2. Assuming Validator0's current height is `current_height` and `current_round`, Validator0 can create a Block for height `current_height + 1` using its own `private_key` and the `partner_key`. This Block can pass validation checks performed by other nodes' `check_next_block` function and be successfully advanced by other nodes using `advance_next_block`. Refer to the code [here](https://github.com/ghostant-1017/snarkOS/commit/540fe2731b1f898e4824135f707e6379f837d2bc#diff-3012fb9a78ce90d5a3f4f79a35f31074e1596fecf075e8cd781fc60b63807868R62).
3. Validator0 will increase the frequency of `PrimaryPing` (interval 5s => 1s), ensuring that other nodes detect the `Block` which is  `current_height + 1`.
4. When Validator0 receives an `Event::BlockRequest` from other nodes, it can send the incorrect `Block` to them.
5. When other nodes receive the `BlockResponse`, the malicious Block will pass all checks and eventually be advanced using [advance_to_next_block](https://github.com/AleoHQ/snarkOS/pull/3217/files#diff-f82e0c9ecc0d4ff1e73050db6da1dc5909f4f8c5afc98d3effd254a440f079e3R368), causing the ledger to be in an incorrect state.

It's worth noting that the malicious Block **may not** pass `sync_certificate_with_blocks`, but since the attack is sustainable, other nodes will **eventually end up in an incorrect state**.

[...]

PR #3232, opened 2024-04-22
[Fix] Perform block advancement after quorum check

raychu86:
## Motivation

This PR updates the final stage of validator syncing by adding an additional quorum check. Instead of blindly adding the block to ledger, we now check that each block's leader certificate reached quorum based on the certificates in the subsequent block. 

This solves 2 problems:

### Naive validator sync
1. A validator synced up to tip and added the last block (`X`) without performing any quorum checks. 
2. The validator would start processing consensus messages and process incoming certificates.
3. It would reach a commit state and try to construct `X` through the standard non-sync procedure. 
4. However the ledger already stored `X`.
5. An error would repeat, halting block advancement on this node until the node was rebooted.

```
`Unable to advance to the next block - Failed to speculate on transactions - Failed to post-ratify - Next round {} must be greater than current round {}`
```

The change now skips adding the latest block because we can't guarantee the block meets the quorum. So we will process the block through consensus instead of the sync process.

### Malicious validator sync

The malicious sync case is defined in a hackerOne finding: https://github.com/AleoHQ/snarkOS/issues/3226

The new quorum checks should ensure that we are not processing blocks blindly.

### DIFF
--- a/.devnet/start_sync_test.sh
+++ b/.devnet/start_sync_test.sh
@@ -1,0 +1,59 @@
+#!/bin/bash
+
+# Determine the number of AWS EC2 instances by checking ~/.ssh/config
+NODE_ID=0
+while [ -n "$(grep "aws-n${NODE_ID}" ~/.ssh/config)" ]; do
+    NODE_ID=$((NODE_ID + 1))
+done
+
+# Read the number of AWS EC2 instances to query from the user
+read -p "Enter the number of AWS EC2 instances to query (default: $NODE_ID): " NUM_INSTANCES
+NUM_INSTANCES="${NUM_INSTANCES:-$NODE_ID}"
+
+echo "Using $NUM_INSTANCES AWS EC2 instances for querying."
+
+# Read the verbosity level from the user (default: 1)
+read -p "Enter the verbosity level (default: 1): " VERBOSITY
+VERBOSITY="${VERBOSITY:-1}"
+
+echo "Using verbosity level $VERBOSITY."
+
+# Get the IP address of NODE 0 from the SSH config for aws-n0
+NODE_0_IP=$(awk '/Host aws-n0/{f=1} f&&/HostName/{print $2; exit}' ~/.ssh/config)
+
+# Define a function to start snarkOS in a tmux session on a node
+start_snarkos_in_tmux() {
+  local NODE_ID=$1
+  local NODE_IP=$2
+
+  # SSH into the node and start snarkOS in a new tmux session
+  ssh -o StrictHostKeyChecking=no aws-n$NODE_ID << EOF
+    # Commands to run on the remote instance
+    sudo -i  # Switch to root user
+    WORKSPACE=~/snarkOS
+    cd \$WORKSPACE
+
+    # Start snarkOS within a new tmux session named "snarkos-session"
+    tmux new-session -d -s snarkos-session
+
+    # Send the snarkOS start command to the tmux session with the NODE_ID
+    tmux send-keys -t "snarkos-session" "snarkos start --client --nocdn --nodisplay --rest 0.0.0.0:3030 --node 0.0.0.0:4130 --verbosity 4 --metrics --logfile "/tmp/snarkos-syncing-range-3.log" --peers 167.71.249.65:4130,157.245.218.195:4130,167.71.249.55:4130" C-m
+
+    exit  # Exit root user
+EOF
+
+  # Check the exit status of the SSH command
+  if [ $? -eq 0 ]; then
+    echo "snarkOS started successfully in a tmux session on aws-n$NODE_ID."
+  else
+    echo "Failed to start snarkOS in a tmux session on aws-n$NODE_ID."
+  fi
+}
+
+# Loop through aws-n nodes and start snarkOS in tmux sessions in parallel
+for NODE_ID in $(seq 0 $(($NUM_INSTANCES - 1))); do
+  start_snarkos_in_tmux $NODE_ID "$NODE_0_IP" &
+done
+
+# Wait for all background jobs to finish
+wait
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3292,7 +3292,7 @@ dependencies = [
 [[package]]
 name = "snarkvm"
 version = "0.16.19"
-source = "git+https://github.com/AleoHQ/snarkVM.git?rev=43abe1b#43abe1ba7cdb4de6a435905a1688cca4caf0c7e3"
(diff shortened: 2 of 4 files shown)


raychu86:
@ghostant-1017 Could you confirm that this addresses your finding - https://github.com/AleoHQ/snarkOS/issues/3226?

ghostant-1017:
> @ghostant-1017 Could you confirm that this addresses your finding - #3226?

Of course! I will test it out.

ghostant-1017:
Hey @raychu86 ,
This PR indeed addresses the issue [#3226](https://github.com/AleoHQ/snarkOS/issues/3226).

However, during my review process, I identified another potential DOS attack vector. It only requires a [minor modification](https://github.com/ghostant-1017/snarkOS/commit/912f81ea2f9fa15a778a5b703deea095d5687426) on top of the existing changes, I haven't conducted practical tests yet.

The main logic is as follows:

1. Upon receiving an invalid Block, we invoke [self.storage.sync_certificate_with_block](https://github.com/AleoHQ/snarkOS/blob/a95e6c7dce7e75f1bc5b9f46c9d76375c5471687/node/bft/src/sync/mod.rs#L354). Regardless of its success, we always call [bft_sender.send_sync_bft(certificate)](https://github.com/AleoHQ/snarkOS/blob/a95e6c7dce7e75f1bc5b9f46c9d76375c5471687/node/bft/src/sync/mod.rs#L365).
2. Upon receiving rx_sync_bft.recv(), [self.update_dag](https://github.com/AleoHQ/snarkOS/blob/043b97ec20232391d6c320fc0929bf97c1fc5e63/node/bft/src/bft.rs#L814) is called.
3. In the `update_dag` function, we [insert the certificate](https://github.com/AleoHQ/snarkOS/blob/043b97ec20232391d6c320fc0929bf97c1fc5e63/node/bft/src/bft.rs#L458) into the subdag.
4. The `BatchCertificate` can contain a maximum of 200 signatures. This means that attackers can continuously send invalid Blocks(different round far in the future), causing honest validator nodes to insert meaningless `BatchCertificates` into memory. Eventually, this could lead to a memory overflow.

I can probably confirm the effectiveness of this attack. Please recheck whether this attack is valid.I won't submit another report for this, I hope you consider this potential DOS attack when assessing the severity of this report.

howardwu:
Thanks for flagging this @ghostant-1017!

I think the attack described may be possible, there's one thing to consider: the sync logic only takes blocks from `BlockResponse` messages/events. I believe we check (when receiving a `BlockResponse`) that we requested the specific block, and then clear this request from the cache (meaning a future `BlockResponse` for the same block will not be allowed in).

If my understanding is correct, this suggests we could have an attacker send this block once, but then it no longer works.

Perhaps one thing we could do is have `sync_storage_with_block` check at the beginning that:
```rs
// Return early if this block has already been processed.
if self.ledger.contains_block_height(block.height()) {
    return Ok(());
}
```

ghostant-1017:
Hey, @howardwu  
Maybe the protection is not enough.
Assume the `latest_height = 10000`, in this situation, the attacker only generate block at `height=10001` and broadcast.
I think it's possible to generate blocks at `height=10002`, `height=10003` and even higher blocks. And the invalid blocks will always pass the height-contain check.

howardwu:
Thanks for the feedback. Wouldn't it require at least `2f+1` private keys to be compromised in order to create a consecutive exploit? We reference the `committee_lookback` to ensure the next subdag is well-formed. I suppose the attacker would need to have compromised a substantial set of the existing validators in order to pull this off. Let me know if you see another angle.

ghostant-1017:
It seems that the `self.update_dag` will be called before quorum check.
`self.update_dag` will be called at: https://github.com/AleoHQ/snarkOS/pull/3232/files#diff-f82e0c9ecc0d4ff1e73050db6da1dc5909f4f8c5afc98d3effd254a440f079e3L355
quorum_check will be called at: https://github.com/AleoHQ/snarkOS/pull/3232/files#diff-f82e0c9ecc0d4ff1e73050db6da1dc5909f4f8c5afc98d3effd254a440f079e3R425

@howardwu Correct me if I'm wrong.

howardwu:
Yes, the idea is that the check in update_dag and subsequently here both enforce availability or quorum threshold. So wouldn't the attacker need to compromise at least f private keys from the committee to allow for the node to sync the block?

(Sorry if I missed the point)

raychu86:
I think Ghostant is saying we are calling `update_dag` before performing any quorum/availability checks. So if malicious nodes send a faulty block with garbage certificates, we add it to storage before performing any quorum checks.

I believe our mitigation for this is in the `BlockSync::construct_request` where we handle the dishonest case.

@ghostant-1017 let us know if our understanding of your concern is correct.

ghostant-1017:
@raychu86 Yes, you get me, that's what I mean.
So you mean the `BlockSync::construct_request` will handle this case? I will take a look into it.

ghostant-1017:
@raychu86 @howardwu 
I don't think BlockRequest::construct_request is secure in this scenario. In the [honest check](https://github.com/AleoHQ/snarkOS/blob/d9bd0db4a8b4ced5da5d9d7ef928fa38f6a85b31/node/sync/src/block_sync.rs#L836) within construct_request, the attacker is the only peer with a height greater than all other nodes, it will pass all checks.

Do you agree? 
I'm not entirely certain about my idea without a PoC. I'd be willing to attempt a PoC to validate it if necessary.

raychu86:
@ghostant-1017 I agree that this could be an issue, and we should have some more mitigating checks to prevent this. However this attack existed prior to this PR, so I'm of the opinion we try to mitigate it in a separate PR after this is merged. 

I don't believe it is quite as straightforward to address in this PR, because we need to perform `update_dag` and `sync_certificate_with_block` in order to even perform the availability threshold checks.

In fact, I believe that this is a more fundamental issue with receiving blocks, as this attack using malicious blocks could be done to clients and validators performing the non-BFT sync. Bitcoin has this issue as well, but i believe they get around this by doing chain-reorgs and rollbacks.

Couple ideas for mitigation:
1. Have nodes declare their trusted sync nodes.
2. Enforce at least `X` other sync nodes has the block before requesting.
3. Initialize community trusted sync nodes.


Let us know if you agree or have other insights.

ghostant-1017:
@raychu86 
Yes! Our views have aligned now.
One more suggestion:  Maybe we can check the certificate's round is not too far when `update_dag`?

And I'm trying finding a way to simplify the syncing process. I'm not quite sure, but I will share with you if I'm ready.

mdelle1 (approved):

PR #3233, opened 2024-04-24
[Fix] Add `is_linked` checks in validator syncing

raychu86:
## Motivation

This PR builds on https://github.com/AleoHQ/snarkOS/pull/3232 and adds an additional `is_linked` check.

The solution in the original PR checks that every block meets availability threshold, however this is not correct. In the case of committing multiple blocks from a single subdag, the sub-subdags do not need to undergo these availability threshold checks. This means the original block sync would come to a halt in this edge case.

The fix in this PR is to add on an additional `is_linked` lookback. 

1. Process individual block responses iteratively
2. If the block does not meet the availability threshold then move on to the next block.
3. The first block we find that meets the availability threshold, we perform `is_linked` checks on the blocks we skipped in step 2. 
4. If the `is_linked` checks pass, then we attempt to add the entire list of contiguous blocks to the ledger.

### DIFF
--- a/node/bft/src/sync/mod.rs
+++ b/node/bft/src/sync/mod.rs
@@ -377,7 +377,6 @@ impl<N: Network> Sync<N> {
 
         // Fetch the latest block height.
         let latest_block_height = self.ledger.latest_block_height();
-        let next_block_height = latest_block_height.saturating_add(1);
 
         // Insert the latest block response.
         latest_block_responses.insert(block.height(), block);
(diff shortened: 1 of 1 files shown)


raychu86:
@ghostant-1017 Could you take a look at this PR as well regarding the fix for https://github.com/AleoHQ/snarkOS/issues/3226

howardwu (approved):

howardwu:
Merging this PR into PR #3232

ghostant-1017:
> @ghostant-1017 Could you take a look at this PR as well regarding the fix for #3226

Yes, of course! It's my pleasure.

PR #3235, opened 2024-04-25
feat: dim logs on shutdown

joske:
This PR implements log dimming (if running in TTY mode) on shutdown (CTRL-C handler).

I've moved the shutdown flag earlier in the startup so it can be reused for the log dimming too.

There is some performance impact as it needs to check the boolean for every log event (and an extra method call).


iamalwaysuncomfortable (commented):
Thanks for this @joske, I'll try to run this later today. Is it possible to add an explicit message in the terminal to wait for shutdown to complete?

cli/src/commands/start.rs
```diff
@@ -569,9 +577,9 @@ impl Start {
 
         // Initialize the node.
         match node_type {
-            NodeType::Validator => Node::new_validator(node_ip, bft_ip, rest_ip, self.rest_rps, account, &trusted_peers, &trusted_validators, genesis, cdn, storage_mode, self.allow_external_peers, dev_txs).await,
-            NodeType::Prover => Node::new_prover(node_ip, account, &trusted_peers, genesis, storage_mode).await,
-            NodeType::Client => Node::new_client(node_ip, rest_ip, self.rest_rps, account, &trusted_peers, genesis, cdn, storage_mode).await,
+            NodeType::Validator => Node::new_validator(node_ip, bft_ip, rest_ip, self.rest_rps, account, &trusted_peers, &trusted_validators, genesis, cdn, storage_mode, self.allow_external_peers, dev_txs, shutdown.clone()).await,
+            NodeType::Prover => Node::new_prover(node_ip, account, &trusted_peers, genesis, storage_mode, shutdown.clone()).await,
+            NodeType::Client => Node::new_client(node_ip, rest_ip, self.rest_rps, account, &trusted_peers, genesis, cdn, storage_mode, shutdown.clone()).await,
```
```suggestion
            NodeType::Client => Node::new_client(node_ip, rest_ip, self.rest_rps, account, &trusted_peers, genesis, cdn, storage_mode, shutdown).await,
```
nit, before `clippy` notices

cli/src/helpers/dynamic_format.rs
```diff
@@ -0,0 +1,101 @@
...
+use tracing::{Event, Subscriber};
+use tracing_subscriber::{
+    fmt::{format::Writer, FmtContext, FormatEvent, FormatFields},
+    registry::LookupSpan,
+};
+
+pub struct DynamicFormatter<'b> {
+    dim_format: DimFormat<'b>,
+    default_format: tracing_subscriber::fmt::format::Format,
+    dim: Arc<AtomicBool>,
```
ljedrz, line 35: I presume that this is the same atomic as the shutdown flag? It would be a good idea to comment on that so that it's clear.

cli/src/helpers/dynamic_format.rs
```diff
@@ -0,0 +1,101 @@
...
+        Self { fmt: format }
+    }
+}
+
+impl<'b, S, N> FormatEvent<S, N> for DimFormat<'b>
+where
+    S: Subscriber + for<'a> LookupSpan<'a>,
+    N: for<'a> FormatFields<'a> + 'static,
+{
+    fn format_event(&self, ctx: &FmtContext<'_, S, N>, mut writer: Writer<'_>, event: &Event<'_>) -> std::fmt::Result {
```
ljedrz, line 82: It might be a good idea to explain what this function does (seems to me that this just dims the regular event log display), in case there is some simpler derive macro (or something else) that can do this in the future.

cli/src/helpers/dynamic_format.rs
```diff
@@ -0,0 +1,101 @@
...
+    fn format_event(&self, ctx: &FmtContext<'_, S, N>, writer: Writer<'_>, event: &Event<'_>) -> std::fmt::Result {
+        if self.dim.load(Ordering::Relaxed) {
+            self.dim_format.format_event(ctx, writer, event)
+        } else {
+            self.default_format.format_event(ctx, writer, event)
+        }
+    }
+}
+
+impl<'b> DynamicFormatter<'b> {
```
ljedrz: This object could use a comment too.

node/src/traits.rs
```diff
@@ -90,6 +90,7 @@ pub trait NodeInterface<N: Network>: Routing<N> {
         tokio::task::spawn(async move {
             match signal_listener().await {
                 Ok(()) => {
+                    warn!("Orderly shutdown started, please wait until shutdown is complete.");
```
ljedrz:
```suggestion
                    warn!("Orderly shutdown started, please wait until it is complete.");
```

node/src/validator/mod.rs
```diff
@@ -502,6 +500,7 @@ mod tests {
             storage_mode,
             false,
             dev_txs,
+            Arc::new(AtomicBool::new(false)),
```
ljedrz:
```suggestion
            Default::default(),
```

node/tests/common/node.rs
```diff
@@ -30,6 +33,7 @@ pub async fn client() -> Client<CurrentNetwork, ConsensusMemory<CurrentNetwork>>
         sample_genesis_block(),
         None, // No CDN.
         StorageMode::Production,
+        Arc::new(AtomicBool::new(false)),
```
ljedrz:
```suggestion
        Default::default(),
```

node/tests/common/node.rs
```diff
@@ -42,6 +46,7 @@ pub async fn prover() -> Prover<CurrentNetwork, ConsensusMemory<CurrentNetwork>>
         &[],
         sample_genesis_block(),
         StorageMode::Production,
+        Arc::new(AtomicBool::new(false)),
```
ljedrz:
```suggestion
        Default::default(),
```

node/tests/common/node.rs
```diff
@@ -61,6 +66,7 @@ pub async fn validator() -> Validator<CurrentNetwork, ConsensusMemory<CurrentNet
         StorageMode::Production,
         true,  // This test requires validators to connect to peers.
         false, // No dev traffic in production mode.
+        Arc::new(AtomicBool::new(false)),
```
ljedrz:
```suggestion
        Default::default(),
```

ljedrz (approved):
LGTM with a few nits.

niklaslong:
CircleCi is having a moment, that failure is unrelated to the changes. Let's try rerunning a little later.

niklaslong (approved):
LGTM (also tested locally) 👍

node/src/traits.rs
```diff
@@ -90,6 +90,7 @@ pub trait NodeInterface<N: Network>: Routing<N> {
         tokio::task::spawn(async move {
             match signal_listener().await {
                 Ok(()) => {
+                    warn!("Orderly shutdown started, please wait until it is complete.");
```
iamalwaysuncomfortable: This message needs to made more urgent and clear about the potential consequences of early termination.

Also these two messages should be made into an alert color like yellow or orange and bolded if possible.

```suggestion
                    warn!("Graceful node shutdown started!");
                    warn!("NOTE: PLEASE WAIT UNTIL THE SHUTDOWN PROCESS COMPLETES TO AVOID DATA CORRUPTION");
```

iamalwaysuncomfortable (changes requested):
The initial message needs to be a bit more clear and should probably be highlighted or bolded. I've included a suggestion for wording.

Also:
- During shutdown, are any further SIGINT/SIGHUP/SIGTERM signals caught & ignored so that graceful shutdown can continue?
- I think this work is a good start, but in a follow-on PR we should maybe consider launching a display which displays graceful shutdown warnings more prominently.

node/src/traits.rs
ljedrz: Note: an unclean shutdown shouldn't cause any db corruption (since we write atomically), and if it does, it implies a bug. I haven't experienced it in a long time regardless of how the shutdown looked like.

iamalwaysuncomfortable: The point of this PR to ensure people avoid sending `SIGKILL` or initiating any other kind of non-graceful shutdown  and leaving the node in an inconsistent state. Thus it should be emphasized very clearly what the potential consequences are.

In theory we write atomically but until there's recorded evidence from tests that repeated non-graceful shutdowns do not corrupt data, we should not be making the assumption that it will never do so.

joske: Subsequent signals are ignored as the signal handler is just sleeping and then exiting.

joske: We can not bold the text of the initial message as we don't know at that point if we are running in TTY mode.

PR #3236, opened 2024-04-26
Reduce the number of lookups in update_peer_locators

ljedrz:
The `update_peer_locators` method is quite prominent in heap profiles, and tweaking it could yield considerable improvements.

This PR proposes a simple change: instead of starting from height 0 and "going up" the received list of locators (which would be most optimal in case of a deep fork) it starts with the last locator and, if it's not found in the local storage, goes over the remaining locators in reverse order, stopping once the most recent ancestor is found (which is much faster in the "happy path" scenario).

In a local `--dev` run with 4 validator nodes, this reduces the number of allocations in a non-tx-producing node by ~1/3 after 1h. Also, while this is presented as a memory-related improvement, this change should result in reduced (peer) block sync time as well.

As for whether this is as safe as the current setup; a malicious node that knows the most recent valid locator would also be able to provide valid past locators (and it would take longer to process those as well). As mentioned before, the one downside that I see is that this change would make this operation slower for very deep forks, where the number of diverging blocks is larger than the number of canonical ones.

### DIFF
--- a/node/sync/src/block_sync.rs
+++ b/node/sync/src/block_sync.rs
@@ -381,11 +381,11 @@ impl<N: Network> BlockSync<N> {
 
         // Compute the common ancestor with this node.
         let mut ancestor = 0;
-        for (height, hash) in locators.clone().into_iter() {
+        for (height, hash) in locators.clone().into_iter().rev() {
             if let Ok(canon_hash) = self.canon.get_block_hash(height) {
-                match canon_hash == hash {
-                    true => ancestor = height,
-                    false => break, // fork
+                if canon_hash == hash {
+                    ancestor = height;
+                    break;
                 }
             }
         }
--- a/node/sync/src/block_sync.rs
+++ b/node/sync/src/block_sync.rs
@@ -381,14 +381,32 @@ impl<N: Network> BlockSync<N> {
 
         // Compute the common ancestor with this node.
         let mut ancestor = 0;
-        for (height, hash) in locators.clone().into_iter().rev() {
-            if let Ok(canon_hash) = self.canon.get_block_hash(height) {
-                if canon_hash == hash {
-                    ancestor = height;
-                    break;
+
+        // The unwrap is safe, as `ensure_is_valid` guarantees that `recents` is not empty.
+        let (latest_locator_height, latest_locator_hash) = locators.recents.last().unwrap();
+
+        // Check if the latest locator matches the canon to quickly determine if it's a fork.
+        // This is an optimization that allows us to skip the traversal of all the locators in
+        // case it's a match.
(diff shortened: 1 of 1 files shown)


node/sync/src/block_sync.rs
```diff
@@ -381,11 +381,11 @@ impl<N: Network> BlockSync<N> {
...
         let mut ancestor = 0;
-        for (height, hash) in locators.clone().into_iter() {
+        for (height, hash) in locators.clone().into_iter().rev() {
             if let Ok(canon_hash) = self.canon.get_block_hash(height) {
-                match canon_hash == hash {
-                    true => ancestor = height,
-                    false => break, // fork
+                if canon_hash == hash {
+                    ancestor = height;
+                    break;
```
howardwu: @ljedrz can you keep the original logic but encapsulate it with rayon (to check each iteration of the look in parallel)? We want to do fork detection, which is what the original for loop does. Please add a comment to denote this for future readers not to remove.

Then, in addition, you can add the new early-return for loop as a way to quickly detect the latest common ancestor.

ljedrz: > can you keep the original logic but encapsulate it with rayon

Unfortunately, this would be tricky due to the reassignment of the `ancestor` and the breaking logic. However, the addition of the loop skip means it's no longer a performance concern, as forks are unusual.

> add a comment to denote this for future readers not to remove.

Done.

> add the new early-return for loop as a way to quickly detect the latest common ancestor.

Done.

node/sync/src/block_sync.rs
```diff
@@ -381,14 +381,32 @@ impl<N: Network> BlockSync<N> {
...
+        let (latest_locator_height, latest_locator_hash) = locators.recents.last().unwrap();
+
+        // Check if the latest locator matches the canon to quickly determine if it's a fork.
+        // This is an optimization that allows us to skip the traversal of all the locators in
+        // case it's a match.
+        if let Ok(canon_hash) = self.canon.get_block_hash(*latest_locator_height) {
+            if *latest_locator_hash == canon_hash {
+                ancestor = *latest_locator_height;
+            }
+        }
```
howardwu, line 395: What happens if a malicious node sends garbage locators except for the most recent height? Then we will skip fork detection and won't discover this case.

ljedrz: I don't know what the goal of such a node would be, but I may be missing the bigger picture. Please feel free to close this PR if this is a viable attack vector or if it invalidates some of the assumptions about our peers; we can always revisit this if block locator spam/DoS proves to be a greater concern.

howardwu:
Closing this PR, and adding an important comment in commit https://github.com/AleoHQ/snarkOS/commit/913db14f08e34d7a27814194a22c9400fe599691 to the file.

PR #3238, opened 2024-04-28
Decrement counter

ghostant-1017:
I remember there was an attack which would send many peer responses to other peers, if without the this line of code, the attack seems will work again.


vicsn (approved):
Indeed great catch, compared to the other two `decrement_outbound_*` functions, this wasn't used yet, tagging @niklaslong just as a double check.

niklaslong (approved):
Good catch, LGTM 👍

PR #3239, opened 2024-04-30
[HackerOne-2452182] Store `Proposal` and `SignedProposals` to disk and remove proposal expiration

raychu86:
## Motivation

This PR has three changes to address https://github.com/AleoHQ/snarkOS/issues/3171: 

1. Reverted _proposal expiration_ introduced in https://github.com/AleoHQ/snarkOS/pull/3202.
    - This approach was unsafe because it violated the non-equivocation properties of our BFT protocol.  
    - A community member @ghostant-1017 also highlighted this issue in a bug bounty finding [HackerOne-2452182].

2. The `ProposalCache` (`latest_round`, pending `Proposal`, and `SignedProposals`) will be stored to a file on shutdown, loaded on bootup, and cleared with `snarkos clean`. 
    - Ensures that honest validators do not create multiple proposals on the same round
    - Ensures that we do not sign additional validator proposals on the same round after a reboot.    
    - Ensures that the node does not create any proposals prior to it's storage round prior to the reboot.
    
3. Periodic attempts to increment the round if we met the quorum requirements in storage.
    - In our current processing of `PrimaryPing`s, we try to advance rounds in `sync_with_batch_header_from_peer`, however we perform this check PRIOR to processing the batch header, meaning it will not be included in the quorum calculation. This prevents us from advancing rounds when there are just enough certificates to meet quorum for that round (since we omit the last one we receive in the quorum calculation).
    
**Note**: An alternative to this `latest_round` approach is to store all of the certificates in `Storage` that have not made it into blocks to the `ProposalCache` file. The `latest_round` approach relies on `PrimaryPing`s to advance the storage state, but does allow the `ProposalCache` file to be relatively small.

### Implications

- If enough (a majority) of validators reboot their nodes at the same time, there could be a halting case.
    - If we signed proposals past our ledger round, but all nodes have thrown away their `Storage`, then we won't have the signatures required to reconstruct the original proposal state. 
    - This will need to be remedied by having all validators delete their `proposal_cache` file and reboot.

- Validators looking to swap machines will need to migrate their `proposal_cache` file between machines to ensure honest behavior.

## Test Plan

Unit tests have been added to for the new types and to ensure that the new checks hold.

Extensive local and burn-in testing will need to be performed.

## Related PRs

Reverts: https://github.com/AleoHQ/snarkOS/pull/3202.
An extension of: https://github.com/AleoHQ/snarkOS/pull/3200

## TODO

- [ ] Consider adding new log messages to inform validators of the importance of the storage.

### DIFF
--- a/node/bft/src/helpers/proposal.rs
+++ b/node/bft/src/helpers/proposal.rs
@@ -28,6 +28,7 @@ use snarkvm::{
 use indexmap::{IndexMap, IndexSet};
 use std::collections::HashSet;
 
+#[derive(Debug, PartialEq, Eq)]
 pub struct Proposal<N: Network> {
     /// The proposed batch header.
     batch_header: BatchHeader<N>,
@@ -200,6 +201,40 @@ impl<N: Network> FromBytes for Proposal<N> {
         }
 
         Ok(Self { batch_header, transmissions, signatures })
+    }
+}
+
+#[cfg(test)]
+pub(crate) mod tests {
+    use super::*;
+    use crate::helpers::storage::tests::sample_transmissions;
+    use snarkvm::{console::network::MainnetV0, utilities::TestRng};
+
+    type CurrentNetwork = MainnetV0;
+
+    const ITERATIONS: usize = 100;
+
+    pub(crate) fn sample_proposal(rng: &mut TestRng) -> Proposal<CurrentNetwork> {
+        let certificate = snarkvm::ledger::narwhal::batch_certificate::test_helpers::sample_batch_certificate(rng);
+        let (_, transmissions) = sample_transmissions(&certificate, rng);
+
+        let transmissions = transmissions.into_iter().map(|(id, (t, _))| (id, t)).collect::<IndexMap<_, _>>();
+        let batch_header = certificate.batch_header().clone();
+        let signatures = certificate.signatures().copied().collect();
+
+        Proposal { batch_header, transmissions, signatures }
+    }
+
+    #[test]
+    fn test_bytes() {
+        let rng = &mut TestRng::default();
+
+        for _ in 0..ITERATIONS {
+            let expected = sample_proposal(rng);
+            // Check the byte representation.
+            let expected_bytes = expected.to_bytes_le().unwrap();
+            assert_eq!(expected, Proposal::read_le(&expected_bytes[..]).unwrap());
+        }
     }
 }
 
--- a/node/bft/src/helpers/proposal.rs
+++ b/node/bft/src/helpers/proposal.rs
@@ -22,11 +22,11 @@ use snarkvm::{
         committee::Committee,
         narwhal::{BatchCertificate, BatchHeader, Transmission, TransmissionID},
     },
-    prelude::{bail, ensure, Itertools, Result},
+    prelude::{bail, ensure, FromBytes, Itertools, Result, ToBytes},
 };
 
 use indexmap::{IndexMap, IndexSet};
-use std::collections::HashSet;
+use std::{collections::HashSet, io};
 
 pub struct Proposal<N: Network> {
     /// The proposed batch header.
@@ -164,6 +164,42 @@ impl<N: Network> Proposal<N> {
         let certificate = BatchCertificate::from(self.batch_header.clone(), self.signatures.clone())?;
         // Return the certificate and transmissions.
         Ok((certificate, self.transmissions.clone()))
+    }
+}
+
+impl<N: Network> ToBytes for Proposal<N> {
+    fn write_le<W: io::Write>(&self, mut writer: W) -> io::Result<()> {
+        self.batch_header.write_le(&mut writer)?;
+        u32::try_from(self.transmissions.len()).map_err(|_| io::ErrorKind::Other)?.write_le(&mut writer)?;
+        for (transmission_id, transmission) in &self.transmissions {
+            transmission_id.write_le(&mut writer)?;
+            transmission.write_le(&mut writer)?;
+        }
+        u32::try_from(self.signatures.len()).map_err(|_| io::ErrorKind::Other)?.write_le(&mut writer)?;
+        for signature in &self.signatures {
+            signature.write_le(&mut writer)?;
+        }
+        Ok(())
+    }
+}
+
+impl<N: Network> FromBytes for Proposal<N> {
+    fn read_le<R: io::Read>(mut reader: R) -> io::Result<Self> {
+        let batch_header = FromBytes::read_le(&mut reader)?;
+        let num_transmissions = u32::read_le(&mut reader)?;
+        let mut transmissions = IndexMap::default();
+        for _ in 0..num_transmissions {
+            let transmission_id = FromBytes::read_le(&mut reader)?;
+            let transmission = FromBytes::read_le(&mut reader)?;
+            transmissions.insert(transmission_id, transmission);
+        }
+        let num_signatures = u32::read_le(&mut reader)?;
+        let mut signatures = IndexSet::default();
+        for _ in 0..num_signatures {
+            signatures.insert(FromBytes::read_le(&mut reader)?);
+        }
+
+        Ok(Self { batch_header, transmissions, signatures })
     }
 }
 
--- a/node/bft/src/helpers/proposal.rs
+++ b/node/bft/src/helpers/proposal.rs
@@ -170,13 +170,18 @@ impl<N: Network> Proposal<N> {
 
 impl<N: Network> ToBytes for Proposal<N> {
     fn write_le<W: Write>(&self, mut writer: W) -> IoResult<()> {
+        // Write the batch header.
         self.batch_header.write_le(&mut writer)?;
+        // Write the number of transmissions.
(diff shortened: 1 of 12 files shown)


node/bft/src/primary.rs
```diff
@@ -576,24 +606,21 @@ impl<N: Network> Primary<N> {
...
-                    true => self.signed_proposals.write().remove(&batch_author),
-                    // If the proposal has not expired, then disconnect the validator.
-                    false => {
-                        self.gateway.disconnect(peer_ip);
-                        bail!("Proposed another batch for the same round ({signed_round}) prior to expiration");
-                    }
-                };
+                // Proceed to disconnect the validator.
+                self.gateway.disconnect(peer_ip);
+                bail!("Malicious peer - proposed another batch for the same round ({signed_round})");
```
howardwu: I wouldn't call these 2 malicious peers. We've seen it happen on honest peers.

node/bft/src/primary.rs
```diff
@@ -295,6 +311,21 @@ impl<N: Network> Primary<N> {
...
+            if self.current_round() < proposal.round()
+                || proposal
+                    .batch_header()
+                    .previous_certificate_ids()
+                    .iter()
+                    .any(|id| !self.storage.contains_certificate(*id))
+            {
+                // TODO (raychu86): Explicitly request the missing certificates from peers.
+                debug!(
+                    "Cannot propose a batch for round {} - the current storage is not caught up to the proposed batch.",
```
howardwu: You should print `the current storage ({current_storage})`

node/bft/src/primary.rs
raychu86: The intention for this PR is to remove that behavior, but we can still be more lenient with the disconnects.

node/bft/src/primary.rs
raychu86: Resolved with [7aaf276](https://github.com/AleoHQ/snarkOS/pull/3239/commits/7aaf2765bc7c53152f87b913c78b62892db54792) and [53eca2f](https://github.com/AleoHQ/snarkOS/pull/3239/commits/53eca2ffd697a50145a8f341bf60d6573872ac03)

node/bft/src/primary.rs
raychu86: Resolved with [7aaf276](https://github.com/AleoHQ/snarkOS/pull/3239/commits/7aaf2765bc7c53152f87b913c78b62892db54792)

node/bft/src/helpers/proposal.rs
```diff
@@ -167,6 +168,86 @@ impl<N: Network> Proposal<N> {
...
+        Ok(())
+    }
+}
+
+impl<N: Network> FromBytes for Proposal<N> {
+    fn read_le<R: Read>(mut reader: R) -> IoResult<Self> {
+        // Read the batch header.
+        let batch_header = FromBytes::read_le(&mut reader)?;
+        // Read the number of transmissions.
+        let num_transmissions = u32::read_le(&mut reader)?;
```
ljedrz, line 197: should we impose a limit on the number of transmissions (same goes for signatures, and the other new objects in this PR), or do we only ever expect to create and load them locally?

node/bft/src/helpers/proposal_cache.rs
```diff
@@ -0,0 +1,195 @@
...
+    }
+
+    /// Returns `true` if a proposal cache exists for the given network and `dev`.
+    pub fn exists(dev: Option<u16>) -> bool {
+        proposal_cache_path(N::ID, dev).exists()
+    }
+
+    /// Load the proposal cache from the file system and ensure that the proposal cache is valid.
+    pub fn load(expected_signer: Address<N>, dev: Option<u16>) -> Result<Self> {
+        // Load the proposal cache from the file system.
```
ljedrz:
```suggestion
        // Construct the proposal cache file system path.
```
nit

node/bft/src/helpers/proposal_cache.rs
```diff
@@ -0,0 +1,195 @@
...
+        let bytes = self.to_bytes_le()?;
+        // Store the proposal cache to the file system.
+        fs::write(&path, bytes)
+            .map_err(|err| anyhow!("Couldn't write the proposal cache to {} - {err}", path.display()))?;
+
+        Ok(())
+    }
+
+    /// Returns the latest round, proposal and signed proposals.
+    pub fn into(self) -> (u64, Option<Proposal<N>>, SignedProposals<N>) {
```
ljedrz, line 114: nit: this would ideally be an `std::convert::Into` impl

node/bft/src/helpers/signed_proposals.rs
```diff
@@ -0,0 +1,154 @@
...
+        self.0.iter().all(|(_, (_, _, signature))| signature.to_address() == expected_signer)
+    }
+}
+
+impl<N: Network> ToBytes for SignedProposals<N> {
+    fn write_le<W: Write>(&self, mut writer: W) -> IoResult<()> {
+        // Write the number of signed proposals.
+        u32::try_from(self.0.len()).map_err(error)?.write_le(&mut writer)?;
+        // Serialize the signed proposals.
+        for (address, (round, batch_id, signature)) in &self.0 {
```
ljedrz, line 43: we might want to consider adding an `impl (From/To)Bytes for (A, B, C)` to `snarkvm-utilities`

node/bft/src/primary.rs
```diff
@@ -94,12 +94,12 @@ pub struct Primary<N: Network> {
...
     latest_proposed_batch_timestamp: Arc<RwLock<i64>>,
-    /// The recently-signed batch proposals (a map from the address to the round, timestamp, batch ID, and signature).
-    signed_proposals: Arc<RwLock<HashMap<Address<N>, (u64, i64, Field<N>, Signature<N>)>>>,
+    /// The recently-signed batch proposals.
+    signed_proposals: Arc<RwLock<SignedProposals<N>>>,
     /// The spawned handles.
     handles: Arc<Mutex<Vec<JoinHandle<()>>>>,
     /// The lock for propose_batch.
-    propose_lock: Arc<TMutex<(u64, i64)>>,
+    propose_lock: Arc<TMutex<u64>>,
```
ljedrz, line 102:
```suggestion
    propose_lock: Arc<AtomicU64>,
```

ljedrz: Though I now see that it also serves as a locking mechanism... `clippy` might complain about it, though.

node/bft/src/primary.rs
```diff
@@ -1081,6 +1105,46 @@ impl<N: Network> Primary<N> {
...
+                let next_round = self_.current_round().saturating_add(1);
+                // Determine if the quorum threshold is reached for the current round.
+                let is_quorum_threshold_reached = {
+                    // Retrieve the certificates for the next round.
+                    let certificates = self_.storage.get_certificates_for_round(next_round);
+                    // If there are no certificates, then skip this check.
+                    if certificates.is_empty() {
+                        continue;
+                    }
+                    let authors = certificates.iter().map(BatchCertificate::author).collect();
```
ljedrz: can we move this assignment under the committee lookback check (i.e. right above the quorum threshold check)? it would avoid an allocation every time that skip is triggered

node/bft/src/primary.rs
```diff
@@ -1546,6 +1603,16 @@ impl<N: Network> Primary<N> {
         self.workers.iter().for_each(|worker| worker.shut_down());
         // Abort the tasks.
         self.handles.lock().iter().for_each(|handle| handle.abort());
+        // Save the current proposal cache to disk.
+        let proposal_cache = {
+            let proposal = self.proposed_batch.write().take();
+            let signed_proposals = self.signed_proposals.write().clone();
```
```suggestion
            let signed_proposals = self.signed_proposals.read().clone();
```

ljedrz (commented):
Left a few comments and suggestions.

node/bft/src/primary.rs
raychu86: Resolved with [c53f5fd](https://github.com/AleoHQ/snarkOS/pull/3239/commits/c53f5fdc50cc47c68b04dad81e380e339e708e7b)

node/bft/src/primary.rs
raychu86: Resolved with [c53f5fd](https://github.com/AleoHQ/snarkOS/pull/3239/commits/c53f5fdc50cc47c68b04dad81e380e339e708e7b)

node/bft/src/helpers/proposal_cache.rs
raychu86: Resolved with [c53f5fd](https://github.com/AleoHQ/snarkOS/pull/3239/commits/c53f5fdc50cc47c68b04dad81e380e339e708e7b)

node/bft/src/helpers/proposal.rs
raychu86: Resolved with [c53f5fd](https://github.com/AleoHQ/snarkOS/pull/3239/commits/c53f5fdc50cc47c68b04dad81e380e339e708e7b)

howardwu (approved):

PR #3246, opened 2024-05-02
[Logs] Update log levels for safe behavior

raychu86:
## Motivation

This PR updates the log messages, primary to de-escalate `WARN` logs that should realistically be safe and be on the `DEBUG` level.

Examples of such logs are:
- `WARN Cannot store a signature from '{peer_ip}' - This batch was already certified`
- `WARN Cannot sign a batch from '{peer_ip}' - Batch for round {round} already exists in storage (gc = {gc_round})`
- `WARN Primary is safely skipping (round {round} was already certified)`

These do not need to be at a `WARN` level as are alarming to node operators, but they demonstrate safe behavior.

### DIFF
--- a/node/bft/src/primary.rs
+++ b/node/bft/src/primary.rs
@@ -384,7 +384,8 @@ impl<N: Network> Primary<N> {
                     }
                 }
             }
-            bail!("Primary is safely skipping {}", format!("(round {round} was already certified)").dimmed());
+            debug!("Primary is safely skipping {}", format!("(round {round} was already certified)").dimmed());
+            return Ok(());
         }
 
         // Retrieve the committee to check against.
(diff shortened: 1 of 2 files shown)


node/bft/src/primary.rs
```diff
@@ -762,7 +773,15 @@ impl<N: Network> Primary<N> {
                     // Ensure the batch ID matches the currently proposed batch ID.
                     if proposal.batch_id() != batch_id {
                         match self_.storage.contains_batch(batch_id) {
-                            true => bail!("This batch was already certified"),
+                            // If this batch was already certified, return early.
+                            true => {
+                                debug!(
+                                    "Primary is safely skipping a a batch signature from {peer_ip} for round {} - batch is already certified",
```
howardwu, line 779:
```suggestion
                                    "Primary is safely skipping a a batch signature from {peer_ip} for round {} - batch was recently certified",
```

howardwu (approved):

PR #3254, opened 2024-05-13
[Fix] Clear cached validator requests on disconnect

niklaslong:
This PR clears the validator requests cache map for a particular peer on disconnect. This ensures the collection is periodically cleaned up despite not having a time-based expiry and is sufficient to avoid infinite growth as the committee size is bounded (but the IP may change).

A small aside for future discussion: when time permits, we may want to revisit our approach to cache invalidation both in the gateway and the router. One option would be to use an LRU with a fixed size that we clear on specific state transitions (e.g. on round increment).

### DIFF
--- a/node/bft/src/gateway.rs
+++ b/node/bft/src/gateway.rs
@@ -1085,6 +1085,10 @@ impl<N: Network> Disconnect for Gateway<N> {
     async fn handle_disconnect(&self, peer_addr: SocketAddr) {
         if let Some(peer_ip) = self.resolver.get_listener(peer_addr) {
             self.remove_connected_peer(peer_ip);
+
+            // We don't clear this map based on time but only on peer disconnect.
+            // This is sufficient to avoid infinite growth as the committee has a fixed number
+            // of members.
             self.cache.clear_outbound_validators_requests(peer_ip);
         }
     }
--- a/node/bft/src/gateway.rs
+++ b/node/bft/src/gateway.rs
@@ -1085,6 +1085,7 @@ impl<N: Network> Disconnect for Gateway<N> {
     async fn handle_disconnect(&self, peer_addr: SocketAddr) {
         if let Some(peer_ip) = self.resolver.get_listener(peer_addr) {
             self.remove_connected_peer(peer_ip);
+            self.cache.clear_outbound_validators_requests(peer_ip);
         }
     }
 }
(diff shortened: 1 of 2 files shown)


node/bft/src/helpers/cache.rs
```diff
@@ -39,6 +39,10 @@ pub struct Cache<N: Network> {
     /// The ordered timestamp map of peer IPs and their cache hits on transmission requests.
     seen_outbound_transmissions: RwLock<BTreeMap<i64, HashMap<SocketAddr, u32>>>,
     /// The map of IPs to the number of validators requests.
+    ///
+    /// Note: we don't clear this map based on time but only on peer disconnect.
+    /// This is sufficient to avoid infinite growth as the committee has a fixed number
+    /// of members.
```
iamalwaysuncomfortable:
```suggestion
    /// The map of IPs to the number of validators requests
```

This shouldn't be explained within the struct member comment, but instead on the invocation of `clear_outbound_validators_requests`

node/bft/src/gateway.rs
```diff
@@ -1085,6 +1085,7 @@ impl<N: Network> Disconnect for Gateway<N> {
     async fn handle_disconnect(&self, peer_addr: SocketAddr) {
         if let Some(peer_ip) = self.resolver.get_listener(peer_addr) {
             self.remove_connected_peer(peer_ip);
+            self.cache.clear_outbound_validators_requests(peer_ip);
```
iamalwaysuncomfortable, line 1092: The comment mentioned below should be added here.

niklaslong: Done!

node/bft/src/helpers/cache.rs
```diff
@@ -39,6 +39,7 @@ pub struct Cache<N: Network> {
     /// The ordered timestamp map of peer IPs and their cache hits on transmission requests.
     seen_outbound_transmissions: RwLock<BTreeMap<i64, HashMap<SocketAddr, u32>>>,
     /// The map of IPs to the number of validators requests.
+    ///
```
raychu86:
```suggestion
```

raychu86 (approved):
LGTM,

The only use of `seen_outbound_validators_requests` is when we check that an incoming `ValidatorsResponse` corresponds to an existing `ValidatorsRequest`.

node/bft/src/helpers/cache.rs
niklaslong: Good catch, thanks!

howardwu (approved):

PR #3256, opened 2024-05-14
[Fix] Guard recursive certificate fetching with `is_synced` check.

raychu86:
## Motivation

This PR prevents nodes from recursively fetching missing certificates, despite syncing via blocks. This should not happen, and the node should focus on syncing via the blocks. 

We noticed this edge case where a node would be considered synced for a moment and then start fetching certificates from peers. This cycle can go on for a long time since there may be many certificates to request, however this should be cut short when the node realizes that it should be syncing via blocks instead.


vicsn (approved):
Changes make sense to me - and it passed yesterday's automated e2e test run which includes syncing.

howardwu (approved):

PR #3257, opened 2024-05-16
[HackerOne-2485380] Prevent DOS by limiting sent requests to one per peer

raychu86:
## Motivation

This PR aims to solve a DOS vector that is outlined in https://github.com/AleoNet/snarkOS/issues/3243, where a malicious validator can fill up the `max_redundant_requests` and not respond. This PR adjusts the rules so that we only send one request for that item to a given peer at a time.

This means that `max_redundant_requests` can only be reached with unique peers. And because `max_redundant_requests` is based on the availability threshold you can expect at least one node to be honest.

## Test Plan

A test has been added to ensure that the functionality of `contains_peer_with_sent_request` is as intended.

### DIFF
--- a/node/bft/src/helpers/pending.rs
+++ b/node/bft/src/helpers/pending.rs
@@ -270,7 +270,7 @@ mod tests {
         let addr_1 = SocketAddr::from(([127, 0, 0, 1], 1234));
         let addr_2 = SocketAddr::from(([127, 0, 0, 1], 2345));
         let addr_3 = SocketAddr::from(([127, 0, 0, 1], 3456));
-        let addr_4 = SocketAddr::from(([127, 0, 0, 1], 3456));
+        let addr_4 = SocketAddr::from(([127, 0, 0, 1], 5678));
 
         // Initialize the callbacks.
         let (callback_sender_1, _) = oneshot::channel();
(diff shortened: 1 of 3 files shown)


ghostant-1017:
LGTM!

node/bft/src/helpers/pending.rs
```diff
@@ -254,24 +264,29 @@ mod tests {
         let solution_id_1 = TransmissionID::Solution(rng.gen::<u64>().into());
         let solution_id_2 = TransmissionID::Solution(rng.gen::<u64>().into());
         let solution_id_3 = TransmissionID::Solution(rng.gen::<u64>().into());
+        let solution_id_4 = TransmissionID::Solution(rng.gen::<u64>().into());
 
         // Initialize the SocketAddrs.
         let addr_1 = SocketAddr::from(([127, 0, 0, 1], 1234));
         let addr_2 = SocketAddr::from(([127, 0, 0, 1], 2345));
         let addr_3 = SocketAddr::from(([127, 0, 0, 1], 3456));
+        let addr_4 = SocketAddr::from(([127, 0, 0, 1], 3456));
```
ljedrz: shouldn't this address (specifically the port) be unique?

raychu86: Good catch. Shouldn't affect the test, but i updated the port.

ljedrz (approved):
LGTM

howardwu (approved):

PR #3264, opened 2024-05-20
[Fix] Proposal fails bcz few transmissions

elderhammer:
This PR is to fix this bug：https://github.com/AleoNet/snarkOS/issues/3263


vicsn:
Let's await this discussion first: https://github.com/AleoNet/snarkOS/issues/3263#issuecomment-2120311895

PR #3265, opened 2024-05-21
Avoid unnecessary calculations in propose_batch

elderhammer:
Determine in advance whether the proposal has expired.
If it expires, return directly and skip the subsequent process.

### DIFF
--- a/node/bft/src/primary.rs
+++ b/node/bft/src/primary.rs
@@ -351,12 +351,6 @@ impl<N: Network> Primary<N> {
             return Ok(());
         }
 
-        // Determine if the current proposal is expired.
-        if round == *lock_guard {
-            warn!("Primary is safely skipping a batch proposal - round {round} already proposed");
-            return Ok(());
-        }
-
         // If there is a batch being proposed already,
         // rebroadcast the batch header to the non-signers, and return early.
         if let Some(proposal) = self.proposed_batch.read().as_ref() {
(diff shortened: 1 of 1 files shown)


vicsn (commented):
Interesting find. Perhaps the following should also be moved up for consistency:
```
// Ditto if the batch had already been proposed and not expired.
ensure!(round > 0, "Round 0 cannot have transaction batches");
```

elderhammer:
> Interesting find. Perhaps the following should also be moved up for consistency:
> 
> ```
> // Ditto if the batch had already been proposed and not expired.
> ensure!(round > 0, "Round 0 cannot have transaction batches");
> ```

Done.

node/bft/src/primary.rs
```diff
@@ -342,12 +342,21 @@ impl<N: Network> Primary<N> {
...
+
         // If the current storage round is below the latest proposal round, then return early.
         if round < *lock_guard {
             warn!("Cannot propose a batch for round {round} - the latest proposal cache round is {}", *lock_guard);
             return Ok(());
         }
 
+        // Determine if the current proposal is expired.
+        if round == *lock_guard {
+            warn!("Primary is safely skipping a batch proposal - round {round} already proposed");
```
raychu86: The placement of this check is incorrect. This will prevent the node from re-sending the `proposed_batch` if one already exists.

This should be moved to (at least) after the
` // Ensure the primary has not proposed a batch for this round before.`  check. Because the call to `send_primary_round_to_bft` is needed in some cases to update the storage round.

elderhammer: Agree. Should be rebroadcast to improve broadcast reliability and try to update the round as early as possible.

howardwu (approved):

raychu86 (approved):
LGTM

PR #3267, opened 2024-05-22
Update for ARC-41: handle credits.aleo changes

evanmarshall:
This PR should be merged after the snarkVM ARC-41 changes: https://github.com/AleoNet/snarkVM/pull/2453

## Summary from snarkVM

Currently, validators will need to have 10M credits in order to self.bond as a validator. The problem is that most or all of these validators won't actually have 10M credits. To bring Aleo in line with other ecosystems, we want to relax 2 existing constraints:

Delegators can delegate to any address (with the same 10,000 credit minimum & maximum of 100,000 delegators) even if that address is not in the committee
Validators only require a self bond of 100 credits to join the committee, given the total delegation for them is >= 10M credits.
Although the changes are significant, the staking abstraction should closely follow what existed before.

## Technical changes

  - These changes only require adding support for adding the commission to the struct within the committee mapping when generating the genesis block.
  - We may also have to bump snarkVM/snarkOS versions

### DIFF
--- a/cli/src/commands/start.rs
+++ b/cli/src/commands/start.rs
@@ -393,10 +393,11 @@ impl Start {
                         );
 
                         // Add or update the validator entry in the list of members
-                        members
-                            .entry(*validator_address)
-                            .and_modify(|(stake, _, _)| *stake += amount)
-                            .or_insert((*amount, true, rng.gen_range(0..100)));
+                        members.entry(*validator_address).and_modify(|(stake, _, _)| *stake += amount).or_insert((
+                            *amount,
+                            true,
+                            rng.gen_range(0..100),
+                        ));
                     }
                     // Construct the committee.
                     let committee = Committee::<N>::new(0u64, members)?;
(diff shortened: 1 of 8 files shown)


vicsn:
More required changes surfaced by clippy: https://github.com/AleoHQ/snarkOS/commit/b97c3f728edec4c41d631022f7ea937f0daf6145

cli/src/commands/start.rs
```diff
@@ -658,9 +658,9 @@ fn load_or_compute_genesis<N: Network>(
     ]?);
 
     // Input the parameters' metadata.
+    preimage.extend(snarkvm::parameters::mainnet::BondValidatorVerifier::METADATA.as_bytes());
```
howardwu: These need to be network specific.

Use this format:
```rs
match N::ID {
    snarkvm::console::network::MainnetV0::ID => { preimage.extend(...) },
    snarkvm::console::network::TestnetV0::ID => { preimage.extend(...) },
    snarkvm::console::network::CanaryV0::ID => { preimage.extend(...) },
}
```

vicsn (approved):
CI passes, review comments look incorporated, approving this so it can be tested in canary. If existing reviewers or auditors encounter issues, they can open new PRs.

apruden2008:
LGTM. Thanks!

PR #3269, opened 2024-05-22
[Feature] Introduce `CanaryV0` Network

raychu86:
## Motivation

This PR adds support for an additional network instance - `CanaryV0` defined in https://github.com/AleoNet/snarkVM/pull/2463.


howardwu:
@raychu86 Network ID = 2 should technically be named `CanaryV0`

howardwu (approved):

PR #3275, opened 2024-05-24
Avoid unnecessary calculations in propose_batch

elderhammer:
Determiner in advance whether the current round is 0 or the current round has been proposed.
If so, skip the subsequent process and return directly.

Note: There was a PR (https://github.com/AleoNet/snarkOS/pull/3265) with the same changes, but it was reverted because it could not be directly merged into mainnet. This PR is to resubmit the changes and prepare to be merged into mainnet-staging.


zosorock (approved):
This was already approved by core devs. Thank you for your contribution!

PR #3281, opened 2024-05-28
Revert revert canary network trait

vicsn:
Re-introduces https://github.com/AleoNet/snarkOS/pull/3269

Previously https://github.com/AleoNet/snarkOS/pull/3272 reverted a number of PRs to ensure they can be tested properly. The time has come to test the PRs, so we're doing a partial revert revert.

With https://github.com/AleoNet/snarkOS/pull/3269 back in place they can all be tested on Canary.

After the sister PR https://github.com/AleoNet/snarkVM/pull/2468 is merged, we should update the ref again.

### DIFF
--- a/node/router/src/lib.rs
+++ b/node/router/src/lib.rs
@@ -401,12 +401,17 @@ impl<N: Network> Router<N> {
                 // TODO: Populate me with Mainnet Beta IP addresses.
             ]
         } else if N::ID == snarkvm::console::network::TestnetV0::ID {
-            // Testnet contains the following bootstrap peers.
+            // TestnetV0 contains the following bootstrap peers.
             vec![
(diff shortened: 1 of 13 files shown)


node/router/src/lib.rs
damons, line 1: Need to make sure the bootstrap peers are configured before merge.

vicsn: Thx for noticing. Relaying from Ray:
> No need for Canary since it’s for internal testing at the moment and not open to the public. We would just coordinate the connected IPs as we did previously.

Which I think makes sense. I'll make sure this comments gets looked at before the move to mainnet.

zosorock: We added the IPs to the other PR (#3282)

zosorock:
Duplicate of #3282

PR #3285, opened 2024-05-31
Disable logging to stdout since it is already logged to a file

elderhammer:
## Motivation
This PR is to disable logging to stdout, since it already logs to a file.
https://github.com/AleoNet/snarkOS/blob/81ca9cf6f923dc8955347b0993beebeaa17161fc/cli/src/helpers/logger.rs#L131
Directly outputting logs to stdout is not convenient for viewing, and the error information is flushed away due to continuous output.

### DIFF
--- a/cli/src/helpers/logger.rs
+++ b/cli/src/helpers/logger.rs
@@ -114,16 +114,7 @@ pub fn initialize_logger<P: AsRef<Path>>(
     };
 
     // Initialize tracing.
-    let _ = tracing_subscriber::registry()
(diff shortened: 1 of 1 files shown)


vicsn:
Although I see and empathize with the reasoning - advising against this PR. Various teams and people are depending on stdout info for quick local testing.

elderhammer:
> Although I see and empathize with the reasoning - advising against this PR. Various teams and people are depending on stdout info for quick local testing.

Ok

PR #3288, opened 2024-06-04
Remove redundant code

ghostant-1017:
The logic is redundant, because we already get the signer at 
https://github.com/AleoNet/snarkOS/pull/3288/files#diff-730a4dec91b426c3e3cb97824dba5b96097304cf2568e9376f3a6ed793080284R784-R787

### DIFF
--- a/node/bft/src/primary.rs
+++ b/node/bft/src/primary.rs
@@ -822,10 +822,6 @@ impl<N: Network> Primary<N> {
                     }
                     // Retrieve the committee lookback for the round.
                     let committee_lookback = self_.ledger.get_committee_lookback_for_round(proposal.round())?;
-                    // Retrieve the address of the validator.
-                    let Some(signer) = self_.gateway.resolver().get_address(peer_ip) else {
-                        bail!("Signature is from a disconnected validator");
-                    };
                     // Add the signature to the batch.
                     proposal.add_signature(signer, signature, &committee_lookback)?;
                     info!("Received a batch signature for round {} from '{peer_ip}'", proposal.round());


apruden2008:
Is this connected to an issue? And if not, can we quantify the impact?

ghostant-1017:
> Is this connected to an issue? And if not, can we quantify the impact?

It's not connected to any issues, it's just redundant code, doesn't make any effects.

node/bft/src/primary.rs
```diff
@@ -822,10 +822,6 @@ impl<N: Network> Primary<N> {
                     }
                     // Retrieve the committee lookback for the round.
                     let committee_lookback = self_.ledger.get_committee_lookback_for_round(proposal.round())?;
-                    // Retrieve the address of the validator.
-                    let Some(signer) = self_.gateway.resolver().get_address(peer_ip) else {
```
raychu86, line 826: I don't believe this code is redundant. The previous check above:
```
        if self.gateway.resolver().get_address(peer_ip).map_or(true, |address| address != signer) {
            // Proceed to disconnect the validator.
            self.gateway.disconnect(peer_ip);
            bail!("Malicious peer - batch signature is from a different validator ({signer})");
        }
```

Returns true if there was no signer associated with a connected peer ip. Removing this check means that we would now allow validators to process the signatures for non-connected validators.

In theory, could alter the above check, but I believe the double-lookup is cheap enough to not warrant any changes.

ljedrz (commented):
There are 2 points to consider:
- the lookup at L787 already returns early for disconnected validators, though the message would indicate that the signer was different
- that particular lookup is in fact quite cheap, but avoidable

My suggestion would be to adjust this change so that we:
1. perform the address lookup and bind it to a variable or - if it fails - return early stating that the validator is disconnected
2. check if the address is the signer and if that fails, return early stating that the signer is different

That way we would avoid the double lookup, and make the logs more clear.

There is one caveat: the validator technically could disconnect between L787 and L826, which would alter the current behavior with such a change (or with the one currently proposed in the PR); this may or may not be acceptable.

vicsn:
Thank you for the suggestion. 🙏 Closing for the reasons mentioned in the above-mentioned comments

PR #3292, opened 2024-06-05
Use parameters to simplify config file updates

robwafle:
Cousin / sister of https://github.com/AleoNet/snarkVM/pull/2477

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,5 +1,13 @@
 version: 2.1
 
+parameters:
+  xlarge:
+    type: string
+    default: xlarge
+  twoxlarge:
+    type: string
+    default: 2xlarge
+
 orbs:
   windows: circleci/windows@5.0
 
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -1,5 +1,13 @@
 version: 2.1
 
+parameters:
+  xlarge:
+    type: string
+    default: anf/xlarge
```
zosorock, line 6: @robwafle can you modify to be just `xlarge` on both (here and on line 9) please?

zosorock (changes requested):

.circleci/config.yml
vicsn: d73667cc1

zosorock:
These changes are already implemented in `mainnet-staging` because we were using them for a while already 😄 so I am closing as unnecessary at this point. Thank you for the changes anyway @robwafle !

PR #3295, opened 2024-06-06
 [Fix] Guard recursive certificate fetching with is_synced check.

### LINKED ISSUE #3293: [Bug] Sync status incorrectly reported
## 🐛 Bug Report
`snarkos_bft_is_synced` metric always reports 0. It appears its criteria to be considered synced is either too strict or flawed. 

I confirmed with client and validator operators on canary net that they also see this issue. 

## Steps to Reproduce

Based on slack threads, any client or validator will see their node as not synced. 

## Expected Behavior

Once the node is synced, or within certain block time of current time the node should report itself as synced. 

## Your Environment

- mainnet-staging (`81ca9cf6f923dc8955347b0993beebeaa17161fc`)
- v1.66+
- Ubuntu 22

vicsn:
## Motivation

Reintroduces https://github.com/AleoNet/snarkOS/pull/3256
Which was reverted.
[CI passed here.](https://app.circleci.com/pipelines/github/ProvableHQ/snarkOS/13303/workflows/e7f49124-79d0-4282-a159-c3a63026631e)

Closes https://github.com/AleoNet/snarkOS/issues/3293

### DIFF
--- a/node/bft/src/primary.rs
+++ b/node/bft/src/primary.rs
@@ -161,7 +161,8 @@ impl<N: Network> Primary<N> {
                         // We use a dummy IP because the node should not need to request from any peers.
                         // The storage should have stored all the transmissions. If not, we simply
                         // skip the certificate.
-                        if let Err(err) = self.sync_with_certificate_from_peer(DUMMY_SELF_IP, certificate).await {
+                        if let Err(err) = self.sync_with_certificate_from_peer::<true>(DUMMY_SELF_IP, certificate).await
+                        {
                             warn!("Failed to load stored certificate {} from proposal cache - {err}", fmt_id(batch_id));
                         }
                     }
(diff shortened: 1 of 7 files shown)


node/bft/src/primary.rs
```diff
@@ -2070,13 +2087,48 @@ mod tests {
 
         // The author must be known to resolver to pass propose checks.
         primary.gateway.resolver().insert_peer(peer_ip, peer_ip, peer_account.1.address());
+        // The primary must be considered synced.
+        primary.sync.block_sync().try_block_sync(&primary.gateway.clone()).await;
```
gluax, line 2091: NIT: Tests should be moved to a different file. Otherwise, it's unclear if this line of code was in a test. Commenting since I would have left a comment about the fact that we are taking a reference to a `clone`, i.e, the `clone` is pointless.

gluax (approved):

evanmarshall (approved):
lgtm

Meshiest (approved):

node/bft/src/primary.rs
vicsn: I think that's a subjective decision which won't go out of this code anytime soon. Fortunately Github indicates `mod tests {` in the blue-grey summary line just above.

PR #3296, opened 2024-06-07
[Feature] Add network ID support to devnet scripts 

vicsn:
## Motivation

Reintroduces https://github.com/AleoNet/snarkOS/pull/3262

Which was reverted

## Test Plan

It is not influenced by CI

## Related PRs

We had a temporary Mirror here: https://github.com/ProvableHQ/snarkOS/pull/12


gluax (approved):
LGTM and makes sense to me. 

I do wish we could consider an endpoint `/api/v1/network`(or something equivalent) that would return the current networks name though.

Meshiest (approved):

PR #3298, opened 2024-06-07
[Perf] Include json serialisation in the get_blocks blocking task

vicsn:
## Motivation

Revives the reverted PR https://github.com/AleoNet/snarkOS/pull/3253
[CI passed](https://app.circleci.com/pipelines/github/ProvableHQ/snarkOS?branch=kp%2Ffix%2Fstress_test_6)

### DIFF
--- a/node/rest/src/routes.rs
+++ b/node/rest/src/routes.rs
@@ -135,16 +135,18 @@ impl<N: Network, C: ConsensusStorage<N>, R: Routing<N>> Rest<N, C, R> {
             )));
         }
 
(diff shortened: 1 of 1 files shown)


gluax (approved):

evanmarshall (approved):
Looks good overall. If the serialization is so slow, we may want to migrate away from json to protobuf. I'm guessing protobuf would also be faster than the homegrown binary serialization that exists in many places as well.

Meshiest (approved):

PR #3306, opened 2024-06-08
Adds a restrictions ID check into handshake protocols

vicsn:
## Motivation

This PR introduces a restrictions ID check into the handshake protocols.

When two nodes are handshaking on the Router or Gateway level, it is expected that they transmit in their ChallengeResponse messages a copy of their current restrictions ID to their peer, who subsequently checks and enforces that the restrictions ID matches what they expect from their node.

As such, going forward, the following node types are responsible to adhere to the same restrictions list:

- NodeType::Validator
- NodeType::Client

Of note, as provers do not validate transitions or transactions, they do not carry ledger state and thus are not subject to the restrictions list. Note that it is critical for clients to also perform the same validation as the validators because the validator rely upon core clients to assist with validation and propagation of transactions to the validators' memory pools.

This design is defined by the foundation to require all handshake connections to enforce the current restrictions ID to be correct and matching at the time of connection; however, it doesn't require consensus to enforce a matching restrictions ID at the time of use.

This PR builds on top of https://github.com/AleoNet/snarkVM/pull/2487 in snarkVM, which introduces the concept of the restrictions list for programs, functions, and arguments.

## Related PRs

Reviewed here: https://github.com/ProvableHQ/snarkOS/pull/14 
[CI passed](https://app.circleci.com/pipelines/github/ProvableHQ/snarkOS/13330/workflows/193a267a-121d-464d-97b2-402437e26501)

### DIFF
--- a/node/bft/ledger-service/src/prover.rs
+++ b/node/bft/ledger-service/src/prover.rs
@@ -20,7 +20,7 @@ use snarkvm::{
         narwhal::{BatchCertificate, Data, Subdag, Transmission, TransmissionID},
         puzzle::{Solution, SolutionID},
     },
-    prelude::{bail, Address, Field, Network, Result},
+    prelude::{bail, Address, Field, Network, Result, Zero},
 };
 
 use indexmap::IndexMap;
@@ -54,6 +54,11 @@ impl<N: Network> LedgerService<N> for ProverLedgerService<N> {
     /// Returns the latest block in the ledger.
     fn latest_block(&self) -> Block<N> {
         unreachable!("Latest block does not exist in prover")
+    }
+
+    /// Returns the latest restrictions ID in the ledger.
+    fn latest_restrictions_id(&self) -> Field<N> {
+        Field::zero()
     }
 
     /// Returns the latest cached leader and its associated round.
--- a/node/src/prover/router.rs
+++ b/node/src/prover/router.rs
@@ -25,7 +25,7 @@ use snarkos_node_router::messages::{
     UnconfirmedTransaction,
 };
 use snarkos_node_tcp::{Connection, ConnectionSide, Tcp};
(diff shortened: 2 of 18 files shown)


node/src/prover/router.rs
```diff
@@ -45,7 +45,8 @@ impl<N: Network, C: ConsensusStorage<N>> Handshake for Prover<N, C> {
         let conn_side = connection.side();
         let stream = self.borrow_stream(&mut connection);
         let genesis_header = *self.genesis.header();
-        self.router.handshake(peer_addr, stream, conn_side, genesis_header).await?;
+        let restrictions_id = Field::zero(); // Provers may bypass restrictions, since they do not validate transactions.
+        self.router.handshake(peer_addr, stream, conn_side, genesis_header, restrictions_id).await?;
```
gluax, line 49: I feel like the handshake method should take an option, then no? Why send extra data if it's unnecessary?

node/bft/ledger-service/src/prover.rs
```diff
@@ -56,6 +56,11 @@ impl<N: Network> LedgerService<N> for ProverLedgerService<N> {
         unreachable!("Latest block does not exist in prover")
     }
 
+    /// Returns the latest restrictions ID in the ledger.
+    fn latest_restrictions_id(&self) -> Field<N> {
+        Field::zero()
+    }
+
```
gluax, line 63: TBH I feel like it should be considered here as well. Rather than return useless unused data this should be an Option and prover's should return `None`

Meshiest (approved):

node/src/prover/router.rs
vicsn: I agree it would have been cleaner, if the PR was not merged yet or we were not in a crunch, I would have pushed the change. Thx for the suggestion

PR #3318, opened 2024-06-16
fix(tcp): fix outbound tcp requests not respecting --node bind port

Meshiest:
## Motivation

These changes force outgoing TCP connections to respect the address configured by the `--node` flag. When not properly bound, peers of multi-home/multi-address hosts are broadcasting incorrect socket addresses, which can increase load on unrelated nodes.

Example scenario (based off of our canary nodes, all 3 nodes running on the same server):

1. client 1 bound on 6.5.4.100
2. client 2 bound on 6.5.4.101
3. client 3 bound on 6.5.4.102
4. clients 2 and 3 make outgoing connection requests on 6.5.4.100, rather than .101 or .102 
5. [the wrong socket address is extrapolated for successive peer requests](https://github.com/AleoNet/snarkOS/blob/mainnet-staging/node/router/src/handshake.rs#L217)

Side-effects of not binding the correct address:
- other nodes in the network will not connect to clients 1, 2 or 3 at the same time because the `is_connected` check is based on this "resolved address"
- peers connected to clients 2 and 3 broadcast the wrong address and clients 2/3 never get new peers from initiated connections (other than ones configured properly via `--peers 6.5.4.101:4130`)
- increased peering traffic is sent to client 1

## Test Plan

We are successfully running this on canary, and this is a minimal change based on [tokio's documentation](https://docs.rs/tokio/latest/tokio/net/struct.TcpSocket.html)

### DIFF
--- a/node/tcp/src/tcp.rs
+++ b/node/tcp/src/tcp.rs
@@ -225,21 +225,31 @@ impl Tcp {
             return Err(io::ErrorKind::AlreadyExists.into());
         }
 
-        let stream =
-            match timeout(Duration::from_millis(self.config().connection_timeout_ms.into()), TcpStream::connect(addr))
-                .await
-            {
-                Ok(Ok(stream)) => Ok(stream),
-                Ok(err) => {
-                    self.connecting.lock().remove(&addr);
-                    err
-                }
-                Err(err) => {
-                    self.connecting.lock().remove(&addr);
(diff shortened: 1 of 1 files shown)


ljedrz (approved):
Looks reasonable to me :+1:.

niklaslong (approved):

apruden2008:
@zosorock this is good to merge pending CI passing

PR #3328, opened 2024-06-24
Clean up verbosity 0 logging messages and display one-line block sync

damons:
## Motivation

Currently verbosity settings of the client are mixed with info, debug, and trace log levels.   When a user starts the node, the mix of messages is confusing and it's not clear if a node is connecting, syncing blocks, and working properly.  It's important to give immediate feedback to users that their node is syncing as quickly as possible without distractions.  This patch does two important things:

1.  When running with `--verbosity 0` it only displays `info` messages that give meaningful information to the user about:  peer connections (initial, drops, and refreshes) and block sync progress.
2. It removes the extra `\n\n`'s from the block advance and instead only prints: [{BLOCK_NUM}] {BLOCK_HASH} which gives a clear message to the user every time a block is added to the ledger.  When run with verbosity 0, the primary information a user sees is blocks being added to the ledger.  This gives immediate and clear feedback to the user that their client node is syncing properly and it shows the performance of the node.

Here's an example of `--verbosity 0` output using this PR:

Screenshot 2024-06-23 at 8 29 11 PM

Several files where changed from `info!` to `debug!` and vice versa.  Some `debug!`s were changed to `trace!` as some of the modules override the usual log levels which causes trace and debug logs to be produced at verbosity 0.  

## Test Plan

Compiled and run using `--verbosity 0` and the client output matches the above screenshot.

### DIFF
--- a/node/bft/ledger-service/src/ledger.rs
+++ b/node/bft/ledger-service/src/ledger.rs
@@ -352,7 +352,7 @@ impl<N: Network, C: ConsensusStorage<N>> LedgerService<N> for CoreLedgerService<
             metrics::update_block_metrics(block);
         }
 
-        tracing::info!("\n\nAdvanced to block {} at round {} - {}\n", block.height(), block.round(), block.hash());
+        tracing::info!("[{}] {}", block.height(), block.hash());
         Ok(())
     }
 }
--- a/node/bft/src/gateway.rs
+++ b/node/bft/src/gateway.rs
@@ -175,7 +175,7 @@ impl<N: Network> Gateway<N> {
         worker_senders: IndexMap<u8, WorkerSender<N>>,
         sync_sender: Option<SyncSender<N>>,
     ) {
-        debug!("Starting the gateway for the memory pool...");
+        trace!("Starting the gateway for the memory pool...");
 
         // Set the primary sender.
         self.primary_sender.set(primary_sender).expect("Primary sender already set in gateway");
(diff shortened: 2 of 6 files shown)


joske (approved):
LGTM, I'm all for cleaner, less noisy but still informative logs.

node/bft/ledger-service/src/ledger.rs
```diff
@@ -352,7 +352,7 @@ impl<N: Network, C: ConsensusStorage<N>> LedgerService<N> for CoreLedgerService<
             metrics::update_block_metrics(block);
         }
 
-        tracing::info!("\n\nAdvanced to block {} at round {} - {}\n", block.height(), block.round(), block.hash());
+        tracing::info!("[{}] {}", block.height(), block.hash());
```
raychu86, line 355: While more verbose, the previous message of `Advanced to lock {} at round {}` has been very valuable for debugging and parsing logs.

node/bft/src/gateway.rs
```diff
@@ -175,7 +175,7 @@ impl<N: Network> Gateway<N> {
         worker_senders: IndexMap<u8, WorkerSender<N>>,
         sync_sender: Option<SyncSender<N>>,
     ) {
-        debug!("Starting the gateway for the memory pool...");
+        trace!("Starting the gateway for the memory pool...");
```
raychu86, line 178:
```suggestion
        info!("Starting the gateway for the memory pool...");
```

This should be an `info` log like all the other module bootup messages.

node/bft/src/gateway.rs
```diff
@@ -524,7 +524,7 @@ impl<N: Network> Gateway<N> {
         // If the event was unable to be sent, disconnect.
         if let Err(e) = &result {
             warn!("{CONTEXT} Failed to send '{name}' to '{peer_ip}': {e}");
-            debug!("{CONTEXT} Disconnecting from '{peer_ip}' (unable to send)");
+            trace!("{CONTEXT} Disconnecting from '{peer_ip}' (unable to send)");
```
raychu86, line 527:
```suggestion
            warn!("{CONTEXT} Disconnecting from '{peer_ip}' - Failed to send '{name}': {e}");
```

We can unify the 2 log messages to a single one that matches the other disconnect warn messages.

raychu86 (changes requested):

PR #3334, opened 2024-06-27
fix: lockfile

gluax:
## Motivation

The reason is that if you build or install with `--locked` it will lock to the version in the `Cargo.lock`, which currently has mixed versions of `snarkVM`.

This is problematic for obvious reasons.

In the future, let `rust-analyzer` or `cargo build`(without `--locked`) automatically fix this, no need to use `cargo update`.

## Test Plan

All tests should hopefully still work with this if they are being run with `--locked,` which they should be.

## Related PRs

N/A

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3336,7 +3336,7 @@ dependencies = [
  "snarkos-node-consensus",
  "snarkos-node-router",
  "snarkvm",
- "snarkvm-synthesizer 0.16.19 (git+https://github.com/AleoNet/snarkVM.git?rev=8a05317)",
+ "snarkvm-synthesizer",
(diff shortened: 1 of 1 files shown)


Meshiest (approved):
These changes also cleaned up the lockfiles of our tooling that imports snarkOS

ljedrz (approved):
LGTM 👍

raychu86:
I would close this in favor of https://github.com/AleoNet/snarkOS/pull/3342, which I believe fixes the issue by actually referencing the correct crate versions.

gluax:
@raychu86 makes sense to me. I see the same lock changes in that PR.

PR #3337, opened 2024-06-29
fix incorrect usage of Ordering::Relaxed

JakkuSakura:
## Motivation
Ordering is important. While Relaxed is atomic on its own, it can't sync cache between threads.

This PR fixes such incorrect usage


ljedrz (dismissed):
The choice of the `Relaxed` ordering may have been due to a more frequent use of this atomic in the past. I don't think this has caused any issues, but using acquire+release is indeed a more correct approach.

niklaslong (dismissed):
LGTM 👍 

We'll just need to revert the formatting changes before merging (we use nightly).

niklaslong:
@JakkuSakura thanks again for the contribution, really appreciate it! Just a heads up, I've updated the PR with a commit to adjust the formatting.

joske (approved):

ljedrz (approved):

An-tol:
gm

PR #3339, opened 2024-07-01
Update README.md

IAmScRay:
## Motivation

Since the build guide does not mention an ability to use snarkOS flags with `.sh`-files (specifically `--network`), it has been added to the guide with appropriate network values.

Also, if anyone wants to NOT rebuild `snarkOS` binary at startup, I've added a straightforward description with **`snarkos`**. A disclaimer is also added if anyone decides to run the node this way.

## Test Plan

No changes to the code were applied. 

## Related PRs

No related PRs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,13 +127,19 @@ This will output a new Aleo account in the terminal.
 
 Next, to start a proving node, from the `snarkOS` directory, run:
 ```
-./run-prover.sh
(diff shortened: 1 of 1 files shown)


vicsn:
Thank you 🙏 The README is already quite large, and there is a very large prover ecosystem with functioning mining pools, closing for now until more folks complain

PR #3350, opened 2024-07-05
Update README.md

NdugbeTochukwu:
## Motivation

Decentralization is my motivation

## Test Plan
I added the codes for opening the ports on the router


raychu86:
Network configurations may be more complex and may not be encompassed by these commands. We also do not want to enforce `sudo` commands; this should be up to the node operation.

In addition, `sudo ufw allow ssh` has no relevance to the note and is also up to the discretion of the node operator.

Closing, feel free to push back

PR #3351, opened 2024-07-05
Update HW requirements, node start commands, and node shell scripts

kpandl:
This PR changes the `README.md` file and the validator/client/prover shell scripts.

[Rendered Version](https://github.com/AleoNet/snarkOS/blob/b183b268b6fb2c0c033ac0bb57af2e9b60880aa1/README.md)

For the `README.md`, the PR specifically:
* provides updated HW requirements for validators/clients
* removes official HW recommendations for provers
* introduces the concept of core and outer clients
* updates the recommended start commands of validators/clients

For the corresponding shell scripts, it:
* makes use of the updated start commands
* provides a shell script for core/outer clients instead of general clients
* removes the automatic GitHub update loop (which could trigger a large number of nodes to go offline around the same time)


run-outer-client.sh
```diff
@@ -0,0 +1,48 @@
...
+then
+  read -r -p "Enter the peers (comma-separated) (e.g., core_client_ip_1:4130,core_client_ip_2:4130,core_client_ip_3:4130,outer_client_ip_1:4130,...): "
+  PEERS=$REPLY
+fi
+
+if [ "${PEERS}" == "" ]
+then
+  echo "Missing peers."
+  exit 1
+fi
```
raychu86: Is this necessary?

The `peers` flag in the start CLI is intended for "trusted peers". If there are none, the client node will still be able to connect to the network via the bootstrap peers.

kpandl: Thanks, good catch! I've updated the file!

raychu86 (approved):

PR #3358, opened 2024-07-09
 Limit parallel processing for clients

### LINKED ISSUE #3341: [Bug] OOM kill because sudden memory increase and requests
## 🐛 Bug Report

snarkOS has presented several OOM Kill incidents with the following characteristics:

**Incident Date: June 30, 2024**
Affected Service: snarkos Service
Nodes Affected: Node A, Node B, Node C
Summary
On June 30, 2024, the snarkos service on testnet beta client nodes Node A, Node B, and Node C experienced an outage due to the OOM (Out of Memory) killer being invoked around the same time on all three nodes.
Incident Details
**Node A:**

- Node type: Core Client
- Timestamp: 9:19:42 PM EDT
- Memory Requested by snarkos: 81.6 GB
- Memory Used by snarkos before kill: 31.6 GB
- RAM Memory on the server: 32GB
- Last Block Advanced before the incident: 55839

**Node B:**

- Node type: Core Client
- Timestamp: 9:19:44 PM EDT
- Memory Requested by snarkos: 90.8 GB
- Memory Used by snarkos before kill: 31.6 GB
- RAM Memory on the server: 32GB
- Last Block Advanced before the incident: 55839

**Node C:**

- Node type: Core Client
- Timestamp: 9:53:52 PM EDT
- Memory Requested by snarkos: 83.5 GB
- Memory Used by snarkos before kill: 31.6 GB
- RAM Memory on the server: 32GB
- Last Block Advanced before the incident: 56105

Effect on Validator(s)
Validator nodes were minimally affected by the client nodes going down, only experiencing a brief spike of 15% more memory usage than normal.

**Incident Date: July 2, 2024**
Affected Service: snarkos Service
Nodes Affected: Node D, Node A
Summary
On June 30, 2024, the snarkos service on testnet beta client nodes Node D, and Node A experienced an outage due to the OOM (Out of Memory) killer being invoked at different times.
Incident Details
**Node D:**

- Node type: Core Client
- Timestamp: 2:53:44 AM EDT
- Memory Requested by snarkos: 95.8 GB
- Memory Used by snarkos before kill: 30.2 GB
- RAM Memory on the server: 32GB
- Last Block Advanced before the incident: 68199

**Node A:**

- Node type: Core Client
- Timestamp: 11:03:21 PM EDT
- Memory Requested by snarkos: 91.5 GB
- Memory Used by snarkos before kill: 30.2 GB
- RAM Memory on the server: 32GB
- Last Block Advanced before the incident: 70926

Effect on Validator(s)
Validator nodes were minimally affected by the client nodes going down, only experiencing a brief spike of 7% more memory usage than normal.

## Steps to Reproduce

Not known, but the blocks close to the last advanced by the clients should indicate what triggered the issue and lead to way to  reproduce.

It's been mentioned that blocks 55840, 56108, 68201, 70928 have complex program deployments.

## Expected Behavior

snarkOS controlling the amount that it requests and not letting the amount of complexity of deployments to dictate memory, so that also it does not becomes a security issue (like DDoS).

## Your Environment

- snarkOS Version testnet-beta-v0.3.0
- Rust Version 1.79.0
- Ubuntu 22.04LTS

vicsn:
## Motivation

Clients run out of memory because they have no limit on how many transactions or solutions they verify in parallel. This PR proposes to queue them (just like the validator does in Consensus) and limit how much parallel verification we do.

We can do a lot of clever things to increase the processing speed, like check how many constraints the incoming transactions have, await on a channel to rapidly start verifying, but the focus for now is simplicity and safety.

Even though it was [recently suggested clients should have at least 128 GiB of memory](https://github.com/AleoNet/snarkOS/pull/3351), the current implementation uses "only" up to 30 GiB for transaction verification. The right number is up for debate.

## Test Plan

[CI passed](https://app.circleci.com/pipelines/github/ProvableHQ/snarkOS?branch=max_concurrent_deploy_verification_2-ci)
Ran a local network shooting some executions.
In canary, more serious concurrent traffic can be shot at the network.

## Related PRs

Potentially closes: https://github.com/AleoNet/snarkOS/issues/3341
Replaces: https://github.com/AleoNet/snarkOS/pull/2970


vicsn:
An open design question is how much memory should be reserved for worst case transaction + solution verification. 

Summoning @evanmarshall @zosorock @Meshiest . [Provable previously recommended](https://github.com/AleoNet/snarkOS/pull/3351) that clients should have 128 GiB of RAM, but I understand you and others want to run with less RAM. So my questions to you:
1. What do you think is a sufficient default amount of RAM which should be assumed to be available for transaction/solution verification?
2. How badly do you want an additional feature flag which lets you increase the amount of RAM used for transaction/solution verification?

Meshiest:
I have an undiversified node operational experience so I don't have a good estimate for 1 and personally haven't run into anything needing the feature flag from 2

The main limiter we've observed in client sync is core count. 16 cores was barely able to keep up with block production on canary after the 15tps tests. Upgrading a client VM from 16 to 32 cores massively increased sync speed. Our servers with more than 64 cores were powerhouses when syncing. RAM seemed less important in our tests, though we weren't using large programs or verifying any solutions.

raychu86 (dismissed):
The code quality and logic LGTM. 

Dialing in a reasonable number for these bounds will likely be an iterative process. 

In theory we could also use the `sysinfo` crate to fetch the total memory of the machine thats running the node.

node/src/client/mod.rs
```diff
@@ -180,6 +229,128 @@ impl<N: Network, C: ConsensusStorage<N>> Client<N, C> {
...
+                // Try to verify transactions.
+                let mut tx_queue = node.transaction_queue.lock();
+                while let Some((_, (peer_ip, serialized, transaction))) = tx_queue.pop_lru() {
+                    // Increment the verification counter.
+                    let previous_counter = node.tx_verification_counter.fetch_add(1, Relaxed);
+                    let _node = node.clone();
+                    // For each transaction, spawn a task to verify it.
+                    tokio::spawn(async move {
+                        // Check the transaction.
+                        if _node.ledger.check_transaction_basic(&transaction, None, &mut rand::thread_rng()).is_ok() {
```
ljedrz: shouldn't this be done in the context of a blocking task?

ljedrz (dismissed):
LGTM with one open point

vicsn:
@raychu86 sorry added a separate execution queue, I couldn't let a simple 200x improvement in max throughput slide (assuming available compute): 9049f3e34

> In theory we could also use the sysinfo crate to fetch the total memory of the machine thats running the node.

Yes thought about it, but this dynamic behaviour will complicate infra planning too much I think, so I think there should rather be a --available-ram flag or something if users have very diverse preferences.

node/src/client/mod.rs
vicsn: Makes sense to me: [799f157](https://github.com/AleoNet/snarkOS/pull/3358/commits/799f157ceaccd768ca26ddab340fae3b3f09bd26)

node/src/client/mod.rs
```diff
@@ -35,21 +35,58 @@ use snarkvm::{
...
 };
-use tokio::task::JoinHandle;
+use tokio::{task::JoinHandle, time::sleep};
+
+/// The maximum number of deployments to verify in parallel.
+/// Note: worst case memory to verify a deployment (MAX_DEPLOYMENT_CONSTRAINTS = 1 << 20) is ~2 GiB.
+const MAX_PARALLEL_DEPLOY_VERIFICATIONS: usize = 5;
+/// The maximum number of executions to verify in parallel.
+/// Note: worst case memory to verify an execution is 0.01 GiB.
+const MAX_PARALLEL_EXECUTE_VERIFICATIONS: usize = 1000;
```
ljedrz, line 71: the default (and the one we're currently using) [blocking task](https://docs.rs/tokio/latest/tokio/task/fn.spawn_blocking.html) count limit is 512; we might need to raise it now, especially considering that there are other users of blocking tasks all around the codebase

vicsn: I will run another benchmark to see how fast executions can be verified, perhaps its just a few ms.

From how many ms would you recommend using blocking tasks? Also could you make a recommendation for the new count limit, assuming scenario's 1. executions are verified using spawns 2. executions are verified using spawn_blocking?

ljedrz: For anything that goes into the millisecond scale we should be using blocking tasks, otherwise we will negatively affect the async executor. In addition, for any operations that aren't just slow but are heavy on the CPU we should be using `rayon` to execute them _in addition to_ using blocking tasks, or utilize `oneshot` channels ([example](https://github.com/AleoNet/snarkOS/commit/a4c905e2e697b7abeebe3aa16c6fbc6e4591c510)).

As for bumping the number of blocking threads... I would say it could be a lot larger (especially if the actual work is offloaded to `rayon` and they are used only to be able to `await` such operations), but we should test how it impacts node memory use etc.

vicsn: As discussed elsewhere: ["After reaching the upper limit, the tasks are put in a queue."](https://docs.rs/tokio/latest/tokio/task/fn.spawn_blocking.html). Given that changing the task limit may have complex side-effects, I'll mark it as a future TODO for further optimization

raychu86 (dismissed):
LGTM

node/src/client/router.rs
```diff
@@ -332,15 +311,23 @@ impl<N: Network, C: ConsensusStorage<N>> Inbound<N> for Client<N, C> {
...
     ) -> bool {
-        // Check that the transaction is not a fee transaction.
-        if transaction.is_fee() {
-            return true; // Maintain the connection.
-        }
-        // Check that the transaction is well-formed and unique.
-        if self.ledger.check_transaction_basic(&transaction, None, &mut rand::thread_rng()).is_ok() {
-            // Propagate the `UnconfirmedTransaction`.
-            self.propagate(Message::UnconfirmedTransaction(serialized), &[peer_ip]);
+        // Try to add the transaction to a verification queue, without changing LRU status of known solutions.
```
d0cd:
```suggestion
        // Try to add the transaction to a verification queue, without changing LRU status of known transactions.
```

raychu86:
@vicsn Is this PR still relevant?

vicsn:
> @vicsn Is this PR still relevant?

Yes I believe this to still be relevant to reduce client downtime. The main alternative would be to recommend the ecosystem to auto-restart their clients if they OOM, but still such events will lead to, in web3 terms, FUD. :D

I finally got around to testing this on a devnet, we can clearly see how the memory usage on the 2nd and 3rd peak (which runs with this PR) is capped at a lower point now. If I would increase the # of deployments, the first peak would keep growing until hitting OOM.

Screenshot 2024-10-03 at 16 06 14

Overall processing speed of deployments does not seem significantly impacted because validators also throttle deployments.

node/src/client/mod.rs
```diff
@@ -181,6 +242,177 @@ impl<N: Network, C: ConsensusStorage<N>> Client<N, C> {
...
     }
 
+    /// Initializes solution verification.
+    fn initialize_solution_verification(&self) {
+        // Start the solution verification loop.
+        let node = self.clone();
+        self.handles.lock().push(tokio::spawn(async move {
+            loop {
+                // If the Ctrl-C handler registered the signal, stop the node.
+                if node.shutdown.load(Acquire) {
```
gluax, line 254: I know this wasn't done in this PR, but I feel like a [CancellationToken](https://docs.rs/tokio-util/latest/tokio_util/sync/struct.CancellationToken.html) could be considered for every use case I've seen this `Arc<AtomicBool>` so far.

gluax (approved):
Code wise LGM

zkxuerb (dismissed):
LGTM

zosorock:
CI link:
https://app.circleci.com/pipelines/github/AleoNet/snarkOS?branch=pull%2F3358%2Fhead

PR #3360, opened 2024-07-10
add configurable metrics listening cli option

HarukaMa:
## Motivation

Add an snarkOS cli option to configure the listening host and port of the metrics exporter.

Although I think listening to 127.0.0.1 by default should be a better default value, I kept 0.0.0.0 to not break existing users (the previous default was hardcoded `0.0.0.0:9000`). Same for the cli option name, although it's different from rest and cdn options. Lemme know if breaking backwards compatibility is acceptable there.

## Test Plan

Tested locally

## Related PRs

(Link any related PRs here)

### DIFF
--- a/cli/src/commands/start.rs
+++ b/cli/src/commands/start.rs
@@ -128,9 +128,13 @@ pub struct Start {
     /// Specify the path to the file where logs will be stored
     #[clap(default_value_os_t = std::env::temp_dir().join("snarkos.log"), long = "logfile")]
     pub logfile: PathBuf,
+
     /// Enables the metrics exporter
     #[clap(default_value = "false", long = "metrics")]
     pub metrics: bool,
+    /// Specify the IP address and port for the metrics exporter
+    #[clap(long = "metrics-ip")]
+    pub metrics_ip: Option<SocketAddr>,
 
     /// Specify the path to a directory containing the storage database for the ledger
     #[clap(long = "storage")]
@@ -558,7 +562,7 @@ impl Start {
 
         // Initialize the metrics.
         if self.metrics {
-            metrics::initialize_metrics();
+            metrics::initialize_metrics(self.metrics_ip);
         }
 
         // Initialize the storage mode.
--- a/node/metrics/src/lib.rs
+++ b/node/metrics/src/lib.rs
@@ -30,6 +30,7 @@ use snarkvm::{
 };
 use std::{
     collections::HashMap,
+    net::SocketAddr,
     sync::{
         atomic::{AtomicUsize, Ordering},
         Arc,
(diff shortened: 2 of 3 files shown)


Meshiest (dismissed):
LGTM

node/metrics/src/lib.rs
```diff
@@ -38,9 +40,12 @@ use std::{
 use time::OffsetDateTime;
 
 /// Initializes the metrics and returns a handle to the task running the metrics exporter.
-pub fn initialize_metrics() {
+pub fn initialize_metrics(ip: Option<SocketAddr>) {
     // Build the Prometheus exporter.
-    metrics_exporter_prometheus::PrometheusBuilder::new().install().expect("can't build the prometheus exporter");
+    metrics_exporter_prometheus::PrometheusBuilder::new()
+        .with_http_listener(ip.unwrap_or(SocketAddr::from_str("0.0.0.0:9000").unwrap()))
```
raychu86: I would prefer not to hard-code the default socketAddr like this - `SocketAddr::from_str("0.0.0.0:9000)`.

Even if it's functionally the same as calling `metrics_exporter_prometheus::PrometheusBuilder::new()` now, if the defaults in PrometheusBuilder change in the future, we would need to manually perform the update.

Ideally, we keep the old syntax if `ip` is not specified here. Maybe with a match case.

raychu86 (changes requested):

cli/src/commands/start.rs
```diff
@@ -128,9 +128,13 @@ pub struct Start {
     /// Specify the path to the file where logs will be stored
     #[clap(default_value_os_t = std::env::temp_dir().join("snarkos.log"), long = "logfile")]
     pub logfile: PathBuf,
+
     /// Enables the metrics exporter
     #[clap(default_value = "false", long = "metrics")]
     pub metrics: bool,
+    /// Specify the IP address and port for the metrics exporter
+    #[clap(long = "metrics-listen")]
```
raychu86: Maybe `metrics-ip` would be a more appropriate name.

```suggestion
    #[clap(long = "metrics-ip")]
```

cli/src/commands/start.rs
```diff
@@ -128,9 +128,13 @@ pub struct Start {
...
+    pub metrics_listen: Option<SocketAddr>,
```
raychu86: Ditto above.

```suggestion
    pub metrics_ip: Option<SocketAddr>,
```

node/metrics/src/lib.rs
HarukaMa: Makes perfect sense, will update after my work shift

ljedrz (dismissed):
LGTM :+1:

HarukaMa:
oops, forgot to clippy and there is an unused import

raychu86 (approved):
LGTM!

PR #3362, opened 2024-07-13
Optimize error handling

ylmin:
## Motivation

Error Handling Improvement: In the `enable_listener` method, use if let Err(e) for error handling instead of directly using expect. This can prevent the program from crashing when an error occurs. The expect method will cause the program to crash and print the specified error message when an error is encountered. This is useful during development and debugging because it helps quickly locate the problem. However, in a production environment, directly crashing may not be the best choice as it can lead to service interruptions.

## Test Plan

## Related PRs

(Link any related PRs here)

### DIFF
--- a/node/router/src/routing.rs
+++ b/node/router/src/routing.rs
@@ -41,7 +41,9 @@ pub trait Routing<N: Network>:
 
     // Start listening for inbound connections.
     async fn enable_listener(&self) {
-        self.tcp().enable_listener().await.expect("Failed to enable the TCP listener");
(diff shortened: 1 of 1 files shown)


vicsn:
Thx for the contribution. No approval from me, the TCP listener is required for startup so using expect is fine. Looking forward to see a more substantial contribution

ylmin:
> 感谢您的贡献。无需我的批准，启动时需要 TCP 侦听器，因此使用 expect 就可以了。期待看到更多实质性的贡献

Ok

PR #3365, opened 2024-07-16
[Feature] Add REST endpoint to get all items in a mapping

raychu86:
## Motivation

This PR adds a new `GET get_mapping_values` REST endpoint to read all the elements in a mapping. 

```
GET /<network>/program/{programID}/mapping/{mappingName}?all={true}&metadata={true}
```

#### Example:

`GET /<network>/program/credits.aleo/mapping/bonded?all=true&metadata=true` will return the full `bonded` mapping of addresses + amounts staked on the network.

### Notes: 
- The new GET endpoint currently requires a query parameters `all` to be set to `true`. In the future, this functionality can be expanded to get subsets of mapping data if required.
- The endpoint is guarded behind the JSON Web Token auth because this call can be expensive for large mappings and may be a DOS vector for nodes. This JWT ensures that only authorized users can call this (potentially) expensive endpoint.

## Related PRs

https://github.com/AleoNet/snarkOS/pull/3219

CI run is [here](https://app.circleci.com/pipelines/github/ProvableHQ/snarkOS/14032/workflows/a9c3acc0-ea17-4a08-84ce-f658ca66d2ad).

### DIFF
--- a/node/rest/src/routes.rs
+++ b/node/rest/src/routes.rs
@@ -263,26 +263,24 @@ impl<N: Network, C: ConsensusStorage<N>, R: Routing<N>> Rest<N, C, R> {
         let height = rest.ledger.latest_height();
 
         // Retrieve all the mapping values from the mapping.
-        let mapping_values = match tokio::task::spawn_blocking(move || {
-            rest.ledger.vm().finalize_store().get_mapping_confirmed(id, name)
-        })
-        .await
+        match tokio::task::spawn_blocking(move || rest.ledger.vm().finalize_store().get_mapping_confirmed(id, name))
+            .await
         {
-            Ok(Ok(mapping_values)) => mapping_values,
-            Ok(Err(err)) => return Err(RestError(format!("Unable to read mapping - {err}"))),
-            Err(err) => return Err(RestError(format!("Unable to read mapping - {err}"))),
-        };
+            Ok(Ok(mapping_values)) => {
+                // Check if metadata is requested and return the mapping with metadata if so.
+                if metadata.metadata.unwrap_or(false) {
+                    return Ok(ErasedJson::pretty(json!({
+                        "data": mapping_values,
+                        "height": height,
+                    })));
+                }
 
-        // Check if metadata is requested and return the mapping with metadata if so.
-        if metadata.metadata.unwrap_or(false) {
-            return Ok(ErasedJson::pretty(json!({
-                "data": mapping_values,
-                "height": height,
-            })));
+                // Return the full mapping without metadata.
+                Ok(ErasedJson::pretty(mapping_values))
+            }
+            Ok(Err(err)) => Err(RestError(format!("Unable to read mapping - {err}"))),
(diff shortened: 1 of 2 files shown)


node/rest/src/routes.rs
```diff
@@ -247,6 +248,44 @@ impl<N: Network, C: ConsensusStorage<N>, R: Routing<N>> Rest<N, C, R> {
...
+    // GET /<network>/program/{programID}/mapping/{mappingName}?all={true}&metadata={true}
+    pub(crate) async fn get_mapping_values(
+        State(rest): State<Self>,
+        Path((id, name)): Path<(ProgramID<N>, Identifier<N>)>,
+        metadata: Query<Metadata>,
+    ) -> Result<ErasedJson, RestError> {
+        // If the `all` query parameter is set, return the full mapping.
+        if metadata.all.unwrap_or(false) {
+            // Retrieve the latest height.
+            let height = rest.ledger.latest_height();
```
ljedrz: nit: the height can be looked up after the mapping values, just in case that query fails

ljedrz: also, since that one is only needed if `metadata.metadata` is `true`, it can happen directly in the applicable `ErasedJson`

node/rest/src/routes.rs
```diff
@@ -247,6 +248,44 @@ impl<N: Network, C: ConsensusStorage<N>, R: Routing<N>> Rest<N, C, R> {
...
     }
 
+    // GET /<network>/program/{programID}/mapping/{mappingName}?all={true}&metadata={true}
+    pub(crate) async fn get_mapping_values(
+        State(rest): State<Self>,
+        Path((id, name)): Path<(ProgramID<N>, Identifier<N>)>,
+        metadata: Query<Metadata>,
+    ) -> Result<ErasedJson, RestError> {
+        // If the `all` query parameter is set, return the full mapping.
+        if metadata.all.unwrap_or(false) {
```
ljedrz: nit: instead of having most of the body in a conditional block, this function could begin with the early error return that's currently at the very end, and then continue without the extra block

ljedrz (dismissed):
Left a few comments, LGTM otherwise :+1:.

node/rest/src/routes.rs
raychu86: I brought it above because the `spawn_blocking` call consumes the `rest` object. 

So it was either move the height call up or to clone the `vm` or `finalize_store` objects (which are Arcs).

node/rest/src/routes.rs
raychu86: Resolved with [87f3bbd](https://github.com/AleoNet/snarkOS/pull/3365/commits/87f3bbd57ef4540b9ce0f99bcac7e0ccfbb703b1)

niklaslong (approved):

node/rest/src/routes.rs
ljedrz: Ah, makes sense; an `Arc` clone is probably a tiny bit faster, but it's unlikely to make any difference in this case.

vicsn (approved):

PR #3366, opened 2024-07-18
[Feature] Introduce IP-level bans for connection spam targeting validators

ljedrz:
This PR primarily targets https://github.com/AleoNet/snarkOS/issues/3311, but my intention is for the new setup to be extensible based on potential future needs (e.g. detecting and banning for message spam).

The 1st commit alters the (currently mostly unused) low-level `KnownPeers` collection to work on IPs only (instead of IP+port pairs) and extends the related `Stats` object with a `timestamp` field which is updated whenever a (low-level) connection is established.

The new IP-level bans are introduced only at the (higher-level) handshake level for the following reasons:
- the BFT setup can use and configure them separately from the `Router` setup
- the `Heartbeat` can be used to clean up expired bans
- the overall setup is simpler that way (e.g. the `Tcp` doesn't need to do any housekeeping)

The downside is that several concurrent low-level connection attempts may still be accepted even post-IP-ban. However, this would also (though to a lesser extent) be the case if the ban check was moved "to the lower level", as we would still have to accept an inbound connection before being able to check its IP. That being said, these are relatively lightweight and still subject to peer limits.

The ban-related `const`s were picked arbitrarily and we may choose any other values instead. Also, this new feature is disabled for tests and `--dev` runs (since they almost exclusively involve a single localhost IP).

CI run is [here](https://app.circleci.com/pipelines/github/ProvableHQ/snarkOS?branch=feat%2Fper_ip_bans_ci)

### DIFF
--- a/node/bft/src/gateway.rs
+++ b/node/bft/src/gateway.rs
@@ -64,7 +64,14 @@ use futures::SinkExt;
 use indexmap::{IndexMap, IndexSet};
 use parking_lot::{Mutex, RwLock};
 use rand::seq::{IteratorRandom, SliceRandom};
-use std::{collections::HashSet, future::Future, io, net::SocketAddr, sync::Arc, time::Duration};
+use std::{
+    collections::{HashMap, HashSet},
+    future::Future,
+    io,
+    net::{IpAddr, SocketAddr},
+    sync::Arc,
+    time::{Duration, Instant},
+};
 use tokio::{
     net::TcpStream,
     sync::{oneshot, OnceCell},
@@ -87,6 +94,12 @@ const RESTRICTED_INTERVAL: i64 = (MAX_CONNECTION_ATTEMPTS as u64 * MAX_BATCH_DEL
 const MIN_CONNECTED_VALIDATORS: usize = 175;
 /// The maximum number of validators to send in a validators response event.
 const MAX_VALIDATORS_TO_SEND: usize = 200;
+
+/// The minimum permitted interval between connection attempts for an IP; anything shorter is considered malicious.
+#[cfg(not(any(test, feature = "test")))]
+const MIN_CONNECTION_INTERVAL_IN_SECS: u64 = 10;
+/// The amount of time an IP address is prohibited from connecting.
+const IP_BAN_TIME_IN_SECS: u64 = 30;
 
 /// Part of the Gateway API that deals with networking.
 /// This is a separate trait to allow for easier testing/mocking.
@@ -119,6 +132,8 @@ pub struct Gateway<N: Network> {
     /// prevent simultaneous "two-way" connections between two peers (i.e. both nodes simultaneously
     /// attempt to connect to each other). This set is used to prevent this from happening.
     connecting_peers: Arc<Mutex<IndexSet<SocketAddr>>>,
+    /// The set of banned IP addresses.
+    banned_ips: Arc<RwLock<HashMap<IpAddr, Instant>>>,
     /// The primary sender.
     primary_sender: Arc<OnceCell<PrimarySender<N>>>,
     /// The worker senders.
@@ -160,6 +175,7 @@ impl<N: Network> Gateway<N> {
             trusted_validators: trusted_validators.iter().copied().collect(),
             connected_peers: Default::default(),
             connecting_peers: Default::default(),
+            banned_ips: Default::default(),
             primary_sender: Default::default(),
             worker_senders: Default::default(),
             sync_sender: Default::default(),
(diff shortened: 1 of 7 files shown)


ljedrz:
[CI run link](https://app.circleci.com/pipelines/github/ProvableHQ/snarkOS?branch=feat%2Fper_ip_bans-ci)

vicsn:
Thx for making this! @kpandl would you like to take the first pass review in the coming week?

kpandl:
Generally looks good to me. Can you explain the setting of the const variables (`MIN_CONNECTION_INTERVAL_IN_SECS` and `IP_BAN_TIME_IN_SECS`) or do you know recommended values from other protocols? I understand they can be set arbitrarily, 30 could seem a bit low for `IP_BAN_TIME_IN_SECS` to me.

I've seen [24 hour bans in the Bitcoin protocol](https://par.nsf.gov/servlets/purl/10407054#:~:text=With%20the%20ban%2Dscore%20mechanism,get%20banned%20for%2024%20hours.), but I've read [other sources](https://github.com/MrChrisJ/fullnode/blob/master/Setup_Guides/bitcoin.conf) that suggest it could primarily limit DOS attacks over Tor.

Either way, setting the values in a config could also be an interesting feature in the long term.

vicsn (approved):

ljedrz:
@kpandl the values were chosen semi-arbitrarily; basically any limit imposed on the number of connection attempts will suffice to avoid a "flood", but we may decide to be as harsh as we like.

Technically, `MIN_CONNECTION_INTERVAL_IN_SECS` should be equal to or (IMO preferably) smaller than the heartbeat interval (so we don't consider legitimate heartbeat re-connection attempts malicious), and `IP_BAN_TIME_IN_SECS` should be larger than the heartbeat interval, and it can be arbitrarily large (in order for it to be discouraging in addition to just limiting).

kpandl (approved):

snarkworld (dismissed):
Additional testing/validation needed

node/bft/src/gateway.rs
```diff
@@ -1105,9 +1141,29 @@ impl<N: Network> OnConnect for Gateway<N> {
...
+            // If the IP is already banned, update the ban timestamp and reject the connection.
+            if self.is_ip_banned(peer_addr.ip()) {
+                self.update_ip_ban(peer_addr.ip());
+                trace!("{CONTEXT} Gateway rejected a connection request from banned IP '{}'", peer_addr.ip());
+                return Err(error(format!("'{}' is a banned IP address", peer_addr.ip())));
+            }
+
+            // Check the previous low-level connection timestamp.
+            if let Some(peer_stats) = self.tcp.known_peers().get(peer_addr.ip()) {
+                if peer_stats.timestamp().elapsed().as_secs() <= MIN_CONNECTION_INTERVAL_IN_SECS {
```
vicsn, line 1159: @Meshiest wrote: "in my test environment this is permanently banning my local validators (they're connecting to eachother while booting up)"

@Meshiest can you describe your test in more detail? Most importantly how many validators, from genesis or not, and what rough machine specs? That would be useful for @ljedrz to reproduce and chime in on an alternative design or constants.

Meshiest: I have an entirely local 4 validator, 4 bootstrap+core clients, and 4 fringe client network for testing another change & made a few snarkos changes to support 127.x.y.z IPs so the peers would have different addresses while still being local and having default ports (can probably push these somewhere). I also updated the bootstrap peers to be the core clients

When booting the nodes all at the same time, validators are starved for peers and are the initiators for all their requests. This results in the validators getting banned both by other validators as well as their own core clients in the non-gateway concurrent con check. 

I'm not certain my IP changes to support other local addresses are contributing but they are part of the equation and it could be related to the other testing I was doing at the time

ljedrz: @Meshiest just to rule out other possible issues, could you try starting the nodes one-by-one and see if this avoids the bans? I agree that we should also support concurrent launch, this is just to double-check that it is this edge case.

Meshiest: Rebuilding via:
1. checkout #3366
2. cherry-pick #3397 (I'm running all my nodes locally)
3. merge mainnet

And I wasn't able to reproduce the parallel booting ban, though after more testing and rapidly restarting nodes I wasn't able to achieve any bans. Not entirely sure what's going on. Out of time until later today (at least 12 hours)

raychu86 (approved):

raychu86:
Can be closed with - https://github.com/AleoNet/snarkOS/pull/3422.

PR #3367, opened 2024-07-20
[Fix] Add checksum to `TransmissionID`

raychu86:
## Motivation

This PR is the sister PR of https://github.com/AleoNet/snarkVM/pull/2521, which introduces the `TransmissionChecksum` and adds additional solution guards during block creation. In the snarkOS world, additional checks are added to enforce that incoming peer transmissions have the correct `TransmissionChecksum`. This prevents malicious validators from performing a malformed solution/transaction attack.

The CI is run [here](https://app.circleci.com/pipelines/github/ProvableHQ/snarkOS?branch=fix%2Ftransmission-checksum-ci).

## Related PRs

https://github.com/AleoNet/snarkVM/pull/2521

### DIFF
--- a/node/bft/src/bft.rs
+++ b/node/bft/src/bft.rs
@@ -597,10 +597,34 @@ impl<N: Network> BFT<N> {
             if !IS_SYNCING {
                 // Initialize a map for the deduped transmissions.
                 let mut transmissions = IndexMap::new();
+                // Initialize a map for the deduped transaction ids.
+                let mut seen_transaction_ids = IndexSet::new();
+                // Initialize a map for the deduped solution ids.
+                let mut seen_solution_ids = IndexSet::new();
                 // Start from the oldest leader certificate.
                 for certificate in commit_subdag.values().flatten() {
                     // Retrieve the transmissions.
                     for transmission_id in certificate.transmission_ids() {
+                        // If the transaction ID or solution ID already exists in the map, skip it.
+                        // Note: This additional check is done to ensure that we do not include duplicate
+                        // transaction IDs or solution IDs that may have a different transmission ID.
+                        match transmission_id {
+                            TransmissionID::Solution(solution_id, _) => {
+                                // If the solution already exists, skip it.
+                                if seen_solution_ids.contains(&solution_id) {
+                                    continue;
+                                }
+                            }
+                            TransmissionID::Transaction(transaction_id, _) => {
+                                // If the transaction already exists, skip it.
+                                if seen_transaction_ids.contains(transaction_id) {
+                                    continue;
+                                }
+                            }
+                            TransmissionID::Ratification => {
+                                bail!("Ratifications are currently not supported in the BFT.")
+                            }
+                        }
                         // If the transmission already exists in the map, skip it.
                         if transmissions.contains_key(transmission_id) {
                             continue;
@@ -619,6 +643,16 @@ impl<N: Network> BFT<N> {
                                 certificate.round()
                             );
                         };
+                        // Insert the transaction ID or solution ID into the map.
+                        match transmission_id {
+                            TransmissionID::Solution(id, _) => {
+                                seen_solution_ids.insert(id);
+                            }
+                            TransmissionID::Transaction(id, _) => {
+                                seen_transaction_ids.insert(id);
+                            }
+                            TransmissionID::Ratification => {}
+                        }
                         // Add the transmission to the set.
                         transmissions.insert(*transmission_id, transmission);
                     }
--- a/node/bft/src/bft.rs
+++ b/node/bft/src/bft.rs
@@ -602,7 +602,7 @@ impl<N: Network> BFT<N> {
                     // Retrieve the transmissions.
                     for transmission_id in certificate.transmission_ids() {
                         // If the transmission already exists in the map, skip it.
-                        if transmissions.contains_key(transmission_id.0) {
+                        if transmissions.contains_key(transmission_id) {
                             continue;
                         }
                         // If the transmission already exists in the ledger, skip it.
@@ -619,7 +619,7 @@ impl<N: Network> BFT<N> {
                             );
                         };
                         // Add the transmission to the set.
-                        transmissions.insert(*transmission_id.0, transmission);
+                        transmissions.insert(*transmission_id, transmission);
                     }
                 }
                 // Trigger consensus, as this will build a new block for the ledger.
--- a/node/bft/src/bft.rs
+++ b/node/bft/src/bft.rs
@@ -602,7 +602,7 @@ impl<N: Network> BFT<N> {
                     // Retrieve the transmissions.
                     for transmission_id in certificate.transmission_ids() {
                         // If the transmission already exists in the map, skip it.
-                        if transmissions.contains_key(transmission_id) {
+                        if transmissions.contains_key(transmission_id.0) {
                             continue;
                         }
                         // If the transmission already exists in the ledger, skip it.
@@ -619,7 +619,7 @@ impl<N: Network> BFT<N> {
                             );
                         };
                         // Add the transmission to the set.
-                        transmissions.insert(*transmission_id, transmission);
+                        transmissions.insert(*transmission_id.0, transmission);
                     }
                 }
                 // Trigger consensus, as this will build a new block for the ledger.
--- a/node/bft/src/bft.rs
+++ b/node/bft/src/bft.rs
@@ -613,8 +613,9 @@ impl<N: Network> BFT<N> {
                         // Retrieve the transmission.
                         let Some(transmission) = self.storage().get_transmission(*transmission_id) else {
                             bail!(
-                                "BFT failed to retrieve transmission '{}' from round {}",
+                                "BFT failed to retrieve transmission '{}.{}' from round {}",
                                 fmt_id(transmission_id),
+                                fmt_id(transmission_id.checksum().unwrap_or_default()).dimmed(),
                                 certificate.round()
                             );
                         };
--- a/node/bft/src/primary.rs
+++ b/node/bft/src/primary.rs
@@ -522,14 +522,39 @@ impl<N: Network> Primary<N> {
                     }
                     // Check the transmission is still valid.
                     match (id, transmission.clone()) {
-                        (TransmissionID::Solution(solution_id), Transmission::Solution(solution)) => {
+                        (TransmissionID::Solution(solution_id, checksum), Transmission::Solution(solution)) => {
+                            // Ensure the checksum matches.
+                            match solution.to_checksum::<N>() {
+                                Ok(solution_checksum) if solution_checksum == checksum => (),
+                                _ => {
+                                    trace!(
+                                        "Proposing - Skipping solution '{}' - Checksum mismatch",
+                                        fmt_id(solution_id)
+                                    );
+                                    continue 'inner;
+                                }
+                            }
                             // Check if the solution is still valid.
                             if let Err(e) = self.ledger.check_solution_basic(solution_id, solution).await {
                                 trace!("Proposing - Skipping solution '{}' - {e}", fmt_id(solution_id));
                                 continue 'inner;
                             }
                         }
-                        (TransmissionID::Transaction(transaction_id), Transmission::Transaction(transaction)) => {
+                        (
+                            TransmissionID::Transaction(transaction_id, checksum),
+                            Transmission::Transaction(transaction),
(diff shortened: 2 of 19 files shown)


node/bft/src/primary.rs
```diff
@@ -522,14 +522,39 @@ impl<N: Network> Primary<N> {
                     }
                     // Check the transmission is still valid.
                     match (id, transmission.clone()) {
-                        (TransmissionID::Solution(solution_id), Transmission::Solution(solution)) => {
+                        (TransmissionID::Solution(solution_id, checksum), Transmission::Solution(solution)) => {
```
vicsn, line 525: I think these checksum checks are superfluous. We already confirm the checksum when we receive the transmission. And in contrast to some other added superfluous checks, there was no precedent to check if the solution_id was consistent. So I suggest removing these.

node/bft/src/primary.rs
```diff
@@ -1212,8 +1237,13 @@ impl<N: Network> Primary<N> {
...
         self.spawn(async move {
             while let Some((solution_id, solution, callback)) = rx_unconfirmed_solution.recv().await {
+                // Compute the checksum for the solution.
+                let Ok(checksum) = solution.to_checksum::<N>() else {
+                    error!("Failed to compute the checksum for the unconfirmed solution");
+                    continue;
+                };
                 // Compute the worker ID.
-                let Ok(worker_id) = assign_to_worker(solution_id, self_.num_workers()) else {
+                let Ok(worker_id) = assign_to_worker((solution_id, checksum), self_.num_workers()) else {
```
vicsn, line 1246: A validator only stores and tries to verify unique `solution_id`s, it will ignore duplicates. So for assignment to workers, we can ignore the checksum. 

Because of https://github.com/AleoNet/snarkVM/pull/2520 there is at least no censorship risk for solutions.

And we can't remove the censorship risk for transactions the same way anyway, because it would entail regenerating a proof.

node/bft/src/worker.rs
```diff
@@ -318,18 +318,20 @@ impl<N: Network> Worker<N> {
         // Construct the transmission.
         let transmission = Transmission::Solution(solution.clone());
         // Compute the checksum.
-        let checksum = solution.to_checksum();
+        let checksum = solution.to_checksum::<N>()?;
```
vicsn, line 328: I know I added this before, but for the same reason as my comment above, I think we should discard these checksum calculations and just stick to tracking by {solution, transaction}_id.

node/consensus/src/lib.rs
```diff
@@ -278,11 +285,15 @@ impl<N: Network> Consensus<N> {
 impl<N: Network> Consensus<N> {
     /// Adds the given unconfirmed solution to the memory pool.
     pub async fn add_unconfirmed_solution(&self, solution: Solution<N>) -> Result<()> {
+        // Calculate the transmission checksum.
+        let checksum = Data::<Solution<N>>::Buffer(solution.to_bytes_le()?.into()).to_checksum::<N>()?;
```
vicsn, line 289: Slight efficiency improvement, the checksum calculation can be propagated one level higher. This function is called from two places, in one of those `fn unconfirmed_solution`, we can call `serialized.solution.to_checksum()`

node/bft/src/bft.rs
```diff
@@ -602,7 +602,7 @@ impl<N: Network> BFT<N> {
                     // Retrieve the transmissions.
                     for transmission_id in certificate.transmission_ids() {
                         // If the transmission already exists in the map, skip it.
-                        if transmissions.contains_key(transmission_id.0) {
+                        if transmissions.contains_key(transmission_id) {
```
vicsn, line 629: The VM requires {transaction, solution}_ids to be unique, and maintains the same order as this for-loop, so its more efficient to skip transmissions by id here already.

Maybe noting that as a comment would also be helpful. O:)

node/bft/src/primary.rs
raychu86: Workers are assigned based on `Into <TransmissionID>`, which is why we need the tuple there. 

An alternative, is to update `assign_to_worker` to intake a `ToBytes` object instead, which allows us to pass in just the solution ID or transaction ID.

node/bft/src/worker.rs
raychu86: That would require overhauling the `pending` and  `ready` objects and the underlying `fn contains_transmission`.

node/bft/src/primary.rs
raychu86: The subsequent calls to `check_solution_basic` and `check_transaction_basic` do enforce that the `{transaction/solution}ID` does match the one in the object. So for consistency, I added the checksum enforcement here as well.

node/bft/src/bft.rs
raychu86: I'm not sure `transmission_id.0` even compiles. In addition, the subsequent call to `tx_consensus_subdag.send` requires the full transmission ID, including the checksum.

The checksum here is already provided, so this does not incur any additional overhead.

node/bft/src/primary.rs
vicsn: Thank you for clarifying.

I see my above comment is wrong. We would significantly reduce (but not eliminate) risk of transmission censorship by just using the full `transmissionID` to determine whether a transmission was seen already. So I think we should rather use the full `transmissionID` as key in the `*_queue` and `seen_*`  caches in `node/consensus/src/lib.rs`, then we can just pass the full TransmissionID along so we don't have to recompute it here.

If you think that's too much for this PR, we can leave this and the below comment for future work.

node/bft/src/primary.rs
vicsn: I see, how about moving the checksum check into `snarkOS::check_{transaction, solution}_basic`? I see it would require changing the `LedgerService` trait slightly by passing in TransmissionId or the checksum argument. What we gain is if we ever want to remove those checks for improved speed/readability, its in one place.

node/bft/src/worker.rs
vicsn: If you think that's too much for this PR, we can leave this and the above comment for future work.

node/bft/src/worker.rs
```diff
@@ -317,17 +324,26 @@ impl<N: Network> Worker<N> {
...
+        // Construct the transmission ID.
+        let transmission_id = TransmissionID::Solution(solution_id, checksum);
         // Remove the solution ID from the pending queue.
-        self.pending.remove(solution_id, Some(transmission.clone()));
+        self.pending.remove(transmission_id, Some(transmission.clone()));
         // Check if the solution exists.
-        if self.contains_transmission(solution_id) {
-            bail!("Solution '{}' already exists.", fmt_id(solution_id));
+        if self.contains_transmission(transmission_id) {
+            bail!("Solution '{}.{}' already exists.", fmt_id(solution_id), fmt_id(checksum));
```
ljedrz:
```suggestion
            bail!("Solution '{}.{}' already exists.", fmt_id(solution_id), fmt_id(checksum).dimmed());
```
nit (this is dimmed for `Transaction`)

node/bft/src/worker.rs
```diff
@@ -340,17 +356,26 @@ impl<N: Network> Worker<N> {
...
         self.ledger.check_transaction_basic(transaction_id, transaction).await?;
         // Adds the transaction to the ready queue.
-        if self.ready.insert(&transaction_id, transmission) {
-            trace!("Worker {} - Added unconfirmed transaction '{}'", self.id, fmt_id(transaction_id));
+        if self.ready.insert(transmission_id, transmission) {
+            trace!(
+                "Worker {}.{} - Added unconfirmed transaction '{}'",
+                self.id,
+                fmt_id(transaction_id),
+                fmt_id(checksum)
```
ljedrz:
```suggestion
                fmt_id(checksum).dimmed()
```
nit: this is dimmed for `Solution`

node/bft/src/worker.rs
raychu86: Ideally, we try to minimize changes, especially since the introduction of checksums seems to already touch so many surfaces.

node/bft/src/primary.rs
raychu86: I thought about that and decided against for a few reasons:

1. Keep code change as minimal as possible.
2. The entry points for `check_*_basic` is usually via raw transactions or solutions from peer messaging rather than via transmissions, so adding a checksum check here may actually be more expensive overall.
3. Optimizations can be added in the future, the current priority is safety and simplicity. Especially since the checksum checks here are fairly cheap.

LMK if that makes sense or there are holes in my logic.

node/bft/src/primary.rs
raychu86: Same as my other comment, I'd prefer to keep the changeset minimal and only change what is required to support the `TransmissionChecksum`s. We can (and should) revisit optimizations after we are sure this is safe and works properly.

vicsn (dismissed):
LGTM, optimizations can be tackled in a future PR

ljedrz (dismissed):
LGTM

niklaslong (dismissed):

node/bft/src/bft.rs
```diff
@@ -613,8 +613,9 @@ impl<N: Network> BFT<N> {
                         // Retrieve the transmission.
                         let Some(transmission) = self.storage().get_transmission(*transmission_id) else {
                             bail!(
-                                "BFT failed to retrieve transmission '{}' from round {}",
+                                "BFT failed to retrieve transmission '{}.{}' from round {}",
                                 fmt_id(transmission_id),
+                                fmt_id(transmission_id.checksum().unwrap_or_default()).dimmed(),
```
Meshiest, line 642: if [my comment](https://github.com/AleoNet/snarkVM/pull/2521#discussion_r1690551279) on the snarkVM side is implemented (base16 or bech32 encoding), this would need to be updated

Meshiest (dismissed):

gluax (dismissed):

ljedrz (dismissed):

raychu86:
Note: [64e58a0](https://github.com/AleoNet/snarkOS/pull/3367/commits/64e58a053b92a1a6ce91452cb88316cc31a8c781) was added to further enforce the uniqueness of solution/transaction IDs that is committed to a block. This is also done on the snarkVM level with -  [a6d175a](https://github.com/AleoNet/snarkVM/pull/2521/commits/a6d175aa0b4a4f738cae4e44aa9405a028041216)

node/bft/src/bft.rs
```diff
@@ -597,10 +597,34 @@ impl<N: Network> BFT<N> {
             if !IS_SYNCING {
                 // Initialize a map for the deduped transmissions.
                 let mut transmissions = IndexMap::new();
+                // Initialize a map for the deduped transaction ids.
+                let mut seen_transaction_ids = IndexSet::new();
+                // Initialize a map for the deduped solution ids.
+                let mut seen_solution_ids = IndexSet::new();
```
ljedrz, line 603:
```suggestion
                let mut seen_transaction_ids = HashSet::new();
                // Initialize a map for the deduped solution ids.
                let mut seen_solution_ids = HashSet::new();
```
nit: if we're only interested in ensuring uniqueness and the order of these collections is insignificant, a plain `HashSet` can be used instead.

node/bft/src/bft.rs
```diff
@@ -597,10 +597,34 @@ impl<N: Network> BFT<N> {
...
+                        match transmission_id {
+                            TransmissionID::Solution(solution_id, _) => {
+                                // If the solution already exists, skip it.
+                                if seen_solution_ids.contains(&solution_id) {
+                                    continue;
+                                }
+                            }
+                            TransmissionID::Transaction(transaction_id, _) => {
+                                // If the transaction already exists, skip it.
+                                if seen_transaction_ids.contains(transaction_id) {
```
ljedrz, line 620:
```suggestion
                                if !seen_solution_ids.contains(&solution_id) {
                                    continue;
                                }
                            }
                            TransmissionID::Transaction(transaction_id, _) => {
                                // If the transaction already exists, skip it.
                                if !seen_transaction_ids.contains(transaction_id) {
```
if the item already exists in the set, `false` is returned

ljedrz: nvm, I mixed it up with `insert`

ljedrz (dismissed):

snarkworld (dismissed):
LGTM

snarkworld (approved):
LGTM

vicsn (approved):

PR #3368, opened 2024-07-25
Skip devnet script client generation when total_clients=0

vicsn:
## Motivation

When starting a devnet with num_clients=0, the current script would still launch a client.

## Test Plan

Ran it locally.


niklaslong (approved):
LGTM (and tested it locally as well) 👍

raychu86 (approved):
LGTM

PR #3369, opened 2024-07-25
[Fix] Cache block requests on the gateway

### LINKED ISSUE #3324: [Proposal] The bft module's gateway should check block request and block response from peers to avoid spam
## 💥 Proposal

Issue https://github.com/AleoNet/snarkOS/issues/3315 points out that malicious validator can DDoS other validator by sending block response spam.
To solve this problem, in addition to putting the deserialization logic into a separate rayon thread https://github.com/AleoNet/snarkOS/pull/3316, we should also:
1. Check the request frequency of block request
2. Check whether the block response has a corresponding block request

In fact, node/router has already implemented the above two checks:
https://github.com/AleoNet/snarkOS/blob/5d4de0286964a72ecd2b0c4ace30a938ef269086/node/router/src/inbound.rs#L87-L92
https://github.com/AleoNet/snarkOS/blob/5d4de0286964a72ecd2b0c4ace30a938ef269086/node/router/src/inbound.rs#L111-L114

niklaslong:
This PR implements caching for block requests so they can be rate limited and matched to inbound block responses. I have also refactored some verbose branching logic to use `match` statements in the second commit. It can be dropped, if a smaller diff is preferred. 

I've opted to clear the outbound block request cache in the same way as the validator requests (https://github.com/AleoNet/snarkOS/pull/3254). Time-based expiry isn't necessary as the committee is bounded and the collection is cleared on peer disconnect. Inbound requests use a time-based expiry.

Fixes #3324.

### DIFF
--- a/node/bft/src/gateway.rs
+++ b/node/bft/src/gateway.rs
@@ -570,6 +570,11 @@ impl<N: Network> Gateway<N> {
             if num_events >= self.max_cache_duplicates() {
                 return Ok(());
             }
+        } else if matches!(&event, &Event::BlockRequest(_)) {
+            let num_events = self.cache.insert_inbound_block_request(peer_ip, CACHE_REQUESTS_INTERVAL);
+            if num_events >= self.max_cache_duplicates() {
+                return Ok(());
+            }
         }
         trace!("{CONTEXT} Received '{}' from '{peer_ip}'", event.name());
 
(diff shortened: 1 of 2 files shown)


Pjrich1313 (commented):
Fix

vicsn (approved):

vicsn:
Operating solid

Screenshot 2024-10-02 at 16 00 20

PR #3370, opened 2024-07-30
Rest: add /{network}/epochProgram/:epoch_number

ghostant-1017:


node/rest/Cargo.toml
```diff
@@ -70,6 +70,10 @@ rev = "d170a9f"
 default-features = false
 optional = true
 
+[dependencies.snarkvm-ledger-puzzle-epoch]
```
raychu86, line 73: Ideally we expose this via `snarkvm::prelude` or some other reference to avoid having to reimport a snarkVM crate here. Especially one that will require updating the `rev`.

node/rest/src/routes.rs
```diff
@@ -23,6 +23,7 @@ use indexmap::IndexMap;
 use rayon::prelude::*;
 use serde::{Deserialize, Serialize};
 use serde_json::json;
+use snarkvm_ledger_puzzle_epoch::EpochProgram;
```
raychu86: nit: Move this up to the import section above.

node/rest/src/routes.rs
```diff
@@ -209,6 +210,20 @@ impl<N: Network, C: ConsensusStorage<N>, R: Routing<N>> Rest<N, C, R> {
         }
     }
 
+    // GET /<network>/epoch_program/{epochNumber}
```
raychu86, line 213: How does the user know what epoch number is the latest?

raychu86:
What is the intention of this PR? Please provide a description.

ghostant-1017:
> What is the intention of this PR? Please provide a description.

There is a scenario where EpochProgram needs to be distributed from the Server to the Provers, we actually have a fork, and it seems reasonable to provide as much information as possible in the snarkOS RPC, so I propose to merge it.

node/rest/src/routes.rs
ghostant-1017: Users can use `/latest/height` or `/latest/block` to get the latest block height and calculate the `epoch_number`

node/rest/Cargo.toml
ghostant-1017: Shall I create another PR for snarkVM to reimport the crate?

node/rest/src/routes.rs
ghostant-1017: Done

node/rest/Cargo.toml
vicsn: Apologies for the very late reply. If this is still of interest, yes please 🙏

PR #3373, opened 2024-08-05
comments about quorum threshold changed from (2f + 1) to (N - f)

bendyarm:
## Motivation

In `snarkvm_ledger_committee`, `quorum_threshold` is correctly calculated as `N - f`.
Since `N` is not always `3f + 1` (it could be `3f + 2` or `3f + 3`),
the quorum threshold is not always `2f + 1`.

NOTE: no need for CI to pass since only comments were changed.

### DIFF
--- a/node/bft/README.md
+++ b/node/bft/README.md
@@ -18,7 +18,8 @@ Each round runs until one of two conditions is met:
 
 #### Advancing Rounds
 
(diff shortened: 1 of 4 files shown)


vicsn (approved):

raychu86 (approved):
LGTM

PR #3378, opened 2024-08-09
feat: test targets

joske:
## Motivation

Add feature flag to be able to lower coinbase and proof targets for testing purposes. This feature flag will lower the proof and coinbase targets to 8 and 32 respectively.

The feature flag needed to be cascaded to all dependencies of snarkVM to be sure everything uses same constants.

To use, compile with `--features test_targets`
e.g.
```
cargo install --locked --features test_targets --path .
```

## Test Plan

tested by using tx-cannon and adding println to verify correct targets.

[CI link](https://app.circleci.com/pipelines/github/ProvableHQ/snarkOS/13686/workflows/2efc80fc-9c83-47cf-88c0-67b27ee9ab53)

## Related PRs

Must first merge https://github.com/AleoNet/snarkVM/pull/2529 and the snarkVM revision must be updated in snarkOS in 2 places before merging this.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,9 +45,9 @@ version = "=0.1.24"
 default-features = false
 
 [workspace.dependencies.snarkvm]
-#path = "../snarkVM"
-git = "https://github.com/AleoNet/snarkVM.git"
-rev = "be171ce" # If this is updated, the rev in `node/rest/Cargo.toml` must be updated as well.
+git = "https://github.com/joske/snarkVM.git"
+branch = "feat/test-targets"
+# rev = "be171ce" # If this is updated, the rev in `node/rest/Cargo.toml` must be updated as well.
 #version = "=0.16.18"
 features = [ "circuit", "console", "rocks" ]
 
@@ -58,6 +58,7 @@ path = "snarkos/main.rs"
 [features]
 metrics = [ "snarkos-node-metrics", "snarkos-node/metrics" ]
 history = [ "snarkos-node/history" ]
+test_targets = [ "snarkos-cli/test_targets" ]
 
 [dependencies.anyhow]
 version = "1.0.79"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -136,9 +136,9 @@ dependencies = [
 
 [[package]]
 name = "anstream"
-version = "0.6.14"
+version = "0.6.15"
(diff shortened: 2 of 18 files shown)


vicsn (dismissed):
LGTM if you update the Cargo.toml to the rev of https://github.com/AleoNet/snarkVM/pull/2529 (which is just for illustrative purposes, as the maintainer will have to update it anyway)

raychu86:
Does this apply to compiled `snarkOS` binaries via `cargo install --path .`? What does that process look like? Additional documentation in the PR description/in the code would be helpful to know the implications.

joske:
@raychu86 I've updated the description. Let me know if still not clear.

raychu86:
So if `cargo install --locked --features test_targets --path .` is used, the `snarkOS` binary will always have the `test_targets` version of the coinbase targets? We should have some failsafes where the node fails to boot-up in production mode if that is the case. 

We need some sort of enforcement that the `test_targets` is being used only for devnets via `--dev`

Cargo.toml
```diff
@@ -45,9 +45,9 @@ version = "=0.1.24"
 default-features = false
 
 [workspace.dependencies.snarkvm]
-#path = "../snarkVM"
-git = "https://github.com/AleoNet/snarkVM.git"
-rev = "be171ce" # If this is updated, the rev in `node/rest/Cargo.toml` must be updated as well.
+git = "https://github.com/joske/snarkVM.git"
```
howardwu, line 48: **Warning**: snarkOS will no longer be using `AleoNet`'s implementation of snarkVM after this PR.

**Proceed with caution.**

zkxuerb: @joske I'm guessing this was changed for testing purposes and we can change it back to `AleoNet`'s snarkVM ?

vicsn: And merge AleoNet/staging + run `cargo fmt` 🙏

vicsn:
Superseded by https://github.com/AleoNet/snarkOS/pull/3424 to resolve merge conflict

Cargo.toml
joske: Yes, this was because the snarkVM PR needs to be merged first. It was always intended that it be changed to an Aleonet snarkVM repo before merging this.

vicsn:
NOTE: this was not actually merged, must be a bug in Github since the branchname is the same as the superseding PR.

Issue #3377, opened 2024-08-06
[Bug] imprecision in test function is_round_reached()


bendyarm:
## 🐛 Bug Report

This bug does not affect mainnet.  It is a bug in a test function
that may cause false positives or negatives from the tests that use it.
However, there are other more complete test suites that
are being run and passed, so this bug is minor.

The bug is in this function:
```
    // Checks if at least 2f + 1 nodes have reached the given round.
    pub fn is_round_reached(&self, round: u64) -> bool {
        let quorum_threshold = self.validators.len() / 2 + 1;
        self.validators.values().filter(|v| v.primary.current_round() >= round).count() >= quorum_threshold
    }
```
There are 12 tests that use this test function [permalink here](https://github.com/AleoNet/snarkOS/blob/22510524b594d1822763e17328a60078a2c2fede/node/bft/tests/common/primary.rs#L282), but this function is not doing what it claims.

I presume validators here are equal (not stake-weighted).
If there are N validators, `f` should be
`(N - 1)/3` (where `/` means integer division rounding down),
and quorum threshold should. be `N - f` (not `2f + 1` as mentioned in the comment),
which can be computed from N as `2N/3 + 1` (same meaning of `/`, integer division).

For example:
For 4 validators, this function will return 3, which is correct.
For 5 validators, this function will return 3, which is incorrect.
When there are 5 validators, the quorum threshold should be 4.

See also the definition of `quorum_threshold()` in snarkVM
ledger/committee/src/lib.rs
[permalink](https://github.com/AleoNet/snarkVM/blob/be171ce0720544c2bc51e27e1674a62fff585adc/ledger/committee/src/lib.rs#L173)
(note the comment at the permalink still has `2f + 1`, but that should be fixed with [this PR](https://github.com/AleoNet/snarkVM/pull/2526).
The code does the correct calculation.

ungaro:
I've submitted a pull request to address this issue:

[Fix: Correct quorum threshold calculation in is_round_reached() #3384
](https://github.com/AleoNet/snarkOS/pull/3384)

This PR updates the quorum threshold calculation to ensure proper Byzantine fault tolerance. The new implementation correctly calculates the quorum as `N - f` where `N` is the number of validators and `f` is the maximum number of faulty validators that can be tolerated.

The PR includes:
- Updated implementation of `is_round_reached()`
- Corrected comments explaining the Byzantine fault tolerance calculation
- Comparison of results between the old and new implementations

Thanks

[...]

PR #3384, opened 2024-08-24
[Fix] Correct quorum threshold calculation in test function is_round_reached()

ungaro:
## Motivation

This PR updates the quorum threshold calculation in the `is_round_reached` function to ensure correct Byzantine fault tolerance. The current implementation uses a simple majority calculation, which is insufficient for Byzantine fault tolerance in certain cases.

### Current implementation:
```rust
let quorum_threshold = self.validators.len() / 2 + 1;
```

### Proposed implementation:
```rust
let quorum_threshold = self.validators.len() - (self.validators.len() - 1) / 3;
```

The new calculation is based on the principle that in a Byzantine fault-tolerant system with N validators, we must be able to tolerate f faulty validators, where N = 3f + 1 + k (0 <= k < 3). The correct quorum threshold is N - f, which our new calculation achieves.
## Quorum Threshold Comparison

Comparison between current implementation (`n / 2 + 1`) and proposed implementation (`n - (n - 1) / 3`) for various ranges of validator counts (n).

### Range: 1 <= n < 10

| n | Current (n/2 + 1) | Proposed (n - (n-1)/3) | Correct? |
|---|-------------------|------------------------|----------|
| 1 | 1                 | 1                      | Yes      |
| 2 | 2                 | 2                      | Yes      |
| 3 | 2                 | 3                      | No       |
| 4 | 3                 | 3                      | Yes      |
| 5 | 3                 | 4                      | No       |
| 6 | 4                 | 5                      | No       |
| 7 | 4                 | 5                      | No       |
| 8 | 5                 | 6                      | No       |
| 9 | 5                 | 7                      | No       |

### Range: 99 < n < 110

| n   | Current (n/2 + 1) | Proposed (n - (n-1)/3) | Correct? |
|-----|-------------------|------------------------|----------|
| 100 | 51                | 67                     | No       |
| 101 | 51                | 68                     | No       |
| 102 | 52                | 68                     | No       |
| 103 | 52                | 69                     | No       |
| 104 | 53                | 70                     | No       |
| 105 | 53                | 70                     | No       |
| 106 | 54                | 71                     | No       |
| 107 | 54                | 72                     | No       |
| 108 | 55                | 72                     | No       |
| 109 | 55                | 73                     | No       |

### Range: 999 < n < 1010

| n    | Current (n/2 + 1) | Proposed (n - (n-1)/3) | Correct? |
|------|-------------------|------------------------|----------|
| 1000 | 501               | 667                    | No       |
| 1001 | 501               | 668                    | No       |
| 1002 | 502               | 668                    | No       |
| 1003 | 502               | 669                    | No       |
| 1004 | 503               | 670                    | No       |
| 1005 | 503               | 670                    | No       |
| 1006 | 504               | 671                    | No       |
| 1007 | 504               | 672                    | No       |
| 1008 | 505               | 672                    | No       |
| 1009 | 505               | 673                    | No       |

### Key Observations:

1. For n < 4, both implementations give the same result, but the current implementation is correct only by coincidence.

2. Starting from n = 3, the proposed implementation consistently requires a higher quorum threshold, which is necessary for Byzantine fault tolerance.

3. As n increases, the difference between the two implementations becomes more pronounced. The current implementation significantly underestimates the required quorum for larger validator sets.

4. The current implementation always returns a value close to 50% of n, which is insufficient for Byzantine fault tolerance.

5. The proposed implementation correctly calculates the Byzantine fault-tolerant quorum (approximately 2/3 of n) for all values of n.

This comparison demonstrates that the proposed implementation provides the correct quorum threshold for Byzantine fault tolerance across all ranges of validator counts, while the current implementation fails to do so in almost all cases where n > 2.


ungaro:
would like to get a review, thanks.
@bendyarm @d0cd

raychu86:
This may be trivial as the changes only affect the assumptions of a **testing only** `TestNetwork`. @niklaslong  Wanted to double check the intention of `2f+1` here.

acoglio:
I don't know the details of this particular testing code, but I confirm, with formal verification certainty, that n-f is the general correct quorum number. 2f+1 only works if n = 3f+1, but it fails (with easy counterexamples) if n = 3f+2 or n = 3f+3. Here f is defined the maximum one that satisfies f < n/3.

bendyarm:
@acoglio @raychu86 @niklaslong 
For more background see my notes on the [issue #3377](https://github.com/ProvableHQ/snarkOS/issues/3377) .
I did not fix this one at the time I did [PR#2526](https://github.com/ProvableHQ/snarkVM/pull/2526) because it wasn't clear to me what we
were using the test for and there were multiple problems with the test code.

acoglio (approved):
This looks fine to me to merge, but I'd like to share an observation. Before the PR, the comment said "2f + 1", but the code actually calculated "n/2 + 1". That makes me wonder whether the purpose of that code was actually to check a majority (i.e. more than half) rather than an actual quorum (n - f). Perhaps it does not matter much, as both can be worthy tests to run, but, after looking at the code in that file, I can't readily tell what the purpose was.

niklaslong:
This is indeed a mistake, thanks for correcting it! 

As for why, I'm fairly certain this was leftover from debugging flaky tests with larger numbers of validators—the asynchronous nature of the tests means they still sometimes fail with the corrected bound on `debug` builds. We also use `N = 4` in tests, so thankfully the bound is equivalent there.

niklaslong (approved):

PR #3386, opened 2024-08-28
[Fix] Coupling block sync to DAG state 

mdelle1:
## Motivation

This PR focuses on coupling block sync to DAG state replication. When a node is syncing via block responses, it will sync its storage and DAG with the certificates contained in the block and attempt to update its ledger. Previously, there were scenarios where a node would commit certificates in its DAG without advancing blocks. Instead, the committal of leader certificates and advancement of blocks during sync should be tightly coupled. 

To achieve this, if a node is syncing, it should forgo committing leader certificates on a rolling basis inside `update_dag`.  Instead, the leader certificate should be committed just before the ledger is ready to advance to the next block during sync. To facilitate this, we use a sender channel that communicates the leader certificate to be committed from the Sync module to the BFT. 

A previous version of this PR can be found here https://github.com/AleoNet/snarkOS/pull/3268 . 
The differences are that the previous PR does not prevent syncing nodes from committing leader certificates within `update_dag`, and that certificates inside block responses are added to the DAG all at once when the availability threshold is met. For contrast, in this PR, we maintain the original method of updating the DAG as soon as the certificates in the block response are processed within `sync_storage_with_block`, but only commit the leader certificate when the ledger is ready to advance to the next block. Updating the DAG as soon as the certificates are processed is necessary to ensure there is no discrepancy between the storage and the DAG state beyond the latest committed block. 

To summarize, this PR makes the following changes: 
- Implements sender-receiver channels for committing leader certificates and determining recently committed certificates between Sync and BFT.
- Prevents syncing nodes from committing leader certificates within `update_dag`. 
- Instead, commits the leader certificates in the BFT using a sender-receiver channel within `sync_storage_with_block` just before the ledger advances blocks.

## Test Plan
Relevant BFT test cases include the following: 
2. BFT-Rebonding 
12. Sync-Invalid-Peers-Attack 
13. Sync-Far-Behind

## Related PRs
https://github.com/AleoNet/snarkOS/pull/3268

### DIFF
--- a/node/bft/src/bft.rs
+++ b/node/bft/src/bft.rs
@@ -814,6 +814,7 @@ impl<N: Network> BFT<N> {
             mut rx_sync_bft_dag_at_bootup,
             mut rx_sync_bft,
             mut rx_commit_bft,
+            mut rx_is_recently_committed,
         } = bft_receiver;
 
         // Process the current round from the primary.
--- a/node/bft/src/bft.rs
+++ b/node/bft/src/bft.rs
@@ -520,11 +520,16 @@ impl<N: Network> BFT<N> {
             return Ok(());
         }
 
-        /* Proceeding to commit the leader. */
-        info!("Proceeding to commit round {commit_round} with leader '{}'", fmt_id(leader));
+        // Commit the leader certificate if the primary is not syncing. 
+        if !IS_SYNCING {
+            /* Proceeding to commit the leader. */
+            info!("Proceeding to commit round {commit_round} with leader '{}'", fmt_id(leader));
+            // Commit the leader certificate, and all previous leader certificates since the last committed round.
+            self.commit_leader_certificate::<ALLOW_LEDGER_ACCESS, IS_SYNCING>(leader_certificate).await?;
+        }
 
-        // Commit the leader certificate, and all previous leader certificates since the last committed round.
-        self.commit_leader_certificate::<ALLOW_LEDGER_ACCESS, IS_SYNCING>(leader_certificate).await
+        Ok(())
+
     }
 
     /// Commits the leader certificate, and all previous leader certificates since the last committed round.
--- a/node/bft/src/bft.rs
+++ b/node/bft/src/bft.rs
@@ -856,6 +856,30 @@ impl<N: Network> BFT<N> {
                 callback.send(result).ok();
             }
         });
+
+        // Process the request to commit the leader certificate. 
+        let self_ = self.clone();
+        self.spawn(async move {
+            while let Some((certificate, callback)) = rx_commit_bft.recv().await {
+                // Update the DAG with the certificate.
+                let result = self_.commit_leader_certificate::<true, true>(certificate).await;
+                // Send the callback **after** updating the DAG.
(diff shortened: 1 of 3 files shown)


node/bft/src/sync/mod.rs
```diff
@@ -484,13 +484,46 @@ impl<N: Network> Sync<N> {
...
+                            match bft_sender.send_sync_is_recently_committed(leader_round, leader_id).await {
+                                Ok(is_recently_committed) => {
+                                    if !is_recently_committed {
+                                        bail!(
+                                            "Sync - Failed to advance blocks - leader certificate with author {leader_author} from round {leader_round} was not recently committed.",
+                                        );
+                                    }
+                                    debug!(
+                                        "Sync - Leader certificate with author {leader_author} from round {leader_round} was recently committed.",
+                                    );
```
gluax, line 511: Maybe a nit but the boolean can be moved into the match statement you get a compile warning if missing a branch and this has better readability IMO.

i.e:

```rust
Ok(true) => {
  debug!("...");
}
Ok(false) => {
  bail!("...");
}
// rest of cases
```

gluax (approved):

node/bft/src/bft.rs
```diff
@@ -855,6 +862,30 @@ impl<N: Network> BFT<N> {
...
+                callback.send(result).ok();
+            }
+        });
+
+        // Process the request to check if the batch certificate was recently committed.
+        let self_ = self.clone();
+        self.spawn(async move {
+            while let Some(((round, certificate_id), callback)) = rx_is_recently_committed.recv().await {
+                // Check if the certificate was recently committed.
+                let is_committed = self_.dag.read().is_recently_committed(round, certificate_id);
```
ljedrz, line 883: seems like a lot of hassle just to perform this one check (in its own task with a dedicated channel), especially with both the `BFT` and `Dag` being clonable :thinking:... it _does_ make it async, but do we expect the `recent_committed_ids` collection to grow large enough to make it blocking?

alzger:
There are still CIs jobs failing @ljedrz . @vicsn , please let us know how you would like to proceed.

vicsn:
Can this PR be changed to draft state? No need to get it into next release.

alzger:
Converted to draft as per your [comment](https://github.com/AleoNet/snarkOS/pull/3386#issuecomment-2452287959) @vicsn

kaimast:
What is the status of this PR? 

The sync code has changed a lot in the last few months. If you are still working on the PR, please rebase it. If not, I will close it.

node/bft/src/bft.rs
```diff
@@ -521,11 +521,16 @@ impl<N: Network> BFT<N> {
             return Ok(());
         }
 
-        /* Proceeding to commit the leader. */
-        info!("Proceeding to commit round {commit_round} with leader '{}'", fmt_id(leader));
+        // Commit the leader certificate if the primary is not syncing.
```
kaimast, line 524: This comment needs to be extended and explain why we don't commit during sync.

vicsn:
@kaimast you can close

PR #3388, opened 2024-08-30
Extend the IP-level ban to Router-bound connections

ljedrz:
Based on https://github.com/AleoNet/snarkOS/pull/3366, only the last 2 commits are new.

This PR moves the IP-level ban piping down to the `Tcp` object level, making it available to both `Gateway` and `Router`, and applies it to the latter.

Filing as a draft, as it depends on the aforementioned base PR.

Cc @vicsn


raychu86:
Can be closed with https://github.com/AleoNet/snarkOS/pull/3422.

Issue #3387, opened 2024-08-28
[Bug] rest api for /blocks?start=...&end=... does not include last block


andrezorzo:
## 🐛 Bug Report

This is a low priority bug. 

The documentation says that  `GET /testnet/blocks?start={start_height}&end={end_height}` should return blocks from the given block range, but when requested with **start=1&end=2**, the output is not including the block 2, and the response is a list with only block 1. Also, if provided start=1&end=1 it will not return any blocks.

## Steps to Reproduce

Just curl the rest api with any block range:
```
curl 'http://127.0.0.1:3030/canary/blocks?start=1&end=2' | jq '.[].header.metadata.height'
1

curl 'http://127.0.0.1:3030/canary/blocks?start=1&end=3' | jq '.[].header.metadata.height'
1
2
```

## Expected Behavior

Expected that when requested a block range, all the blocks in the range should be returned.

raychu86:
Same issue as #3394.

Here is some context - https://github.com/AleoNet/snarkOS/pull/3394#issuecomment-2334523679

PR #3394, opened 2024-09-05
[Fix] change get_blocks range to closed interval

xG0Ian:
## Motivation
Close issue: #3387

We say `GET /testnet/blocks?start={start_height}&end={end_height}` Returns the blocks for **the given block range** ( [api doc](https://developer.aleo.org/testnet/public_endpoints/get_blocks) )

**the given block range**  was not described clearly enough.

`/testnet/blocks` returns the blocks of `from start_height to end_height - 1`. But I think it should return `from start_height to end_height` with closed interval which is more friendly with human intuition, cuz we always use current block height as end_height, easily misunderstood if we need give it **current block height + 1**.

## Test Plan
change
https://github.com/AleoNet/snarkOS/blob/0cc6fe3d9ad1f8c43c2037dac670a1fd417b2c99/node/rest/src/routes.rs#L140
to
```rust
let blocks = cfg_into_iter!((start_height..=end_height)) 
```

### DIFF
--- a/node/rest/src/routes.rs
+++ b/node/rest/src/routes.rs
@@ -137,7 +137,7 @@ impl<N: Network, C: ConsensusStorage<N>, R: Routing<N>> Rest<N, C, R> {
 
         // Prepare a closure for the blocking work.
         let get_json_blocks = move || -> Result<ErasedJson, RestError> {
-            let blocks = cfg_into_iter!((start_height..end_height))
+            let blocks = cfg_into_iter!((start_height..=end_height))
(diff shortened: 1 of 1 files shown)


raychu86:
We are following the `inclusive start` and `exclusive end` standard that is present in many blockchain and non-blockchain APIs.

The API documentation could definitely be more clear, however changing this standard will have a lot of downstream effects. This seems like an innocent change, but will break explorers, wallets, services, etc.

xG0Ian:
got it, so we need to make the documentation clearer.

PR #3395, opened 2024-09-05
Cache transmissions for faster processing

kpandl:
## Motivation

The `prepare_advance_to_next_quorum_block` function incurs significant processing time, which contributes to increased block times. Based on a flamegraph analysis, time-consuming operations are the deserialization of objects, or more general, retrieving transmissions from storage.

Therefore, this PR introduces an LRU cache for transmissions inside the `BFTPersistentStorage`. The transmissions are inserted into the cache in the `insert_transmissions` function, alongside with the insertion into the persistent storage. The `get_transmission` function will first try to retrieve the transmissions from the LRU cache, and if it fails, try to retrieve it from the persistent storage.

## Test Plan

Ran it locally, and also ran a load test (high tps of `transfer_public` executions) on aws. Thereby, we observed a ~50% reduction in block times.

Please note: The cache aims to cover two rounds of transmissions. As such, it can store `MAX_COMMITTEE_SIZE * MAX_TRANSMISSIONS_PER_BATCH * 2 = 1500`  transmissions. At the current `MAX_TRANSACTION_SIZE` of 128 kB, this results in approximately 188 MB additional memory use.


node/bft/storage-service/src/persistent.rs
```diff
@@ -46,24 +50,40 @@ pub struct BFTPersistentStorage<N: Network> {
     transmissions: DataMap<TransmissionID<N>, (Transmission<N>, IndexSet<Field<N>>)>,
     /// The map of `aborted transmission ID` to `certificate IDs` entries.
     aborted_transmission_ids: DataMap<TransmissionID<N>, IndexSet<Field<N>>>,
+    /// The LRU cache for `transmission ID` to `(transmission, certificate IDs)` entries.
+    cache_transmissions: RwLock<LruCache<TransmissionID<N>, (Transmission<N>, IndexSet<Field<N>>)>>,
+    /// The LRU cache for `aborted transmission ID` to `certificate IDs` entries.
+    cache_aborted_transmission_ids: RwLock<LruCache<TransmissionID<N>, IndexSet<Field<N>>>>,
```
ljedrz: two comments here:
1. the docs should explicitly state that these caches are for entries that exist in the persistent storage, so as not to confuse them with other, network-related caches
2. since `LruCache::peek` is not used at all, there is no reason to use an `RwLock` instead of the simpler `Mutex`

node/bft/storage-service/src/persistent.rs
```diff
@@ -152,66 +179,84 @@ impl<N: Network> StorageService<N> for BFTPersistentStorage<N> {
...
                     }
+
+                    // Also, update the cache.
+                    self.cache_transmissions.write().put(transmission_id, (transmission, certificate_ids));
                 }
                 Ok(None) => {
-                    // Retrieve the missing transmission.
-                    let Some(transmission) = missing_transmissions.remove(&transmission_id) else {
-                        if !aborted_transmission_ids.contains(&transmission_id)
-                            && !self.contains_transmission(transmission_id)
```
ljedrz, line 176: note to reviewers: apparently this call (`contains_transmission`) was redundant (as it checks `self.transmissions` and `self.aborted_transmissions_id`, which would have already been done by now)

node/bft/storage-service/src/persistent.rs
```diff
@@ -152,66 +179,84 @@ impl<N: Network> StorageService<N> for BFTPersistentStorage<N> {
...
-                    continue 'outer;
                 }
             }
         }
-        // Inserts the aborted transmission IDs.
+
+        // Next, handle the aborted transmission IDs.
         for aborted_transmission_id in aborted_transmission_ids {
-            // Retrieve the transmission entry.
             match self.aborted_transmission_ids.get_confirmed(&aborted_transmission_id) {
```
ljedrz: this is not coming from this PR, but I think it would be nice to squeeze this block by returning `certificate_ids` from its branches and performing the insertion to persistent storage and its cache below (which would deduplicate those calls)

ljedrz (commented):
Left a few suggestions, other than that LGTM in general :+1:.

ljedrz (dismissed):
LGTM :+1:

vicsn:
CI is failing. I suggest running `node` and `node-bft` locally to find out if its actually failing. For `node-router`, consider using a more lenient `deadline` instead of the `sleep` (see other tests using `deadline!` for examples)

node/bft/storage-service/src/persistent.rs
vicsn: I'm investigating a validator halting on a devnet running with this PR, made me look deeper into this. I don't think `contains_transmission` is redundant, as it additionally check `self.aborted_transmissions`. Without doing a full analysis, @kpandl can you revert to the original logic? 

```
let Some(transmission) = missing_transmissions.remove(&transmission_id) else {
                        if !aborted_transmission_ids.contains(&transmission_id)
                            && !self.contains_transmission(transmission_id)
```

node/bft/storage-service/src/persistent.rs
vicsn: Nice find! @kpandl want to apply the same trick for the first loop above?

node/router/tests/disconnect.rs
```diff
@@ -33,8 +34,12 @@ async fn test_disconnect_without_handshake() {
 
     // Connect node0 to node1.
     node0.connect(node1.local_ip());
-    // Sleep briefly.
-    tokio::time::sleep(Duration::from_millis(200)).await;
+    // Await both nodes being connected.
+    let node0_ = node0.clone();
+    let node1_ = node1.clone();
+    deadline!(Duration::from_secs(5), move || {
```
niklaslong: If these test cases usually run in milliseconds, we can probably lower the upper bound here a little?

node/bft/storage-service/Cargo.toml
```diff
@@ -29,6 +29,9 @@ workspace = true
 version = "2.1"
 features = [ "serde", "rayon" ]
 
+[dependencies.lru]
+version = "0.12.1"
```
niklaslong, line 36: No need to change it but just as a note to reviewers, the latest version (which will be used by cargo when building) is `0.12.4`.

vicsn (dismissed):

node/router/tests/disconnect.rs
kpandl: Thanks! I've reduced it to 1 second.

kpandl:
Thanks for the feedback! I've updated the code. The `node` test ran successfully locally. The `node-bft` test did not, but that was likely due to constraint local resources, not an issue with the code itself.

vicsn (approved):

zosorock:
@kpandl this branch has conflicts. Could you please resolve them before we attempt to merge?

node/bft/storage-service/src/persistent.rs
```diff
@@ -47,24 +51,40 @@ pub struct BFTPersistentStorage<N: Network> {
...
+    cache_aborted_transmission_ids: Mutex<LruCache<TransmissionID<N>, IndexSet<Field<N>>>>,
 }
 
 impl<N: Network> BFTPersistentStorage<N> {
     /// Initializes a new BFT persistent storage service.
     pub fn open(storage_mode: StorageMode) -> Result<Self> {
+        let capacity = NonZeroUsize::new(
+            (Committee::<N>::MAX_COMMITTEE_SIZE as usize) * (BatchHeader::<N>::MAX_TRANSMISSIONS_PER_BATCH) * 2,
+        )
+        .unwrap();
```
gluax: Is this something we never have to worry about erroring? If so might be good to leave a comment above explaining why it wouldn't

gluax (approved):

node/bft/storage-service/src/persistent.rs
vicsn: Hmm good catch, @kpandl can you change both unwraps by `?`? If the error type does not match the function signature, I usually use e.g.: `.map_err(|_| anyhow!("Could not construct NonZeroUsize"))?;`

kpandl:
@zosorock thanks, I've merged the latest `staging` into this branch, there should be no conflicts now

node/bft/storage-service/src/persistent.rs
kpandl: Thanks, I've added a commit addressing this

Issue #3403, opened 2024-09-23
[Bug] Most recent `curl-sys` crate update causes SnarkOS build to fail due to SnarkOS MSRV being at `v1.76.0`


iamalwaysuncomfortable:
## 🐛 Bug Report

The `curl-sys` library version was recently bumped to `0.4.76` which updated the Rust MSRV in that crate to `1.80.0`. This currently causes SnarkOS to fail to build using `cargo build` or `cargo install` to fail. 

This is resolved if the MSRV in SnarkOS is updated. 

## Steps to Reproduce

1. Navigate to the SnarkOS main directory
2. Execute `cargo build` or `cargo install --path .`
3. The build fails with the error:

```
error: unsupported output in build script of `curl-sys v0.4.76+curl-8.10.1`: `cargo::rustc-check-cfg=cfg(libcurl_vendored,link_libnghttp2,link_libz,link_openssl,)`
Found a `cargo::key=value` build directive which is reserved for future use.
Either change the directive to `cargo:key=value` syntax (note the single `:`) or upgrade your version of Rust.
See https://doc.rust-lang.org/cargo/reference/build-scripts.html#outputs-of-the-build-script for more information about build script outputs.
```

## Potential Fixes

* Update the MSRV in SnarkOS/SnarkVM to `Rust v1.80.0` or forcing usage of `Rust >1.80.0` fixes the issue. As of this bug report, the full consequences of updating the Rust MSRV in SnarkOS haven't been weighed, but it likely should be updated as more libraries may start to fail in the future.

zosorock:
@iamalwaysuncomfortable could you please confirm this is fixed with the merging of #3413?

raychu86:
Should be resolved with #3413

PR #3405, opened 2024-09-24
Fix requests per second, proper error responses, compilation related to the curl-sys library

### LINKED ISSUE #3402: [Bug] Rest request per second is actually burst per second, and returns 200s instead of errors
## 🐛 Bug Report

1. The `GovernorConfigBuilder` is configured to only replenish 1 element of the quota every second per ip address, regardless of the `rest_rps` variable passed in. The `rest_rps` variable is being used to set the burst_size per ip address, but assuming the `burst_size` is exhausted, the effective rate limit of a client node is 1 request per second per ip. My suggested fix would be to use `.per_nanosecond((1_000_000_000 / rest_rps) as u64)` in place of `.per_second(1)`, as this will replenish the `rest_rps` fully every second. Bug location: https://github.com/AleoNet/snarkOS/blob/mainnet-staging/node/rest/src/lib.rs#L114
2. 429s (too many requests) (and any other errors from the rate-limiting library) end up being swallowed and returned as 200s. As a patch, I'm going to use this in place of the current error handler:
```.error_handler(|error|  {
                // Properly return a 429 Too Many Requests error
                let error_message = error.to_string();
                let mut response = Response::new(error_message.clone().into());
                *response.status_mut() = StatusCode::INTERNAL_SERVER_ERROR;
                if error_message.contains("Too Many Requests") {
                  *response.status_mut() = StatusCode::TOO_MANY_REQUESTS;
                }
                response
            })
 ```
Location: https://github.com/AleoNet/snarkOS/blob/mainnet-staging/node/rest/src/lib.rs#L116

## Steps to Reproduce

Run a local devnet with a low rps, like 2. Hit an endpoint twice, and observe you can only hit the rest endpoint once per second thereafter.

## Expected Behavior

Errors should contain error codes, not 200s. Rest rps should be the actual requests per second and not just the burst metric.

## Your Environment

- snarkOS Version 
- v2.2.7
- Rust Version
- 1.81.0
- Computer OS
- observed locally on macOS, as well as on the demox client running on ubuntu

fulltimemike:
## Motivation

Fixes https://github.com/AleoNet/snarkOS/issues/3402. This PR includes a patch for https://github.com/AleoNet/snarkOS/issues/3403, in order to get this code to compile.

## Test Plan

Manually tested with low rps, verified that the requests per second did not become 1 after the burst quota was exhausted.

## Additional Details
This is our patch fix, so feel free to adapt this code as necessary. We use `per_nanosecond` instead of `per_second` for the replenishment rate, with the goal of replenishing the burst rate in full every second. Our primary concern was the rps, which is why the 429 (Too many requests) error in particular is handled but other errors are defaulted to a 500 (Internal Server Error). Furthermore, related to issue #3403, I had to bump the rust version in order to get snarkos to compile.


vicsn:
Thx! Mind removing the MSRV bump? It'll need to be updated in `Cargo.toml` and `.circleci/config.yml` as well, as well as for snarkVM. Better to make those separate PRs to be readable for everyone. 

Also perhaps someone will object using the latest MSRV, so a separate PR is a better ground for discussion. Will also ask some people directly about their opinion.

fulltimemike:
Sure, removed it. I couldn't get it to compile without updating the rust version, so you may need to update that before this is merged in

vicsn (approved):
LGTM!

evanmarshall (approved):

vicsn:
Can you merge `aleonet/staging` and fix the fmt issue?

fulltimemike:
> Can you merge `aleonet/staging` and fix the fmt issue?

done -- feel free to change this anyway you all need to get it merged

PR #3407, opened 2024-09-28
Use dev_id to prevent db lock contention during dev genesis creation

vicsn:
## Motivation

During testing, we often experienced lock contention, apparently because all dev nodes temporarily initialized a DB using `dev(0)`.

Also shortened `development_private_keys` to `dev_keys` to prevent `cargo fmt` from adding many lines of code.

## Test Plan

Tested locally. The adjusted `load_or_compute_genesis` function is only used in dev mode.


zkxuerb (approved):
LGTM

PR #3410, opened 2024-10-03
Update syncing logic to fix duplicate block requests

### LINKED ISSUE #3404: [Bug] Duplicate block requests when validators sync
## 🐛 Bug Report

When a validator syncs, it sends out a duplicate block request for a certain height, despite already having received a response for that height from the earlier request.

Specifically, it requests blocks of a certain range (e.g., `1` to `n+m` with `m>1`), and then advances to block `n`. Then, it sends out a block request for height `n+1` again, despite it already being in the initial block range request.

The bug slows down the syncing of validators.

## Steps to Reproduce

Run the `./devnet.sh` script with 4 validators, turn off 1 validator. Let it run for ~30 blocks. Then, start the fourth validator and let it sync. Consider adding logs to snarkOS for sync profiling.

## Expected Behavior

The validator should use the block already sent to it and advance to it (and further).

## Your Environment

- snarkOS branch `kp/profile/sync/` in `ProvableHQ/snarkOS`
- macOS

## Debugging notes

- In `sync_storage_with_blocks`, there are two places where we advance with blocks - without and with BFT checks.
- The issue is for cases where we advance with BFT checks. `sync_storage_with_block` [does not always](https://github.com/AleoNet/snarkOS/blob/2349f3edab28ff57cf551f3b96febba5aa120b8b/node/bft/src/sync/mod.rs#L422-L423) update the canon ledger. Thus, `self.canon` can't necessarily tell until which height we already parsed responses.
- While the block is still in `self.latest_block_responses`, this member is in `Sync` and not in `BlockSync`. A refactor should consider the entire syncing logic, and how `Sync` and `BlockSync` relate.

kpandl:
## Motivation

As issue #3404 described, validators send out duplicate block requests when syncing. This PR updates the syncing logic to make sure we only remove a block response after it was added to the ledger or if we encountered an error.

For background, currently, syncing works as follows: `BlockSync` requests the blocks, and stores it in the internal `responses` map. `Sync` calls `block_sync.process_next_block(current_height)` to get the next block to process. This removes it from `BlockSync`’s `responses` and returns it to `Sync`. `Sync` performs the checks, and if successful, adds the block to the `ledger`. This also updates BlockSync’s `canon`.

The issue is that `BlockSync`‘s view of the `canon` is only updated after `Sync` is done processing blocks (and adding to the ledger). There’s a window where `BlockSync` is unaware of blocks that are pending validation in `Sync`. Thus, it re-requests them.

This PR adjusts the block processing interface to the `Sync` module as follows:
* `process_next_block` READS a block from a particular height
* `remove_block_response` REMOVES a block from a particular height, which MUST be called after advancing completed or failed

Note: in the latest commit, `process_next_block` was renamed to `peek_next_block`, following the reviewer discussion.

## Test Plan

Ran it locally and verified it removes the duplicate block request.
Also ran it with light load in a cloud devnet, and verified syncing works for validators and clients.

The following figure shows a syncing process before the fix, with a longer pause until a duplicate request is generated:
syncing_not_fixed

The following figure shows the syncing process with the fix (note there is no double request and no delay, as such, the syncing is much faster):
syncing_fixed

Closes #3404


node/sync/src/block_sync.rs
```diff
@@ -325,8 +325,21 @@ impl<N: Network> BlockSync<N> {
     /// Returns the next block to process, if one is ready.
     #[inline]
     pub fn process_next_block(&self, next_height: u32) -> Option<Block<N>> {
-        // Try to advance the ledger with a block from the sync pool.
-        self.remove_block_response(next_height)
+        // Acquire the requests write lock.
+        // Note: This lock must be held across the entire scope, due to asynchronous block responses
+        // from multiple peers that may be received concurrently.
+        let requests = self.requests.write();
```
```suggestion
        let requests = self.requests.read();
```
reasoning: we don't need to alter the collection in this scope, and I don't see a reason to disallow concurrent readers if by the end of it we're going to produce the same result anyway

node/sync/src/block_sync.rs
```diff
@@ -351,20 +364,33 @@ impl<N: Network> BlockSync<N> {
...
+            let advanced = match self.canon.check_next_block(&block) {
+                Ok(_) => match self.canon.advance_to_next_block(&block) {
+                    Ok(_) => true,
+                    Err(err) => {
+                        warn!("{err}");
+                        false
+                    }
+                },
+                Err(err) => {
+                    warn!("{err}");
```
ljedrz: note: this is a change that might surprise users who relied on the format of [the old log](https://github.com/AleoNet/snarkOS/pull/3410/files#diff-c2e724c7020cd2ef76ea842b503a4cb85f626f19d85d60e9c685fee70e3725ddL362)

ljedrz (commented):
LGTM with some nits

node/sync/src/block_sync.rs
kpandl: Thanks, good find! I've added more expressive warnings similar to the previous code here: https://github.com/AleoNet/snarkOS/pull/3410/commits/d6e4f737e99361dd2ffdff6c8abf5621689545ca

zkxuerb (dismissed):

node/sync/src/block_sync.rs
```diff
@@ -609,32 +635,25 @@ impl<N: Network> BlockSync<N> {
...
         // Remove the response entry for the given height.
-        self.responses.write().remove(&height)
+        self.responses.write().remove(&height);
     }
 
     /// Removes the block request for the given peer IP, if it exists.
     #[allow(dead_code)]
     fn remove_block_request_to_peer(&self, peer_ip: &SocketAddr, height: u32) {
-        let mut can_revoke = self.responses.read().get(&height).is_none();
+        let mut can_revoke = self.process_next_block(height).is_none();
```
raychu86: Why is `remove_block_request_to_peer` now calling `process_next_block` instead of just removing it from the `responses`?

raychu86: Nevermind, I answered my own question. You are now nesting the response removal inside `process_next_block` and adding some additional checks.

ljedrz: while I also believe it to be fine, an adjustment to the function name or extending the doc comment could be beneficial here

ljedrz (approved):

raychu86 (dismissed):
LGTM! I think I'll feel a bit more comfortable with the change after we run our gambit of sync tests and malice tests, but otherwise logic looks good

node/sync/src/block_sync.rs
kpandl: Thanks, good discussion! I've renamed `process_next_block` to `peek_next_block` and adjusted the doc comments here: https://github.com/AleoNet/snarkOS/pull/3410/commits/778961a1b61ade1c28bfa9b34b3c127418a3d170
Indeed, we are no longer removing the block from the responses in the peek_next_block/process_next_block function.

vicsn:
> LGTM! I think I'll feel a bit more comfortable with the change after we run our gambit of sync tests and malice tests, but otherwise logic looks good

Note that the PR description mentions sync tests were run on deployed network. We don't have any malicious sync tests yet, but we will one day!

vicsn (approved):

PR #3413, opened 2024-10-15
[Update] MSRV and clippy.

d0cd:
This PR updates the minimum Rust version to v1.81.0 and addresses clippy issues associated with the upgrade.
As a result, this PR:
    - removes unused conditional compilation
    - refactors code
    - updates CI configurations accordingly
These changes should be carefully reviewed to ensure that behavior is not changed.

This PR also introduces an update to the naming convention for `cache_key`s. 
The proposed convention is `v<SNARKOS_VERSION>-rust-<RUST_VERSION>-<NAME>`.

@reviewers If we are considering changes to the key naming convention, consider:
- including a checksum of the `Cargo.lock` file which would capture any dependency updates
- using a variable for Rust version and snarkOS version

CI runs are [here](https://app.circleci.com/pipelines/github/ProvableHQ/snarkOS?branch=update%2Frust-version%2Bclippy-ci)


node/tcp/src/helpers/config.rs
```diff
@@ -73,12 +73,6 @@ impl Config {
 impl Default for Config {
     /// Initializes a new Tcp configuration with the default values.
     fn default() -> Self {
-        #[cfg(feature = "test")]
```
vicsn, line 76: Was this called out as unused by the compiler?

vicsn (approved):
LGTM, just one question for my understanding

gluax (approved):

node/tcp/src/helpers/config.rs
d0cd: That's right, the `tcp` crate doesn't have a `test` feature.

zosorock (approved):

PR #3417, opened 2024-10-22
Add dockerfile

Raphexion:
## Motivation

I've added a Dockerfile to add support containerization. Especially useful for ephemeral networks.

## Test Plan

none

## Related PRs

none

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,18 @@
+# Ignore everything
+*
+
+!account/
+!build.rs
+!Cargo.toml
+!Cargo.lock
+!cli/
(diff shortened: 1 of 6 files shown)


zosorock:
hi @Raphexion - thank you for the contribution but we already have a Dockerfile internally (for a few years actually) that we are in the process of making public. You can see the old one in this branch:
https://github.com/AleoNet/snarkOS/tree/docker-builds

I do like your `.dockerignore` though, we may use it 😁 I will ping you on the PR when we merge it in. Thanks again for contributing!

PR #3420, opened 2024-10-23
[Perf] Fetch transmissions and certificates simultaneously

vicsn:
## Motivation

Fetching transmissions poses a significant bottleneck in consensus. This simple change can avoid waiting for a synchronous roundtrip of certificate fetching, which happens almost every round.

## Test Plan

Tested on a local devnet with traffic.

### DIFF
--- a/node/bft/src/primary.rs
+++ b/node/bft/src/primary.rs
@@ -1560,14 +1560,17 @@ impl<N: Network> Primary<N> {
             self.try_increment_to_the_next_round(batch_round).await?;
         }
 
+        // Ensure the primary has all of the transmissions.
+        let transmissions_promise = self.fetch_missing_transmissions(peer_ip, batch_header);        
+
         // Ensure the primary has all of the previous certificates.
         let missing_previous_certificates =
             self.fetch_missing_previous_certificates(peer_ip, batch_header).await.map_err(|e| {
                 anyhow!("Failed to fetch missing previous certificates for round {batch_round} from '{peer_ip}' - {e}")
             })?;
 
-        // Ensure the primary has all of the transmissions.
-        let missing_transmissions = self.fetch_missing_transmissions(peer_ip, batch_header).await.map_err(|e| {
+        // Collect any missing transmissions.
+        let missing_transmissions = transmissions_promise.await.map_err(|e| {
             anyhow!("Failed to fetch missing transmissions for round {batch_round} from '{peer_ip}' - {e}")
         })?;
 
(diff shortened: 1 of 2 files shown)


node/bft/src/primary.rs
```diff
@@ -1560,14 +1560,17 @@ impl<N: Network> Primary<N> {
...
         // Ensure the primary has all of the previous certificates.
         let missing_previous_certificates =
             self.fetch_missing_previous_certificates(peer_ip, batch_header).await.map_err(|e| {
                 anyhow!("Failed to fetch missing previous certificates for round {batch_round} from '{peer_ip}' - {e}")
             })?;
 
-        // Ensure the primary has all of the transmissions.
-        let missing_transmissions = self.fetch_missing_transmissions(peer_ip, batch_header).await.map_err(|e| {
+        // Collect any missing transmissions.
+        let missing_transmissions = transmissions_promise.await.map_err(|e| {
```
d0cd: Does it make sense to `await` at the very end of the function, since we only return the result of the promise?

vicsn: Also had that thought! But `sync_with_certificate_from_peer` below recursively calls back into this function, which could then cause duplicate transmissions to be fetched. Under certain assumptions, that might actually be faster, but it seems cleaner to await fetching transmissions before syncing with new certificates.

node/bft/src/primary.rs
```diff
@@ -1560,14 +1560,17 @@ impl<N: Network> Primary<N> {
             self.try_increment_to_the_next_round(batch_round).await?;
         }
 
+        // Ensure the primary has all of the transmissions.
+        let transmissions_promise = self.fetch_missing_transmissions(peer_ip, batch_header);
```
ljedrz: in order for this call do do anything, it would need to be spawned within its own task where it would be awaited; then, the handle to that task would be awaited in the `missing_transmissions` binding; otherwise this change is ineffective, isn't it (since this binding only creates an inert `Future`)?

d0cd: Ah that's right. An `.await` is necessary to tell the async runtime to poll the future, which then calls the underlying function.

d0cd: @ljedrz does it make sense to use a `join` here?

ljedrz: yes, that would work, though since errors trigger early returns, [`try_join`](https://docs.rs/tokio/latest/tokio/macro.try_join.html) should be used instead

vicsn: Thx gents, looks much cleaner too: d0b93272d

zkxuerb (approved):

PR #3421, opened 2024-10-23
Periodically process the unconfirmed transmissions in the memory pool

vicsn:
## Motivation

To ensure transmissions do not remain stuck in the mempool when there is no traffic, trigger processing of transmissions periodically.

## Test Plan

Tested on local devnet with transaction traffic.

### DIFF
--- a/node/consensus/src/lib.rs
+++ b/node/consensus/src/lib.rs
@@ -21,6 +21,7 @@ extern crate tracing;
 use snarkos_account::Account;
 use snarkos_node_bft::{
     BFT,
+    MAX_BATCH_DELAY_IN_MS,
     Primary,
     helpers::{
         ConsensusReceiver,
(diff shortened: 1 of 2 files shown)


node/consensus/src/lib.rs
```diff
@@ -455,6 +468,23 @@ impl<N: Network> Consensus<N> {
...
+                // Sleep briefly.
+                tokio::time::sleep(Duration::from_millis(MAX_BATCH_DELAY_IN_MS)).await;
+                // Process the unconfirmed transactions in the memory pool.
+                if let Err(e) = self_.process_unconfirmed_transactions().await {
+                    warn!("Cannot process unconfirmed transactions - {e}");
+                }
+                // Process the unconfirmed solutions in the memory pool.
+                if let Err(e) = self_.process_unconfirmed_solutions().await {
+                    warn!("Cannot process unconfirmed solutions - {e}");
+                }
```
ljedrz, line 485: if these can be processed concurrently, `tokio::join` might be a good idea

ljedrz, line 485: if these can be processed concurrently, `tokio::join` might be a good idea

ljedrz (commented):
LGTM in general, left one suggestion

PR #3422, opened 2024-10-24
Spam connections and forged block locators fix

zkxuerb:
## Motivation

The motivation behind this PR is to have a mechanism of banning peer IPs as outlined [here](https://github.com/AleoNet/snarkOS/pull/3388) to mitigate cases that might stall client/validators.

## Approach
A banned peers list was implemented to store peers acting maliciously that are banned for 300 seconds. If a peer sends more than 10 connection requests in under 10 seconds they get banned and if a peer does not respond to block requests within 600 seconds ( can be changed ) they get banned.

## Test Plan

This was tested on an isolated network with a different ip/machine per peer. Plan is to test this out on the Canary network with the relevant attacks.

## Related PRs

https://github.com/AleoNet/snarkOS/pull/3366
https://github.com/AleoNet/snarkOS/pull/3388

### DIFF
--- a/node/bft/src/gateway.rs
+++ b/node/bft/src/gateway.rs
@@ -1142,7 +1142,6 @@ impl<N: Network> Handshake for Gateway<N> {
         if self.dev().is_none() && peer_side == ConnectionSide::Initiator {
             // If the IP is already banned, update the ban timestamp and reject the connection.
             if self.is_ip_banned(peer_addr.ip()) {
-                self.update_ip_ban(peer_addr.ip());
                 trace!("{CONTEXT} Gateway rejected a connection request from banned IP '{}'", peer_addr.ip());
                 return Err(error(format!("'{}' is a banned IP address", peer_addr.ip())));
             }
--- a/node/bft/src/gateway.rs
+++ b/node/bft/src/gateway.rs
@@ -896,6 +896,8 @@ impl<N: Network> Gateway<N> {
         self.handle_unauthorized_validators();
         // If the number of connected validators is less than the minimum, send a `ValidatorsRequest`.
         self.handle_min_connected_validators();
+        // Unban any addresses whose ban time has expired.
+        self.handle_banned_ips();
     }
 
     /// Logs the connected validators.
@@ -975,6 +977,11 @@ impl<N: Network> Gateway<N> {
                 });
             }
         }
+    }
+
+    // Remove addresses whose ban time has expired.
+    fn handle_banned_ips(&self) {
+        self.tcp.banned_peers().remove_old_bans(IP_BAN_TIME_IN_SECS);
     }
 }
 
@@ -1129,6 +1136,33 @@ impl<N: Network> Handshake for Gateway<N> {
         // Perform the handshake.
         let peer_addr = connection.addr();
         let peer_side = connection.side();
+
+        // Check (or impose) IP-level bans.
+        #[cfg(not(any(test, feature = "test")))]
+        if self.dev().is_none() && peer_side == ConnectionSide::Initiator {
+            // If the IP is already banned, update the ban timestamp and reject the connection.
+            if self.is_ip_banned(peer_addr.ip()) {
+                self.update_ip_ban(peer_addr.ip());
+                trace!("{CONTEXT} Gateway rejected a connection request from banned IP '{}'", peer_addr.ip());
+                return Err(error(format!("'{}' is a banned IP address", peer_addr.ip())));
+            }
+
+            // Check the previous low-level connection timestamp.
+            if let Some(peer_stats) = self.tcp.known_peers().get(peer_addr.ip()) {
+                let num_attempts = self.cache.insert_inbound_connection(peer_addr.ip(), 10);
+
+                debug!("Number of connection attempts from '{}': {}", peer_addr.ip(), num_attempts);
+                debug!("Seconds since connection attempt: {}", peer_stats.timestamp().elapsed().as_secs());
+                if peer_stats.timestamp().elapsed().as_secs() <= MIN_CONNECTION_INTERVAL_IN_SECS
+                    && num_attempts >= MAX_CONNECTION_ATTEMPTS
+                {
+                    self.update_ip_ban(peer_addr.ip());
+                    trace!("{CONTEXT} Gateway rejected a consecutive connection request from IP '{}'", peer_addr.ip());
+                    return Err(error(format!("'{}' appears to be spamming connections", peer_addr.ip())));
+                }
+            }
+        }
+
         let stream = self.borrow_stream(&mut connection);
 
         // If this is an inbound connection, we log it, but don't know the listening address yet.
--- a/node/bft/src/gateway.rs
+++ b/node/bft/src/gateway.rs
@@ -65,6 +65,8 @@ use futures::SinkExt;
 use indexmap::{IndexMap, IndexSet};
 use parking_lot::{Mutex, RwLock};
 use rand::seq::{IteratorRandom, SliceRandom};
+#[cfg(not(any(test, feature = "test")))]
+use std::net::IpAddr;
 use std::{collections::HashSet, future::Future, io, net::SocketAddr, sync::Arc, time::Duration};
 use tokio::{
     net::TcpStream,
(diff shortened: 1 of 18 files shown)


node/sync/Cargo.toml
```diff
@@ -66,6 +66,10 @@ version = "=3.0.0"
 path = "locators"
 version = "=3.0.0"
 
+[dependencies.snarkos-node-tcp]
+path = "../tcp"
+version = "=2.2.7"
```
ljedrz:
```suggestion
version = "=3.0.0"
```

node/sync/src/block_sync.rs
```diff
@@ -700,6 +708,17 @@ impl<N: Network> BlockSync<N> {
             // If the request has timed out, or is obsolete, then remove it.
             if is_timeout || is_obsolete {
                 trace!("Block request {height} has timed out: is_time_passed = {is_time_passed}, is_request_incomplete = {is_request_incomplete}, is_obsolete = {is_obsolete}");
+
+                if let Some((_, _, peer_ips)) = requests.get(height) { peer_ips.iter().for_each(|peer_ip| {
+                        debug!("Removing peer {peer_ip} from block request {height}");
+                        // Remove the locators entry for the given peer IP.
+                        locators.swap_remove(peer_ip);
```
ljedrz: does the order of the peer block locators not matter? since it's an `IndexMap` and I see it being iterated over, I'd use `shift_remove` to be on the safe side (we typically don't care about map indices, but the relative order is of interest); then again, the key is just the IP in this case, but then I'd prefer for the collection to be a plain `HashMap` instead

node/sync/src/block_sync.rs
```diff
@@ -711,6 +730,18 @@ impl<N: Network> BlockSync<N> {
...
 
+        // After the retain loop, handle the banning of peers
+        for peer_ip in peers_to_ban {
+            trace!("Banning peer {peer_ip} for timing out on block requests");
+            self.tcp.banned_peers().update_ip_ban(peer_ip.ip());
+
+            let tcp = self.tcp.clone();
+            tokio::spawn(async move {
+                tcp.disconnect(peer_ip).await;
+                trace!("Peer disconnected!");
```
ljedrz: nit: we're already getting `DEBUG`-level information on the peer being disconnected [here](https://github.com/AleoNet/snarkOS/blob/b0b89c3d1c5c3c6690472801b9b283edeb46b4d3/node/tcp/src/tcp.rs#L298)

node/tcp/src/helpers/banned_peers.rs
```diff
@@ -0,0 +1,66 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::{collections::HashMap, net::IpAddr, time::Instant};
+
+use parking_lot::RwLock;
+
+/// Contains the ban details for a banned peer.
+#[derive(Clone)]
+pub struct BanConfig {
```
ljedrz: nit: since there isn't much room for per-ban configuration, `BanDetails` might be more fitting here

node/bft/src/gateway.rs
```diff
@@ -1125,6 +1152,28 @@ impl<N: Network> Handshake for Gateway<N> {
...
+            // If the IP is already banned reject the connection.
+            if self.is_ip_banned(peer_addr.ip()) {
+                trace!("{CONTEXT} Gateway rejected a connection request from banned IP '{}'", peer_addr.ip());
+                return Err(error(format!("'{}' is a banned IP address", peer_addr.ip())));
+            }
+
+            let num_attempts = self.cache.insert_inbound_connection(peer_addr.ip(), CONNECTION_ATTEMPTS_SINCE_SECS);
+
+            debug!("Number of connection attempts from '{}': {}", peer_addr.ip(), num_attempts);
+            if num_attempts >= MAX_CONNECTION_ATTEMPTS
```
ljedrz: nit: `MAX_*` implies that `num_attempts == MAX_CONNECTION_ATTEMPTS` is permissible (i.e. that `>` would be a better fit here)

node/router/src/handshake.rs
```diff
@@ -101,6 +101,26 @@ impl<N: Network> Router<N> {
...
+            // If the IP is already banned reject the connection.
+            if self.is_ip_banned(peer_addr.ip()) {
+                trace!("Rejected a connection request from banned IP '{}'", peer_addr.ip());
+                return Err(error(format!("'{}' is a banned IP address", peer_addr.ip())));
+            }
+
+            let num_attempts = self.cache.insert_inbound_connection(peer_addr.ip(),  Router::<N>::CONNECTION_ATTEMPTS_SINCE_SECS);
+
+            debug!("Number of connection attempts from '{}': {}", peer_addr.ip(), num_attempts);
+            if num_attempts >= Router::<N>::MAX_CONNECTION_ATTEMPTS
```
ljedrz: same off-by-one nit as with the `Gateway`

node/router/src/heartbeat.rs
```diff
@@ -233,11 +237,19 @@ pub trait Heartbeat<N: Network>: Outbound<N> {
         if num_deficient > 0 {
             // Initialize an RNG.
             let rng = &mut OsRng;
+            let banned_ips = self.tcp().banned_peers().get_banned_ips();
 
             // Attempt to connect to more peers.
-            for peer_ip in self.router().candidate_peers().into_iter().choose_multiple(rng, num_deficient) {
+            for peer_ip in self
+                .router()
+                .candidate_peers()
```
ljedrz: I wonder if it would make sense to exclude the banned IPs directly from within `candidate_peers` (whenever it's called, so ban expiration is honored)

Pjrich1313 (changes requested):
This isn't a dupe white board needs to debug

vicsn (approved):

node/tcp/src/tcp.rs
```diff
@@ -283,13 +295,6 @@ impl Tcp {
                 task.abort();
             }
 
-            // If the (owning) Tcp was not the initiator of the connection, it doesn't know the listening address
-            // of the associated peer, so the related stats are unreliable; the next connection initiated by the
-            // peer could be bound to an entirely different port number
-            if conn.side() == ConnectionSide::Initiator {
```
howardwu, line 289: @ljedrz can you confirm safety?

node/src/validator/mod.rs
```diff
@@ -133,6 +131,9 @@ impl<N: Network, C: ConsensusStorage<N>> Validator<N, C> {
         )
         .await?;
 
+        // Initialize the sync module.
+        let sync = BlockSync::new(BlockSyncMode::Gateway, ledger_service, router.tcp().clone());
```
howardwu, line 135: @raychu86 can you review to confirm this reordering is safe? IIRC this has initialization implications to the node's state.

node/src/prover/mod.rs
```diff
@@ -119,6 +117,10 @@ impl<N: Network, C: ConsensusStorage<N>> Prover<N, C> {
             matches!(storage_mode, StorageMode::Development(_)),
         )
         .await?;
+
+        // Initialize the sync module.
+        let sync = BlockSync::new(BlockSyncMode::Router, ledger_service.clone(), router.tcp().clone());
```
howardwu, line 122: @raychu86 can you review to confirm this reordering is safe? IIRC this has initialization implications to the node's state.

node/src/client/mod.rs
```diff
@@ -123,6 +121,10 @@ impl<N: Network, C: ConsensusStorage<N>> Client<N, C> {
             matches!(storage_mode, StorageMode::Development(_)),
         )
         .await?;
+
+        // Initialize the sync module.
+        let sync = BlockSync::new(BlockSyncMode::Router, ledger_service.clone(), router.tcp().clone());
```
howardwu, line 126: @raychu86 can you review to confirm this reordering is safe? IIRC this has initialization implications to the node's state.

node/tcp/src/tcp.rs
ljedrz: this is safe; until now, the known (low-level) peers were distinguished by IP+port, and since inbound connections are assigned generic ports, this wasn't suitable for long-term peer tracking; however, moving to IPs only for that collection means that there is no need for this special-casing anymore

Pjrich1313 (commented):
Debug and rerun nobes

node/src/validator/mod.rs
raychu86: This particular reordering should be safe - `BlockSync::new` now requires an objects from `Router`, which is why it is initialized after Router. `Router` initialization is not affected and does not affect node state.

node/src/prover/mod.rs
raychu86: Ditto the above - 

This particular reordering should be safe - `BlockSync::new` now requires an objects from `Router`, which is why it is initialized after Router. `Router` initialization is not affected and does not affect node state.

node/src/client/mod.rs
raychu86: Ditto the above - 

This particular reordering should be safe - `BlockSync::new` now requires an objects from `Router`, which is why it is initialized after Router. `Router` initialization is not affected and does not affect node state.

ljedrz (approved):
LGTM :+1:

PR #3424, opened 2024-10-29
feat: test targets

vicsn:
## Motivation

Add feature flag to be able to lower coinbase and proof targets for testing purposes. This feature flag will lower the proof and coinbase targets to 8 and 32 respectively.

The feature flag needed to be cascaded to all dependencies of snarkVM to be sure everything uses same constants.

To use, compile with --features test_targets
e.g.

`cargo install --locked --features test_targets --path .`

## Test Plan

tested by using tx-cannon and adding println to verify correct targets, and by checking the genesis block.

Only `testnet` and `canary` in `--dev` mode have lower targets.

[CI link](https://app.circleci.com/pipelines/github/ProvableHQ/snarkOS/13686/workflows/2efc80fc-9c83-47cf-88c0-67b27ee9ab53)

## Related PRs

Replaces: https://github.com/AleoNet/snarkOS/pull/3378


vicsn:
Tested again, works as expected

ljedrz (approved):
Looks like this just allows snarkOS to utilize a feature implemented in snarkVM.
> The feature flag needed to be cascaded to all dependencies of snarkVM to be sure everything uses same constants.

Note: this would also be achieved with workspace-wide feature, but only for full compilations - not for tests (i.e. this is the correct way of introducing this feature).

LGTM with the conflict resolved :+1:.

gluax (approved):

zkxuerb (approved):
LGTM

PR #3425, opened 2024-10-30
Add support for account new --save-to-file

Raphexion:
## Motivation
When creating ephemeral networks for integration testing, it is convenient to be able to generate accounts programmatically and save the account information into files, especially the private-key-file for a new account.

Therefore, we introduce the command line switch `--save-to-file`.

```
snarkos account new --save-to-file /path/to/test-account-1
```

This feature, will create good synergy with the existing feature to use the command-line switch `--private-key-file`.

The PR is split into two commits

1. Add the feature to use --save-to-file for account new
1. Add tempfile as a dependency

## Test Plan

Make sure that you can use the switches `--save-to-file` and `--private-key-file` together when using the command-line interface (cli).

### DIFF
--- a/cli/src/commands/account.rs
+++ b/cli/src/commands/account.rs
@@ -30,9 +30,11 @@ use rand::SeedableRng;
 use rand_chacha::ChaChaRng;
 use rayon::prelude::*;
 use std::{
+    fs::File,
     io::{Read, Write},
     path::PathBuf,
 };
+
 use zeroize::Zeroize;
 
 /// Commands to manage Aleo accounts.
@@ -52,6 +54,9 @@ pub enum Account {
         /// Print sensitive information (such as the private key) discreetly in an alternate screen
         #[clap(long)]
         discreet: bool,
+        /// Specify the path to a file where to save the account in addition to printing it
+        #[clap(long = "save-to-file")]
+        save_to_file: Option<String>,
     },
     Sign {
         /// Specify the network of the private key to sign with
@@ -97,10 +102,18 @@ fn aleo_literal_to_fields<N: Network>(input: &str) -> Result<Vec<Field<N>>> {
 impl Account {
     pub fn parse(self) -> Result<String> {
         match self {
-            Self::New { network, seed, vanity, discreet } => {
+            Self::New { network, seed, vanity, discreet, save_to_file } => {
                 // Ensure only the seed or the vanity string is specified.
                 if seed.is_some() && vanity.is_some() {
                     bail!("Cannot specify both the '--seed' and '--vanity' flags");
+                }
+
+                if save_to_file.is_some() && vanity.is_some() {
+                    bail!("Cannot specify both the '--save-to-file' and '--vanity' flags");
+                }
+
+                if save_to_file.is_some() && discreet {
+                    bail!("Cannot specify both the '--save-to-file' and '--discreet' flags");
                 }
 
                 match vanity {
@@ -113,9 +126,9 @@ impl Account {
                     },
                     // Generate a seeded account for the specified network.
                     None => match network {
-                        MainnetV0::ID => Self::new_seeded::<MainnetV0>(seed, discreet),
-                        TestnetV0::ID => Self::new_seeded::<TestnetV0>(seed, discreet),
-                        CanaryV0::ID => Self::new_seeded::<CanaryV0>(seed, discreet),
+                        MainnetV0::ID => Self::new_seeded::<MainnetV0>(seed, discreet, save_to_file),
+                        TestnetV0::ID => Self::new_seeded::<TestnetV0>(seed, discreet, save_to_file),
+                        CanaryV0::ID => Self::new_seeded::<CanaryV0>(seed, discreet, save_to_file),
                         unknown_id => bail!("Unknown network ID ({unknown_id})"),
                     },
                 }
@@ -227,7 +240,7 @@ impl Account {
     }
 
     /// Generates a new Aleo account with an optional seed.
-    fn new_seeded<N: Network>(seed: Option<String>, discreet: bool) -> Result<String> {
+    fn new_seeded<N: Network>(seed: Option<String>, discreet: bool, save_to_file: Option<String>) -> Result<String> {
         // Recover the seed.
         let seed = match seed {
             // Recover the field element deterministically.
@@ -242,6 +255,13 @@ impl Account {
             PrivateKey::try_from(seed).map_err(|_| anyhow!("Failed to convert the seed into a valid private key"))?;
         // Construct the account.
         let account = snarkos_account::Account::<N>::try_from(private_key)?;
+        // Save to file in addition to printing it back to the user
+        if let Some(path) = save_to_file {
+            crate::check_parent_permissions(&path)?;
+            let mut file = File::create_new(path)?;
+            file.write_all(account.private_key().to_string().as_bytes())?;
+            crate::set_user_read_only(&file)?;
+        }
         // Print the new Aleo account.
         if !discreet {
             return Ok(account.to_string());
@@ -331,13 +351,15 @@ fn wait_for_keypress() {
 #[cfg(test)]
 mod tests {
     use crate::commands::Account;
+    use std::{fs, fs::Permissions, io::Write};
+    use tempfile::{NamedTempFile, TempDir};
 
     use colored::Colorize;
 
     #[test]
     fn test_new() {
         for _ in 0..3 {
-            let account = Account::New { network: 0, seed: None, vanity: None, discreet: false };
+            let account = Account::New { network: 0, seed: None, vanity: None, discreet: false, save_to_file: None };
             assert!(account.parse().is_ok());
         }
     }
@@ -363,7 +385,7 @@ mod tests {
         );
 
         let vanity = None;
-        let account = Account::New { network: 0, seed, vanity, discreet: false };
+        let account = Account::New { network: 0, seed, vanity, discreet: false, save_to_file: None };
         let actual = account.parse().unwrap();
         assert_eq!(expected, actual);
     }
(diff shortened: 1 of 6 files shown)


cli/src/commands/account.rs
```diff
@@ -97,12 +101,20 @@ fn aleo_literal_to_fields<N: Network>(input: &str) -> Result<Vec<Field<N>>> {
...
                     bail!("Cannot specify both the '--seed' and '--vanity' flags");
                 }
 
+                if save_to_file.is_some() && vanity.is_some() {
+                    bail!("Cannot specify both the '--save-to-file' and '--vanity' flags");
+                }
+
+                if save_to_file.is_some() && discreet {
+                    bail!("Cannot specify both the '--save-to-file' and '--discreet' flags");
+                }
```
ljedrz, line 117: this would work fine, but I think [conflicts_with_all](https://docs.rs/clap/latest/clap/struct.Arg.html#method.conflicts_with_all) could be used instead together with the `clap` attribute on `save_to_file` for brevity

cli/src/commands/account.rs
```diff
@@ -242,6 +254,12 @@ impl Account {
             PrivateKey::try_from(seed).map_err(|_| anyhow!("Failed to convert the seed into a valid private key"))?;
         // Construct the account.
         let account = snarkos_account::Account::<N>::try_from(private_key)?;
+        // Save to file instead of printing it back to the user
+        if let Some(path) = save_to_file {
+            let mut file = File::create(path)?;
```
ljedrz: I wonder if [create_new](https://doc.rust-lang.org/std/fs/struct.File.html#method.create_new) wouldn't be a bit safer, just to avoid an unintentional overwrite

cli/src/commands/account.rs
```diff
@@ -242,6 +254,12 @@ impl Account {
...
+            file.write_all(account.private_key().to_string().as_bytes())?;
+            return Ok(String::from(""));
```
ljedrz: shouldn't the account value also be returned here, especially considering the incompatibility with `--discreet`?

cli/src/commands/account.rs
Raphexion: After our discussion, I will keep the code as-is in this PR. Mostly because it was none-trivial to rework the code to `conflicts_with_all` combined with `--discreet`.

cli/src/commands/account.rs
Raphexion: Great catch :fishing_pole_and_fish: I have updated the code and added a unit-test for this issue. We don't want to accidentally overwrite a private-key-file. :sun_with_face:

cli/src/commands/account.rs
Raphexion: Great catch :fishing_pole_and_fish: I've changed the behavior and the test-cases. Now, we will not return early but continue.

Now the code continues like normal. Technically, it `save-to-file` could co-run with discreet. But I don't think it is a realistic use-case. Because one is dependent on the user looking at the screen, the other is for a programmatic solution.

cli/src/commands/account.rs
```diff
@@ -389,11 +410,70 @@ mod tests {
...
+        assert!(actual.contains(expected));
+
+        let content = fs::read_to_string(file).expect("Failed to read private-key-file");
+        assert_eq!(expected, content);
+    }
+
+    #[test]
+    fn test_new_prevent_overwrite_existing_file() {
+        let mut file = NamedTempFile::new().expect("Failed to create temp file");
+        write!(file, "{}", "don't overwrite me").expect("Failed to write secret to file");
```
ljedrz:
```suggestion
        write!(file, "don't overwrite me").expect("Failed to write secret to file");
```
nit

ljedrz:
One more thing I forgot before: we need to adjust the file's permissions too; see [this PR](https://github.com/AleoNet/snarkOS/pull/2998) for reference.

cli/src/commands/account.rs
Raphexion: Thanks :sun_with_face:  Fixed in d2d142f6e

Raphexion:
@ljedrz Thank you for the link to the PR. I've implemented the permissions feature and unit test in 8ce298296

cli/src/commands/account.rs
```diff
@@ -258,6 +260,7 @@ impl Account {
         if let Some(path) = save_to_file {
             let mut file = File::create_new(path)?;
             file.write_all(account.private_key().to_string().as_bytes())?;
+            file.set_permissions(PermissionsExt::from_mode(0o400))?;
```
ljedrz: this looks fine, but we'll also need to check the [parent](https://doc.rust-lang.org/std/path/struct.Path.html#method.parent) folder's permissions are `0o700`; this can be done right before creating the new file

cli/src/commands/account.rs
```diff
@@ -430,14 +434,19 @@ mod tests {
         let expected = "APrivateKey1zkp2n22c19hNdGF8wuEoQcuiyuWbquY6up4CtG5DYKqPX2X";
         assert!(actual.contains(expected));
 
-        let content = fs::read_to_string(file).expect("Failed to read private-key-file");
+        let content = fs::read_to_string(file.clone()).expect("Failed to read private-key-file");
```
ljedrz:
```suggestion
        let content = fs::read_to_string(&file).expect("Failed to read private-key-file");
```

cli/src/commands/account.rs
```diff
@@ -389,11 +414,75 @@ mod tests {
...
-        let account = Account::New { network: 0, seed, vanity, discreet: false };
+        let account = Account::New { network: 0, seed, vanity, discreet: false, save_to_file: None };
         let actual = account.parse().unwrap();
         assert_eq!(expected, actual);
     }
 
+    #[test]
+    fn test_new_save_to_file() {
+        let dir = TempDir::new().expect("Failed to create temp folder");
+        let file = dir.path().join("my-private-key-file").to_string_lossy().to_string();
```
ljedrz:
```suggestion
        let file = dir.path().to_owned().push("my-private-key-file");
```
this would retain strong path typing

cli/src/commands/account.rs
```diff
@@ -389,11 +414,75 @@ mod tests {
...
+
+        let seed = Some(1231275789u64.to_string());
```
ljedrz, line 432:
```suggestion
        let seed = Some("1231275789u64");
```
nit: most of these test strings don't need to be heap-allocated `String` objects, they could all just be literals

cli/src/commands/account.rs
Raphexion: Great idea. Fixed in ef05e86454fafea8803b84e7c182d5b9ec777bd5 :sunny:

cli/src/commands/account.rs
Raphexion: Fixed in 98b68f456976a582b851cfba02c24e359f38c6f6 :sunny:

cli/src/commands/account.rs
```diff
@@ -328,16 +349,40 @@ fn wait_for_keypress() {
...
+        let path = Path::new(path_str);
+        if let Some(parent) = path.parent() {
+            let permissions = parent.metadata()?.permissions().mode();
+            ensure!(permissions & 0o777 == 0o700, "The folder {:?} must be readable only by the owner (0700)", parent);
+        } else {
+            bail!("Parent does not exist for path={}", path_str);
+        }
+    }
+    Ok(())
+}
```
ljedrz: isn't this already a part of `commands/start.rs`? in that case it should be moved to `cli/src/lib.rs`, made `pub`lic, and imported in both `start.rs` and here

Raphexion: Fixed in 9d9eac039 :sunny:

cli/src/lib.rs
```diff
@@ -21,3 +21,27 @@ extern crate thiserror;
 
 pub mod commands;
 pub mod helpers;
+
+use anyhow::Result;
+use std::path::PathBuf;
+
+pub fn check_parent_permissions(path: &PathBuf) -> Result<()> {
```
ljedrz:
```suggestion
pub fn check_parent_permissions<T: AsRef<Path>>(path: T) -> Result<()> {
```
this should allow both `&str` (or `&String`) and `&PathBuf` to work, making `check_parent_permissions_for_string` redundant

Raphexion: Great suggestion. Fixed :sun_with_face:

cli/src/commands/account.rs
Raphexion: The parameter needs to be `Some(String)` this suggestion doesn't work in this exact example. However, great find and I will remember it in future changes.

ljedrz (approved):
LGTM :ok_hand:

zkxuerb (dismissed):
LGTM

zosorock:
it is failing the Windows CI for some reason:
https://app.circleci.com/pipelines/github/AleoNet/snarkOS/630/workflows/7000df5d-0288-41b9-b93a-287cdf3f8099/jobs/7806

Raphexion:
I can see the problem in the CI. I will fix :+1:

cli/src/lib.rs
```diff
@@ -23,20 +23,42 @@ pub mod commands;
...
-            bail!("Parent does not exist for path={:?}", path);
-        }
+    use anyhow::{bail, ensure};
+    use std::os::unix::fs::PermissionsExt;
+
+    if let Some(parent) = path.as_ref().parent() {
+        let permissions = parent.metadata()?.permissions().mode();
+        ensure!(permissions & 0o777 == 0o700, "The folder {:?} must be readable only by the owner (0700)", parent);
+    } else {
+        bail!("Parent does not exist for path={:?}", path);
```
ljedrz: nit: if you added a new binding like `let path = path.as_ref();` at the beginning of this function, you could remove the `std::fmt::Debug` trait bound

ljedrz (approved):
Left one suggestion, other than that LGTM :+1:.

cli/src/lib.rs
Raphexion: Great idea :star: Fixed :sunny:

zkxuerb (approved):

Raphexion:
@zkxuerb I fixed the formatting that was detected by the CI (`check-fmt`). I am sorry about that.

zkxuerb (approved):
LGTM (once Ci is re-run and is clear).

ljedrz (approved):
LGTM :+1:

Pjrich1313 (commented):
Debug

Pjrich1313 (commented):
Anything I can contribute to the cause I will do

PR #3427, opened 2024-10-31
[Perf] Rate-limit unique Certificates from unique peers

vicsn:
## Motivation

Certificates take a long time to process (up to `MAX_CERTIFICATES` signatures). Moreover, primary certificates are rebroadcasted every 5 seconds. Under heavy load, when rounds take longer, this means a lot of work is spent processing duplicate certificates. This PR implements some simple rate limiting checks.

Perhaps, in the future the id check can be encapsulated within the `Data` object.

## Test Plan

- [x] Local test
- [ ] Remote test

### DIFF
--- a/node/bft/events/src/primary_ping.rs
+++ b/node/bft/events/src/primary_ping.rs
@@ -19,6 +19,7 @@ use super::*;
 pub struct PrimaryPing<N: Network> {
     pub version: u32,
     pub block_locators: BlockLocators<N>,
+    pub certificate_id: Field<N>,
     pub primary_certificate: Data<BatchCertificate<N>>,
 }
 
@@ -27,16 +28,17 @@ impl<N: Network> PrimaryPing<N> {
     pub const fn new(
         version: u32,
         block_locators: BlockLocators<N>,
+        certificate_id: Field<N>,
         primary_certificate: Data<BatchCertificate<N>>,
     ) -> Self {
-        Self { version, block_locators, primary_certificate }
+        Self { version, block_locators, certificate_id, primary_certificate }
     }
 }
 
 impl<N: Network> From<(u32, BlockLocators<N>, BatchCertificate<N>)> for PrimaryPing<N> {
     /// Initializes a new ping event.
     fn from((version, block_locators, primary_certificate): (u32, BlockLocators<N>, BatchCertificate<N>)) -> Self {
-        Self::new(version, block_locators, Data::Object(primary_certificate))
+        Self::new(version, block_locators, primary_certificate.id(), Data::Object(primary_certificate))
     }
 }
 
@@ -54,6 +56,8 @@ impl<N: Network> ToBytes for PrimaryPing<N> {
         self.version.write_le(&mut writer)?;
         // Write the block locators.
         self.block_locators.write_le(&mut writer)?;
+        // Write the certificate id.
+        self.certificate_id.write_le(&mut writer)?;
         // Write the primary certificate.
         self.primary_certificate.write_le(&mut writer)?;
 
@@ -67,11 +71,13 @@ impl<N: Network> FromBytes for PrimaryPing<N> {
         let version = u32::read_le(&mut reader)?;
         // Read the block locators.
         let block_locators = BlockLocators::read_le(&mut reader)?;
+        // Read the certificate id.
+        let certificate_id = Field::read_le(&mut reader)?;
         // Read the primary certificate.
         let primary_certificate = Data::read_le(&mut reader)?;
 
         // Return the ping event.
-        Ok(Self::new(version, block_locators, primary_certificate))
+        Ok(Self::new(version, block_locators, certificate_id, primary_certificate))
     }
 }
 
--- a/node/bft/src/gateway.rs
+++ b/node/bft/src/gateway.rs
@@ -17,13 +17,14 @@ use crate::{
     CONTEXT,
     MAX_BATCH_DELAY_IN_MS,
     MEMORY_POOL_PORT,
+    PRIMARY_PING_IN_MS,
     Worker,
-    events::{EventCodec, PrimaryPing},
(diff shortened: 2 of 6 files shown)


ljedrz:
> Perhaps, in the future the id check can be encapsulated within the Data object.

Hmm, I wonder if something like
```
impl<T: FromBytes + ToBytes + Send + 'static> Data<T> {
    (...)

    pub fn deserialize_blocking_partial<U: FromBytes>(&self) -> Result<U> {
        match self {
            Self::Object(_) => Err(error("Can't partially deserialize unserialized objects").into()),
            Self::Buffer(bytes) => U::from_bytes_le(bytes),
        }
    }
}
```
wouldn't suffice? You'd then just need to call
```
// a blocking task would likely be an overkill just to read a single field
let certificate_id = certificate.deserialize_blocking_partial::<Field<N>>(&batch_certified)?;
```
[here](https://github.com/AleoNet/snarkOS/pull/3427/files#diff-8cf4dc35dbc765800ac3ba1ae72c96036548f24681f36c88ad606b84ef31d252R617-R618) in order to read just the ID.

vicsn:
> Hmm, I wonder if something like wouldn't suffice? You'd then just need to call

Fancy idea! Would that depend on the assumption that `batch_id` remains the first item in the Struct? If so, seems like a bit of an opaque assumption, hard for future developers to take that into account.

ljedrz:
> Would that depend on the assumption that batch_id remains the first item in the Struct?

It does indeed, though it isn't much different from any similar changes - due to the way we parse things, we might end up with corrupt (but fully parsed) objects if someone were to reorder the members; this could be documented.

To expand on it a little bit, note that the way parsing is currently done is by stringing calls to `Object::read_le(..)` without any member identifiers (which would most likely be an obligatory component of the `FromBytes` trait).

node/bft/events/src/primary_ping.rs
```diff
@@ -19,6 +19,7 @@ use super::*;
 pub struct PrimaryPing<N: Network> {
     pub version: u32,
     pub block_locators: BlockLocators<N>,
+    pub certificate_id: Field<N>,
```
raychu86, line 22: Changes to PrimaryPIng or messages in general will necessitate an update to the `Event` Version.

This will require a coordinated restart of validators.

node/bft/src/gateway.rs
```diff
@@ -561,6 +566,21 @@ impl<N: Network> Gateway<N> {
...
+                let certificate_id = match &event {
+                    Event::PrimaryPing(PrimaryPing { certificate_id, .. }) => *certificate_id,
+                    Event::BatchCertified(BatchCertified { certificate_id, .. }) => *certificate_id,
+                    _ => unreachable!(),
+                };
+                // Skip processing duplicate primary certificates received from the same peer.
+                let num_events = self
+                    .cache
+                    .insert_inbound_primary_certificate((peer_ip, certificate_id), CACHE_PRIMARY_CERTIFICATES_INTERVAL);
+                if num_events >= 2 {
```
raychu86, line 580: Current attributes are (for messages with the same certiifcate id):

- Only 1 can be processed at a time
- Can be processed again after 15 seconds

How can we dial this number to be safe? I feel like malicious validators might be able to prevent other nodes from syncing certificates by:

1. Send a certain batch certificate to an honest node.
2. Withholding the internal data to prevent the honest node from accurately processing it.
3. Repeat every 15 seconds OR send many of the same message to fill up the cache.
4. This will prevent the honest node from processing the certificate from other honest validators.

raychu86:
I like this approach! Currently the only check we use to stop duplicate processing in `process_batch_certificate_from_peer` is the following:
```
        // Ensure storage does not already contain the certificate.
        if self.storage.contains_certificate(certificate.id()) {
            return Ok(());
        }
```

This PR goes further and prevents duplicate `PrimaryPing`s or `BatchCertified` from being deserialized and processed if the node is already processing a certain number of these objects.

I have 2 questions:

1. Is there a way to do this without having to change the `Event::PrimaryPing` types? This requires coordinated restarts which could be a bit disruptive to network operations. I would prefer to reduce this as much as necessary.

2. How can we test that this improves functionality? In theory it's clear that it should prevent excess load from duplicate certificate processing, but is there a way to gauge this? This is a bit more involved of a question, but some sort of measure could be useful for future optimizations as well

node/bft/src/gateway.rs
vicsn: This is indeed the type of attack I'm worried about. But the attack you mention won't work because we use `(peer_ip, certificate_id)` as key for the rate limiting.

Made it a bit more lenient from 15 to 10 seconds: 084d71f10

vicsn:
> Is there a way to do this without having to change the Event::PrimaryPing types? This requires coordinated restarts which could be a bit disruptive to network operations. I would prefer to reduce this as much as necessary.

Building on Lukasz' suggestion above, how do you feel about having a `fn from_bytes_le_partial` of `fn id_from_bytes` for both `BatchCertificate` and `BatchHeader`?

> How can we test that this improves functionality? In theory it's clear that it should prevent excess load from duplicate certificate processing, but is there a way to gauge this? This is a bit more involved of a question, but some sort of measure could be useful for future optimizations as well

- Flamegraphs on unit tests with 40 validators have shown deserialization of BatchCertificates outweighs most other computation (except for transaction verification)
- I applied some awesome TCP Gateway metrics feature from @miazn on a private staging branch (they were considered too spammy for mainnet). Deployed networks showed Certificates to be a significant portion of bytes received:
Screenshot 2024-11-22 at 17 41 55
- Manual review of logs showed that at under stress, with a blocktime of 30+ seconds, the same certificate is received 4+ times.

vicsn:
Closing this PR for now, networks of 10+ validators with this PR halt. 

Besides, the performance benefit only really shines in the bad case (long blocktimes), which empirically has not happened yet. And there are other more important design changes required to reduce the risk and impact of long blocktimes. I'm supportive of efforts to revive this PR in the future.

PR #3430, opened 2024-11-05
Optimize BlockLocators Iterator with Parallel Processing

meetrick:
## Motivation

This PR aims to optimize the BlockLocators into_iter method by introducing parallel processing. The existing sequential processing approach can be time-consuming when handling large datasets of block data. By leveraging rayon for parallel processing, this change seeks to improve performance in merging checkpoints and recents maps, providing a more efficient solution for large-scale data handling.

## Test Plan

Generate a dataset of 1000 items to compare processing speeds.

## Related PRs

(Link any related PRs here)

### DIFF
--- a/node/sync/locators/src/block_locators.rs
+++ b/node/sync/locators/src/block_locators.rs
@@ -15,10 +15,13 @@
 
 use snarkvm::prelude::{FromBytes, IoResult, Network, Read, ToBytes, Write, error, has_duplicates};
 
+use rayon::prelude::*;
+use rayon::iter::IntoParallelIterator;
 use anyhow::{Result, bail, ensure};
 use indexmap::{IndexMap, indexmap};
 use serde::{Deserialize, Serialize};
 use std::collections::{BTreeMap, btree_map::IntoIter};
+use std::iter::FromIterator;
 
 /// The number of recent blocks (near tip).
 pub const NUM_RECENT_BLOCKS: usize = 100; // 100 blocks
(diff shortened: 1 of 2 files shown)


meetrick:
hey guys!

I expected that processing speed would increase with parallel processing. However, after testing, I observed that up to a certain number of items, parallel processing introduced overhead, which caused it to take longer. Based on my test code, sequential processing performed better up to approximately 100,000 items. The performance is also affected by the machine’s capabilities.

I would appreciate reviewers’ opinions on this. It seems that additional changes may be needed in the backend to fully accommodate these updates.

node/sync/locators/src/block_locators.rs
```diff
@@ -61,11 +64,14 @@ impl<N: Network> IntoIterator for BlockLocators<N> {
...
-    // TODO (howardwu): Consider using `BTreeMap::from_par_iter` if it is more performant.
-    //  Check by sorting 300-1000 items and comparing the performance.
-    //  (https://docs.rs/indexmap/latest/indexmap/map/struct.IndexMap.html#method.from_par_iter)
     fn into_iter(self) -> Self::IntoIter {
-        BTreeMap::from_iter(self.checkpoints.into_iter().chain(self.recents)).into_iter()
+        BTreeMap::from_iter(
+            self.checkpoints
+            .into_par_iter()
+            .chain(self.recents.into_par_iter())
+            .collect::<Vec<_>>(),
```
ljedrz, line 72: can't this be collected directly into a `BTreeMap` via `from_par_iter`?

ljedrz:
@meetrick in general, I'm not convinced that parallelizing the iteration itself is likely to be a performance improvement at any scale, _unless_ we can utilize `rayon` to do the sorting that's involved when creating the `BTreeMap`; in other words, the operation that we should consider is [BTreeMap::from_par_iter](https://docs.rs/rayon/latest/rayon/iter/trait.FromParallelIterator.html#tymethod.from_par_iter), as indicated in the removed comment.

meetrick:
@ljedrz  It seems I misunderstood the TODO. I’ll work on it again. Thanks for your input.

ljedrz:
In addition, it is possible that sorting is not always necessary; I don't recall the related logic very well, but it seems to me that `checkpoints` and `recents` might already be in the right order at least in some scenarios. in such cases, avoiding the iteration and collecting to a `BTreeMap` would have a positive impact on performance.

PR #3432, opened 2024-11-08
feat: Add support for both ip and hostname for peers

Raphexion:
## Motivation

It is convinent to be able to specify both IPs and hostnames as peers. Especially, when you are building ephemeral testnets and want to specify peers as

```
./snarkos start --peers "validator0:4130,client0:4130"
```

Where you can populate the `/etc/hosts` file with subnet hostnames.

## Test Plan

Added unit-tests but can be tested on the commandline

## Related PRs

None.

### DIFF
--- a/cli/src/commands/start.rs
+++ b/cli/src/commands/start.rs
@@ -42,7 +42,7 @@ use rand::{Rng, SeedableRng};
 use rand_chacha::ChaChaRng;
 use serde::{Deserialize, Serialize};
 use std::{
-    net::SocketAddr,
+    net::{SocketAddr, ToSocketAddrs},
     path::PathBuf,
     sync::{Arc, atomic::AtomicBool},
 };
@@ -230,11 +230,18 @@ impl Start {
             false => Ok(self
                 .peers
                 .split(',')
-                .flat_map(|ip| match ip.parse::<SocketAddr>() {
-                    Ok(ip) => Some(ip),
-                    Err(e) => {
-                        eprintln!("The IP supplied to --peers ('{ip}') is malformed: {e}");
-                        None
+                .flat_map(|ip_or_hostname| {
+                    let trimmed = ip_or_hostname.trim();
+                    match trimmed.to_socket_addrs() {
+                        Ok(mut ip_iter) => {
+                            // A hostname might resolve to multiple IP addresses. We will use only the first one,
+                            // assuming this aligns with the user's expectations.
+                            ip_iter.next()
+                        },
+                        Err(e) => {
+                            eprintln!("The hostname or IP supplied to --peers ('{trimmed}') is malformed: {e}");
+                            None
+                        }
                     }
                 })
                 .collect()),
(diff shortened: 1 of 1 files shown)


cli/src/commands/start.rs
```diff
@@ -228,11 +228,14 @@ impl Start {
...
-                        eprintln!("The IP supplied to --peers ('{ip}') is malformed: {e}");
-                        None
+                .flat_map(|ip_or_hostname| {
+                    let trimmed = ip_or_hostname.trim();
+                    match trimmed.to_socket_addrs() {
+                        Ok(mut addr) => addr.next(),
+                        Err(e) => {
+                            eprintln!("The hostname or IP supplied to --peers ('{trimmed}') is malformed: {e}");
+                            None
+                        }
```
ljedrz: this is a little bit convoluted; what we should be doing is first collect the peers (once split by `,`) to a vector, and then call `to_socket_addrs()` on it, and collect it into a `Result<Vec<SocketAddr, _>>`, which would be the final output

ljedrz: discussed this offline, and it's fine as long as only the 1st resolved IP is of interest and it's documented :+1:

Raphexion: Great suggestion (incl offline discussion). I've added a comment to clarify. Moreover, I've change the variable name as you suggested. To make it easier to read the code. :sun_with_face:

cli/src/commands/start.rs
```diff
@@ -1039,4 +1046,60 @@ mod tests {
...
+
+    #[test]
+    fn parse_peers_when_ips() {
+        let arg_vec = vec!["snarkos", "start", "--peers", "127.0.0.1:3030,127.0.0.2:3030"];
+        let cli = CLI::parse_from(arg_vec);
+
+        if let Command::Start(start) = cli.command {
+            let peers = start.parse_trusted_peers();
+            assert!(peers.is_ok());
+            assert!(peers.unwrap().len() == 2, "Expected two peers");
```
ljedrz: nit: `assert_eq` could be used here

cli/src/commands/start.rs
```diff
@@ -1039,4 +1046,60 @@ mod tests {
...
+
+    #[test]
+    fn parse_peers_when_hostnames() {
+        let arg_vec = vec!["snarkos", "start", "--peers", "www.example.com:4130,www.google.com:4130"];
+        let cli = CLI::parse_from(arg_vec);
+
+        if let Command::Start(start) = cli.command {
+            let peers = start.parse_trusted_peers();
+            assert!(peers.is_ok());
+            assert!(peers.unwrap().len() == 2, "Expected two peers");
```
ljedrz: ditto

cli/src/commands/start.rs
```diff
@@ -1039,4 +1046,60 @@ mod tests {
...
+
+    #[test]
+    fn parse_peers_when_mixed_and_with_whitespaces() {
+        let arg_vec = vec!["snarkos", "start", "--peers", "  127.0.0.1:3030,  www.google.com:4130 "];
+        let cli = CLI::parse_from(arg_vec);
+
+        if let Command::Start(start) = cli.command {
+            let peers = start.parse_trusted_peers();
+            assert!(peers.is_ok());
+            assert!(peers.unwrap().len() == 2, "Expected two peers");
```
ljedrz: ditto

cli/src/commands/start.rs
```diff
@@ -1039,4 +1046,60 @@ mod tests {
...
+
+    #[test]
+    fn parse_peers_when_unknown_hostname_gracefully() {
+        let arg_vec = vec!["snarkos", "start", "--peers", "banana.cake.eafafdaeefasdfasd.com"];
+        let cli = CLI::parse_from(arg_vec);
+
+        if let Command::Start(start) = cli.command {
+            let peers = start.parse_trusted_peers();
+            assert!(peers.is_ok());
+            assert!(peers.unwrap().len() == 0, "Expected zero peers");
```
```suggestion
            assert!(peers.unwrap().is_empty());
```

cli/src/commands/start.rs
Raphexion: Thanks :sun_with_face: Fixed :heavy_check_mark:

cli/src/commands/start.rs
Raphexion: Thanks :sun_with_face: Fixed :heavy_check_mark:

cli/src/commands/start.rs
Raphexion: Thanks :sun_with_face: Fixed :heavy_check_mark:

cli/src/commands/start.rs
Raphexion: Thanks :sun_with_face: Fixed :heavy_check_mark:

ljedrz (approved):
LGTM :ok_hand:

cli/src/commands/start.rs
```diff
@@ -228,11 +228,18 @@ impl Start {
...
-                .flat_map(|ip| match ip.parse::<SocketAddr>() {
-                    Ok(ip) => Some(ip),
-                    Err(e) => {
-                        eprintln!("The IP supplied to --peers ('{ip}') is malformed: {e}");
-                        None
+                .flat_map(|ip_or_hostname| {
+                    let trimmed = ip_or_hostname.trim();
+                    match trimmed.to_socket_addrs() {
+                        Ok(mut ip_iter) => {
+                            // A hostname might resolve to multiple IP addresses. We will use only the first one,
```
vicsn, line 237: To avoid any surprises for node operators, is there any known syntax for selecting a particular index? Should we invent our own, e.g. <hostname:port:1>

vicsn:
Though its a small and useful change, let's make sure there's buy-in and awareness from all dev orgs that the nodes now also rely on DNS, incurring a small new privacy and security risk.

Also worth looking at geth's recent addition, which apparently also picks the first ip if multiple are defined. :) https://github.com/ethereum/go-ethereum/pull/18524

cli/src/commands/start.rs
Raphexion: An alternative solution would be to enforce that only one IP is expected and allowed. So if there are more than one IP found, then we will print a warning and ignore the hostname.

The main idea behind this feature is to allow for a more flexible setup during tests. And in those cases, we can make sure that there is only one "validator-1" and only one "client-42". So I think it is a good trade-off, to enforce that only one IP is expected and found.

Raphexion:
@vicsn I agree :100: , we should make sure that this feature is wanted and accepted :+1: 

And great link to geth :pray: :sunny:

PR #3434, opened 2024-11-13
Resolves clippy issue with cfg test feature

zkxuerb:
## Motivation
Clippy was having issues with `#[cfg(not(any(test, feature = "test")))]` as the `test` feature does not exist. So this PR changes it to `#[cfg(not(any(test)))]`.


vicsn:
>  So this PR changes it to #[cfg(not(any(test)))]

Can you explain this in more depth? I don't see it in the PR contents.

raychu86:
I second @vicsn's question. The PR description do not match the changes itself.

Why is it safe to remove the cfg flags for these changes? Will this not invalidate tests?

zkxuerb:
Apologies used the wrong branch for PR.

PR #3435, opened 2024-11-13
Remove test feature from cfg

zkxuerb:
## Motivation

Clippy was having issues with `#[cfg(not(any(test, feature = "test")))]` as the `test` feature does not exist. So this PR changes it to `#[cfg(not(any(test)))]`.


vicsn (approved):
While `any` is superfluous, fine to leave as is, LGTM

raychu86 (approved):

PR #3437, opened 2024-11-14
Process transactions and solutions in parallel

vicsn:
## Motivation

As suggested by @ljedrz in [this comment](https://github.com/AleoNet/snarkOS/pull/3421/files#r1837739260), processes transactions and solutions in parallel. This is a small cleanup and speedup.

## Test Plan

- [x] Ran a local network

### DIFF
--- a/node/consensus/src/lib.rs
+++ b/node/consensus/src/lib.rs
@@ -479,7 +479,7 @@ impl<N: Network> Consensus<N> {
                 if let Err(e) =
                     tokio::try_join!(self_.process_unconfirmed_transactions(), self_.process_unconfirmed_solutions(),)
                 {
-                    warn!("Cannot process unconfirmed transactions and solutions - {e}");
+                    warn!("Cannot process unconfirmed transactions or solutions - {e}");
                 }
             }
         });
--- a/node/consensus/src/lib.rs
+++ b/node/consensus/src/lib.rs
@@ -475,13 +475,11 @@ impl<N: Network> Consensus<N> {
             loop {
                 // Sleep briefly.
                 tokio::time::sleep(Duration::from_millis(MAX_BATCH_DELAY_IN_MS)).await;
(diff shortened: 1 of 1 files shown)


node/consensus/src/lib.rs
```diff
@@ -475,13 +475,11 @@ impl<N: Network> Consensus<N> {
...
-                    warn!("Cannot process unconfirmed transactions - {e}");
-                }
-                // Process the unconfirmed solutions in the memory pool.
-                if let Err(e) = self_.process_unconfirmed_solutions().await {
-                    warn!("Cannot process unconfirmed solutions - {e}");
+                // Process the unconfirmed transactions and solutions in the memory pool.
+                if let Err(e) =
+                    tokio::try_join!(self_.process_unconfirmed_transactions(), self_.process_unconfirmed_solutions(),)
+                {
+                    warn!("Cannot process unconfirmed transactions and solutions - {e}");
```
```suggestion
                    warn!("Cannot process unconfirmed transactions or solutions - {e}");
```
nit

ljedrz (approved):
LGTM; also, a tiny technical note - this _may_ cause these operations to be performed in parallel, but they are bound to be run concurrently.

node/consensus/src/lib.rs
vicsn: Thx, I went back and forth on this one 😅

Pjrich1313 (approved):

zkxuerb (approved):

raychu86:
I think this will still run them concurrently unless the underlying function has it's own thread: https://dtantsur.github.io/rust-openstack/tokio/macro.try_join.html.

Also, I think this will lead to some unwanted consequences - if the first call fails, it will not run the other one.

vicsn:
> Also, I think this will lead to some unwanted consequences - if the first call fails, it will not run the other one.

Good catch that there's an early return on `Err`, don't think any potential speedup will be worth the risk - at this stage.

PR #3439, opened 2024-11-18
[Perf] Avoid fetching already seen Certificates

vicsn:
## Motivation

To reduce the chance validators have to fetch any certificates, this PR ensures they don't just check their storage of confirmed certificates, but also a cache of certificates which are currently being processed.

This PR is meant not to change any consensus logic.

The second commit is a small optimization to reduce temporary allocations.

## Test Plan

- [x] Local network passed.
- [x] Deployed network passed


node/bft/src/primary.rs
```diff
@@ -1668,14 +1667,23 @@ impl<N: Network> Primary<N> {
...
                 continue;
             }
-            // If we do not have the certificate, request it.
-            if !self.storage.contains_certificate(*certificate_id) {
+            // Check if the certificate already exists in storage.
+            if self.storage.contains_certificate(*certificate_id) {
+                continue;
+            }
+            // If we have not fully processed the certificate yet, store it.
+            if let Some(certificate) = self.storage.get_unprocessed_certificate(*certificate_id) {
```
vicsn, line 1686: NOTE: we can additionally resolve any outstanding certificate requests for this particular certificate.

node/bft/src/primary.rs
```diff
@@ -916,9 +916,18 @@ impl<N: Network> Primary<N> {
...
+            // Proceed to disconnect the validator.
+            self.gateway.disconnect(peer_ip);
+            bail!("Malicious peer - Received a batch certificate from an unauthorized validator IP ({peer_ip})");
+        }
         // Ensure storage does not already contain the certificate.
         if self.storage.contains_certificate(certificate.id()) {
             return Ok(());
+        // Otherwise, ensure ephemeral storage contains the certificate.
+        } else if !self.storage.contains_unprocessed_certificate(certificate.id()) {
+            self.storage.insert_unprocessed_certificate(certificate.clone())?;
```
raychu86, line 930: Why is the certificate not removed from the `unprocessed_certificates` cache when this function is completed?

vicsn: It is removed from the cache if the certificate is inserted into storage, so even sooner than waiting for the function to complete.

You might wonder whether a malicious validator could fill up the cache, and they could. The cache helps in the optimistic case to speed up processing.

raychu86 (approved):
Overall code and concept LGTM! 

I think some additional testing/brainstorming would give more confidence that this could not be exploited by malicious parties. Or if there are cases where it may be detrimental (regressions);

PR #3440, opened 2024-11-18
[Perf] Add cache for most recent blocks

vicsn:
## Motivation

Deserialization is one of the most expensive operations, because we're validating all data. We already have a `Committee` and `Transmission` cache in memory, we don't have a `Block` cache yet. This PR should have a significant performance impact, noting:
- internal testing showed that validators under load request on average 1 block every 5 blocks.
- clients, of course, request *all* blocks.

I decided against an `LruCache` because its more important we can always return **recent** blocks quickly to validators participating in consensus.

## Test Plan

- [x] local network passed
- [x] Deployed network passed


node/bft/ledger-service/src/ledger.rs
```diff
@@ -41,12 +42,15 @@ use std::{
 
 /// The capacity of the LRU holding the recently queried committees.
 const COMMITTEE_CACHE_SIZE: usize = 16;
+/// The capacity of the cache holding the highest blocks.
+const BLOCK_CACHE_SIZE: usize = 10;
```
ljedrz, line 47: for such a small cache I think something like a `CricularQueue` might be faster to query (even with linear search) and iterate, and pushing beyond capacity would automatically drop the oldest entry

vicsn: I'd rather not add a dependency for such a small performance gain.

Just in case you don't have to do too much research, and in case you like it, think a `VecDeque` would be worth it for similar reasons?

ljedrz: it could be, yes

vicsn: Benchmarked BTreeMap locally, does not seem worth it to refactor:
- Elapsed time 10_000 iterations of 100MB Get: 84.875µs
- Elapsed time 10_000 iterations of 100MB Insert: 136.667µs

vicsn: Definitely appreciate the recommendation and learning about the datastructure trade-offs

ljedrz: sounds good, if it's not a hot path we can probably spare several microseconds here :smiley:

node/bft/ledger-service/src/ledger.rs
```diff
@@ -120,13 +125,47 @@ impl<N: Network, C: ConsensusStorage<N>> LedgerService<N> for CoreLedgerService<
...
-        self.ledger.get_blocks(heights)
+        // Prepare blocks to collect.
+        let mut blocks = Vec::with_capacity(heights.len());
+        // Determine the lowest height.
+        let start_height = heights.start;
+        // Prepare missing heights to collect.
+        let mut missing_heights = BTreeSet::from_iter(heights.clone());
+        // For each height in the range, check if the block is in the cache.
+        for height in heights {
+            if let Some(block) = self.block_cache.read().get(&height) {
```
ljedrz: it would probably be more perf-friendly to keep a read guard for the entire duration of this loop

ljedrz: it could also be used to quickly skip the attempt to obtain blocks from the cache in case its 1st and last entries are outside of the requested range

node/bft/ledger-service/src/ledger.rs
```diff
@@ -120,13 +125,47 @@ impl<N: Network, C: ConsensusStorage<N>> LedgerService<N> for CoreLedgerService<
...
+                blocks.insert(index, block.clone());
+                missing_heights.remove(&height);
+            }
+        }
+        // Check if there are any blocks not found in the cache.
+        let missing_heights = match (missing_heights.first(), missing_heights.last()) {
+            // All blocks found in the cache.
+            (None, None) => return Ok(blocks),
+            // Some blocks found in the cache.
+            (Some(&start), Some(&end)) => start..end + 1,
```
ljedrz:
```suggestion
            (Some(&start), Some(&end)) => start..=end,
```

vicsn:
```
error[E0308]: mismatched types
   --> node/bft/ledger-service/src/ledger.rs:167:45
    |
167 |         for block in self.ledger.get_blocks(missing_heights)? {
    |                                  ---------- ^^^^^^^^^^^^^^^ expected `Range<u32>`, found `RangeInclusive<u32>`
    |                                  |
    |                                  arguments to this method are incorrect
```

node/bft/ledger-service/src/ledger.rs
vicsn: > it would probably be more perf-friendly to keep a read guard for the entire duration of this loop

Great idea: bb00286a4

> it could also be used to quickly skip the attempt to obtain blocks from the cache in case its 1st and last entries are outside of the requested range

Interesting idea, though I'd rather keep it as is and make the happy path (which matters for honest validators) fast and readable.

raychu86:
Mostly LGTM, 1 question I had would be what happens when the latest blocks are max size. Could this lead to a OOM issue if 10 of the largest possible blocks are held in memory?

vicsn:
@raychu86 

Thx for noting, I think right now we're safe, and this particular cache doesn't increase the risk significantly, but really appreciate the mention because we should include this in the criteria for increasing the validator set - I just updated Foundation documentation.

Current average case scenario: 25 validators, 50 unique transmissions, 5kb txs, 4 round blocks are common: 60MB/block. For the gentle price of 2 credits per tx, or 2400 credits per block, an attacker can create 100kb transactions, creating 20x larger memory footprint.

Besides this new cache, at any given time, a validator can be expected to have ~15 blocks in memory anyway (3 clients syncing 5 blocks at a time). If a client is far behind, they'll request 10x more in parallel. Even with very large blocks, this still falls within the memory requirements. But this will become an issue once the validator set increases, at which point the maximum tx size can be considered to be reduced.

raychu86 (dismissed):
LGTM

node/bft/ledger-service/src/ledger.rs
```diff
@@ -365,6 +379,12 @@ impl<N: Network, C: ConsensusStorage<N>> LedgerService<N> for CoreLedgerService<
         }
         // Advance to the next block.
         self.ledger.advance_to_next_block(block)?;
+        // Add the block to the block cache.
+        self.block_cache.write().insert(block.height(), block.clone());
+        // Prune the block cache if it exceeds the maximum size.
+        if self.block_cache.read().len() > BLOCK_CACHE_SIZE {
+            self.block_cache.write().pop_first();
+        }
```
ljedrz:
```suggestion
        {
            let mut block_cache = self.block_cache.write();
            block_cache.insert(block.height(), block.clone());
            // Prune the block cache if it exceeds the maximum size.
            if block_cache.len() > BLOCK_CACHE_SIZE {
                block_cache.pop_first();
            }
        }
```
considering the amount of potential interest in this shared resource, I think it'd be faster on average if we kept the write guard open throughout these operations; if I'm not mistaken, we will also always pop from the cache after the first few (currently 10) blocks

ljedrz (approved):
LGTM with a small suggestion

raychu86 (approved):

PR #3441, opened 2024-11-18
perf: Change link-time optimization in dev and test

Raphexion:
## Motivation

This change improves the development cycle in both dev and test, when running the same command multiple times in a row.

On my machine I can see improvements from minutes to seconds. For example when running `cargo run` and `cargo test`.

## Test Plan

Compare running this commands 3 times in a row with or without this patch

```
cargo test
```

```
cargo run accounts new
```

## Related PRs

None

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -150,7 +150,7 @@ debug-assertions = false
 
 [profile.dev]
 opt-level = 2
-lto = "thin"
+lto = "off"
 incremental = true
 
 [profile.test]


ljedrz (commented):
I'm in favor of this as long as it doesn't have a negative impact on CI test times (in which case it would still make sense for the `dev` profile).

vicsn:
Now that CI works again, do you mind rebasing on the latest `staging` branch to trigger it? And make an equivalent PR for snarkVM? Then we can see by how much the CI times are impacted. :)

Raphexion:
@vicsn Thanks for helping with this PR. I've created a similar PR for SnarkVM as you suggested:

https://github.com/ProvableHQ/snarkVM/pull/2600

vicsn:
Seems like the snarkOS CircleCI tests do run slower, 23 minutes on this PR instead of 17 or 18 minutes on recent PRs.

Can you remove the link-time optimization for the test profile? We can keep it as the default on the dev profile. :)

Also the `snarkos` job OOM'd but I haven't seen that before, so hopefully adjusting the lto option reduces compiler memory usage and CI will be green again.

vicsn (approved):
Thanks!

PR #3443, opened 2024-11-19
Clarify that only missing transmissions are checked

vicsn:
## Motivation

I spent some time looking to avoid deserialization of seen transmissions, before realising we already filter for seen transmissions inside of `sync_with_batch_header_from_peer`

## Test Plan

No testing needed, this is only a variable name change.


raychu86 (approved):
LGTM. Maybe more documentation would help to make this more clear.

PR #3444, opened 2024-11-19
[Perf] Avoid deserialization of previously added blocks

vicsn:
## Motivation

Validators often optimistically request blocks (given that `MAX_BLOCKS_BEHIND` is only 1). As a result, validators often receive a block which is incredibly expensive to deserialize, even though they already advanced.

This PR avoids deserialization if we already advanced to `request.end_height`.

As a future optimization, we can also meddle with the deserialization of `DataBlocks` so we can deserialize only *some* blocks if we already advanced.

## Test Plan

- [ ] Local network failed, seemingly at the point where we switch from Block syncing to Consensus syncing
- [ ] Ran deployed network

### DIFF
--- a/node/bft/src/gateway.rs
+++ b/node/bft/src/gateway.rs
@@ -661,6 +661,10 @@ impl<N: Network> Gateway<N> {
                     if !self.cache.remove_outbound_block_request(peer_ip, &request) {
                         bail!("Unsolicited block response from '{peer_ip}'")
                     }
+                    // Return early if we already advanced past the block request.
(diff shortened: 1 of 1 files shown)


ljedrz (approved):
LGTM 👌

vicsn:
This PR breaks our fragile syncing logic. After resetting a validator ledger, they're unable to sync to tip. Given that the performance benefit is relatively small, this topic can be revisited when syncing is more extensively documented and modeled.

PR #3448, opened 2024-11-26
[Fix] Update Github Action permissions for `release.yaml`

raychu86:
## Motivation

This PR updates the job permissions for the Jobs in `release.yaml` to allow for automated version releases when specific tags are created. 

The current repo permissions are intended to be restrictive, so this change allows **only** this job to have updated permissions.


zosorock (approved):
lgtm 👍

PR #3451, opened 2024-11-27
[Perf] Greedily verify Executions before block generation

vicsn:
## Motivation

When processing hundreds of transmissions during block generation, transaction verification is a large bottleneck, taking up 90% of compute. By verifying transactions ahead of time, verification will be cached in `partially_verified_transactions`, moving compute out of the hot path.

Some useful details:
- The current PR does not mean to change any consensus logic. We still sign proposals without awaiting succesful verification of transmissions, just like before.
- Solutions are not verified because there are only 4 per block, and https://github.com/AleoNet/ARCs/discussions/77 might change the game entirely.
- Deployments are not verified, because in the honest case there won't be a lot of them, and they consume a lot of memory so would require a lot more complexity (a queue and thread steadily popping from it)

## Test Plan

- [x] Deployed a network succesfully.

### DIFF
--- a/node/bft/src/worker.rs
+++ b/node/bft/src/worker.rs
@@ -305,6 +305,18 @@ impl<N: Network> Worker<N> {
             // All other combinations are clearly invalid.
             _ => false,
         };
+        // If the transmission is a deserialized execution, verify it immediately.
+        // This takes heavy transaction verification out of the hot path during block generation.
+        if let (TransmissionID::Transaction(tx_id, _), Transmission::Transaction(tx)) = (transmission_id, &transmission)
+        {
+            if let Data::Object(Transaction::Execute(..)) = tx {
+                let self_ = self.clone();
+                let tx_ = tx.clone();
+                tokio::spawn(async move {
+                    let _ = self_.ledger.check_transaction_basic(tx_id, tx_).await;
+                });
+            }
+        }
         // If the transmission ID and transmission type matches, then insert the transmission into the ready queue.
         if is_well_formed && self.ready.insert(transmission_id, transmission) {
             trace!(


raychu86:
Relevant PR - https://github.com/AleoNet/snarkOS/pull/3098. 

We initially did not verify transactions from other validator's proposals to prevent a block halting issue. In your case, I believe this is safe because it doesn't interfere with the consensus rules. 

However, it would be interesting to see the implications of this kind of change. I worry that this may open up avenues for malicious nodes to slow down other validators by spamming them with transactions.

vicsn:
> I worry that this may open up avenues for malicious nodes to slow down other validators by spamming them with transactions.

Valid concern, but note that:
- this only impacts execution verification, which is light (takes a couple ms).
- validators currently are only at 50% average CPU saturation. It is theoretically possible that the early tx verification eats up resources from, say, certificate deserialization. But in practice my E2E stress test shows a large speedup. Happy to share grafana links etc.
- assuming the cache works as intended, we are not increasing compute, merely moving it (you used the term frontloading before).

raychu86 (approved):
LGTM as long as we feel confident that this shouldn't introduce any new attack vectors!

niklaslong (approved):
LGTM 👍 

> But in practice my E2E stress test shows a large speedup

If the difference is substantial, we might want to consider handling more non-immediate expensive operations and the caching of their results in a more structured way. 

Other than this, I can think of one caveat: this approach assumes the spawned task completes before the hotpath queries the cache, otherwise the expensive operations will run twice. Though, I don't think this can be easily leveraged to put additional pressure on a node.

kpandl (approved):
LGTM!

As far as I understand, critical parts of transactions are finally verified again before inclusion into a block, in the speculation phase . So there shouldn't be an error if a tx is in `partially_verified_transactions`, and in the meantime, another tx is included in a block that invalidates the original tx in `partially_verified_transactions` (e.g., by spending credits or consuming a record).

PR #3452, opened 2024-12-03
fix content type for metrics

Raphexion:
## Motivation

When prometheus updates to v3, it is becoming stricter with the Content-Type of the response.
    
 https://prometheus.io/docs/prometheus/3.0/migration/#scrape-protocols
    
```
Prometheus v3 is more strict concerning the Content-Type header received when scraping. 
Prometheus v2 would default to the standard Prometheus text protocol if the target being 
scraped did not specify a Content-Type header or if the header was unparsable or unrecognised. 
This could lead to incorrect data being parsed in the scrape. Prometheus v3 will now fail the scrape in such cases.
```

For snarkOS to work with prometheus 3.0, we need to fix this. Fortunately, there is already a fix for this if we update our dependency on metrics-exporter-prometheus. So all we have to do is update the version.
    
From the metrics/metrics-exporter-prometheus CHANGELOG.md:
  
```
- Update the `Content-Type` response header to `text/plain`, matching the Exposition format
  specification. ([#496](https://github.com/metrics-rs/metrics/pull/496))
```

## Test Plan

Test with prometheus 3

## Related PRs

None

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -255,10 +255,42 @@ dependencies = [
 ]
 
 [[package]]
+name = "atomic-waker"
+version = "1.1.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1505bd5d3d116872e7271a6d4e16d81d0c8570876c8de68093a09ac269d8aac0"
+
+[[package]]
 name = "autocfg"
 version = "1.3.0"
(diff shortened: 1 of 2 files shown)


Raphexion:
I accidentally deleted the branch :facepalm: Sorry about that. I've pushed the branch again.

PR #3462, opened 2025-01-06
[Fix] Configure metrics compilation in root toml

### LINKED ISSUE #3379: [Bug] Clean up metrics feature flag
## 🐛 Bug Report

Lower level crates enable metrics by default, making the higher level `metrics` feature useless. See the appended diff for an improved setup [diff.txt](https://github.com/user-attachments/files/16597419/diff.txt).

This simple cleanup is not high priority by itself, but it is key that features are handled uniformly so that we don't mess up implementing consensus-critical features like [test_targets](https://github.com/AleoNet/snarkOS/pull/3378)

## Expected Behavior

Ideally the top-level `Cargo.toml` has metrics in its `default` feature list, which can be used to enable or disable all metrics functionality

niklaslong:
The `cli` crate currently enables metrics as a default feature, this PR moves the feature's configuration to the root level toml, enables it by default and introduces a `metrics` feature on `cli`. In addition, `snarks-node-metrics` is made optional. 

Correct configuration can be verified with a reverse-search for the metrics crate by feature:
- `cargo tree -e features -i snarkos-node-metrics` (which should show matches, metrics being enabled by default)
- `cargo tree -e features -i snarkos-node-metrics --no-default-features` (which should show no matches)

Fixes #3379. 

The last commit realigns the dependencies in the tcp crate, lmk if I should drop it or move it someplace else. 

Edit: prometheus metric sanity check was also successful with a small devnet.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -56,7 +56,7 @@ name = "snarkos"
 path = "snarkos/main.rs"
 
 [features]
-metrics = [ "snarkos-node-metrics", "snarkos-node/metrics" ]
+default = [ "snarkos-node-metrics", "snarkos-node/metrics", "snarkos-cli/metrics" ]
 history = [ "snarkos-node/history" ]
 test_targets = [ "snarkos-cli/test_targets" ]
 
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -17,7 +17,7 @@ license = "Apache-2.0"
 edition = "2021"
 
 [features]
-default = [ "snarkos-node/metrics" ]
+metrics = [ "snarkos-node/metrics" ]
 test_targets = [ "snarkvm/test_targets" ]
 
 [dependencies.aleo-std]
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -17,7 +17,7 @@ license = "Apache-2.0"
 edition = "2021"
 
 [features]
-metrics = [ "snarkos-node/metrics" ]
+metrics = [ "dep:metrics", "snarkos-node/metrics" ]
 test_targets = [ "snarkvm/test_targets" ]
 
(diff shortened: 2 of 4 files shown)


Cargo.toml
```diff
@@ -56,7 +56,7 @@ name = "snarkos"
 path = "snarkos/main.rs"
 
 [features]
-metrics = [ "snarkos-node-metrics", "snarkos-node/metrics" ]
+default = [ "snarkos-node-metrics", "snarkos-node/metrics", "snarkos-cli/metrics" ]
```
niklaslong, line 59: Adding `snarkos-cli/metrics` here isn't strictly necessary for the feature to be compiled in (as they are additive across shared dependencies) but it is more complete.

vicsn (dismissed):
LGTM! Want to un-draft, and mention in the PR message as well the result of a sanity check: compile snarkOS, run a small local devnet using `devnet.sh` and see if the `/metrics` endpoint has non-zero outputs.

niklaslong:
Done, sanity check sample output from the devnet prometheus metrics:
```
# TYPE snarkos_bft_leaders_elected_total counter
snarkos_bft_leaders_elected_total 169

# TYPE snarkos_bft_connecting_total gauge
snarkos_bft_connecting_total 1

# TYPE snarkos_bft_primary_certified_batches gauge
snarkos_bft_primary_certified_batches 215

# TYPE snarkos_blocks_coinbase_target gauge
snarkos_blocks_coinbase_target 536870911

# TYPE snarkos_bft_last_committed_round gauge
snarkos_bft_last_committed_round 220
```

vicsn:
@niklaslong can you merge in staging and resolve the conflicts? O:)

raychu86 (approved):
Fairly straightforward. LGTM

PR #3463, opened 2025-01-07
[Perf] Use rayon to obtain pending certificates

ljedrz:
Migrated from https://github.com/ProvableHQ/snarkOS/pull/1.

The certificate-related queries are all reads, so they should parallelize well. Other than substituting the now-serial calls to potentially (and by-default) parallel ones, this PR involves a simple refactoring related to how the pending certificates are collected.


niklaslong (approved):
LGTM 👍

raychu86 (approved):
Logic LGTM. 

I think the PR linked in the description is incorrect.

PR #3469, opened 2025-01-20
Fix flaky test_storage_coherence

vicsn:
## Motivation

This test failed in [recent CI](https://github.com/ProvableHQ/snarkOS/runs/35793859049) and sometimes locally too on macbook M2 Max. Let's make it a bit more lenient.

A more complete refactor would entail only checking if at least a quorum's worth of certificates is the same in a quorum's worth of nodes, but the current heuristic approach should be sufficient as a test case.


niklaslong (approved):
LGTM 👍

PR #3470, opened 2025-01-22
Add support for GET /{network}/block/{height}/header

Raphexion:
## Motivation

Currently one must query the whole block in order to get the information in the header. This commit mimics the existing endpoint to only get the transactions for a block.

With this commit it is possible to query only the headers for blocks. This will reduce both bandwidth and payload sizes for the one parsing the result.

## Test Plan

Start a local validator node and query the header of the first block. 

```
curl http://127.0.0.1:3030/canary/block/0/header
```

Second, query a non existing block

```
curl http://127.0.0.1:3030/canary/block/999999999/header
```

## Related PRs

None.

### DIFF
--- a/node/rest/src/lib.rs
+++ b/node/rest/src/lib.rs
@@ -153,6 +153,7 @@ impl<N: Network, C: ConsensusStorage<N>, R: Routing<N>> Rest<N, C, R> {
             .route(&format!("/{network}/block/:height_or_hash"), get(Self::get_block))
             // The path param here is actually only the height, but the name must match the route
             // above, otherwise there'll be a conflict at runtime.
+            .route(&format!("/{network}/block/:height_or_hash/header"), get(Self::get_block_header))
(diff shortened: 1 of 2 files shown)


vicsn (approved):

raychu86:
I recall this route pattern was a special case and there were some issues due to the overlap with `"/{network}/block/:height_or_hash"`. It's why the following comment was added:
```
            // The path param here is actually only the height, but the name must match the route
            // above, otherwise there'll be a conflict at runtime.
```

Adding functionality looks good to me, but can you confirm that the existing routes still work for the following:
- `/{network}/block/:height_or_hash`
- `/{network}/block/:height_or_hash/transactions`

Raphexion:
@raychu86 Thank you for the comment. I've double checked on my local machine and all three works for me

```
http://localhost:3030/canary/block/0
http://localhost:3030/canary/block/0/transactions
http://localhost:3030/canary/block/0/header
```

PR #3471, opened 2025-01-22
[Feature] Batch proposal spend limits

niklaslong:
This PR introduces spend limit checks on batch proposals at construction and prior to signing. ~~This requires some changes to batch construction: the workers are now drained only _after_ the transmissions have been checked for inclusion in a batch. This, to avoid reinserting transmissions into the memory pool once the spend limit is surpassed.~~ I have modified the `Ready` queue to support `O(1)` operations at the front, as well as inclusion checks and regular inserts, so the batch proposal logic could be simplified. We no longer need to defer draining and can instead optimistically remove the transmission from the queue for processing. There's an occasional `O(n)` cost to reset the internally tracked offset but this should happen at most once per epoch, when solutions are cleared from the queue. It was also necessary to expose a `compute_cost` function on the `LedgerService` trait—its internals might still be moved into snarkVM. 

This changeset will need to be refined and tested, hence the draft. CI is currently expected to fail.

Related PRs: 
- https://github.com/ProvableHQ/snarkVM/pull/2601 (required for merge)
- https://github.com/ProvableHQ/snarkVM/pull/2565 (previous discussion)

### DIFF
--- a/node/bft/ledger-service/src/ledger.rs
+++ b/node/bft/ledger-service/src/ledger.rs
@@ -375,6 +375,7 @@ impl<N: Network, C: ConsensusStorage<N>> LedgerService<N> for CoreLedgerService<
         Ok(())
     }
 
+    /// Computes the execution cost in microcredits for a transaction.
     fn compute_cost(&self, _transaction_id: N::TransactionID, transaction: Transaction<N>) -> Result<u64> {
         self.ledger.vm().compute_cost(&transaction)
     }
--- a/node/bft/ledger-service/src/ledger.rs
+++ b/node/bft/ledger-service/src/ledger.rs
@@ -379,7 +379,11 @@ impl<N: Network, C: ConsensusStorage<N>> LedgerService<N> for CoreLedgerService<
     /// Returns the storage and compute cost for a transaction in microcredits.
     /// This is used to limit the amount of compute in the block generation hot
     /// path. This does NOT represent the full costs which a user has to pay.
-    fn compute_cost(&self, _transaction_id: N::TransactionID, transaction: Transaction<N>) -> Result<u64> {
+    fn transaction_spent_cost_in_microcredits(
+        &self,
+        _transaction_id: N::TransactionID,
+        transaction: Transaction<N>,
+    ) -> Result<u64> {
         match &transaction {
             // Include the synthesis cost and storage cost for deployments.
             Transaction::Deploy(_, _, _, deployment, _) => {
--- a/node/bft/ledger-service/src/ledger.rs
+++ b/node/bft/ledger-service/src/ledger.rs
@@ -381,4 +381,29 @@ impl<N: Network, C: ConsensusStorage<N>> LedgerService<N> for CoreLedgerService<
         tracing::info!("\n\nAdvanced to block {} at round {} - {}\n", block.height(), block.round(), block.hash());
         Ok(())
     }
+
+    fn compute_cost(&self, _transaction_id: N::TransactionID, transaction: Data<Transaction<N>>) -> Result<u64> {
+        // TODO: move to VM or ledger?
+        let process = self.ledger.vm().process();
+
+        // Deserialize the transaction. If the transaction exceeds the maximum size, then return an error.
+        let transaction = match transaction {
+            Data::Object(transaction) => transaction,
+            Data::Buffer(bytes) => Transaction::<N>::read_le(&mut bytes.take(N::MAX_TRANSACTION_SIZE as u64))?,
+        };
+
+        // Collect the Optional Stack corresponding to the transaction if its an Execution.
+        let stack = if let Transaction::Execute(_, ref execution, _) = transaction {
+            // Get the root transition from the execution.
+            let root_transition = execution.peek()?;
+            // Get the stack from the process.
+            Some(process.read().get_stack(root_transition.program_id())?.clone())
+        } else {
+            None
+        };
+
+        use snarkvm::prelude::compute_cost;
+
+        compute_cost(&transaction, stack)
+    }
 }
--- a/node/bft/ledger-service/src/ledger.rs
+++ b/node/bft/ledger-service/src/ledger.rs
@@ -393,7 +393,7 @@ impl<N: Network, C: ConsensusStorage<N>> LedgerService<N> for CoreLedgerService<
         };
 
         // Collect the Optional Stack corresponding to the transaction if its an Execution.
-        let stack = if let Transaction::Execute(_, ref execution, _) = transaction {
+        let stack = if let Transaction::Execute(_, _, ref execution, _) = transaction {
             // Get the root transition from the execution.
             let root_transition = execution.peek()?;
             // Get the stack from the process.
--- a/node/bft/ledger-service/src/ledger.rs
+++ b/node/bft/ledger-service/src/ledger.rs
@@ -376,7 +376,7 @@ impl<N: Network, C: ConsensusStorage<N>> LedgerService<N> for CoreLedgerService<
         Ok(())
     }
 
-    /// Returns the storage and compute cost for a transaction in microcredits.
+    /// Returns the spent cost for a transaction in microcredits.
     /// This is used to limit the amount of compute in the block generation hot
     /// path. This does NOT represent the full costs which a user has to pay.
     fn transaction_spent_cost_in_microcredits(
--- a/node/bft/ledger-service/src/ledger.rs
+++ b/node/bft/ledger-service/src/ledger.rs
@@ -310,15 +310,8 @@ impl<N: Network, C: ConsensusStorage<N>> LedgerService<N> for CoreLedgerService<
     async fn check_transaction_basic(
         &self,
         transaction_id: N::TransactionID,
-        transaction: Data<Transaction<N>>,
+        transaction: Transaction<N>,
     ) -> Result<()> {
-        // Deserialize the transaction. If the transaction exceeds the maximum size, then return an error.
-        let transaction = spawn_blocking!({
-            match transaction {
-                Data::Object(transaction) => Ok(transaction),
-                Data::Buffer(bytes) => Ok(Transaction::<N>::read_le(&mut bytes.take(N::MAX_TRANSACTION_SIZE as u64))?),
-            }
-        })?;
         // Ensure the transaction ID matches in the transaction.
         if transaction_id != transaction.id() {
             bail!("Invalid transaction - expected {transaction_id}, found {}", transaction.id());
@@ -382,15 +375,9 @@ impl<N: Network, C: ConsensusStorage<N>> LedgerService<N> for CoreLedgerService<
         Ok(())
     }
 
-    fn compute_cost(&self, _transaction_id: N::TransactionID, transaction: Data<Transaction<N>>) -> Result<u64> {
+    fn compute_cost(&self, _transaction_id: N::TransactionID, transaction: Transaction<N>) -> Result<u64> {
         // TODO: move to VM or ledger?
         let process = self.ledger.vm().process();
-
-        // Deserialize the transaction. If the transaction exceeds the maximum size, then return an error.
-        let transaction = match transaction {
-            Data::Object(transaction) => transaction,
-            Data::Buffer(bytes) => Transaction::<N>::read_le(&mut bytes.take(N::MAX_TRANSACTION_SIZE as u64))?,
-        };
 
         // Collect the Optional Stack corresponding to the transaction if its an Execution.
         let stack = if let Transaction::Execute(_, _, ref execution, _) = transaction {
--- a/node/bft/ledger-service/src/ledger.rs
+++ b/node/bft/ledger-service/src/ledger.rs
@@ -23,9 +23,10 @@ use snarkvm::{
         puzzle::{Solution, SolutionID},
         store::ConsensusStorage,
     },
-    prelude::{Address, Field, FromBytes, Network, Result, bail, cfg_into_iter},
+    prelude::{Address, Field, FromBytes, Network, Result, bail, cfg_into_iter, deployment_cost, execution_cost_v2},
 };
 
+use anyhow::anyhow;
 use indexmap::IndexMap;
 use parking_lot::RwLock;
 use rayon::prelude::*;
@@ -375,8 +376,25 @@ impl<N: Network, C: ConsensusStorage<N>> LedgerService<N> for CoreLedgerService<
         Ok(())
     }
 
-    /// Computes the execution cost in microcredits for a transaction.
+    /// Returns the storage and compute cost for a transaction in microcredits.
+    /// This is used to limit the amount of compute in the block generation hot
+    /// path. This does NOT represent the full costs which a user has to pay.
     fn compute_cost(&self, _transaction_id: N::TransactionID, transaction: Transaction<N>) -> Result<u64> {
-        self.ledger.vm().compute_cost(&transaction)
+        match &transaction {
+            // Include the synthesis cost and storage cost for deployments.
+            Transaction::Deploy(_, _, _, deployment, _) => {
+                let (_, (storage_cost, synthesis_cost, _)) = deployment_cost(deployment)?;
+                storage_cost
+                    .checked_add(synthesis_cost)
+                    .ok_or(anyhow!("The storage and synthesis cost computation overflowed for a deployment"))
+            }
+            // Include the finalize cost and storage cost for executions.
+            Transaction::Execute(_, _, execution, _) => {
+                let (total_cost, (_, _)) = execution_cost_v2(&self.ledger.vm().process().read(), execution)?;
+                Ok(total_cost)
+            }
+            // Fee transactions are internal to the VM, they do not have a compute cost.
+            Transaction::Fee(..) => Ok(0),
+        }
     }
 }
--- a/node/bft/ledger-service/src/ledger.rs
+++ b/node/bft/ledger-service/src/ledger.rs
@@ -376,21 +376,6 @@ impl<N: Network, C: ConsensusStorage<N>> LedgerService<N> for CoreLedgerService<
     }
 
     fn compute_cost(&self, _transaction_id: N::TransactionID, transaction: Transaction<N>) -> Result<u64> {
-        // TODO: move to VM or ledger?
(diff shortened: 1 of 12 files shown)


node/bft/ledger-service/src/ledger.rs
```diff
@@ -404,4 +404,29 @@ impl<N: Network, C: ConsensusStorage<N>> LedgerService<N> for CoreLedgerService<
         tracing::info!("\n\nAdvanced to block {} at round {} - {}\n", block.height(), block.round(), block.hash());
         Ok(())
     }
+
+    fn compute_cost(&self, _transaction_id: N::TransactionID, transaction: Data<Transaction<N>>) -> Result<u64> {
+        // TODO: move to VM or ledger?
+        let process = self.ledger.vm().process();
+
+        // Deserialize the transaction. If the transaction exceeds the maximum size, then return an error.
+        let transaction = match transaction {
```
vicsn: Deserialization is extremely expensive. I would consider moving this calculation into `ledger.rs:check_transaction_basic`, where we already deserialize. Perhaps that function can return the compute cost? Or some other design?

node/bft/ledger-service/src/mock.rs
```diff
@@ -235,4 +235,9 @@ impl<N: Network> LedgerService<N> for MockLedgerService<N> {
         self.height_to_round_and_hash.lock().insert(block.height(), (block.round(), block.hash()));
         Ok(())
     }
+
+    /// TODO: is this reasonable?
```
vicsn: > /// TODO: is this reasonable?

I guess if our mock ledger doesn't have the `compute_cost` function, yes its reasonable, no need to extend every test abstraction which exists.

node/bft/ledger-service/src/prover.rs
```diff
@@ -186,4 +186,8 @@ impl<N: Network> LedgerService<N> for ProverLedgerService<N> {
     fn advance_to_next_block(&self, block: &Block<N>) -> Result<()> {
         bail!("Cannot advance to next block in prover - {block}")
     }
+
+    fn compute_cost(&self, transaction_id: N::TransactionID, _transaction: Data<Transaction<N>>) -> Result<u64> {
+        bail!("Transaction '{transaction_id}' doesn't exist in prover")
```
vicsn, line 196: Not sure where this is used, will this break prover nodes?

node/bft/src/primary.rs
```diff
@@ -490,88 +490,100 @@ impl<N: Network> Primary<N> {
...
-            'outer: while num_transmissions_included_for_worker < num_transmissions_per_worker {
-                // Determine the number of remaining transmissions for the worker.
-                let num_remaining_transmissions =
-                    num_transmissions_per_worker.saturating_sub(num_transmissions_included_for_worker);
-                // Drain the worker.
-                let mut worker_transmissions = worker.drain(num_remaining_transmissions).peekable();
-                // If the worker is empty, break early.
-                if worker_transmissions.peek().is_none() {
-                    break 'outer;
+            let mut worker_transmissions = worker.transmissions().into_iter();
```
vicsn: So correct me if I'm wrong, but the previous behaviour was to **selectively drain** transmissions, and if they are duplicates or are incorrect, we discard them.

Your current implementation **completely drains** transmissions, so besides discaring duplicates or invalid ones, once we hit the size or compute limit, remaining transmissions are discarded.

Some options:
1. preprocess the compute spend of a transmission and to cache it in the worker's `Ready` queue. But this would unfortunately require a lot of code changes and wouldn't make sense everywhere.
2. We drain transmissions one by one. If we notice post-deserialization that we're above the compute limit, we reinsert that one transmission back into the queue using `worker.reinsert`. This has strictly less overhead than draining all transmissions.

node/bft/src/primary.rs
```diff
@@ -490,88 +490,100 @@ impl<N: Network> Primary<N> {
...
+
+            // Check the transactions for inclusion in the batch proposal.
+            while num_transmissions_included_for_worker < num_transmissions_per_worker {
```
vicsn: Wondering whether it is possible and desirable to put this check in the same place as the compute cost check, because they're both indicating "the batch proposal is too full now".

The loop could then be something like `While let Some(transmission) = worker.drain(1)`.

Up to you, just a suggestion

vicsn (commented):
Hope it helps

node/bft/ledger-service/src/ledger.rs
niklaslong: I did consider it and should revisit the idea but intuitively we might want to avoid coupling the cost calculation with `check_transaction_basic` as it's only called in `propose_batch` and not in `process_batch_propose_from_peer`, at least not directly.

node/bft/ledger-service/src/prover.rs
niklaslong: I don't think so, this function is called directly in the primary, which provers don't implement. Returning an error if a prover should attempt to call `compute_cost` is probably reasonable?

node/bft/src/primary.rs
niklaslong: Previously the primary would preemptively drain the workers by batches, sized based on the remaining proposal space. This was fine as transactions were either included or dropped altogether but we need more nuance: we might choose to save a transaction for the following batch if it pushes the batch cost over the spend limit.

For the PoC, I opted to clone the transmissions (happens in `worker.transmissions()`), to avoid reinsertion logic initially, leaving the underlying collection untouched until the number of valid transactions is determined and draining it only at the end of the worker loop. Ideally, I'd want to `peek` the next item in the collection, and it might still be the most efficient. The next best thing would be to reinsert but that's still O(n) and requires more logic.

node/bft/ledger-service/src/ledger.rs
vicsn: Ah good point... That is unfortunate, because in the case of `process_batch_propose_from_peer`, we might be retrieving the transmission from disk, in which case we'll most certainly have to incur the deserialization cost.

Maybe if we let `fn compute_cost` cost take in a `Transaction` instead of `Data<Transaction>`, it can at least be made explicit. For our own proposal we call it from within `check_transaction_basic`, for incoming proposals we'll need to deserialize before calling it.

And if it turns out to be a bottleneck, we can always refactor the locations where we deserialize more comprehensively, and potentially create a cache for the compute_cache if needed.

node/bft/src/primary.rs
vicsn: Peeking one by one would work!

Or draining one by one. If we optionally have to reinsert the very last one we drained if it crosses the compute threshold, I'm not sure what would be O(n)

niklaslong: O(n) comes from shifting all the transmissions by one in the collection when reinserting at index 0 (to preserve fairness in the ordering of transmissions). Peeking directly likely won't work because it requires either holding the lock (not a good idea) or cloning. But perhaps doing the lookup one-by-one, followed by a removal could work. Though any operation that isn't inserting or removing values from the tail of the map will incur an O(n) shift cost.

vicsn: I see, thank you for explaining.

I take everything back, your approach seems close to optimal (assuming we don't change or introduce additional data structures).

> But perhaps doing the lookup one-by-one, followed by a removal could work. 

Indeed you can clone and lookup one-by-one (to avoid cloning the entire ready queue). And then batch drain at the end like you're already doing.

node/bft/ledger-service/src/mock.rs
```diff
@@ -235,4 +235,10 @@ impl<N: Network> LedgerService<N> for MockLedgerService<N> {
         self.height_to_round_and_hash.lock().insert(block.height(), (block.round(), block.hash()));
         Ok(())
     }
+
+    /// Computes the execution cost in microcredits for a transaction.
+    fn compute_cost(&self, _transaction_id: N::TransactionID, _transaction: Data<Transaction<N>>) -> Result<u64> {
+        // Return 1 credit so this function can be used to test spend limits.
```
raychu86: nit: 10_000_000 microcredits is 10 credits

node/bft/src/primary.rs
```diff
@@ -488,90 +488,99 @@ impl<N: Network> Primary<N> {
...
                                 _ => {
                                     trace!(
-                                        "Proposing - Skipping transaction '{}' - Checksum mismatch",
+                                        "Proposing - Skipping transaction '{}' - Batch spend limit surpassed",
                                         fmt_id(transaction_id)
                                     );
-                                    continue 'inner;
+
+                                    // Reinsert the transmission into the worker, O(n).
+                                    worker.shift_insert_front(id, transmission);
```
raychu86: If there is a huge transmission that is 99% of the batch spend limit, isn't your worker never going to create a proposal with a larger utilization?

raychu86: Might be smarter to pull other transactions that might fill the remaining space of the proposal, with smaller transactions. 

This also might be an opportune moment to finally implement the use of `priority_fee`, which was previously not factored into the ordering

vicsn: I think the priority fee should be tackled independently, it'll be a decent change to review:
- need to insert ahead of the transactions_queue (draft PR exists)
- need to insert ahead of the ready queue is a good extra idea.

niklaslong: > If there is a huge transmission that is 99% of the batch spend limit, isn't your worker never going to create a proposal with a larger utilization?

Indeed, we don't currently attempt to fill the batch with smaller transmissions after the huge one is skipped. We could maintain _some_ degree of fairness by continuing to fill batch `n` until the spend limit or the batch size is reached and prioritise the large skipped transmission(s) in batch `n + 1`. 

Not quite sure if new vectors are introduced with this approach or if it should be bundled with the priority fee changes but happy to make the adjustment here as well.

niklaslong:
Note to reviewers: I have modified the ready queue to support O(1) operations at the front, giving us an efficient fifo. This makes batch building logic simpler because we can optimistically remove the transmission from the queue before processing it for inclusion, with reinsertions at the front being required **only** for valid transmissions that bring the batch above the spend limit. This may also be useful for implementing priority fees in future.

node/bft/src/primary.rs
```diff
@@ -768,6 +787,47 @@ impl<N: Network> Primary<N> {
         // Inserts the missing transmissions into the workers.
         self.insert_missing_transmissions_into_workers(peer_ip, missing_transmissions.into_iter())?;
 
+        // Ensure the transaction doesn't bring the proposal above the spend limit.
+        let block_height = self.ledger.latest_block_height() + 1;
+        if N::CONSENSUS_VERSION(block_height)? >= ConsensusVersion::V4 {
```
raychu86, line 819: Note that the Consensus/BFT modules do not care about block heights and only care about rounds. 

This means different validators may be at same round, but different block height if one of the validators have not received enough certificates to advance their ledger. This means they may process incoming proposals differently, but neither are technically wrong from their point of view.

raychu86: Intuitively, this likely won't be an issue if the validator from behind syncs up with other certificates and catches their ledger up before processing the incoming proposal.

niklaslong: Let me check my understanding: the edge case is a straggling node processing transmissions into a batch without limiting spend when the limit is already enforced by other validators? Eventually all nodes should process the same subdag into a block at the same height so unless there is a quorum of nodes behind (impossible by definition?) there's no chance of a spend limit going unenforced because of asynchrony during the consensus upgrade? 

In other words, despite rounds not matching, block height is still dependable as a trigger for new logic in these modules? 🤔

acoglio: Perhaps we could use a lookback approach for how (the heights of) blocks determine the consensus version active at each round, similarly to how blocks determine the active (= lookback) committee at each round.

niklaslong:
Rebased

node/bft/src/primary.rs
```diff
@@ -488,90 +488,114 @@ impl<N: Network> Primary<N> {
...
-                            match transaction.to_checksum::<N>() {
-                                Ok(transaction_checksum) if transaction_checksum == checksum => (),
+
+                        // Ensure the transaction doesn't bring the proposal above the spend limit.
+                        //
+                        // Validators start following the new consensus rules early, in order to guarantee a smooth transition.
+                        // The activation height is offset by the maximum number of blocks that can be produced within gc to
+                        // ensure honest validators can't mismatch due to differing views of the dag.
+                        let block_height =
+                            self.ledger.latest_block_height() + 1 + (BatchHeader::<N>::MAX_GC_ROUNDS as u32 / 2);
```
raychu86: nit: use `saturating_*` for adds and divs

raychu86: Also maybe add further explanation on why the specific range was selected.

niklaslong: Added in [17ddf90](https://github.com/ProvableHQ/snarkOS/pull/3471/commits/17ddf908b06dae6aac0f812a6df7cc8338fecc07).

zklimaleo:
I'm actually curious whether if this new limit will reject large block like this from mainnet outage recover? And does that means it can never recover if so? https://aleoscan.io/block?h=5213428 @vicsn

vicsn:
> I'm actually curious whether if this new limit will reject large block like this from mainnet outage recover? And does that means it can never recover if so? https://aleoscan.io/block?h=5213428 @vicsn

Thx for asking! The new limit applies to **proposals** and will limit the maximum compute used in a block. It is likely that the block you refer to still falls within the margin. If it doesn't, it is not an issue, as the limit only becomes effective from a certain future block height onwards.

zklimaleo:
> Thx for asking! The new limit applies to **proposals** and will limit the maximum compute used in a block. It is likely that the block you refer to still falls within the margin. If it doesn't, it is not an issue, as the limit only becomes effective from a certain future block height onwards.

@vicsn Thanks! It's more of a concern for future outage, maybe it's helpful to also make sure the validators are able to spread out transactions to different blocks in case of massive pending queue if not yet.

raychu86:
> It's more of a concern for future outage, maybe it's helpful to also make sure the validators are able to spread out transactions to different blocks in case of massive pending queue if not yet.

Thank you for bringing up this topic, as it is good to think about in terms of possible edge cases.

 I don't believe this will cause an issue, because this enforcement is on the proposal/certificate level (which are the building components of a block). This just means that each valid certificate would be limited, which implies that a large pending queue will take longer to clear out, but should not invalidate any block production. Blocks can still grow larger if the subdag takes many rounds to commit, however now there's a stronger limiter to prevent blocks from ballooning too quickly.
 
 Another point to note is that block production itself does not care about the number of transactions, it just cares that a proper subdag is found.

zklimaleo:
> > It's more of a concern for future outage, maybe it's helpful to also make sure the validators are able to spread out transactions to different blocks in case of massive pending queue if not yet.
> 
> Thank you for bringing up this topic, as it is good to think about in terms of possible edge cases.
> 
> I don't believe this will cause an issue, because this enforcement is on the proposal/certificate level (which are the building components of a block). This just means that each valid certificate would be limited, which implies that a large pending queue will take longer to clear out, but should not invalidate any block production. Blocks can still grow larger if the subdag takes many rounds to commit, however now there's a stronger limiter to prevent blocks from ballooning too quickly.
> 
> Another point to note is that block production itself does not care about the number of transactions, it just cares that a proper subdag is found.

I see thanks for the clarification!

niklaslong:
Rebased.

niklaslong:
After the recent changes on the VM side, I have moved cost calculation up into the ledger. The two spend limit test cases are expected to fail as no limit is currently enforced.

vicsn:
You can merge in `origin/staging` and open the PR now!

acoglio (approved):
I'll just note here, for the record, the issue we already discussed. Two validators may happen to have different block heights when they check the same batch proposal, and thus have a slightly different idea of whether the batch proposal is valid or not. I don't think that this can cause forking, because a quorum of signatures, if achieved, is still a quorum. But it would be a cleaner story if the notion of whether a batch is valid at a given round only depended on the round number, which we could achieve by using round numbers instead of block heights to define different versions of batch proposal validity (such as cost limit).

node/bft/src/helpers/ready.rs
```diff
@@ -45,105 +47,152 @@ impl<N: Network> Default for Ready<N> {
...
-        // Drain the transmission IDs.
-        transmissions.drain(range).collect::<IndexMap<_, _>>()
+    /// Removes and returns the transmission at the front of the queue.
+    pub fn remove_front(&mut self) -> Option<(TransmissionID<N>, Transmission<N>)> {
+        if let Some((transmission_id, transmission)) = self.transmissions.pop_front() {
+            self.transmission_ids.remove(&transmission_id);
+
+            if self.transmission_ids.is_empty() {
+                debug_assert!(self.transmissions.is_empty());
+                self.offset = 0;
```
vicsn, line 168: Could we theoretically also have a debug_assert that it should equal 0?

node/bft/src/primary.rs
```diff
@@ -493,90 +493,131 @@ impl<N: Network> Primary<N> {
...
+                }
+
+                // Check if the ledger already contains the transmission.
+                if self.ledger.contains_transmission(&id).unwrap_or(true) {
+                    trace!("Proposing - Skipping transmission '{}' - Already in ledger", fmt_id(id));
+                    continue;
+                }
+
+                // Check if the storage already contain the transmission.
+                // Note: We do not skip if this is the first transmission in the proposal, to ensure that
```
vicsn, line 526: Want to make an issue to get rid of this in the future? Empty proposals are allowed, and I assume if its in storage it'll likely be added to ledger.

vicsn (dismissed):
LGTM! Two nits. Might want to update the snarkVM rev one more time before merging.

node/bft/src/helpers/ready.rs
niklaslong: Almost (I thought this might be the case as we don't currently have a method for removing from the back) but consider the following: an equal number of inserts (no change to the offset when adding to the back of the queue) followed by removals at the front (increments offset) would result in a positive offset when the queue is empty again.

node/bft/src/primary.rs
niklaslong: I've opened https://github.com/ProvableHQ/snarkOS/issues/3551.

vicsn (approved):

PR #3472, opened 2025-01-27
Fix/inbound logging

vicsn:
## Motivation

> Currently when self.inbound() fails, sometimes, nothing is logged, and you don't see that anything went wrong.
> This tripped me up when diagnosing sync issues.

## Test Plan

No need.

## Related PRs

Replaced https://github.com/ProvableHQ/snarkOS-fork/pull/13


ljedrz (approved):

raychu86 (approved):

raychu86:
`ProtocolViolation` is meant to be a catch-all for issues. Code looks fine, one small nit would be to be more specific in the logs rather than relying on `ProtocolViolation`

PR #3475, opened 2025-02-04
[Deps] Run cargo update, update the toolchain

ljedrz:
A lockfile bump, plus an update to the toolchain, as it was done in snarkVM (1.81 > 1.83).


vicsn:
You can now rebase and update the dependency

ljedrz:
Note: the update may include some dependencies that are newer than their snarkVM counterparts due to it having been run later; they will become aligned with the next update.

niklaslong:
CI breaks because of the changes introduced in https://github.com/ProvableHQ/snarkVM/pull/2605, I've just encountered this in local dev as well.

ljedrz:
Going back to draft version in favor of snarkVM's new update, https://github.com/ProvableHQ/snarkVM/pull/2612.

node/bft/src/gateway.rs
```diff
@@ -157,7 +154,7 @@ impl<N: Network> Gateway<N> {
             (Some(ip), _) => ip,
         };
         // Initialize the TCP stack.
-        let tcp = Tcp::new(Config::new(ip, Committee::<N>::MAX_COMMITTEE_SIZE));
+        let tcp = Tcp::new(Config::new(ip, N::MAX_CERTIFICATES.last().unwrap().1));
```
vicsn: Can `LATEST_MAX_CERTIFICATES` be used here for consistency? And since we return a Result anyway, can we properly handle the error?

node/bft/src/gateway.rs
```diff
@@ -1696,4 +1689,15 @@ mod prop_tests {
...
     }
+
+    #[test]
+    fn ensure_sufficient_rw_queue_depth() {
+        let desired_rw_queue_depth = 2
+            * BatchHeader::<MainnetV0>::MAX_GC_ROUNDS
+            * MainnetV0::LATEST_MAX_CERTIFICATES().unwrap() as usize
+            * BatchHeader::<MainnetV0>::MAX_TRANSMISSIONS_PER_BATCH;
+
+        assert!(<Gateway<MainnetV0> as Reading>::MESSAGE_QUEUE_DEPTH >= desired_rw_queue_depth);
```
vicsn, line 1738: Why `>=` instead of `==`?

We'll have to update `MESSAGE_QUEUE_DEPTH` now manually every time we increase `MAX_CERTIFICATES` right? I have a slight preference for just turning the const into a runtime function like we do in snarkVM now.

In any case, I'll make a TODO to analyse whether that value should be fixed at some point (since we don't expect the number of peers to increase forever either).

node/bft/src/helpers/proposal.rs
```diff
@@ -210,7 +210,7 @@ impl<N: Network> FromBytes for Proposal<N> {
         // Read the number of signatures.
         let num_signatures = u32::read_le(&mut reader)?;
         // Ensure the number of signatures is within bounds (this is an early safety check).
-        if num_signatures as usize > Committee::<N>::MAX_COMMITTEE_SIZE as usize {
+        if num_signatures as usize > N::LATEST_MAX_CERTIFICATES().unwrap() as usize {
```
vicsn: Can we handle the error like we do in snarkVM `read_le` functions?

node/bft/src/helpers/pending.rs
```diff
@@ -34,11 +34,14 @@ pub(crate) const CALLBACK_EXPIRATION_IN_SECS: i64 = MAX_FETCH_TIMEOUT_IN_MS.div_
...
     // Determine the number of validators in the committee lookback for the given round.
-    let num_validators = ledger
-        .get_committee_lookback_for_round(round)
-        .map(|committee| committee.num_members())
-        .ok()
-        .unwrap_or(Committee::<N>::MAX_COMMITTEE_SIZE as usize);
+    let num_validators =
+        ledger.get_committee_lookback_for_round(round).map(|committee| committee.num_members()).ok().unwrap_or_else(
+            || {
+                let max_committee_size =
```
vicsn: Sorry for nitpicking, can we handle the error?

node/bft/src/worker.rs
```diff
@@ -623,7 +623,7 @@ mod tests {
         let ledger: Arc<dyn LedgerService<CurrentNetwork>> = Arc::new(mock_ledger);
 
         // Ensure the maximum number of redundant requests is correct and consistent across iterations.
-        assert_eq!(max_redundant_requests(ledger, 0), 34, "Update me if the formula changes");
+        assert_eq!(max_redundant_requests(ledger, 0), 6, "Update me if the formula changes");
```
vicsn: Hmm is it possible to test the `34` value as well? By using a higher round or height? Or does the test ledger not simulate having higher block heights?

Same for other similar tests.

vicsn (changes requested):

node/bft/src/gateway.rs
ljedrz: > Why >= instead of ==?

That value can be arbitrarily large (it's technically only limited by memory), so it doesn't have to be strictly equal.

> We'll have to update MESSAGE_QUEUE_DEPTH now manually every time we increase MAX_CERTIFICATES right? I have a slight preference for just turning the const into a runtime function like we do in snarkVM now.

Indeed, as long as the value increases; we could turn it into a function, though I'm not sure how much fallout there would be.

> In any case, I'll make a TODO to analyse whether that value should be fixed at some point (since we don't expect the number of peers to increase forever either).

Yes, and this should probably also consider the amount of memory that it would require, as I'm pretty sure the maximum is already going into hundreds of MB.

node/bft/src/gateway.rs
niklaslong: Is the error recoverable (i.e. occasionally expected)? If not, panicking may be appropriate to stop the node from continuing in an unexpected bad state? Just a thought.

node/bft/src/worker.rs
ljedrz: > Or does the test ledger not simulate having higher block heights?

That's precisely the issue; our tests usually pick a fresh ledger and start at round 0.

node/bft/src/gateway.rs
ljedrz: This operation is infallible (unless we ever intend to completely prune the entries in the list of maximums of certificates), so this change is mostly "cosmetic", and in this particular spot, painless :smiley:.

vicsn: It is not recoverable and not expected... don't feel *too* strongly either way

node/bft/src/worker.rs
vicsn: Not sure if this is going too deep into the rabbithole, but maybe for at least one test, you can see if with `mock_ledger.expect_*` you can return a larger committee.

vicsn: Noting for posteriority: we could also adjust the function signature of `max_redundant_requests` to take in `num_validators` to test different aspects.

Given how simple the function is, let's defer.

node/bft/src/gateway.rs
vicsn: Can you add all of this context as a comment in the test to make it easier for future developers?

vicsn (dismissed):
1 nit left

node/bft/storage-service/src/persistent.rs
```diff
@@ -61,10 +60,10 @@ pub struct BFTPersistentStorage<N: Network> {
 impl<N: Network> BFTPersistentStorage<N> {
     /// Initializes a new BFT persistent storage service.
     pub fn open(storage_mode: StorageMode) -> Result<Self> {
-        let capacity = NonZeroUsize::new(
-            (Committee::<N>::MAX_COMMITTEE_SIZE as usize) * (BatchHeader::<N>::MAX_TRANSMISSIONS_PER_BATCH) * 2,
-        )
-        .ok_or_else(|| anyhow!("Could not construct NonZeroUsize"))?;
+        let max_committee_size = N::LATEST_MAX_CERTIFICATES().unwrap();
```
niklaslong: I know these are equivalent but should we be using `Committee::<N>::max_committee_size()` here?

niklaslong (dismissed):
Left one question, otherwise LGTM 🚀

node/bft/storage-service/src/persistent.rs
ljedrz: Hmm, do we even need `Committee::<N>::max_committee_size()` if this is a `Network`-level value?

niklaslong: Here's where we [define the function](https://github.com/ProvableHQ/snarkVM/blob/staging/ledger/committee/src/lib.rs#L66). I don't think our consensus is likely to decouple these values soon but I'd suggest using whatever is clearer based on context for now.

ljedrz: I don't mind either way; should I change them all? @vicsn I recall having asked in the related snarkVM-side PR if we intend to remove the `Committee` and `BatchHeader` consts in favor of a `Network`-level config, which could be contained in its dedicated `struct`; is this the future plan, or do we still expect to be using config bits from different objects?

ljedrz:
As for the CI failure:
```
thread 'test_storage_coherence' panicked at node/bft/tests/narwhal_e2e.rs:163:5:
assertion failed: network.is_certificate_round_coherent(1..TARGET_ROUND - 2)
```
I can't reproduce it locally; has this test been flaky in the past? Or perhaps this PR has the potential to make it flakier?

node/bft/storage-service/src/persistent.rs
vicsn: Let's keep the architecture unchanged - despite being a bit wonky. I agree with Niklas its better for clarity to use the `max_committee_size` where its more directly applicable.

vicsn:
> I can't reproduce it locally; has this test been flaky in the past? Or perhaps this PR has the potential to make it flakier?

This has been flaky in the past. I'm deferring fixing it because Ray is making some adjustments to certificate fetching at the moment which *might* fix it too.

ljedrz:
I think everything has been addressed now; I haven't rebased so as not to complicate reviewing, but I think this can still be merged cleanly.

vicsn (approved):

PR #3476, opened 2025-02-06
fix: Increase the max number of peers to aid consensus

Raphexion:
## Motivation

Previously, we only had 16 validator nodes, the it was sufficient with 21 peers. However, as we have increased the number of validators to 26, we need to increase the number of maximum peers.

## Test Plan

Full integration and acceptance tests needed.

## Related PRs

None.


Raphexion:
I missed that we override this value for validators:
https://github.com/ProvableHQ/snarkOS/blob/staging/node/src/validator/router.rs#L154-L157

PR #3477, opened 2025-02-06
Fix sync request timeout off-by-one

raychu86:
## Motivation

This PR fixes an edge case where the sync module was not removing requests that should be deemed obsolete. If your ledger advances to a block height X, there is no need for block X to be requested again. 

NOTICE: This PR is a hotfix into `mainnet` and will be rolled into `staging` post release.

## Test Plan

Tests have been added


vicsn (approved):
LGTM! Sync tests will be running today

niklaslong (approved):

PR #3482, opened 2025-02-11
Merge mainnet release back into staging

vicsn:
## Motivation

This is a routine PR to make sure that the mainnet release finds its way back into the staging branch.


vicsn:
Superseded by: https://github.com/ProvableHQ/snarkOS/pull/3483 because both `staging` and `mainnet` have branch protection enabled, which is good practice

PR #3484, opened 2025-02-13
Remove redundant check.

acoglio:
The diff looks more substantial, but the only change is to remove the check `if self.sync.is_gateway_mode()` and to reformat the code.

As discussed on Slack, and as can be seen examining the code, the test is always true, since a validator is always in gateway mode.

Although not a big change, it's one less thing to see and remember.

### DIFF
--- a/node/bft/src/primary.rs
+++ b/node/bft/src/primary.rs
@@ -1090,23 +1090,21 @@ impl<N: Network> Primary<N> {
         });
 
         // Start the worker ping(s).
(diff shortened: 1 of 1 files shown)


acoglio:
I found a similar check in the same function, and pushed another commit.

vicsn (approved):
This is an excellent find! Great step towards cleaning and tightening up the sync logic.

vicsn:
The CI failure is unrelated, I'll rerun

PR #3490, opened 2025-02-19
Fix missing certificate fetching

raychu86:
## Motivation

This PR fixes an edge case in `fetch_missing_certificates` where the call will terminate early without results if all of the certificate IDs already exist in the `unprocessed_certificate` cache. The fix ensures that we no longer ignore certificates in the cache if there are no certificates to send via the network.

### Example

The following scenario is an example of how the edge case could occur:

1. Validator node receives `BatchCertificate` via `process_batch_certificate_from_peer` 
    - This adds the certificate to the `unprocessed_certificate` cache
3. Validator node starts processing the data and requests missing transmissions/certificates
4. Validator node receives **ALL** other `BatchCertificates` of the missing certificates in step 3 from other sources (like the `BatchCertificate` message in **Step 1**)
5. Validator never finishes processing the `BatchCertificate` in **Step 1**
6. Validator is now stuck in a loop of being unable to process incoming certificates with a `Missing previous certificate ...` message

## Related PRs

https://github.com/ProvableHQ/snarkOS/pull/3439

### DIFF
--- a/node/bft/src/primary.rs
+++ b/node/bft/src/primary.rs
@@ -1697,9 +1697,9 @@ impl<N: Network> Primary<N> {
             }
         }
 
-        // If there are no missing certificates, return early.
+        // If there are no certificates to fetch, return early with the existing unprocessed certificates.
         match fetch_certificates.is_empty() {
-            true => return Ok(Default::default()),
+            true => return Ok(missing_certificates),
             false => trace!(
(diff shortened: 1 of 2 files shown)


vicsn (approved):

kpandl (approved):

vicsn:
E2E tests succeeded.

PR #3492, opened 2025-02-20
Clarify proposal ordering comment

vicsn:
## Motivation

This is not malicious behaviour, it happens regularly (once a day) on mainnet, and is theoretically valid.


acoglio (approved):

raychu86:
What is the case that this would be valid? The check is intended to ensure that a validator does not send another proposal from a previous round `X` after their proposal for round `Y` has already been signed (`Y > X`).

raychu86:
Two possible non-malicious cases:

1. Your validator node somehow signs and processes proposal X+1 before receiving proposal X due to network latency.
2. A validator node restarted and cleared it's cache file

niklaslong:
> Your validator node somehow signs and processes proposal X+1 before receiving proposal X due to network latency.

I don't think this is possible iff messages are sent on the same connection as tcp guarantees in-order delivery of data and I don't think validators can send proposal `x + 1` before proposal `x` for signing? 

> it happens regularly (once a day) on mainnet

@vicsn are there any clues as to why this might currently be happening on mainnet?

vicsn:
A third hypothesis is that *processing* does not happen in order. E.g. because proposal N is much bigger and is slower to deserialize compared to proposal N-1.

acoglio:
Could disconnections and re-connections play a role?

niklaslong:
> Could disconnections and re-connections play a role?

Could be! If a new connection is opened, there's no ordering guarantee w.r.t. a previous one. 

>A third hypothesis

Also very possible with our current setup as we defer deserialisation, it doesn't happen when reading messages from the network.

niklaslong (approved):

raychu86 (approved):

Issue #2185, opened 2022-12-16
[Bug] Block locator check has incorrect assumption


HarukaMa:
## 🐛 Bug Report

I noticed this when I was reading the code:

https://github.com/AleoHQ/snarkOS/blob/dbca32ddfb55853f1131c74b53086ad41e6c3d3c/node/messages/src/helpers/block_locators.rs#L201-L203

Seems like this check is assuming `RECENT_INTERVAL` is 1. If the interval is not 1, then that check will conflict with:

https://github.com/AleoHQ/snarkOS/blob/dbca32ddfb55853f1131c74b53086ad41e6c3d3c/node/messages/src/helpers/block_locators.rs#L194-L196

and when the latest height is less than `RECENT_INTERVAL * NUM_RECENTS`, no `recent_blocks` could pass the check except empty `BlockLocator`s.

EDIT: also 

https://github.com/AleoHQ/snarkOS/blob/dbca32ddfb55853f1131c74b53086ad41e6c3d3c/node/messages/src/helpers/block_locators.rs#L191-L193

seems superfluous to me as we already have interval checks.

## Steps to Reproduce

Didn't really test this.

## Expected Behavior

The check should be `... != last_height * RECENT_INTERVAL`.

## Your Environment

N/A

acoglio:
Indeed, various parts of the code implicitly assume that `RECENT_INTERVAL` is 1. In PR #3495, I've marked all those spots with TODOs.

The check `*current_height <= last_height` ensures that the subtraction `*current_height - last_height` in the interval check does not wrap around. Block locators are untrusted data coming from the outside and so the recents map may have out-of-order keys.

PR #3495, opened 2025-02-21
[fix] Fixes related to block locators.

acoglio:
Rectify documentation of `BlockLocators` about validation during serialization. Also introduce more explicit notion of well-formedness, since subsequent documentation refers to that.

Make `BlockLocators::new_unchecked()` constructor, which does not perform validation, non-public, leveraging language-level protection. Also document that it is only used in tests.

As discussed on Slack, fix the check that the checkpoints map cover as many blocks as possible. Also explain the check.

As discussed on Slack, fix the check about overlapping checkpoints and recents. Also explain the check.

Remove a redundant check to validate block locators in `BlockSync`, as discussed on Slack, since deserialization already checks that.

Add TODOs related to either generalizing some of the code to `RECENT_INTERVAL > 1`, or removing such TODOs if we decide to hardwire `RECENT_INTERVAL` to 1 and eliminate that constant.

### DIFF
--- a/node/sync/locators/src/block_locators.rs
+++ b/node/sync/locators/src/block_locators.rs
@@ -195,13 +195,19 @@ impl<N: Network> BlockLocators<N> {
 
         // Ensure that `last_checkpoint_height` is
         // the largest multiple of `CHECKPOINT_INTERVAL` that does not exceed `last_recent_height`.
-        // That is, `last_checkpoint_height + CHECKPOINT_INTERVAL`
-        // must be greater than `last_recent_height`.
-        // We do not worry about this addition overflowing,
-        // because our use of `u32` for block heights implies that we do not expect to run out.
-        if last_checkpoint_height + CHECKPOINT_INTERVAL <= last_recent_height {
+        // That is, we must have
+        // `last_checkpoint_height <= last_recent_height < last_checkpoint_height + CHECKPOINT_INTERVAL`.
+        // Although we do not expect to run out of `u32` for block heights,
+        // `last_checkpoint_height` is an untrusted value that may come from a faulty validator,
+        // and thus we use a saturating addition;
+        // only a faulty validator would send block locators with such high block heights,
+        // under the assumption that the blockchain is always well below the `u32` limit for heights.
+        if !(last_checkpoint_height..last_checkpoint_height.saturating_add(CHECKPOINT_INTERVAL))
+            .contains(&last_recent_height)
+        {
             bail!(
-                "Last checkpoint height ({last_checkpoint_height}) is too small for last recent height ({last_recent_height})"
+                "Last checkpoint height ({last_checkpoint_height}) is not the largest multiple of ".to_string()
+                    + "{CHECKPOINT_INTERVAL} that does not exceed the last recent height ({last_recent_height})"
             )
         }
 
--- a/node/sync/locators/src/block_locators.rs
+++ b/node/sync/locators/src/block_locators.rs
@@ -541,7 +541,7 @@ mod tests {
         let mut recents = IndexMap::new();
         for i in 0..NUM_RECENT_BLOCKS {
             recents.insert(i as u32, (Field::<CurrentNetwork>::from_u32(i as u32)).into());
-            let block_locators = BlockLocators::<CurrentNetwork>::new(recents.clone(), checkpoints.clone()).unwrap();
+            let block_locators = BlockLocators::<CurrentNetwork>::new_unchecked(recents.clone(), checkpoints.clone());
             block_locators.ensure_is_valid().unwrap();
         }
         // Ensure NUM_RECENT_BLOCKS + 1 is not valid.
--- a/node/sync/locators/src/block_locators.rs
+++ b/node/sync/locators/src/block_locators.rs
@@ -206,8 +206,8 @@ impl<N: Network> BlockLocators<N> {
             .contains(&last_recent_height)
         {
             bail!(
-                "Last checkpoint height ({last_checkpoint_height}) is not the largest multiple of ".to_string()
-                    + "{CHECKPOINT_INTERVAL} that does not exceed the last recent height ({last_recent_height})"
+                "Last checkpoint height ({last_checkpoint_height}) is not the largest multiple of \
+                 {CHECKPOINT_INTERVAL} that does not exceed the last recent height ({last_recent_height})"
             )
         }
 
--- a/node/sync/locators/src/block_locators.rs
+++ b/node/sync/locators/src/block_locators.rs
@@ -379,6 +379,8 @@ pub mod test_helpers {
     type CurrentNetwork = snarkvm::prelude::MainnetV0;
 
     /// Simulates a block locator at the given height.
+    ///
+    /// The returned block locator is checked to be well-formed.
     pub fn sample_block_locators(height: u32) -> BlockLocators<CurrentNetwork> {
         // Create the recent locators.
         let mut recents = IndexMap::new();
@@ -401,6 +403,8 @@ pub mod test_helpers {
     }
 
     /// Simulates a block locator at the given height, with a fork within NUM_RECENT_BLOCKS of the given height.
+    ///
+    /// The returned block locator is checked to be well-formed.
     pub fn sample_block_locators_with_fork(height: u32, fork_height: u32) -> BlockLocators<CurrentNetwork> {
         assert!(fork_height <= height, "Fork height must be less than or equal to the given height");
         assert!(
@@ -448,7 +452,8 @@ pub mod test_helpers {
             assert_eq!(block_locators.checkpoints.len(), expected_num_checkpoints as usize);
             assert_eq!(block_locators.recents.len(), expected_num_recents as usize);
             assert_eq!(block_locators.latest_locator_height(), expected_height);
-            assert!(block_locators.is_valid());
+            // Note that `sample_block_locators` always returns well-formed block locators,
+            // so we don't need to check `is_valid()` here.
         }
     }
 }
--- a/node/sync/locators/src/block_locators.rs
+++ b/node/sync/locators/src/block_locators.rs
@@ -557,7 +557,7 @@ mod tests {
         let checkpoints = IndexMap::from([(0, zero), (CHECKPOINT_INTERVAL, checkpoint_1)]);
         check_is_valid(
             checkpoints,
-            (CHECKPOINT_INTERVAL - NUM_RECENT_BLOCKS as u32)..(CHECKPOINT_INTERVAL * 2 - NUM_RECENT_BLOCKS as u32),
+            (CHECKPOINT_INTERVAL - NUM_RECENT_BLOCKS as u32 + 1)..(CHECKPOINT_INTERVAL * 2 - NUM_RECENT_BLOCKS as u32),
         );
     }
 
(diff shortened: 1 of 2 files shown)


acoglio:
Closes #2185.

node/sync/locators/src/block_locators.rs
```diff
@@ -86,7 +88,9 @@ impl<N: Network> BlockLocators<N> {
     }
 
     /// Initializes a new instance of the block locators, without checking the validity of the block locators.
-    pub fn new_unchecked(recents: IndexMap<u32, N::BlockHash>, checkpoints: IndexMap<u32, N::BlockHash>) -> Self {
+    /// This is only used for testing; note that it is non-public.
+    #[allow(dead_code)]
```
kaimast: It is better to use `#[cfg(test)]` here.

node/sync/locators/src/block_locators.rs
```diff
@@ -189,27 +193,38 @@ impl<N: Network> BlockLocators<N> {
...
+        // otherwise, the overlap is at the last checkpoint,
+        // which is exactly at the last recent height minus its distance from the last checkpoint.
+        // All of this also works if the last checkpoint is 0:
+        // in this case, there is an overlap (at 0) exactly when the last recent height,
+        // which is the same as its distance from the last checkpoint (0),
+        // is less than `NUM_RECENT_BLOCKS`.
+        // All of this only works if `NUM_RECENT_BLOCKS < CHECKPOINT_INTERVAL`,
+        // because it is only under this condition that there is at most one overlapping height.
+        // TODO: generalize check for RECENT_INTERVAL > 1, or remove this comment if we hardwire that to 1
+        let last_recent_to_last_checkpoint_distance = last_recent_height % CHECKPOINT_INTERVAL;
```
kaimast, line 226: Can we add a unit test where recents and checkpoints overlap?

node/sync/locators/src/block_locators.rs
```diff
@@ -189,27 +193,38 @@ impl<N: Network> BlockLocators<N> {
...
-            && recents.get(&0).copied().unwrap_or_default() != checkpoints.get(&0).copied().unwrap_or_default()
-        {
-            bail!("Recent genesis hash and checkpoint genesis hash mismatch at height {last_recent_height}")
+        // Ensure that `last_checkpoint_height` is
+        // the largest multiple of `CHECKPOINT_INTERVAL` that does not exceed `last_recent_height`.
+        // That is, `last_checkpoint_height + CHECKPOINT_INTERVAL`
+        // must be greater than `last_recent_height`.
+        // We do not worry about this addition overflowing,
+        // because our use of `u32` for block heights implies that we do not expect to run out.
+        if last_checkpoint_height + CHECKPOINT_INTERVAL <= last_recent_height {
```
kaimast: Shouldn't we also check that last_checkpoint_height is less or equal to last_recent_height, or is that already checked somewhere else?

Maybe we could do the following:
```rust
if !(last_checkpoint_height..last_checkpoint_height+CHECKPOINT_INTERVAL).contains(last_recent_height) {
       bail!(..);
}
```

node/sync/locators/src/block_locators.rs
```diff
@@ -248,10 +263,12 @@ impl<N: Network> BlockLocators<N> {
         }
 
         // If the last height is below NUM_RECENT_BLOCKS, ensure the number of recent blocks matches the last height.
+        // TODO: generalize check for RECENT_INTERVAL > 1, or remove this comment if we hardwire that to 1
         if last_height < NUM_RECENT_BLOCKS as u32 && recents.len().saturating_sub(1) as u32 != last_height {
```
kaimast: I think we can just remove these checks and the TODO?

We already know that height increments by `RECENT_INTERVAL` (line 259) and that if `recents.len() < NUM_RECENT_BLOCKS` it holds that the first entry has height 0 (line 253).

From that, assuming `RECENT_INTERVAL` is positive, we can guarantee that either`recents.len() > NUM_RECENT_BLOCKS`, in which case `last_height >= NUM_RECENT_BLOCKS`, or `last_height == (recents.len()-1) * RECENT_INTERVAL`.

node/sync/locators/src/block_locators.rs
```diff
@@ -248,10 +263,12 @@ impl<N: Network> BlockLocators<N> {
         }
```
kaimast, line 269: In line 253, could we write `*current_height > 0` instead of `*current_height != last_height`? The error message makes it clear that the assumption is that `last_height` is 0 at this point, which is indeed the case.

node/sync/src/block_sync.rs
```diff
@@ -420,15 +421,13 @@ impl<N: Network> BlockSync<N> {
...
     /// This function checks that the given block locators are well-formed, however it does **not** check
-    /// that the block locators are consistent the peer's previous block locators or other peers' block locators.
+    /// that the block locators are consistent with the peer's previous block locators or other peers' block locators.
     pub fn update_peer_locators(&self, peer_ip: SocketAddr, locators: BlockLocators<N>) -> Result<()> {
         // If the locators match the existing locators for the peer, return early.
         if self.locators.read().get(&peer_ip) == Some(&locators) {
             return Ok(());
         }
 
-        // Ensure the given block locators are well-formed.
```
kaimast, line 430: Why was this removed? Maybe you can leave a comment here that states why we do not need to check for the validity of the block locators.

kaimast (changes requested):
Looks good to me, aside from the minor notes in my review.

We should add a unit test, though. Ideally, something that would have failed before this fix. Considering there are already unit tests for this struct, adding another test should not take long.

node/sync/locators/src/block_locators.rs
acoglio: Done.

node/sync/locators/src/block_locators.rs
acoglio: I was minimizing changes, but yes, that makes the code clearer and simpler. Pushed.

node/sync/src/block_sync.rs
acoglio: Yes, I updated the doc comment for the function.

node/sync/locators/src/block_locators.rs
acoglio: I think you are right. I've eliminated the check, and documented that it holds.

node/sync/locators/src/block_locators.rs
acoglio: Yes, that check was missing. Independently from that, I also realized that, since last_checkpoint_height is untrusted, we need to use a saturating add. So I used your nice suggested code, with a saturating add.

acoglio:
The failing CI was due to an off-by-one error in the testing code itself. It was creating a BlockLocators structure with checkpoints map [0, 10000] and recents map [9900, ..., 9999], which is invalid, according to the recently added missing test.

node/sync/locators/src/block_locators.rs
acoglio: [This](https://github.com/ProvableHQ/snarkOS/blob/4d4805796036709c367add54e3ab31f5900fa196/node/sync/locators/src/block_locators.rs#L531-L536) is already checking various situations with overlap.

acoglio:
> We should add a unit test, though. Ideally, something that would have failed before this fix. Considering there are already unit tests for this struct, adding another test should not take long.

@kaimast I've added such a unit test.

acoglio:
I've also made some improvements to the testing code (the last two commits).

kaimast (approved):

node/sync/locators/src/block_locators.rs
```diff
@@ -189,27 +193,44 @@ impl<N: Network> BlockLocators<N> {
...
+        // and thus we use a saturating addition;
+        // only a faulty validator would send block locators with such high block heights,
+        // under the assumption that the blockchain is always well below the `u32` limit for heights.
+        if !(last_checkpoint_height..last_checkpoint_height.saturating_add(CHECKPOINT_INTERVAL))
+            .contains(&last_recent_height)
         {
-            bail!("Recent genesis hash and checkpoint genesis hash mismatch at height {last_recent_height}")
+            bail!(
+                "Last checkpoint height ({last_checkpoint_height}) is not the largest multiple of ".to_string()
+                    + "{CHECKPOINT_INTERVAL} that does not exceed the last recent height ({last_recent_height})"
```
raychu86: nit: Why is a string concatenation required here?

```suggestion
                "Last checkpoint height ({last_checkpoint_height}) is not the largest multiple of {CHECKPOINT_INTERVAL} that does not exceed the last recent height ({last_recent_height})"
```

node/sync/locators/src/block_locators.rs
```diff
@@ -189,27 +193,44 @@ impl<N: Network> BlockLocators<N> {
...
+        // the largest multiple of `CHECKPOINT_INTERVAL` that does not exceed `last_recent_height`.
+        // That is, we must have
+        // `last_checkpoint_height <= last_recent_height < last_checkpoint_height + CHECKPOINT_INTERVAL`.
+        // Although we do not expect to run out of `u32` for block heights,
+        // `last_checkpoint_height` is an untrusted value that may come from a faulty validator,
+        // and thus we use a saturating addition;
+        // only a faulty validator would send block locators with such high block heights,
+        // under the assumption that the blockchain is always well below the `u32` limit for heights.
+        if !(last_checkpoint_height..last_checkpoint_height.saturating_add(CHECKPOINT_INTERVAL))
+            .contains(&last_recent_height)
```
raychu86, line 206: Rust compiler may optimize this away, but why do a `contains` check in a range rather than 2 comparator checks?

raychu86:
Left a few comments, mainly stylistic changes. Changes to the checks seem to make sense, but I would like to have this tested further.

@kpandl Could you run the sync tests on this branch to ensure there are no regressions or unwanted behavior?

node/sync/locators/src/block_locators.rs
acoglio: I thought we might get a clippy complaint for exceeding 120 columns?

node/sync/locators/src/block_locators.rs
acoglio: I followed a suggestion of another review. Either way is fine with me, so long as there's agreement.

node/sync/locators/src/block_locators.rs
raychu86: You can make it more concise w/ something like:
``` 
bail!("Invalid checkpoint ({last_checkpoint_height}): must be the largest multiple of {CHECKPOINT_INTERVAL} ≤ {last_recent_height}");
```

Should be fine as long as clippy does not complain. Alternatively you can use the "\" character to split the lines. 

For example:
```
bail!(
    "Checkpoint height ({last_checkpoint_height}) must be the largest multiple of \
     {CHECKPOINT_INTERVAL} not exceeding recent height ({last_recent_height})"
);
```

acoglio: Yes, I should have used `\`. Pushed.

raychu86 (approved):

raychu86 (commented):
Logic looks good, but do not merge until we have confirmation of sync tests

kpandl (approved):
LGTM, syncing works [[Grafana](https://aleostresstest.grafana.net/d/single-region-tests-kp/single-region-tests-kp?from=2025-03-21T12:43:18.054Z&to=2025-03-21T13:06:05.035Z&refresh=5m&timezone=browser&orgId=1)]

PR #3499, opened 2025-02-24
[doc] Extend/clarify/fix some documentation.

acoglio:
Clarify that a sync sender is always present in production.

Explain some indirect flows of control that are not easy to follow unless one already knows or still remembers them. The added documentation should make them easy to follow in an IDE.

Explain the relation between blocks, rounds, and garbage collection better and correctly.

Clarify that block authorities are alway sub-DAGs in production.


niklaslong (approved):
LGTM 👍

vicsn (approved):

Issue #3026, opened 2024-01-22
Do not disconnect from synced peers


kyle-u410:
## 🚀 Feature

When snarkOS does a `periodic refresh of peers`, only select peers that are in sync when there are more than one.

## Motivation

When only some peers are caught up with chain, it could be that only one or two peers can help keep snarkOS in sync. When these peers are removed due to periodic refresh, then the current node becomes out of sync. The periodic cycling is good, but the loss of only peer that is in sync is bad. So only remove an in sync peer when there are a number of them in the peer list.

raychu86:
Should be addressed with https://github.com/AleoNet/snarkOS/pull/3223

PR #3504, opened 2025-02-26
Prioritize removing clients over validators in the heartbeat logic

### LINKED ISSUE #3479: [Bug] Clients may disconnect from trusted validators
## 🐛 Bug Report

During stress tests, it became possible for all clients to not be connected to any validators, causing the clients to halt.

## Steps to Reproduce

Start a network with 16 validators and 2 clients per validator. All nodes have all other nodes as peer.

Over 4 test runs,  clients reliably halt within 4 hours.

Note that clients only have 21 peer slots. So the above test causes 32 clients to fill that up instead of the 16 validators).

heartbeat.rs should be fixed.

## Expected Behavior

The affected logic is all in files called heartbeat.rs

### Option 1

Clients should stay connected to nodes in their trusted peers list which advertise themselves as validator. In more detail:
Clients should not disconnect from nodes in their trusted peers list which are also validators. Given the current "core client" model, this should in production only be a single validator. However, other nodes can advertise themselves as validators, so there are edge cases to take into consideration.

### Option 2
Another option is for peer rotation to *continue* when the clients are not synced.

kaimast:
## Motivation

This pull request fixes #3479 and cleans up the heartbeat logic a little. The logic now prioritizes disconnecting nodes over validators to improve connectivity between validators and ensure new blocks propagate to clients. 

## Test Plan

After talking to Victor I decided against adding tests for this PR and believe we can rely on code review to ensure correctness of the changes. The code can only affect liveness of the protocol, not safety, and tests would add lot of additional code that might need to be refactored later.

## Notes
There are few things that might need to be improved in the future.
- `remove_oldest_connected_peer`does not actually remove the peer with the oldest TCP connection, but the one that we have not received any message from in the longest time.
- Some parts of the code use  "ip" to refer to a socket address (which is a tuple of IP and network port). I already changed some variable names from peer_ip to peer_addr, to address this.
- Virtually all the logic resides in the trait declaration right now. We should either actually specialize the different implementations, or simplify the code by converting Heartbeat to a struct.


node/router/src/heartbeat.rs
```diff
@@ -290,11 +299,11 @@ pub trait Heartbeat<N: Network>: Outbound<N> {
...
     fn handle_trusted_peers(&self) {
         // Ensure that the trusted nodes are connected.
-        for peer_ip in self.router().trusted_peers() {
+        for peer_addr in self.router().trusted_peers() {
             // If the peer is not connected, attempt to connect to it.
-            if !self.router().is_connected(peer_ip) {
-                // Attempt to connect to the trusted peer.
-                self.router().connect(*peer_ip);
+            if !self.router().is_connected(peer_addr) {
+                debug!("Attempting to (re-)connect to trusted peer `{peer_addr}`");
```
kaimast: I added this log message, so it is a little easier to understand what is going on when running the devnet. Otherwise you see a lot of "connection refused" messages, but it is unclear why.

node/router/src/heartbeat.rs
```diff
@@ -117,32 +161,15 @@ pub trait Heartbeat<N: Network>: Outbound<N> {
...
-        if let Some(oldest) = oldest_peer {
+        if let Some(oldest) = self.get_removable_peers().pop().map(|p| p.ip()) {
             info!("Disconnecting from '{oldest}' (periodic refresh of peers)");
             let _ = self.send(oldest, Message::Disconnect(DisconnectReason::PeerRefresh.into()));
             // Disconnect from this peer.
             self.router().disconnect(oldest);
         }
     }
 
-    /// TODO (howardwu): If the node is a validator, keep the validator.
```
kaimast, line 145: I think this has been addressed, but let me know if the TODO should be kept.

node/router/src/heartbeat.rs
```diff
@@ -104,8 +107,49 @@ pub trait Heartbeat<N: Network>: Outbound<N> {
...
+        let trusted = self.router().trusted_peers();
+        // The hardcoded bootstrap nodes
+        let bootstrap = self.router().bootstrap_peers();
+
+        // All peers that can be removed
+        let mut removable: Vec<_> = self
+            .router()
+            .get_connected_peers()
+            .into_iter()
+            .filter(|peer| !trusted.contains(&peer.ip())&& !bootstrap.contains(&peer.ip()))
```
niklaslong: A single `filter` call here might be more succinct, it's more of a style choice so up to you but just wanted to mention it as an option.

node/router/src/heartbeat.rs
```diff
@@ -104,8 +107,49 @@ pub trait Heartbeat<N: Network>: Outbound<N> {
...
+            .filter(|peer| !self.router().cache.contains_inbound_block_request(&peer.ip())) // Skip if the peer is syncing.
+            .filter(|peer| self.is_block_synced() || self.router().cache.num_outbound_block_requests(&peer.ip()) == 0) // Skip if you are syncing from this peer.
+            .collect();
+
+        // Sort by priority
+        // 'greater' means earlier in the list and, thus, lower priority
+        removable.sort_by(|peer1, peer2| {
```
niklaslong: I think you could probably do something like this here:
```
removable.sort_by_key(|peer| (peer.is_client(), Reverse(peer.last_seen)));
```

node/router/src/heartbeat.rs
```diff
@@ -104,8 +107,49 @@ pub trait Heartbeat<N: Network>: Outbound<N> {
...
-    /// This function removes the oldest connected peer, to keep the connections fresh.
-    /// This function only triggers if the router is above the minimum number of connected peers.
+    /// Returns a sorted list of removable connected peers
+    /// where the first entry has the lowest priority and
+    /// the last one the highest.
+    ///
+    /// Rules:
+    ///     - Trusted peers and bootstrap nodes are not removable
+    ///     - Peers that we are currently syncing with are not removable
+    ///     - Validators are considered higher priority than clients
```
niklaslong: I think this comment (or the beginning of this doc comment) might be out of date w.r.t. the code as the current implementation sorts validators at the front of the returned vector (and from most recently seen to least recently seen)?

node/router/src/heartbeat.rs
```diff
@@ -104,8 +107,49 @@ pub trait Heartbeat<N: Network>: Outbound<N> {
...
+        let bootstrap = self.router().bootstrap_peers();
+
+        // All peers that can be removed
+        let mut removable: Vec<_> = self
+            .router()
+            .get_connected_peers()
+            .into_iter()
+            .filter(|peer| !trusted.contains(&peer.ip())&& !bootstrap.contains(&peer.ip()))
+            .filter(|peer| !self.router().cache.contains_inbound_block_request(&peer.ip())) // Skip if the peer is syncing.
+            .filter(|peer| self.is_block_synced() || self.router().cache.num_outbound_block_requests(&peer.ip()) == 0) // Skip if you are syncing from this peer.
```
vicsn: Nit: slight preference for computing `self.is_block_synced()` just once ahead of time.

node/router/src/heartbeat.rs
```diff
@@ -104,8 +107,49 @@ pub trait Heartbeat<N: Network>: Outbound<N> {
...
+            .collect();
```
vicsn: Can we sort before collecting to reduce allocations? The impact here won't be too big, but its good practice. 

Even better, can we return sth like an `impl iterator` to avoid ever collecting the full set?

node/router/src/heartbeat.rs
```diff
@@ -189,30 +216,12 @@ pub trait Heartbeat<N: Network>: Outbound<N> {
...
-                       (self.is_block_synced() || (!self.is_block_synced() && self.router().cache.num_outbound_block_requests(&peer.ip()) == 0))
-                    {
-                        Some(peer_ip)
-                    } else {
-                        None
-                    }
-                })
-                .choose_multiple(rng, num_surplus_clients_validators);
+            let peer_ips_to_disconnect =
+                self.get_removable_peers().into_iter().map(|p| p.ip()).take(num_surplus_clients_validators);
```
vicsn: The previous code had an additional `!peer.is_prover() && // Skip if the peer is a prover.`, can we keep that filter after the call to `get_removable_peers()`.

node/router/src/heartbeat.rs
```diff
@@ -104,8 +107,49 @@ pub trait Heartbeat<N: Network>: Outbound<N> {
...
+            .into_iter()
+            .filter(|peer| !trusted.contains(&peer.ip())&& !bootstrap.contains(&peer.ip()))
+            .filter(|peer| !self.router().cache.contains_inbound_block_request(&peer.ip())) // Skip if the peer is syncing.
+            .filter(|peer| self.is_block_synced() || self.router().cache.num_outbound_block_requests(&peer.ip()) == 0) // Skip if you are syncing from this peer.
+            .collect();
+
+        // Sort by priority
+        // 'greater' means earlier in the list and, thus, lower priority
+        removable.sort_by(|peer1, peer2| {
+            if peer1.is_client() && !peer2.is_client() {
```
vicsn: So there are actual **three** types of peers: validators, clients, and provers (which for consensus purposes are glorified clients). With that in mind, it might make more sense to sort on whether peers are **validators** or not.

vicsn (changes requested):

node/router/src/heartbeat.rs
vicsn: The list returned by `get_removable_peers` still contains validators.

The only code I see really keeping validators is under the branch `if self.router().node_type().is_prover()`. I think we change that with minimal side-effects to `if self.router().node_type().is_prover || self.router().node_type().is_validator()` to that.

kaimast: This was more or less my question. I don't think we should keep all validators, otherwise the set of peers could get very big. The behavior now is to prioritize keeping them over clients, but if that causes problems we can change it.

node/router/src/heartbeat.rs
kaimast: I changed it to sort first.

My initial plan was to return impl Iterator first, but I don't think it is possible to sort something without putting it into some buffer first. The heartbeat logic also only runs twice a minute, so I am not sure if optimizing this piece of code much makes a huge difference.

One thing we might want to change the future is to wrap Peer in an Arc, so it does not have to be cloned as often.

node/router/src/heartbeat.rs
kaimast: It now sorts by `(peer.is_validator(), peer.last_seen())`, so older peers and clients should have lower priority as described in the comment. 
Let me know if you still think there is an inconsistency.

node/router/src/heartbeat.rs
kaimast: This was, more or less, the original code, but I changed it to a single filter call. Either looks fine IMHO.

node/router/src/heartbeat.rs
vicsn: Validators can have 200 peers, so ensuring that validators stay connected is not a concern. However, router messages don't contain consensus-critical content anyway, so its fine, we can leave it as is.

vicsn:
Logic LGTM! Approving once a deployed devnet passes with this code.

node/router/src/heartbeat.rs
```diff
@@ -117,32 +154,15 @@ pub trait Heartbeat<N: Network>: Outbound<N> {
...
-            .iter()
-            .filter(|peer| !trusted.contains(&peer.ip()) && !bootstrap.contains(&peer.ip()))
-            .filter(|peer| !self.router().cache.contains_inbound_block_request(&peer.ip())) // Skip if the peer is syncing.
-            .filter(|peer| self.is_block_synced() || self.router().cache.num_outbound_block_requests(&peer.ip()) == 0) // Skip if you are syncing from this peer.
-            .min_by_key(|peer| peer.last_seen())
-            .map(|peer| peer.ip());
-
         // Disconnect from the oldest connected peer, if one exists.
-        if let Some(oldest) = oldest_peer {
+        if let Some(oldest) = self.get_removable_peers().pop().map(|peer| peer.ip()) {
```
niklaslong: I might be misreading this but if validators are sorted to the end of the vector, is `pop` what we want here?

kaimast: You are right. It should be fixed now.

node/router/src/heartbeat.rs
```diff
@@ -117,32 +154,16 @@ pub trait Heartbeat<N: Network>: Outbound<N> {
...
-            .filter(|peer| self.is_block_synced() || self.router().cache.num_outbound_block_requests(&peer.ip()) == 0) // Skip if you are syncing from this peer.
-            .min_by_key(|peer| peer.last_seen())
-            .map(|peer| peer.ip());
-
-        // Disconnect from the oldest connected peer, if one exists.
-        if let Some(oldest) = oldest_peer {
+        // Disconnect from the oldest connected peer, which is the first entry in the list
+        // of removable peers.
+        // Do nothing, if the list is empty.
+        if let Some(oldest) = self.get_removable_peers().into_iter().map(|peer| peer.ip()).next() {
```
niklaslong: Really appreciate the addition of the comment here! I don't mean to nit but this might be a little more idiomatic:
```suggestion
        if let Some(oldest) = self.get_removable_peers().first().map(Peer::ip) {
```

I might also suggest we add a unit test for `get_removable_peers` as getting the order wrong there would lead to a more unstable network? I'm not sure how involved test writing might be here, we could perhaps include this in future improvements to our test coverage, wdyt?

node/router/src/heartbeat.rs
```diff
@@ -182,50 +203,36 @@ pub trait Heartbeat<N: Network>: Outbound<N> {
...
                     }
                 }
 
-                info!("Disconnecting from '{peer_ip}' (exceeded maximum connections)");
-                self.send(peer_ip, Message::Disconnect(DisconnectReason::TooManyPeers.into()));
+                info!("Disconnecting from '{peer_addr}' (exceeded maximum connections)");
+                self.send(peer_addr, Message::Disconnect(DisconnectReason::TooManyPeers.into()));
                 // Disconnect from this peer.
-                self.router().disconnect(peer_ip);
+                self.router().disconnect(peer_addr);
```
niklaslong: I think `peer_ip` here was correct? `disconnect` internally calls `resolve_to_ambiguous` which maps the listener to the ephemeral address (ip + port). 

You might have noticed this already but just in case, the convention is:
- `peer_ip` (granted, a slight misnomer): the peer's listener `SocketAddr` (ip + port)
- `peer_addr`: the peer's ephemeral `SocketAddr` (ip + port, though only the port should be different from the listener)

node/router/src/heartbeat.rs
vicsn: Well now you're just showing off! 😛

And yes since `get_removable_peers` peers is not influenced by timeouts, it can be unit tested.

node/router/src/heartbeat.rs
```diff
@@ -290,11 +297,11 @@ pub trait Heartbeat<N: Network>: Outbound<N> {
     /// This function attempts to connect to any disconnected trusted peers.
     fn handle_trusted_peers(&self) {
         // Ensure that the trusted nodes are connected.
-        for peer_ip in self.router().trusted_peers() {
+        for peer_addr in self.router().trusted_peers() {
```
niklaslong: This should also be `peer_ip` as iirc they are the pre-configured trusted listener addresses the nodes can connect to.

niklaslong (changes requested):
Could you double check the renaming of `peer_ip` to `peer_addr`? In most collections we track the (rather unfortunately named) "`peer_ip`" aka the listener address (ip + port) instead of the ephemeral address when a connection is received 🙏 

See:
- [the resolver module](https://github.com/ProvableHQ/snarkOS/blob/staging/node/router/src/helpers/resolver.rs#L40-L48)
- [peer insert logic on the validator's router (also the same for clients and provers)](https://github.com/ProvableHQ/snarkOS/blob/staging/node/src/validator/router.rs#L66-L68)

node/router/src/heartbeat.rs
```diff
@@ -104,8 +105,44 @@ pub trait Heartbeat<N: Network>: Outbound<N> {
...
+        let mut peers = self.router().get_connected_peers();
+        peers.sort_by_key(|peer| (peer.is_validator(), peer.last_seen()));
+
+        // Deterimine which of the peers can be removed
+        peers
+            .into_iter()
+            .filter(|peer| {
+                !trusted.contains(&peer.ip()) // Always keep trusted nodes
+                  && !bootstrap.contains(&peer.ip()) // Always keep bootstrap nodes
+                  && !self.router().cache.contains_inbound_block_request(&peer.ip()) // This peer is currently syncing from us
```
raychu86: Can this be exploited in a way where a malicious validator just needs to keep sending block requests to prevent the peer disconnects?

node/router/src/heartbeat.rs
vicsn: Is this naming convention documented anywhere? If not, where would be a good place?

node/router/src/heartbeat.rs
vicsn: Yes, same for malicious clients, this was already an issue before this PR.

@kaimast want to create a new issue for this? A simple solution [as noted here](https://github.com/ProvableHQ/snarkOS/issues/3026) would be to refresh a random selection of syncing peers under some conditions, I bet with further analysis we can avoid hurting sync performance of honest nodes too much.

kaimast: Even without this check, couldn't a malicious peer send Ping messages frequently to appear more responsive and receive higher priority?

kaimast:
> Could you double check the renaming of `peer_ip` to `peer_addr`? In most collections we track the (rather unfortunately named) "`peer_ip`" aka the listener address (ip + port) instead of the ephemeral address when a connection is received 🙏
> 
> See:
> 
> * [the resolver module](https://github.com/ProvableHQ/snarkOS/blob/staging/node/router/src/helpers/resolver.rs#L40-L48)
> * [peer insert logic on the validator's router (also the same for clients and provers)](https://github.com/ProvableHQ/snarkOS/blob/staging/node/src/validator/router.rs#L66-L68)

There was an earlier comment for this part that mentioned I found `peer_ip` confusing for the same reason you mentioned (it holds the full address not just the ip). 
Using "ip" to refer to the public/listening address is maybe not the best approach, but I will change it back for now and this can be something we might address in the future.

node/router/src/heartbeat.rs
kaimast: We could add some more documentation to `Resolver`. It should also mention that the peer_id / listen address is used to uniquely identify a peer in some parts of the code.

node/router/src/heartbeat.rs
kaimast: I'm surprised there isn't a lint this specific error. 

Let me look into how involved unit tests would be, but I generally agree that they are a good idea here.

kaimast:
@niklaslong I added tests as you suggested. I think all remaining comments are addressed as well.

node/router/src/heartbeat.rs
vicsn: Go for it.

node/router/src/heartbeat.rs
vicsn: > Even without this check, couldn't a malicious peer send Ping messages frequently to appear more responsive and receive higher priority?

Oe nice one. We can think of some potential mitigations, e.g. reverting to `choose_multiple`, perhaps weighing clients heavier than validators, and banning when seeing very frequent ping messages (which we might already do).

You can add ideas in the new issue but I still think out of scope for this PR.

raychu86: Agreed, not in scope for this PR. Can be addressed in the future for changes directly related to handling malicious peers.

raychu86 (dismissed):
We should run devnets and observe connection behavior is in line with expectations.

Otherwise, logic looks good overall.

node/router/src/heartbeat.rs
vicsn: Can be done in a separate PR.

vicsn (dismissed):
Two devnets ran with correct observed behaviour (the clients never get stuck in a client bubble, despite having an overwhelming majority clients in their trusted peers list). In the future, we hope to do a comprehensive analysis as to what the appropriate (dis)connection rate should be for peers.

Screenshot 2025-03-10 at 19 48 08

node/router/tests/heartbeat.rs
```diff
@@ -0,0 +1,175 @@
...
+    router.update_last_seen_for_connected_peer(validator_peer2.local_ip());
+
+    let heartbeat = HeartbeatTest { router };
+
+    let removable_peers = heartbeat.get_removable_peers();
+
+    // Ensure connections were established.
+    assert_eq!(heartbeat.router().get_connected_peers().len(), 3);
+
+    // There are no trusted or bootstrap peers, so everything should be remoable.
```
niklaslong:
```suggestion
    // There are no trusted or bootstrap peers, so everything should be removable.
```

niklaslong (dismissed):
LGTM! 🚀

node/sync/src/block_sync.rs
```diff
@@ -1518,7 +1518,9 @@ mod tests {
         }
 
         // Duplicate a new sync module with a different height to simulate block advancement.
-        let ledger_height = rng.gen_range(0..locator_height);
+        // This range needs to be inclusive, so that the range is never empty,
+        // even with a locator height of 0.
+        let ledger_height = rng.gen_range(0..=locator_height);
```
kaimast, line 1523: @raychu86 The previous CI run for my branch failed this test, and could reproduce it failing every 100 times or so on my local machine. It is unrelated to the PR, but I think we should just fix it now.

Could you check if the fix is correct? It looks like you originally wrote the test.

raychu86: This is correct, thanks for finding this issue.

I assume this would create an empty range and fail whenever `let locator_height = rng.gen_range(0..50);` resolved to 0.

niklaslong (approved):

PR #3508, opened 2025-02-26
[Mainnet] Update release version to 3.3.2

vicsn:
## Motivation

This was an oversight and should have been merged into canary v3.3.2 already. This will ensure `snarkOS -V` displays v3.3.2


vicsn:
Closed because a proper version update would require a snarkVM update as well.

PR #3513, opened 2025-02-28
[Fix] Proposal cache storage path configuration

niklaslong:
This PR fixes the storage path configuration for the proposal cache. Previously, it was always stored in the default Aleo ledger directory and wasn't modified by the `--storage` cli option.


cli/src/commands/clean.rs
```diff
@@ -27,6 +27,9 @@ pub struct Clean {
     /// Specify the network to remove from storage.
     #[clap(default_value = "0", long = "network")]
     pub network: u16,
+    /// Specify the path to a directory containing the storage database for the ledger
+    #[clap(long = "storage")]
```
kaimast: This might be a style thing, but, in case you didn't know, you can just write `#[clap(long)] if the long name and the field name are identical.

node/bft/examples/simple_node.rs
```diff
@@ -170,14 +171,14 @@ pub async fn start_primary(
         BatchHeader::<CurrentNetwork>::MAX_GC_ROUNDS as u64,
     );
     // Initialize the gateway IP and dev mode.
```
kaimast: Should the comment say "storage mode" now?

kaimast (approved):
I left some minor comments. The PR looks like it is ready to merge to me, but I only have a limited understanding of this part of the code.

vicsn:
@kaimast can you confirm the result from manually testing the functioning of the flag? You can adjust and use the `devnet.sh` script to spin up a quick network to see if the proposal cache is stored in the expected location.

raychu86:
There were some changes to `aleo-std` in a new release version `v1.0.1`. Related PR https://github.com/ProvableHQ/snarkVM/pull/2590

@ljedrz Can you confirm if this PR is sufficient to match any changes?

cli/src/commands/clean.rs
niklaslong: Good eye! We use both styles in the codebase though the verbose version seems to be the most common. Open to addressing it in this PR, otherwise we can shelve this as a follow-up to minimise the diff.

node/bft/examples/simple_node.rs
niklaslong: And fixed! [1b6ed14](https://github.com/ProvableHQ/snarkOS/pull/3513/commits/1b6ed1415a2bd2c03aa231d8d44a8fae19f51018)

ljedrz:
The updated `aleo-std` would require some adjustments to this PR, but it shouldn't be too tricky. I would recommend that it is merged after the test storage changes for both snarkVM and snarkOS, as they are quite prone to bitrot, especially since merges involve lockfile updates.

cli/src/commands/clean.rs
```diff
@@ -30,35 +30,38 @@ pub struct Clean {
     /// Enables development mode, specify the unique ID of the local node to clean.
     #[clap(long)]
     pub dev: Option<u16>,
-    /// Specify the path to a directory containing the ledger
-    #[clap(long = "path")]
-    pub path: Option<PathBuf>,
+    /// Specify the path to a directory containing the storage database for the ledger.
```
ljedrz:
```suggestion
    /// Specify the path to a directory containing the storage for the ledger.
```
nit: "storage database" is a bit redundant IMO

cli/src/commands/clean.rs
```diff
@@ -30,35 +30,38 @@ pub struct Clean {
...
+    #[clap(long = "storage")]
```
ljedrz: as much as it's vague, we might want to not change this, as it could be part of scripts, which would then break

node/bft/examples/simple_node.rs
```diff
@@ -129,10 +129,10 @@ pub async fn start_bft(
         Arc::new(BFTMemoryService::new()),
         BatchHeader::<CurrentNetwork>::MAX_GC_ROUNDS as u64,
     );
-    // Initialize the gateway IP and dev mode.
-    let (ip, dev) = match peers.get(&node_id) {
-        Some(ip) => (Some(*ip), None),
-        None => (None, Some(node_id)),
+    // Initialize the gateway's storage mode.
+    let (ip, storage_mode) = match peers.get(&node_id) {
+        Some(ip) => (Some(*ip), StorageMode::Production),
```
ljedrz: since this is an example, I wonder if the `Test` storage wouldn't be more fitting here; the `Development` mode might be a redundant option too

node/bft/examples/simple_node.rs
```diff
@@ -171,14 +171,14 @@ pub async fn start_primary(
         BatchHeader::<CurrentNetwork>::MAX_GC_ROUNDS as u64,
     );
     // Initialize the gateway IP and dev mode.
-    let (ip, dev) = match peers.get(&node_id) {
-        Some(ip) => (Some(*ip), None),
-        None => (None, Some(node_id)),
+    let (ip, storage_mode) = match peers.get(&node_id) {
+        Some(ip) => (Some(*ip), StorageMode::Production),
```
ljedrz: same principle: we wouldn't want to write to the production storage when running an example; that being said, that depends on the purpose of the example, and I'm not sure what it currently is

node/bft/src/bft.rs
```diff
@@ -973,7 +975,7 @@ mod tests {
         // Initialize the account.
         let account = Account::new(rng)?;
         // Initialize the BFT.
-        let bft = BFT::new(account.clone(), storage.clone(), ledger.clone(), None, &[], None)?;
+        let bft = BFT::new(account.clone(), storage.clone(), ledger.clone(), None, &[], StorageMode::Production)?;
```
ljedrz: all the tests should only use `StorageMode::Test`; please see the snarkVM changes to see how they were set up

node/bft/src/primary.rs
```diff
@@ -1807,7 +1812,7 @@ mod tests {
         let storage = Storage::new(ledger.clone(), Arc::new(BFTMemoryService::new()), 10);
 
         // Initialize the primary.
-        let mut primary = Primary::new(account, storage, ledger, None, &[], None).unwrap();
+        let mut primary = Primary::new(account, storage, ledger, None, &[], StorageMode::Production).unwrap();
```
ljedrz: ditto: we need to use the `Test` mode in all tests

node/bft/tests/common/primary.rs
```diff
@@ -164,11 +164,26 @@ impl TestNetwork {
...
 
             let (primary, bft) = if config.bft {
-                let bft = BFT::<CurrentNetwork>::new(account, storage, ledger, None, &[], Some(id as u16)).unwrap();
+                let bft = BFT::<CurrentNetwork>::new(
+                    account,
+                    storage,
+                    ledger,
+                    None,
+                    &[],
+                    StorageMode::Development(id as u16),
```
ljedrz: were we actually creating persistent artifacts with tests using this object? we might want to avoid that so as not to mix them up with any preexisting `--dev` runs

ljedrz (changes requested):
Left some suggestions for changes, primarily to spots which would have been correct if the logic behind the use of `StorageMode` hadn't changed recently (tests).

cli/src/commands/clean.rs
niklaslong: fair point, reverted 👍

cli/src/commands/clean.rs
niklaslong: (also reverted)

node/bft/examples/simple_node.rs
```diff
@@ -129,19 +129,20 @@ pub async fn start_bft(
         Arc::new(BFTMemoryService::new()),
         BatchHeader::<CurrentNetwork>::MAX_GC_ROUNDS as u64,
     );
-    // Initialize the gateway IP and dev mode.
-    let (ip, dev) = match peers.get(&node_id) {
-        Some(ip) => (Some(*ip), None),
-        None => (None, Some(node_id)),
+    // Initialize the gateway's storage mode.
```
ljedrz: nit: the IP is also established here

node/bft/src/helpers/proposal_cache.rs
```diff
@@ -108,8 +106,13 @@ impl<N: Network> ProposalCache<N> {
     }
 
     /// Store the proposal cache to the file system.
-    pub fn store(&self, dev: Option<u16>) -> Result<()> {
-        let path = proposal_cache_path(N::ID, dev);
+    pub fn store(&self, storage_mode: &StorageMode) -> Result<()> {
+        // Do not write the proposal cache to the filesystem when in test mode.
+        if matches!(storage_mode, StorageMode::Test(_)) {
+            return Ok(());
```
ljedrz: it's probably not needed in tests, but it technically could be stored at the temporary location (just for future reference)

ljedrz (approved):
LGTM :+1:

node/bft/src/helpers/proposal_cache.rs
niklaslong: Potentially, though with the caveat that multiple test runs will result in the same file being used at the moment (as it is explicitly named for the same ID).

node/bft/examples/simple_node.rs
niklaslong: https://github.com/ProvableHQ/snarkOS/pull/3513/commits/6732a2499d4955399402d43159f02ab05854efbb

node/bft/src/helpers/proposal_cache.rs
vicsn: @ljedrz Is the temporary test folder cleaned up in between tests / do different tests use different ones? If so, we can just store it and save ourselves the headache if we ever build a test which uses it.

ljedrz: every test uses its own random temporary folder, unless specifically configured to share it (which may be useful within some specific test in the future); it's necessary in order to be able to run tests in parallel without worrying about storage clashes

vicsn:
~Can you do a local test of running a validator in a network before this PR, stop all the nodes, move the proposal cache to desired path, start validator with this PR with the desired proposal cache path.~ Please do this on Canarynet with help from Brent V.

vicsn:
Cann you make an issue: do not start if mixing `--storage` path and `--dev` if it doesn't work.

node/bft/src/helpers/proposal_cache.rs
vicsn: @niklaslong please fix.

node/bft/examples/simple_node.rs
```diff
@@ -129,19 +129,20 @@ pub async fn start_bft(
...
         BatchHeader::<CurrentNetwork>::MAX_GC_ROUNDS as u64,
     );
-    // Initialize the gateway IP and dev mode.
-    let (ip, dev) = match peers.get(&node_id) {
-        Some(ip) => (Some(*ip), None),
-        None => (None, Some(node_id)),
+    // Initialize the gateway IP and storage mode.
+    let ip = match peers.get(&node_id) {
+        Some(ip) => Some(*ip),
+        None => Some(SocketAddr::from_str(&format!("127.0.0.1:{}", MEMORY_POOL_PORT + node_id)).unwrap()),
```
kaimast: Instead of `from_str`, it would be a littler nicer to do the following:

 ```rust
SocketAddr::new(IpAddr::V4(IPv4::LOCALHOST), MEMORY_POOL_PORT + node_id)
```
That also avoids the need for `unwrap()`.

niklaslong: Fixed in https://github.com/ProvableHQ/snarkOS/pull/3513/commits/534a29bbcf8d66c9aec23da3ffaa4ce3cc0f0695 along with a few other instances.

node/bft/src/helpers/proposal_cache.rs
niklaslong: The proposal cache is only written to the fs on shutdown, I'm not convinced this is something we'd ever want in tests or examples? I encountered it specifically when running one of the examples followed by tests.

niklaslong: Perhaps I got unlucky, I've tested it again with a devnet run + test runs and haven't seen the same issue without the change, so I have reverted it for now. If we spot it again, the fix is trivial.

vicsn (approved):

PR #3515, opened 2025-02-28
Add releases gantt

vicsn:
## Motivation

Currently, releases are tracked in a spreadsheet, which looks untrustworthy to a new reader and is not easy to interpret.

This PR adds a file rendering expected release dates in pretty Markdown: https://github.com/ProvableHQ/snarkOS/blob/ef1482334b68e6b3127122f8382ee5a26262c824/RELEASES.md

Like any release schedule, it will need to be maintained, which should not be too much work, adjusting 3 lines roughly once a month.

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -9,19 +9,19 @@ gantt
 
     section Canary Releases
     v3.4.0 (03/05) :active, 03/05/2025, 1d
-    v3.5.0 :active, c2, 04/01/2025, 1d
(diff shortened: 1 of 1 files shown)


raychu86:
Content looks good, but is this the correct surface for release topics? I feel like releases may have different circumstances (hotfixes, delays, etc) and having to create a new PR to manage the changes seem like unnecessary overhead

vicsn:
We are live: https://provablehq.github.io/

PR #3516, opened 2025-02-28
[Metrics] Introduce validator telemetry

raychu86:
## Motivation

This PR is an initial iteration of a validator telemetry implementation used for tracking metrics and performance of other validators. This will allow validators to better monitor the behavior of its peers (and itself) and react accordingly - directly via connections or off-chain via alerts.

This new feature is currently guarded with a new `telemetry` feature flag for optional use by validators. We have a rudimentary concept of `participation score` that is exposed via logs and a new REST api:
```
    // GET /{network}/validators/participation
    // GET /{network}/validators/participation?metadata={true}
```

The current metrics being tracked are:
- The number of signatures a validator participated in.
- The number of certificates a validator has created.

There are many other metrics that can/should be tracked in the future. For instance:
- Response time
- Sync rate
- Percentage of proposals that are converted into certificates
- Fullness of proposals
- Connectivity (how many other validators are they connected to)
- Various stake weight considerations
- The latest seen IP address of each validator (useful for debugging purposes)
- etc.


node/bft/src/helpers/telemetry.rs
```diff
@@ -0,0 +1,160 @@
+// Copyright 2024 Aleo Network Foundation
```
kaimast: You probably need to change this to `2024-2025`

node/bft/src/helpers/telemetry.rs
```diff
@@ -0,0 +1,160 @@
...
+        // Insert the certificate author.
+        validator_signatures
+            .entry(certificate_author)
+            .and_modify(|counts| {
+                counts.entry(certificate_round).and_modify(|count| *count += 1).or_insert(1);
+            })
+            .or_insert_with(|| IndexMap::from([(certificate_round, 1)]));
+
+        // Insert the certificate signatures
+        for signature in certificate.signatures() {
```
kaimast: This is the same code as before, right? It would be good to merge those two into a single loop, for example, by doing something like this:
```rust
[certificate_author].iter().chain(certficate.signatures())
```
(I didn't test if this compiles)

node/bft/src/helpers/telemetry.rs
```diff
@@ -0,0 +1,160 @@
...
+        });
+    }
+
+    /// Insert a certificate to the telemetry tracker.
+    pub fn insert_certificate(&self, certificate: &BatchCertificate<N>) {
+        // Acquire the lock for `tracked_certificates`.
+        let mut tracked_certificates = self.tracked_certificates.write();
+
+        // Retrieve the certificate round, author, and ID.
+        let certificate_round = certificate.round();
```
kaimast, line 130: This is purely based on personal preference, but do we need to create dedicated variables here if their names are not significantly shorter than just invoking the method?

node/bft/src/helpers/telemetry.rs
```diff
@@ -0,0 +1,160 @@
...
+                .or_insert_with(|| IndexMap::from([(certificate_round, 1)]));
+        }
+
+        // Acquire the lock for `validator_certificates`.
+        let mut validator_certificates = self.validator_certificates.write();
+
+        // Insert the certificate
+        validator_certificates
+            .entry(certificate_author)
+            .and_modify(|cert_rounds| {
```
kaimast: For code like this, I generally find using `or_default()` cleaner.

```rust
validator_certificates.entry(certificate_author).or_default().insert(certificate_round());
```

Again, this is personal preference, and I am still learning what style people prefer.

node/bft/src/helpers/telemetry.rs
```diff
@@ -0,0 +1,160 @@
...
+use parking_lot::RwLock;
+use rayon::prelude::*;
+use std::sync::Arc;
+
+/// Tracker for the participation metrics of validators.
+#[derive(Clone, Debug)]
+pub struct Telemetry<N: Network> {
+    /// The certificates seen for each round
+    /// A mapping of `round` to set of certificate IDs.
+    tracked_certificates: Arc<RwLock<IndexMap<u64, IndexSet<Field<N>>>>>,
```
kaimast, line 44: It would be great to document why an `IndexMap` is used here. I assume it is to keep the order of metrics consistent?

node/bft/src/helpers/telemetry.rs
```diff
@@ -0,0 +1,160 @@
...
+impl<N: Network> Default for Telemetry<N> {
+    /// Initializes a new instance of telemetry.
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl<N: Network> Telemetry<N> {
+    /// Initializes a new instance of telemetry.
+    pub fn new() -> Self {
```
kaimast, line 67: It might be nicer to `#[derive(Default)]` and write

```rust
pub fn new() -> Self {
       Self::default()
} 
```

node/bft/src/helpers/telemetry.rs
```diff
@@ -0,0 +1,160 @@
...
+        Self {
+            tracked_certificates: Default::default(),
+            validator_signatures: Default::default(),
+            validator_certificates: Default::default(),
+        }
+    }
+
+    // TODO: Determine if this should be pending signatures/certificates, or only certificates included in a fully formed subdag.
+    pub fn insert_subdag(&self, subdag: &Subdag<N>) {
```
kaimast, line 107: This method is missing a doc comment (`///`)

kaimast (changes requested):

kaimast:
I added a few (mostly nitpicky) comments. Could you point me to where this class is used, or will that be a separate PR?

node/bft/src/helpers/telemetry.rs
raychu86: This is copied from the current license_header. Should be handled in a future License PR.

vicsn: A license update PR was merged recently, hence the need for existing open PRs to update.

node/bft/src/helpers/telemetry.rs
raychu86: I kept it the way it was, to stay consistent with how we are handling `new` and `default` on other structs.

node/bft/src/helpers/telemetry.rs
raychu86: I agree, much cleaner with `or_default`.

Resolved with [fefd20b](https://github.com/ProvableHQ/snarkOS/pull/3516/commits/fefd20b09ab52cfe23f55e2eb65c77828125eea4)

node/bft/src/helpers/telemetry.rs
raychu86: Although the rust compiler will optimize it away, we try to reduce calls to the same getter multiple times.

node/bft/src/helpers/telemetry.rs
raychu86: Cleaned up with [fefd20b](https://github.com/ProvableHQ/snarkOS/pull/3516/commits/fefd20b09ab52cfe23f55e2eb65c77828125eea4)

raychu86:
> I added a few (mostly nitpicky) comments. Could you point me to where this class is used, or will that be a separate PR?

The integration logic has been introduced in the latest commits

node/bft/src/helpers/telemetry.rs
raychu86: The intention is to maintain ordering of entries, but in it's current use, HashMaps would suffice as well.

I'd opt to not include specific documentation on that level here since we would have to do so for all of the other instances of Indexmap use

node/bft/src/bft.rs
```diff
@@ -692,7 +692,7 @@ impl<N: Network> BFT<N> {
                     // Initialize a callback sender and receiver.
                     let (callback_sender, callback_receiver) = oneshot::channel();
                     // Send the subdag and transmissions to consensus.
-                    consensus_sender.tx_consensus_subdag.send((subdag, transmissions, callback_sender)).await?;
+                    consensus_sender.tx_consensus_subdag.send((subdag.clone(), transmissions, callback_sender)).await?;
```
ljedrz: hmm, cloning the `Subdag` may get expensive; could we apply the telemetry on the receiving end of the channel that's used here?

node/bft/src/gateway.rs
```diff
@@ -131,6 +131,9 @@ pub struct Gateway<N: Network> {
     /// prevent simultaneous "two-way" connections between two peers (i.e. both nodes simultaneously
     /// attempt to connect to each other). This set is used to prevent this from happening.
     connecting_peers: Arc<Mutex<IndexSet<SocketAddr>>>,
+    #[cfg(feature = "telemetry")]
+    /// The validator telemetry.
```
ljedrz:
```suggestion
    /// The validator telemetry.
    #[cfg(feature = "telemetry")]
```
the established Rust convention

node/bft/src/gateway.rs
```diff
@@ -300,6 +305,12 @@ impl<N: Network> Gateway<N> {
         &self.resolver
     }
 
+    #[cfg(feature = "telemetry")]
+    /// Returns the validator telemetry.
```
ljedrz:
```suggestion
    /// Returns the validator telemetry.
    #[cfg(feature = "telemetry")]
```
ditto

node/bft/src/bft.rs
raychu86: We could, just need to extend the feature flag to the `Consensus` crate.

I also noticed that we are cloning the `commit_subdag` above as well.

node/bft/src/gateway.rs
ljedrz: (applies to all the remaining cases)

node/bft/src/gateway.rs
```diff
@@ -970,6 +985,20 @@ impl<N: Network> Gateway<N> {
...
 
+    #[cfg(feature = "telemetry")]
+    // Logs the validator participation scores.
+    fn log_participation_scores(&self) {
+        if let Ok(current_committee) = self.ledger.current_committee() {
+            // Retrieve the participation scores.
+            let participation_scores = self.validator_telemetry().get_participation_scores(&current_committee);
+            // Log the participation scores.
+            debug!("Participation Scores (in the last {} rounds):", self.storage.max_gc_rounds());
+            for (address, score) in participation_scores {
```
ljedrz, line 998: nit: if we expect to not always be running with debug logs, it would be best to make the execution of this loop conditional based on the log level

node/bft/src/helpers/mod.rs
```diff
@@ -46,6 +46,9 @@ pub use signed_proposals::*;
 pub mod storage;
 pub use storage::*;
 
+pub mod telemetry;
```
ljedrz, line 50: the entire module could be feature-gated to speed up compilation when it's disabled

node/bft/src/helpers/telemetry.rs
```diff
@@ -0,0 +1,368 @@
...
+//  - Latest height of each validator
+//  - Percentage of proposals that are converted into certificates
+//  - Fullness of proposals
+//  - Connectivity (how many other validators are they connected to)
+//  - Various stake weight considerations
+//  - The latest seen IP address of each validator (useful for debugging purposes)
+
+/// Tracker for the participation metrics of validators.
+#[derive(Clone, Debug)]
+pub struct Telemetry<N: Network> {
```
ljedrz, line 41: nit: with all the members `Arc`ed, if this object is expected to be cloned often, that operation would be faster if it was wrapped in a single `Arc` (and not to derive `Clone`); that way every clone would only involve a single atomic op instead of (currently) 4

node/bft/src/bft.rs
raychu86: Resolved with [2ba93d8](https://github.com/ProvableHQ/snarkOS/pull/3516/commits/2ba93d833ab3a0017bca6a071d8260838eac4521)

node/bft/src/gateway.rs
raychu86: Resolved with [2ba93d8](https://github.com/ProvableHQ/snarkOS/pull/3516/commits/2ba93d833ab3a0017bca6a071d8260838eac4521)

node/bft/src/gateway.rs
raychu86: Resolved with [2ba93d8](https://github.com/ProvableHQ/snarkOS/pull/3516/commits/2ba93d833ab3a0017bca6a071d8260838eac4521)

node/bft/src/helpers/mod.rs
raychu86: Resolved with [2ba93d8](https://github.com/ProvableHQ/snarkOS/pull/3516/commits/2ba93d833ab3a0017bca6a071d8260838eac4521)

node/bft/src/helpers/telemetry.rs
```diff
@@ -0,0 +1,368 @@
...
+        // Insert the certificate
+        validator_certificates.entry(certificate_author).or_default().insert(certificate_round);
+    }
+
+    /// Calculate and update the participation scores for each validator.
+    pub fn update_participation_scores(&self) {
+        // Fetch the certificates and signatures.
+        let tracked_certificates = self.tracked_certificates.read();
+        let validator_signatures = self.validator_signatures.read();
+        let validator_certificates = self.validator_certificates.read();
```
ljedrz, line 169: these 3 collections appear to always change together; it would be more performant and less prone to incoherence if they were all under a single `RwLock`

node/bft/src/helpers/telemetry.rs
```diff
@@ -0,0 +1,368 @@
...
+
+        // Fetch the total number of certificates.
+        let num_certificate_rounds = tracked_certificates.len();
+        let total_certificates = validator_certificates.values().map(|rounds| rounds.len()).sum::<usize>();
+
+        // Calculate the signature participation scores for each validator.
+        let signature_participation_scores = Arc::new(RwLock::new(IndexMap::new()));
```
ljedrz: also applies to the other 2 instances below: couldn't we use `collect` to populate these collections instead of locks and individual inserts? `indexmap` has a `rayon` feature, this should be possible

ljedrz (commented):
LGTM in general, left some potentially impactful suggestions for performance adjustments

niklaslong (commented):
Did a first pass, LGTM overall and I have tested the rest endpoint, will wait to approve as I know you have some pending tweaks.

node/bft/src/helpers/telemetry.rs
raychu86: The intention in the future is to add additional metrics that could be independent. Which is why I kept the trackers as individual Arcs.

node/bft/src/helpers/telemetry.rs
raychu86: This object should not be cloned too often, as it should perpetually exist within gateway rather than passed around.

I think unifying Arcs can be in a subsequent PR if we decide to add additional metrics

node/bft/src/helpers/telemetry.rs
raychu86: Resolved with [674a213](https://github.com/ProvableHQ/snarkOS/pull/3516/commits/674a2139c08b1a64652310e17e870c46c05a6122)

raychu86:
Should be ready now, I won't make further tweaks to add functionality, only fixes.

node/bft/src/helpers/telemetry.rs
ljedrz: we can always put those 3 under a single lock later on in case we need extra performance

niklaslong (approved):

ljedrz (approved):
LGTM

PR #3524, opened 2025-03-04
Fix and extend some documentation.

acoglio:
This is for the syncing of blocks using the DAG, which exhibits some complexity and needs more explanation. Some references to blocks also needed to be fixed.

### DIFF
--- a/node/bft/src/sync/mod.rs
+++ b/node/bft/src/sync/mod.rs
@@ -61,6 +61,8 @@ pub struct Sync<N: Network> {
     /// The sync lock.
     sync_lock: Arc<TMutex<()>>,
     /// The latest block responses.
+    /// This is used in [`Sync::sync_storage_with_block()`] to accumulate blocks
+    /// whose addition to the ledger is deferred until certain checks pass.
     latest_block_responses: Arc<TMutex<HashMap<u32, Block<N>>>>,
 }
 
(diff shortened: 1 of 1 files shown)


node/bft/src/sync/mod.rs
```diff
@@ -471,10 +481,15 @@ impl<N: Network> Sync<N> {
...
-        // Ensure that the previous block's leader certificate meets the availability threshold
-        // based on the certificates in the current block.
-        // If the availability threshold is not met, process the next block and check if it is linked to the current block.
+        // Check if each block response, from the contiguous sequence just constructed,
+        // is ready to be added to the ledger.
+        // Ensure that the block's leader certificate meets the availability threshold
+        // based on the certificates in the DAG just after the block's round.
+        // If the availability threshold is not met,
+        // process the next block and check if it is linked to the current block,
+        // in the sense that there is a path in the DAG
```
kaimast, line 492: We currently set rustfmt to 120 character lines IIRC. So you can probably reduce the amount of line breaks you have in these comments, or is this how rustfmt formatted it?

kaimast:
Do we have a style guide for doc comments? I think [this RFC](https://github.com/rust-lang/rfcs/blob/master/text/1574-more-api-documentation-conventions.md) is the most up-to-date guide for Rust docs, and we can just follow that if there is no project specific rule.

How I understand it is that you generally have one sentence that give a high level explanation of what the function does, then there is an empty line, and after that we add any additional explanation or context.

node/bft/src/sync/mod.rs
acoglio: I normally put line breaks according to sentence "structure", since I find that easier to read and write, and it helps minimize diffs if there's any future change (e.g. removing a word and reformatting text makes it look like a much bigger change). But I can conform to a different convention, if there's agreement on one.

acoglio:
It would be great to have a uniform convention. I added two blank lines in the `///` comments I added in this PR, per the RFC.

kaimast:
LGTM!

kaimast (approved):

vicsn (approved):

PR #3528, opened 2025-03-06
Fix two committee retrievals.

acoglio:
When calculating stake of (authors of) certificates, the lookback committee must be the one for the round of the certificates, not the one before it, although currently they happen to be equal.

A longer explanation/discussion was on Slack.

Also make some comments (about that code) more clear.


node/bft/src/sync/mod.rs
```diff
@@ -489,11 +489,12 @@ impl<N: Network> Sync<N> {
             let commit_round = leader_certificate.round();
             let certificate_round = commit_round.saturating_add(1);
```
kaimast: I am curious why we use `saturating_add` in places like this, and if a `checked_add` makes more sense. If the round number overflows, something is really wrong, but this code will just silently continue.

node/bft/src/sync/mod.rs
```diff
@@ -489,11 +489,12 @@ impl<N: Network> Sync<N> {
...
 
-            // Get the committee lookback for the commit round.
-            let committee_lookback = self.ledger.get_committee_lookback_for_round(commit_round)?;
-            // Retrieve all of the certificates for the **certificate** round.
+            // Get the committee lookback for the round just after the leader.
+            let committee_lookback = self.ledger.get_committee_lookback_for_round(certificate_round)?;
```
kaimast: Should we also name this `certificate_committee_lookback`?

kaimast:
Let some minor comments (one is more of a question) but it looks fine and mergeable to me.

kaimast (approved):

niklaslong (dismissed):
LGTM 👍

node/bft/src/sync/mod.rs
acoglio: Yes, a plain `+` would be simpler, faster, and more readable, here and in many other places. The whole thing would fail catastrophically if we ever ran out of round numbers, or block heights, so there is an underlying assumption that we'll always be well below the max's, for the lifetime of the implementation.

acoglio: I guess a `checked_add` could protect against potential implementation errors in release mode, although I don't think that the invariants about always being well below the max should be difficult to establish (informally).

node/bft/src/sync/mod.rs
acoglio: Sure. Pushed.

node/bft/src/sync/mod.rs
raychu86: It's fine to to checked_add here and error if there is an overflow.

raychu86 (dismissed):
LGTM

node/bft/src/sync/mod.rs
acoglio: I've pushed a commit that changes it to a checked_add.

vicsn (dismissed):

node/bft/src/sync/mod.rs
```diff
@@ -487,13 +487,14 @@ impl<N: Network> Sync<N> {
                 _ => bail!("Received a block with an unexpected authority type."),
             };
             let commit_round = leader_certificate.round();
-            let certificate_round = commit_round.saturating_add(1);
+            let certificate_round = commit_round.checked_add(1).unwrap();
```
raychu86: Do not include `unwraps` in non-testing context. 

```suggestion
            let certificate_round = commit_round.checked_add(1).ok_or_else(|| anyhow!("Integer overflow on round number"))?;
```

raychu86 (changes requested):

acoglio:
@raychu86 If this PR looks good, could you approve it?

raychu86 (dismissed):

node/bft/src/bft.rs
```diff
@@ -506,6 +506,11 @@ impl<N: Network> BFT<N> {
             // TODO (howardwu): Investigate how many certificates we should have at this point.
             bail!("BFT failed to retrieve the certificates for certificate round {certificate_round}");
         };
+        // Retrieve the committee lookback for the certificate round (i.e. the round just after the commit round).
+        let Ok(certificate_committee_lookback) = self.ledger().get_committee_lookback_for_round(certificate_round)
+        else {
+            bail!("BFT failed to retrieve the committee with lag for certificate round {certificate_round}");
```
raychu86:
```suggestion
            bail!("BFT failed to retrieve the committee lookback for certificate round {certificate_round}");
```

niklaslong (dismissed):

vicsn (approved):

kaimast (approved):
Looks great!

raychu86 (approved):

PR #3530, opened 2025-03-10
Fix bogus format string.

mikebenfield:
This was apparently an attempt to use formatting syntax outside of a formatting macro.


vicsn (approved):

mikebenfield:
FYI I don't have write access, so I can't merge this.

PR #3537, opened 2025-03-12
[Fix] Check certificate validity early.

### LINKED ISSUE #2935: [Bug] A malicious validator can broadcast invalid BatchCertificate or Propose that cause other validators stack overflow

acoglio:
Before recursively fetching previous certificates.

Closes #2935.

Since it looks like we had working code to run the exploit, it would be good to revive it and make sure it no  longer crashes the validator.

### DIFF
--- a/node/bft/src/helpers/storage.rs
+++ b/node/bft/src/helpers/storage.rs
@@ -22,14 +22,13 @@ use snarkvm::{
         narwhal::{BatchCertificate, BatchHeader, Transmission, TransmissionID},
     },
     prelude::{Address, Field, Network, Result, anyhow, bail, ensure},
-    utilities::{cfg_into_iter, cfg_sorted_by},
+    utilities::{cfg_into_iter, cfg_iter, cfg_sorted_by},
 };
 
 use indexmap::{IndexMap, IndexSet, map::Entry};
 use lru::LruCache;
 use parking_lot::RwLock;
 use rayon::iter::{IntoParallelIterator, IntoParallelRefIterator, ParallelIterator};
-use snarkvm::algorithms::cfg_iter;
 use std::{
     collections::{HashMap, HashSet},
     num::NonZeroUsize,
--- a/node/bft/src/helpers/storage.rs
+++ b/node/bft/src/helpers/storage.rs
@@ -28,7 +28,8 @@ use snarkvm::{
 use indexmap::{IndexMap, IndexSet, map::Entry};
 use lru::LruCache;
 use parking_lot::RwLock;
-use rayon::iter::{IntoParallelIterator, ParallelIterator};
+use rayon::iter::{IntoParallelIterator, IntoParallelRefIterator, ParallelIterator};
+use snarkvm::algorithms::cfg_iter;
 use std::{
     collections::{HashMap, HashSet},
     num::NonZeroUsize,
@@ -492,6 +493,49 @@ impl<N: Network> Storage<N> {
             }
         }
         Ok(missing_transmissions)
+    }
+
+    /// Check the validity of a certificate coming from another validator.
+    ///
+    /// It suffices to check that the signers (author and endorsers) are members of the applicable committee
+    /// and that they form a quorum in the committee.
+    /// Under the fundamental fault tolerance assumption of at most `f` (stake of) faulty validators,
+    /// the quorum check on signers guarantees that at least one correct validator
+    /// has ensured the validity of the proposal contained in the certificate,
+    /// either by construction (by the author) or by checking (by an endorser):
+    /// given `N > 0` total stake, and `f` the largest integer `< N/3` (where `/` is exact rational division),
+    /// we have `N >= 3f + 1`, which implies `N - f >= 2f + 1`, which is always `> f`;
+    /// `N - f` is the quorum stake.
+    pub fn check_incoming_certificate(&self, certificate: &BatchCertificate<N>) -> Result<()> {
+        // Retrieve the certificate author and round.
+        let certificate_author = certificate.author();
+        let certificate_round = certificate.round();
+
+        // Retrieve the committee lookback.
+        let committee_lookback = self.ledger.get_committee_lookback_for_round(certificate_round)?;
+
+        // Ensure that the signers of the certificate reach the quorum threshold.
+        // Note that certificate.signatures() only returns the endorsing signatures, not the author's signature.
+        let mut signers: HashSet<Address<N>> =
+            certificate.signatures().map(|signature| signature.to_address()).collect();
+        signers.insert(certificate_author);
+        ensure!(
+            committee_lookback.is_quorum_threshold_reached(&signers),
+            "Certificate '{}' for round {certificate_round} does not meet quorum requirements",
+            certificate.id()
+        );
+
+        // Ensure that the signers of the certificate are in the committee.
+        cfg_iter!(signers).try_for_each(|signer| {
+            ensure!(
+                committee_lookback.is_committee_member(*signer),
+                "Signer '{signer}' of certificate '{}' for round {certificate_round} is not in the committee",
+                certificate.id()
+            );
+            Ok(())
+        })?;
+
+        Ok(())
     }
 
     /// Checks the given `certificate` for validity, returning the missing transmissions from storage.
--- a/node/bft/src/primary.rs
+++ b/node/bft/src/primary.rs
@@ -955,14 +955,50 @@ impl<N: Network> Primary<N> {
             bail!("Received a batch certificate for myself ({author})");
         }
 
+        // Retrieve the committee lookback.
+        let committee_lookback = self.ledger.get_committee_lookback_for_round(certificate_round)?;
+
+        // Ensure that the signers of the certificate reach the quorum threshold.
+        let signers = certificate.signatures().map(|signature| signature.to_address()).collect();
+        ensure!(
+            committee_lookback.is_quorum_threshold_reached(&signers),
+            "Certificate '{}' for round {certificate_round} does not meet quorum requirements",
+            certificate.id()
+        );
+
+        // Ensure that the signers of the certificate are in the committee.
+        for signer in signers {
+            ensure!(
+                committee_lookback.is_committee_member(signer),
+                "Signer '{signer}' of certificate '{}' for round {certificate_round} is not in the committee",
+                certificate.id()
+            )
+        }
+
+        // Under the fundamental fault tolerance assumption of at most f (stake of) faulty validators,
+        // the above check on signers guarantees that at least one correct validator
+        // has ensured the validity of the proposal,
+        // either by construction (by the author) or by checking (by an endorser):
+        // given N > 0 and f the largest integer < N/3 (where / is exact rational division),
+        // we have N >= 3f + 1, which implies N - f >= 2f + 1, which is always > f.
+        // With the above check that the signers are in the committee,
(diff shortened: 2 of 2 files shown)


kaimast (approved):
Looks good to me!

node/bft/src/primary.rs
```diff
@@ -955,14 +955,53 @@ impl<N: Network> Primary<N> {
...
+        // Note that certificate.signatures() only returns the endorsing signatures, not the author's signature.
+        let mut signers: HashSet<Address<N>> =
+            certificate.signatures().map(|signature| signature.to_address()).collect();
+        signers.insert(author);
+        ensure!(
+            committee_lookback.is_quorum_threshold_reached(&signers),
+            "Certificate '{}' for round {certificate_round} does not meet quorum requirements",
+            certificate.id()
+        );
+
```
raychu86: We should already be performing this check in the subsequent call to `sync_with_certificate_from_peer` -> `insert_certificate` -> `check_certificate`.

Why not add the following `is_committee_member` checks into `check_certificate`?

node/bft/src/primary.rs
```diff
@@ -955,14 +955,53 @@ impl<N: Network> Primary<N> {
...
+        // Ensure that the signers of the certificate are in the committee.
+        for signer in signers {
```
raychu86: This check can be parallelized with `cfg_iter`

node/bft/src/primary.rs
acoglio: We perform the quorum signer check in check_certificate, but that's after the recursive fetching of certificates, which is the reason why the attack in #2935 works. By performing the check before the recursive fetching, we prevent that attack. Then the same check in check_certificate becomes redundant, along with essentially all of the checks done there.

Regarding the separate check that the signers are committee members, I think we could maybe safely perform it later, but it seemed cleaner to do all the necessary checks in one place (in the code added by this commit).

raychu86: That makes sense, but then it seems like we are doing these signature checks multiple times, which can get expensive. 

Additionally, for cleanliness, it might be better to create a helper method for these checks since they are done in multiple places.

acoglio: Yes, we should eliminate those other checks, but I was thinking of doing that in a separate PR?

node/bft/src/primary.rs
acoglio: Done.

node/bft/src/primary.rs
acoglio: I added a `Storage::check_incoming_certificate()` method that could be used to check all incoming certificates from other validators. Certificates created by the current validator must already satisfy the checks by construction. So we should be able to eliminate several redundant checks, as net effect, compared to before. But we need to examine the callers a bit better, which is why I'm thinking that a separate PR to remove those redundant checks would be better.

raychu86 (approved):

vicsn:
Asked Kai some questions in DM about testing this.

kaimast:
Quick update: I was able to reproduce the attack. I will verify that the attack does not work anymore with the proposed changes soon.

Here are some observations I made:
1. There are already some checks in place that reject batches that contain the author as one of the endorsers. The original attack code does that because a certificate needs at least one signer. But this can be circumvented by having a random other address or a second malicious validator sign the batch. There is no way to verify the signer is valid before retrieving the full certificate chain, unless I am missing something.
2. SnarkVM also rejects batches that have timestamps too close together. The original attack set the same timestamp for all fake certificates, and correct nodes will reject these.
3. What seems to work is setting timestamps in the future. Nodes probably should reject batches that are more than a second in the future. On my branch, nodes seem to (at least initially) accept batches that are multiple minutes in the future.

node/bft/src/helpers/storage.rs
```diff
@@ -494,6 +494,49 @@ impl<N: Network> Storage<N> {
...
+    /// It suffices to check that the signers (author and endorsers) are members of the applicable committee
+    /// and that they form a quorum in the committee.
+    /// Under the fundamental fault tolerance assumption of at most `f` (stake of) faulty validators,
+    /// the quorum check on signers guarantees that at least one correct validator
+    /// has ensured the validity of the proposal contained in the certificate,
+    /// either by construction (by the author) or by checking (by an endorser):
+    /// given `N > 0` total stake, and `f` the largest integer `< N/3` (where `/` is exact rational division),
+    /// we have `N >= 3f + 1`, which implies `N - f >= 2f + 1`, which is always `> f`;
+    /// `N - f` is the quorum stake.
+    pub fn check_incoming_certificate(&self, certificate: &BatchCertificate<N>) -> Result<()> {
```
kaimast, line 508: I think we need to change the name of either this function or `check_certificate` to make it more clear which one is used when.

node/bft/src/helpers/storage.rs
```diff
@@ -494,6 +494,49 @@ impl<N: Network> Storage<N> {
...
+        // Retrieve the committee lookback.
+        let committee_lookback = self.ledger.get_committee_lookback_for_round(certificate_round)?;
+
+        // Ensure that the signers of the certificate reach the quorum threshold.
+        // Note that certificate.signatures() only returns the endorsing signatures, not the author's signature.
+        let mut signers: HashSet<Address<N>> =
+            certificate.signatures().map(|signature| signature.to_address()).collect();
+        signers.insert(certificate_author);
+        ensure!(
+            committee_lookback.is_quorum_threshold_reached(&signers),
```
kaimast, line 522: If we check the threshold here, do we still need to retrieve the`committee_lookback` in `process_batch_certificate_from_peer`?

node/bft/src/primary.rs
```diff
@@ -955,14 +955,28 @@ impl<N: Network> Primary<N> {
             bail!("Received a batch certificate for myself ({author})");
         }
 
-        // Store the certificate, after ensuring it is valid.
+        // Ensure that the incoming certificate is valid.
+        self.storage.check_incoming_certificate(&certificate)?;
```
kaimast, line 959: If the certificate is more than 100 rounds in the future, would we just fail here? Could that happen with a correct/honest peer if there is a network issue?

node/bft/src/helpers/storage.rs
kaimast, line 508: I also think we should test the attack and, ideally, add some other unit tests to this function.

If you want, I can try to write a unit test from the attack code I have and push it to this branch.

kaimast (changes requested):
I could verify the fix works (yay!), but, now that I have a better understanding of the code, I also left a few more notes on the PR.

To verify, this is the error message I get when honest nodes receive the attack: 
```
Cannot store a certificate from '127.0.0.1:5004' - Certificate '2671926863586933361463317741031072687577528356414354042732451240158089381291field' for round 5 does not meet quorum requirements
```

Another thing we should consider, not in this PR but sometime in the future, is removing the recursion entirely, because there could still be valid cases where you receive a fairly long certificate chain. Alternatively, we could ensure that the number of recursions always stays below some small bound.
For example, it might be more elegant to have a data structure that maps certificate ID to unprocessed certificates that depend on it. Whenever we accept a batch, we check that mapping for certificates to process next.

node/bft/src/primary.rs
```diff
@@ -955,14 +955,28 @@ impl<N: Network> Primary<N> {
...
+        // The following call recursively fetches and stores
+        // the previous certificates referenced from this certificate.
+        // It is critical to make the following call this after validating the certificate above.
+        // The reason is that a sequence of malformed certificates,
+        // with references to previous certificates with non-decreasing rounds,
+        // cause the recursive fetching of certificates to crash the validator due to resource exhaustion.
+        // Note that if the following call, if not returning an error, guarantees the backward closure of the DAG
+        // (i.e. that all the referenced previous certificates are in the DAG before storing this one),
+        // then all the validity checks in [`Storage::check_certificate`] should be redundant.
+        // TODO: eliminate those redundant checks
```
kaimast, line 971: Are we removing the redundant checks in another PR?

acoglio:
> To verify, this is the error message I get when honest nodes receive the attack:
> 
> ```
> Cannot store a certificate from '127.0.0.1:5004' - Certificate '2671926863586933361463317741031072687577528356414354042732451240158089381291field' for round 5 does not meet quorum requirements
> ```

Nice! That's indeed the error I was expecting to happen: the certificate doesn't have enough signatures. Under the correct super-majority assumption, the validator would be unable to obtain enough signatures.

node/bft/src/helpers/storage.rs
acoglio: I agree that we should change the names. I was thinking that we could do that as we refactor and simplify the certificate checks in subsequent PRs.

node/bft/src/helpers/storage.rs
acoglio: `process_batch_certificate` from peer uses the committee for a different purpose, namely deciding whether to advance the round.

node/bft/src/primary.rs
acoglio: Yes, we would fail because the validator would be unable to calculate the active committee for the certificate's round, given that its blockchain is too behind compared to the round. This ties in with more general liveness issues, which need further study.

node/bft/src/helpers/storage.rs
acoglio: That would be great, if you are willing to do that.

node/bft/src/primary.rs
acoglio: Yes, there are currently several redundant checks (not just this one), and they are a bit complicated (I documented them in the code notes doc), so my thinking was to tackle those in a series of PRs each of which makes a very simple change that is easy to evaluate.

node/bft/src/helpers/storage.rs
kaimast: I am looking into this now, but it might require adding some code to snarkVM to generate test certificates that have valid (not just random) signatures.

kaimast:
I have a branch [here](https://github.com/acoglio/snarkOS/compare/fix-cert-fetch...kaimast:snarkOS:fix-cert-fetch) that adds unit tests, but it requires [a small change](https://github.com/ProvableHQ/snarkVM/pull/2665) to snarkVM. 

We could merge this PR now and I will add the tests with another PR. I tested it extensively, and it should land in staging sooner than later, considering it prevents a possible attack.

PR #3538, opened 2025-03-13
[Feat] Introduce a lock accounting feature

### LINKED ISSUE #3533: [Feature] Enable locktick with a feature flag
## 🚀 Feature

It has proven very useful, we should keep it around so we can enable it on the fly.

ljedrz:
The snarkOS counterpart to https://github.com/ProvableHQ/snarkVM/pull/2647.

Note: this PR will remain in a draft state until the snarkVM PR is merged.

Closes https://github.com/ProvableHQ/snarkOS/issues/3533.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,7 +47,7 @@ default-features = false
 [workspace.dependencies.snarkvm] # If this is updated, the rev in `node/rest/Cargo.toml` must be updated as well.
 # path = "../snarkVM"
 git = "https://github.com/ProvableHQ/snarkVM.git"
-rev = "c6558fb"
+rev = "69f97fa"
 #version = "=1.3.0"
 features = [ "circuit", "console", "rocks" ]
 
@@ -68,6 +68,16 @@ cuda = [
   "snarkos-node-rest/cuda",
   "snarkos-node-router/cuda",
   "snarkos-node-sync/cuda"
+]
+locktick = [
+  "snarkos-node/locktick",
+  "snarkos-node-bft/locktick",
+  "snarkos-node-consensus/locktick",
+  "snarkos-node-metrics/locktick",
+  "snarkos-node-rest/locktick",
+  "snarkos-node-router/locktick",
+  "snarkos-node-sync/locktick",
+  "snarkos-node-tcp/locktick"
 ]
 test_targets = [ "snarkos-cli/test_targets" ]
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -70,6 +70,8 @@ cuda = [
   "snarkos-node-sync/cuda"
 ]
 locktick = [
+  "dep:locktick",
+  "dep:tracing",
   "snarkos-node/locktick",
   "snarkos-node-bft/locktick",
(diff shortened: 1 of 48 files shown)


Cargo.toml
```diff
@@ -139,6 +154,9 @@ version = "2"
 opt-level = 3
 lto = "thin"
 incremental = true
+# uncomment the 2 lines below when building with the locktick feature
```
vicsn, line 173: Should this be in the dev mode as well in case we want to use it there?

vicsn (commented):
You are the GOAT. Really excited to be carrying this around in case shit hits the fan one day.

Cargo.toml
ljedrz: It can be used with any profile, though I don't think we're typically running nodes with the `dev` one (i.e. without `--release`), as they'd be rather slow; I looked into ways of enabling profile overrides when specific features are enabled, but the stable solutions are currently somewhat hacky. That's why the 2 required extra options are commented out; they can be moved under any profile when they're being enabled, though.

niklaslong (commented):
LGTM, pending the merge + VM update once the other PR is merged 🚀

kaimast (approved):
Looks good to me.

I was wondering if there is a way to avoid the repeated conditional imports., but cannot think of a nicer way to do this.
My worry is that a future PR might add new locks and not know about the locktick feature.

Cargo.toml
```diff
@@ -70,6 +82,10 @@ features = [ "derive" ]
 [dependencies.crossterm]
 version = "0.27.0"
 
+[dependencies.locktick]
+version = "0.3"
+optional = true
+
```
kaimast, line 99: Not sure if it makes a huge difference, but you can add `default-features=false` here, as we don't use the `std::sync` locks as far as I can tell.

ljedrz:
> I was wondering if there is a way to avoid the repeated conditional imports., but cannot think of a nicer way to do this.

Same here; I was considering TOML overrides and additional reexports, but it's hacky and troublesome with different kinds of locks at once.

Cargo.toml
ljedrz: The difference would be marginal, and compile-time only; that feature is meant to not be enabled in `nostd` conditions. In addition, since the `std` locks are (these days) more likely to have a better performance than the `parking_lot` ones, we might want to consider using them for some especially "hot" objects in the future, and having this feature around would make it easier to make such swaps for comparison.

vicsn:
This PR can now be rebased, update snarkVM rev and opened.

vicsn (approved):

vicsn:
> I was wondering if there is a way to avoid the repeated conditional imports., but cannot think of a nicer way to do this.
> My worry is that a future PR might add new locks and not know about the locktick feature.

snarkVM has `cfg_iter` macros which enable parallelization only if a feature flag is enabled.

similarly, we could hide the lock primitives into snarkVM... Is this something we want? Maybe a nice first contributor issue? CC @ljedrz @kaimast

PR #3540, opened 2025-03-14
[Fix] Harden the router's resolver

ljedrz:
While investigating a potential issue with some trusted peers being periodically dropped, I've noticed a lot of instances of `Unable to resolve the (...) address` in the log extracts from different networks. I believe most of them are triggered unnecessarily, but we need to be sure, and this PR aims to address this.

The proposed changes are as follows:
- 1e8dc493916cce871686dcdb71bc1db2b6d3091a - changes the dual-lock setup of the resolver to a single-lock one in order to avoid any possibility of mismatch between the address maps; it should also slightly improve its performance
- 260e84bc2d9949e53c4beaa270c905834859c711 - the `inbound` method is "fed" from a lower-level queue which doesn't have an awareness of the address resolver, so the entries that fail to resolve there are basically guaranteed to be post-disconnect "stragglers" and may be ignored (instead of triggering potentially many redundant disconnect attempts, which result in further resolver-related warnings)
- 6bb8a74b95149d4df04033586e6e0237d6de21fe - this swaps the order of disconnect-related operations, altering the resolver only after a peer is no longer marked as connected; this will avoid situations where an outbound message is greenlit to be sent to a peer (who is marked as connected) only to fail at address resolution right afterwards, triggering a bogus warning
- bccf29a99753eaf4e1b204666e7051651451677a - this is a loosely-related drive-by; we should clear any peer-related cache before marking them as a candidate for connections, in order to avoid a (highly unlikely) scenario where the peer is reconnected to while having outdated cache entries, or even having new and applicable cache entries cleared
- 7ee66b44d318dc1363f633cd4ce8052958c89d92 - when a peer sends us a `Message::Disconnect`, we shouldn't report it as a protocol violation; this is mostly a cleanup of one or two misleading logs
- b673d7b1b8ccd9ce49f8202a3056674341f5dc2c - since I've seen some instances of the heartbeat process reporting lingering inactive peers, we should have a fallback cleanup of high-level connection artifacts in case the resolver can't find the physically connected address

Update: the PR was rebased due to a conflict, and while these commit hashes have changed, their contents or order haven't.

### DIFF
--- a/node/router/src/handshake.rs
+++ b/node/router/src/handshake.rs
@@ -211,7 +211,7 @@ impl<N: Network> Router<N> {
         // Finalize the connecting peer information.
         self.connecting_peers.lock().insert(peer_ip, Some(Peer::new(peer_ip, peer_addr, &peer_request)));
         // Adds a bidirectional map between the listener address and (ambiguous) peer address.
-        self.resolver.insert_peer(peer_ip, peer_addr);
+        self.resolver.write().insert_peer(peer_ip, peer_addr);
 
         Ok((peer_ip, framed))
     }
@@ -296,7 +296,7 @@ impl<N: Network> Router<N> {
         // Finalize the connecting peer information.
         self.connecting_peers.lock().insert(peer_ip, Some(Peer::new(peer_ip, peer_addr, &peer_request)));
         // Adds a bidirectional map between the listener address and (ambiguous) peer address.
-        self.resolver.insert_peer(peer_ip, peer_addr);
+        self.resolver.write().insert_peer(peer_ip, peer_addr);
 
         Ok((peer_ip, framed))
     }
--- a/node/router/src/helpers/resolver.rs
+++ b/node/router/src/helpers/resolver.rs
@@ -13,18 +13,14 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#[cfg(feature = "locktick")]
-use locktick::parking_lot::RwLock;
-#[cfg(not(feature = "locktick"))]
-use parking_lot::RwLock;
 use std::{collections::HashMap, net::SocketAddr};
 
 #[derive(Debug)]
 pub struct Resolver {
     /// The map of the listener address to (ambiguous) peer address.
-    from_listener: RwLock<HashMap<SocketAddr, SocketAddr>>,
+    from_listener: HashMap<SocketAddr, SocketAddr>,
     /// The map of the (ambiguous) peer address to listener address.
-    to_listener: RwLock<HashMap<SocketAddr, SocketAddr>>,
+    to_listener: HashMap<SocketAddr, SocketAddr>,
 }
 
 impl Default for Resolver {
@@ -42,24 +38,24 @@ impl Resolver {
 
     /// Returns the listener address for the given (ambiguous) peer address, if it exists.
     pub fn get_listener(&self, peer_addr: &SocketAddr) -> Option<SocketAddr> {
-        self.to_listener.read().get(peer_addr).copied()
+        self.to_listener.get(peer_addr).copied()
     }
 
     /// Returns the (ambiguous) peer address for the given listener address, if it exists.
     pub fn get_ambiguous(&self, peer_ip: &SocketAddr) -> Option<SocketAddr> {
-        self.from_listener.read().get(peer_ip).copied()
+        self.from_listener.get(peer_ip).copied()
     }
 
     /// Inserts a bidirectional mapping of the listener address and the (ambiguous) peer address.
-    pub fn insert_peer(&self, listener_ip: SocketAddr, peer_addr: SocketAddr) {
-        self.from_listener.write().insert(listener_ip, peer_addr);
-        self.to_listener.write().insert(peer_addr, listener_ip);
+    pub fn insert_peer(&mut self, listener_ip: SocketAddr, peer_addr: SocketAddr) {
+        self.from_listener.insert(listener_ip, peer_addr);
+        self.to_listener.insert(peer_addr, listener_ip);
     }
 
     /// Removes the bidirectional mapping of the listener address and the (ambiguous) peer address.
-    pub fn remove_peer(&self, listener_ip: &SocketAddr) {
-        if let Some(peer_addr) = self.from_listener.write().remove(listener_ip) {
-            self.to_listener.write().remove(&peer_addr);
+    pub fn remove_peer(&mut self, listener_ip: &SocketAddr) {
+        if let Some(peer_addr) = self.from_listener.remove(listener_ip) {
+            self.to_listener.remove(&peer_addr);
         }
     }
 }
--- a/node/router/src/helpers/resolver.rs
+++ b/node/router/src/helpers/resolver.rs
@@ -15,7 +15,7 @@
 
 use std::{collections::HashMap, net::SocketAddr};
 
-#[derive(Debug)]
+#[derive(Debug, Default)]
 pub struct Resolver {
     /// The map of the listener address to (ambiguous) peer address.
     from_listener: HashMap<SocketAddr, SocketAddr>,
@@ -23,19 +23,7 @@ pub struct Resolver {
     to_listener: HashMap<SocketAddr, SocketAddr>,
 }
 
-impl Default for Resolver {
-    /// Initializes a new instance of the resolver.
-    fn default() -> Self {
-        Self::new()
-    }
-}
-
 impl Resolver {
-    /// Initializes a new instance of the resolver.
-    pub fn new() -> Self {
-        Self { from_listener: Default::default(), to_listener: Default::default() }
-    }
-
     /// Returns the listener address for the given (ambiguous) peer address, if it exists.
     pub fn get_listener(&self, peer_addr: &SocketAddr) -> Option<SocketAddr> {
         self.to_listener.get(peer_addr).copied()
--- a/node/router/src/helpers/resolver.rs
+++ b/node/router/src/helpers/resolver.rs
@@ -15,6 +15,9 @@
 
 use std::{collections::HashMap, net::SocketAddr};
 
+/// The `Resolver` provides the means to map the connected address (used in the lower-level
+/// `tcp` module internals, and provided by the OS) to the listener address (used as the
+/// unique peer identifier in the higher-level functions), and the other way around.
 #[derive(Debug, Default)]
 pub struct Resolver {
     /// The map of the listener address to (ambiguous) peer address.
--- a/node/router/src/helpers/resolver.rs
+++ b/node/router/src/helpers/resolver.rs
@@ -19,7 +19,7 @@ use std::{collections::HashMap, net::SocketAddr};
 /// `tcp` module internals, and provided by the OS) to the listener address (used as the
 /// unique peer identifier in the higher-level functions), and the other way around.
 #[derive(Debug, Default)]
-pub struct Resolver {
+pub(crate) struct Resolver {
     /// The map of the listener address to (ambiguous) peer address.
     from_listener: HashMap<SocketAddr, SocketAddr>,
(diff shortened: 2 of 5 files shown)


niklaslong (commented):
A nice tightening up of peer tracking! Did a first pass and the current changeset looks good 👍

node/router/src/inbound.rs
```diff
@@ -65,7 +65,10 @@ pub trait Inbound<N: Network>: Reading + Outbound<N> {
         // Retrieve the listener IP for the peer.
         let peer_ip = match self.router().resolve_to_listener(&peer_addr) {
             Some(peer_ip) => peer_ip,
-            None => bail!("Unable to resolve the (ambiguous) peer address '{peer_addr}'"),
+            None => {
+                // No longer connected to the peer.
+                return Ok(());
+            }
```
howardwu: @ljedrz @raychu86 can you double check this? `Ok(())` maintains the connection

node/router/src/inbound.rs
```diff
@@ -142,7 +145,10 @@ pub trait Inbound<N: Network>: Reading + Outbound<N> {
                 bail!("Peer '{peer_ip}' is not following the protocol")
             }
             Message::Disconnect(message) => {
-                bail!("{:?}", message.reason)
+                // The peer informs us that they had disconnected. Disconnect from them too.
+                debug!("Peer '{peer_ip}' decided to disconnect due to '{:?}'", message.reason);
+                self.router().disconnect(peer_ip);
+                Ok(())
```
howardwu: @ljedrz @raychu86 can you double check this? `Ok(())` maintains the connection

node/router/src/lib.rs
```diff
@@ -227,6 +227,12 @@ impl<N: Network> Router<N> {
                 }
                 disconnected
             } else {
+                // FIXME (ljedrz): this shouldn't be necessary; it's a double-check
+                //  that the higher-level collection is consistent with the resolver.
+                if router.is_connected(&peer_ip) {
+                    warn!("Fallback connection artifact cleanup (report this to @ljedrz)");
+                    router.remove_connected_peer(peer_ip);
+                }
```
howardwu, line 269: @ljedrz this approach seems like a hack. Can you confirm you've seen empirically ghost IPs/peers left over in the router?

ljedrz: yes, I've seen them in the Canary network, hence this adjustment; last time I checked the logs there were only a handful or less, but it warrants a closer inspection, and justifies this fallback

howardwu:
@ljedrz Do we need to apply the same changes from `Router` to `Gateway`?

ljedrz:
@howardwu not necessarily; my recommendation would be to first introduce  these changes, and then perform a new analysis of the logs, looking for protocol violation false positives and potential connection stability issues. These changes will make the picture a lot more clear.

node/router/src/inbound.rs
raychu86: The `bail` leads back to calling `self.router().resolve_to_listener(&peer_addr)`, so this logic doesn't actually change any real behavior (except for skipping a log message).

node/router/src/inbound.rs
raychu86: This new logic skips the call to 

`Outbound::send(self, peer_ip, Message::Disconnect(DisconnectReason::ProtocolViolation.into()));`. I believe this is fine because in theory the peer has already disconnected you since they are sending a Disconnect message.

joske:
@ljedrz @niklaslong Which logs are you talking about? Were you able to reproduce the issue yourself?

ljedrz:
@joske I was analyzing the logs of one of the Canarynet clients before and after these changes.

joske:
Could you share those logs?

vicsn:
Since Lukasz was driving this and has a lot of context, and because its not a critical issue, I'd suggest deferring until he's back from leave

vicsn:
> Could you share those logs?

I recall very often seeing the errors Lukasz mentioned. I suggest you can just run your own [local canary client](https://github.com/ProvableHQ/snarkOS/?tab=readme-ov-file#312-run-an-aleo-outer-client) as he suggests, if you don't pass any peers you should connect to bootstrap nodes who will connect you to others.

node/router/src/inbound.rs
bendyarm: I agree with the analysis:
Because `resolve_to_listener(&peer_addr)` just returned `None`,
returning `Ok(())` instead of `bail!(..)` from this point to either of the callers
`validator::router::process_message_inner` or `client::router::process_message_inner`
means a warning and an extra failing call to `resolve_to_listener(&peer_addr)` are skipped.
In the case of `prover::router::process_message` there was no warning, so
just the extra failing call to `resolve_to_listener(&peer_addr)` is skipped,
The end result is otherwise identical; in either case the message is not handled.

However:
The confusing part is, what does it mean for `inbound()` to return `Ok(())`?
Normally it means "This inbound message is fine; we handled it; don't try to disconnect.".
In this new particular case it means "Although the peer is not fine and we didn't handle the message,
don't give a warning or try to disconnect because it is no longer connected
and probably already got a warning."

I think the code would be cleaner if the meaning of returning `Ok(())` were not overloaded.

node/router/src/inbound.rs
bendyarm: I agree it should work but like [the other change](https://github.com/ProvableHQ/snarkOS/pull/3540#discussion_r2105443846) in `inbound` it complicates the interface.  In this case the `self.router().disconnect(peer_ip)` has to get copied from `process_message{_internal}` to `inbound`.  This is the only call to `disconnect()` in `inbound`.

node/router/src/inbound.rs
```diff
@@ -65,7 +65,10 @@ pub trait Inbound<N: Network>: Reading + Outbound<N> {
         // Retrieve the listener IP for the peer.
         let peer_ip = match self.router().resolve_to_listener(&peer_addr) {
             Some(peer_ip) => peer_ip,
-            None => bail!("Unable to resolve the (ambiguous) peer address '{peer_addr}'"),
+            None => {
+                // No longer connected to the peer.
```
kaimast, line 84: Could/should we still log here (trace or debug level)?

node/router/src/lib.rs
```diff
@@ -552,14 +558,14 @@ impl<N: Network> Router<N> {
 
     /// Removes the connected peer and adds them to the candidate peers.
     pub fn remove_connected_peer(&self, peer_ip: SocketAddr) {
-        // Removes the bidirectional map between the listener address and (ambiguous) peer address.
-        self.resolver.remove_peer(&peer_ip);
         // Remove this peer from the connected peers, if it exists.
         self.connected_peers.write().remove(&peer_ip);
```
kaimast, line 619: Would it make sense to hold the lock throughout the function, so we update `connected_peers` and `candidate_peers` atomically?

node/router/src/handshake.rs
```diff
@@ -296,7 +296,7 @@ impl<N: Network> Router<N> {
         // Finalize the connecting peer information.
         self.connecting_peers.lock().insert(peer_ip, Some(Peer::new(peer_ip, peer_addr, &peer_request)));
         // Adds a bidirectional map between the listener address and (ambiguous) peer address.
-        self.resolver.insert_peer(peer_ip, peer_addr);
```
kaimast, line 299: Same comment as for `remove_connected_peer`: Would it make sense to hold the lock throughout the function, so we update `connected_peers` and `candidate_peers` atomically?

Note, here we modify the two in a different order. So you would need to lock the resolver before changing connected peers to prevent deadlocks.

node/router/src/helpers/resolver.rs
```diff
@@ -13,15 +13,14 @@
...
 pub struct Resolver {
     /// The map of the listener address to (ambiguous) peer address.
-    from_listener: RwLock<HashMap<SocketAddr, SocketAddr>>,
+    from_listener: HashMap<SocketAddr, SocketAddr>,
     /// The map of the (ambiguous) peer address to listener address.
-    to_listener: RwLock<HashMap<SocketAddr, SocketAddr>>,
+    to_listener: HashMap<SocketAddr, SocketAddr>,
 }
 
 impl Default for Resolver {
```
kaimast: Minor nit unrelated to your PR, but this is a good opportunity to clean up. Can we derive `Default` in line 19 and then either get rid of `new()` or have new just call `Self::default()`?

node/router/src/helpers/resolver.rs
```diff
@@ -13,15 +13,14 @@
...
 // limitations under the License.
 
-use parking_lot::RwLock;
 use std::{collections::HashMap, net::SocketAddr};
 
 #[derive(Debug)]
 pub struct Resolver {
     /// The map of the listener address to (ambiguous) peer address.
-    from_listener: RwLock<HashMap<SocketAddr, SocketAddr>>,
+    from_listener: HashMap<SocketAddr, SocketAddr>,
```
kaimast, line 24: Not a change request, but one thing I wanted to do eventually is introduce a type `struct PeerId(SocketAddr)`
This would make it clearer that we use the listener address as a unique identifier and prevent accidental/implicit conversions between addresses and IDs.

node/router/src/helpers/resolver.rs
```diff
@@ -13,15 +13,14 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use parking_lot::RwLock;
 use std::{collections::HashMap, net::SocketAddr};
 
 #[derive(Debug)]
```
kaimast: It would be good to add a doc comment for the resolver struct. (I thought I did that at some point, but maybe that change never got merged)

node/router/src/inbound.rs
kaimast: To avoid future confusion, could you add a comment to the function documentation on what returning an error vs Ok(()) here means?

node/router/src/lib.rs
```diff
@@ -81,7 +81,7 @@ pub struct InnerRouter<N: Network> {
     /// The cache.
     cache: Cache<N>,
     /// The resolver.
-    resolver: Resolver,
```
kaimast, line 99: Is the resolver only used in the `router` crate? We currently make it public (see the `pub use helpers::*;` above).
If it is only used, and intended to be used, here, we should reduce its visibility.

node/router/src/lib.rs
kaimast: We should open an issue for that, so we don't forget to fix it and to eventually remove those lines.

kaimast (changes requested):
Let a few comments. Sorry if they are too nitpicky...

Could you resolve the conflicts with staging, as well? Hopefully we can get this merged in the coming days.

node/router/src/helpers/resolver.rs
ljedrz: I would actually prefer to use something else entirely than the `SocketAddr` as a means of distinguishing peers; the split between the listener addr vs. connected addr in TCP is basically a footgun.

node/router/src/lib.rs
ljedrz: https://github.com/ProvableHQ/snarkOS/issues/3705

ljedrz:
Rebased (there were no changes to the original commits), and applied the review comments.

ljedrz:
The devnet CI job has failed, but I can't reproduce it locally; can't restart it either.

Update: it was a deadlock which I avoided locally due to starting the nodes less quickly. Fixed.

node/router/src/lib.rs
ljedrz: Actually, it seems like we can't currently do this due to deadlock potential; however, I'm open to tackling this in a follow-up.

kaimast: Sounds good!

kaimast (approved):
LGTM!

PR #3542, opened 2025-03-18
Run devnet in CI and update snarkVM ref

### LINKED ISSUE #3539: [Feature] Create CI job which runs devnet.sh
## 🚀 Feature

This is a very easy low hanging E2E test which would have caught an issue with a recent PR.

Need to add a job in .circleci/config.yml
- Add a job (e.g. copy `node-router`)
- Add a new version of `run_serial` which doesn't run `cargo test` but runs `devnet.sh`

Suggested parameters:
- large machine
- 4 validators
- 2 clients
- max 15 minute runtime
- mainnet
- check whether we advanced at least X blocks on all 6 nodes. 
  - Need to determine what is highly likely to be reached in this timeframe on this machine to avoid false negatives.
  - If the network doesnt advance, do we just wait until the CI times out after 60 minutes? Or do we force a timeout in the curl script!

kpandl:
## Motivation

We recently experienced a merged PR where the devnet doesn't run. Running a devnet in CI would have prevented this.

This PR adds a devnet run to CI, and expects it to reach at least height 45 within 15 minutes of running (assuming a 20 second block time, which would already be quite slow, but we also need to account for some snarkos startup overhead, the duration of it is currently unclear and also depends on the network connection).

To clarify: The test checks the block height for up to 15 minutes after a 1-minute startup period, with a safety mechanism to terminate the job if there's no output for 20 consecutive minutes.

Note that it adds a new non-interactive devnet script inside the CI config to avoid required user inputs and to also avoid a tmux dependency, which can mess with CircleCI. The default values are 4 vals, 2 clients, mainnet.

## Test Plan

Ran local tests using `circleci local execute devnet-test`. Note that it needs to be run in CI next for testing purposes. Also note that in the current staging branch, the CI job is expected to fail. UPDATE: after updating the snarkVM ref to the latest snarkVM staging, it is expected to pass.

Ran the local test on a previous commit hash where the devnet script runs, and this CI test also reports success there.

Closes #3539

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -166,7 +166,7 @@ commands:
           cache_key: << parameters.cache_key >>
 
   run_devnet:
-    description: "Build and run devnet for integration testing"
+    description: "Run devnet for integration testing"
     parameters:
       workspace_member:
         type: string
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -165,6 +165,151 @@ commands:
       - clear_environment:
           cache_key: << parameters.cache_key >>
 
+  run_devnet:
+    description: "Build and run devnet for integration testing"
+    parameters:
+      workspace_member:
+        type: string
+      cache_key:
+        type: string
+      validators:
+        type: integer
+        default: 4
+      clients:
+        type: integer
+        default: 2
+      network_id:
+        type: integer
+        default: 0  # mainnet, note URL below needs to be adjusted if this is changed
+      min_height:
+        type: integer
+        default: 45
+    steps:
+      - checkout
+      - setup_environment:
+          cache_key: << parameters.cache_key >>
+      - run:
+          name: "Install snarkos"
+          command: |
+            cargo install --locked --path .
+      - run:
+          name: "Create non-interactive devnet script"
+          command: |
+            cat > devnet_ci.sh \<<EOF
+            #!/bin/bash
+            
+            # Set parameters directly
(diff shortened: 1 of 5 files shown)


ljedrz (commented):
Could we commit the new script as its own file? It could be stored under `.circleci` in order to avoid any confusion about its purpose.

ljedrz (commented):
I'm not an expert in scripting, but LGTM in general.

Is the reason why this job is currently expected to break that current `staging` has some bug? If so, we might want to hold off the merge until we rebase once it's fixed.

kpandl:
Yes currently staging has a bug, so the devnet script doesn't run. I think it's the snarkVM bug that you found from here: https://github.com/ProvableHQ/snarkVM/pull/2615

I agree it would be good to get a passing run here from CI. Either we can do it once the snarkVM ref is updated in snarkOS and this here is rebased, or if the CI also runs on other branch PRs, I could open a "do not merge" test PR there.

vicsn:
@kpandl that bug is no longer in staging, try running a local devnet on your machine to be sure.

vicsn:
You can update the rev in this PR, feel free to do so in future PRs too if useful.

ljedrz (approved):
LGTM :+1:

vicsn (approved):

kpandl:
Note: the PR is merged, there seems to be some GitHub bug that shows it as "Closed"

PR #3543, opened 2025-03-18
Deduplicate blocksync

kaimast:
## Motivation

There are two different code paths that perform block synchronization: one for validators, and one for clients/provers.
In the case of validators, there exist two different instances of `BlockSync`. 

This PR aims to address this by moving code that specific to validators or clients. It also improves documentation of the block synchronization logic.
Finally, the PR serves as the basis for upcoming changes to block synchronization that will harden security and improve performance of the protocol.

## Overview

The following is an outline of the structs affected by this PR. It also highlights which structs received significant changes. 

Outline

## Changes

There have been a multiple updates to this PR, and I would prefer if it gets squashes before a merge to not clutter the commit history too much.

Below is a list of the first set of commits that implemented the vast majority of the PR.
*  [509c266](https://github.com/ProvableHQ/snarkOS/pull/3543/commits/509c2666b39db5d9edf1dfc6ba913eaf56fb925e) changes the validator code to only use one instance of `BlockSync`.
*  [258ffcb](https://github.com/ProvableHQ/snarkOS/pull/3543/commits/258ffcb2e82033f0a1cffd7f85831dffcf1e4bb2) and [173e8e4](https://github.com/ProvableHQ/snarkOS/pull/3543/commits/173e8e4c311c417e92a513885082343dfcadd9f9) perform minor clean-ups and improve code documentation/comments.
* [4adaf15](https://github.com/ProvableHQ/snarkOS/pull/3543/commits/4adaf15234b2b4575cad1c5d343f76669d34bc0b) removes duplicated code from testing, because we need to pass `BlockSync` every time due to the very first commit.
* Finally, [0c75996](https://github.com/ProvableHQ/snarkOS/pull/3543/commits/0c75996490770d0ef6de5d263e97146714293cfb) significantly refactors BlockSync and moves all client specific code to the client.

## Test Plan

Existing tests should suffice, as this PR does not change any functionality.


kaimast:
@acoglio feel free to take a look at this draft already. I still need to run a testnet to ensure nothing broke, but should be done soon.

node/bft/src/bft.rs
```diff
@@ -934,6 +938,20 @@ mod tests {
         (committee, account, ledger, storage)
     }
 
+    // Helperfunction to set up BFT for testing
```
acoglio:
```suggestion
    // Helper function to set up BFT for testing
```

node/bft/src/helpers/dag.rs
```diff
@@ -22,9 +22,12 @@ use snarkvm::{
 use indexmap::IndexSet;
 use std::collections::{BTreeMap, HashMap};
 
+/// Maintains an acyclic directed graph (DAG) of batches, from which we build a totally-ordered blockchain.
+/// The DAG is updated in rounds, where each validator adds at most one new batch.
 #[derive(Debug)]
 pub struct DAG<N: Network> {
     /// The in-memory collection of certificates that comprise the DAG.
+    /// For each round, there is a mapping from node address to batch.
```
acoglio, line 30:
```suggestion
    /// For each round, there is a mapping from node address to batches.
```

node/bft/src/primary.rs
```diff
@@ -78,11 +78,13 @@ use tokio::{
 /// A helper type for an optional proposed batch.
 pub type ProposedBatch<N> = RwLock<Option<Proposal<N>>>;
 
+/// The primary logic of a node.
+/// AleoBFT adapts a primary-worker architecture as described in the Narwhal and Tusk paper (Section 4.2).
```
acoglio: I think 'adapts' works, but did you mean 'adopts'?

node/bft/src/primary.rs
```diff
@@ -78,11 +78,13 @@ use tokio::{
...
 #[derive(Clone)]
 pub struct Primary<N: Network> {
-    /// The sync module.
+    /// The sync modules enables fetching data from other validators.
```
```suggestion
    /// The sync module enables fetching data from other validators.
```

node/bft/src/sync/mod.rs
```diff
@@ -58,17 +61,20 @@ pub struct Sync<N: Network> {
     handles: Arc<Mutex<Vec<JoinHandle<()>>>>,
     /// The response lock.
     response_lock: Arc<TMutex<()>>,
-    /// The sync lock.
+    /// The sync lock. Ensures that only one tasks syncs the ledger at a time.
```
acoglio:
```suggestion
    /// The sync lock. Ensures that only one task syncs the ledger at a time.
```

vicsn:
Exciting, looking forward to this improvement!

The project is constrained by reviewer time, and every security review always takes longer than we want, so please revert variable name changes or refactors which have nothing to do with this the core goal of this PR. The comments are welcome, thank you for adding those.

kaimast:
I ran a devnet locally for a while and did not see any issues. I think it is ready to be merged.

node/bft/src/sync/mod.rs
```diff
@@ -97,13 +103,36 @@ impl<N: Network> Sync<N> {
...
+    ///
+    /// When this function returns sucessfully, the sync module will have spawned background tasks
+    /// that fetch blocks from other validators.
     pub async fn run(&self, sync_receiver: SyncReceiver<N>) -> Result<()> {
         info!("Starting the sync module...");
 
         // Start the block sync loop.
         let self_ = self.clone();
-        self.handles.lock().push(tokio::spawn(async move {
+        self.spawn(async move {
```
acoglio, line 166: I don't actually know the details of these handles/locks, but I was wondering why this change?

node/sync/src/block_sync.rs
```diff
@@ -221,96 +181,69 @@ impl<N: Network> BlockSync<N> {
...
-
-            // Acquire the lock to ensure try_advancing_with_block_responses is called only once at a time.
-            // If the lock is already acquired, return early.
-            let Some(_lock) = self.advance_with_sync_blocks_lock.try_lock() else {
-                trace!(
-                    "Skipping a call to try_block_sync() as a block advance is already in progress (at block {current_height})"
-                );
-                return;
-            };
+    /// Are there any pending responses to bock requests that need to be processed?
```
acoglio:
```suggestion
    /// Are there any pending responses to block requests that need to be processed?
```

acoglio (approved):

vicsn:
@kaimast can you add images to the PR description similar in spirit as was done in: https://github.com/ProvableHQ/snarkVM/pull/2628

node/bft/src/bft.rs
```diff
@@ -79,6 +80,7 @@ pub struct BFT<N: Network> {
 impl<N: Network> BFT<N> {
     /// Initializes a new instance of the BFT.
     pub fn new(
+        block_sync: Arc<BlockSync<N>>,
```
ljedrz: while I'm in favor of keeping the `BlockSync` in a single `Arc`, this should be followed by the removal of the individual `Arc`s in its members, and its `Clone` derivation; otherwise this will reduce the number of atomic ops related to its cloning, but it could suggest that `BlockSync` itself or its members don't derive `Clone`

node/bft/src/sync/mod.rs
```diff
@@ -97,13 +103,36 @@ impl<N: Network> Sync<N> {
         self.sync_storage_with_ledger_at_bootup().await
     }
 
+    /// Performs one iteration of the block synchronization.
+    #[inline]
+    pub async fn try_block_sync(&self) {
+        // Prepare the block requests, if any.
+        // In the process, we update the state of `is_block_synced` for the sync module.
+        let (block_requests, sync_peers) = self.block_sync.prepare_block_requests();
```
ljedrz, line 142: is this operation fast? otherwise it should be executed in the context of a blocking task, in order to not block the async worker thread

ljedrz (commented):
A very desirable change; I've left some suggestions.

node/bft/src/bft.rs
kaimast: I actually didn't know about the `derive(Clone)` in many places. Do we prefer `Arc` or making structs clonable? Or is both okay?

ljedrz: 1. if we need the members to be have shared ownership individually, they need their own `Arc` wrappers
2. it is more clear and efficient to have one such wrapper over an object that doesn't derive `Clone` and whose members don't have `Arc`s (if they don't need to individually)
3. having multiple "levels" of `Arc` isn't inherently inefficient, but it may be counterintuitive
4. if there's an object that derives `Clone` and has multiple `Arc`ed members, cloning it involves several atomic operations

Apologies if this is a bit convoluted; my preference is for fewer atomic operations on cloning (fewer `Arc`s involved in a single clone op), and clarity (i.e. not `Arc`ing something that is never handled individually).

kaimast: Thanks! I have the same preference, but saw code that has multiple arced members and derives clone, so I wasn't sure which is the desired code style.

node/bft/src/sync/mod.rs
kaimast: This was running in the regular tokio threads before this change. I am a little unclear at what point we move stuff to the blocking threads, but I can time it if needed.

ljedrz: ah, in this case it's unlikely it's an issue, especially since we've profiled BFT performance with `tokio console`; that being said, we might want to do a new run sometime soon, just to be sure we're not blocking anywhere

kaimast:
@vicsn below is an incomplete diagram. Not sure how helpful it is...
block-sync

node/bft/src/sync/mod.rs
kaimast: `fn spawn(fut)` takes a future and calls `self.handles.lock().push(tokio::spawn(fut)))`.
The idea is that handles contains all background tasks started by the Sync module.

Strangely, this one line did not use the helper function. I assume it was just an oversight and changed it.

node/bft/src/helpers/dag.rs
```diff
@@ -22,9 +22,12 @@ use snarkvm::{
 use indexmap::IndexSet;
 use std::collections::{BTreeMap, HashMap};
 
+/// Maintains an acyclic directed graph (DAG) of batches, from which we build a totally-ordered blockchain.
```
howardwu:
```suggestion
/// Maintains a directed acyclic graph (DAG) of batches, from which we build a totally-ordered blockchain.
```
Minor nit to match the terminology with the acronym

node/bft/src/primary.rs
```diff
@@ -110,6 +112,7 @@ impl<N: Network> Primary<N> {
 
     /// Initializes a new primary instance.
     pub fn new(
+        block_sync: Arc<BlockSync<N>>,
```
howardwu: nit: I would move `block_sync` to be after `ledger` (so that the related items are close to one another, and the order of arguments maintains a relative dependency tree of hierarchy/layers)

howardwu: (with an IDE, you should be able to reorder these arguments and all declarations/instantiations of this method with 1 refactor button-click)

node/bft/src/sync/mod.rs
```diff
@@ -58,17 +61,20 @@ pub struct Sync<N: Network> {
...
     latest_block_responses: Arc<TMutex<HashMap<u32, Block<N>>>>,
 }
 
 impl<N: Network> Sync<N> {
     /// Initializes a new sync instance.
-    pub fn new(gateway: Gateway<N>, storage: Storage<N>, ledger: Arc<dyn LedgerService<N>>) -> Self {
-        // Initialize the block sync module.
-        let block_sync = BlockSync::new(BlockSyncMode::Gateway, ledger.clone(), gateway.tcp().clone());
+    pub fn new(
+        block_sync: Arc<BlockSync<N>>,
```
howardwu: ditto nit: let's move `block_sync` to be after `ledger` to keep topically related items relatively next to one another and in a relative dependency tree hierarchy.

node/sync/src/block_sync.rs
```suggestion
    /// TODO (kaimast): Are there any pending responses to bock requests that need to be processed?
```

node/src/validator/router.rs
```diff
@@ -194,29 +194,21 @@ impl<N: Network, C: ConsensusStorage<N>> Inbound<N> for Validator<N, C> {
...
-                // Check the block locators are valid, and update the peer in the sync pool.
-                if let Err(error) = self.sync.update_peer_locators(peer_ip, block_locators) {
-                    warn!("Peer '{peer_ip}' sent invalid block locators: {error}");
-                    return false;
-                }
-            }
-        }
+    /// Processes the ping and sends back a `Pong` message.
+    fn ping(&self, peer_ip: SocketAddr, _message: Ping<N>) -> bool {
+        // In gateway mode, we do not need to process block locators
```
howardwu: Can you help me understand why this is the case? Do we broadcast block locators in validator Ping messages?

howardwu:
```suggestion
        // In gateway mode, we do not need to process block locators.
```

howardwu:
^You can add this to the main PR description for improved visibility.

Follow up questions for main PR description:
- What memory or time performance improvements to syncing (if any) are improved in this PR?
- What unit or integration tests were introduced to validate the changes made to syncing in this PR?

node/src/validator/router.rs
kaimast: Validators receive BlockLocators from other validators through `PrimaryPing` messages, which are processed in `Gateway`. This code is only invoked at a validator when receiving a ping from a client or prover, and validators ignore those locators. 

I will add a longer comment here.

howardwu: @niklaslong @ljedrz as you have better state on this, can you check that this is the case and that we are safe here?

ljedrz: looks reasonable to me - the validators call `update_peer_locators` when they receive the dedicated `PrimaryPing` messages

node/bft/src/sync/mod.rs
```diff
@@ -237,14 +262,40 @@ impl<N: Network> Sync<N> {
...
         Ok(())
     }
 }
 
+// Callbacks used when receiving messages from the Gateway
+impl<N: Network> Sync<N> {
+    /// We received a block response and can (possibly) advance synchronization.
+    async fn advance_with_sync_blocks(&self, peer_ip: SocketAddr, blocks: Vec<Block<N>>) -> Result<()> {
+        // Verify that the response is valid.
+        self.block_sync.insert_block_responses(peer_ip, blocks)?;
```
raychu86, line 278: Is there a reason we are avoiding the previous abstraction of `self_.block_sync.process_block_response(peer_ip, blocks)` and instead calling `insert_block_responses` directly?

raychu86:
@kpandl Could you run the sync tests on this PR to check if there are any improvements/regressions to syncing after the change?

node/src/validator/router.rs
niklaslong: Perhaps @kaimast could confirm but it seems to me that validators don't ever instantiate a `BlockSyncMode::Router`, so this is a branch that is never reached in the current code? And it also makes intuitive sense as we want validators to check their state against the locators of other validators, not those of other node types.

kaimast: Yes, that is also my understanding of the current code.

kaimast:
I updated the PR description to better reflect the current state of this PR and added a simplified diagram. Hope we can get it merged this week. @vicsn

node/bft/src/primary.rs
```diff
@@ -1538,7 +1541,7 @@ impl<N: Network> Primary<N> {
             // Ensure that the previous certificate was created at least `MIN_BATCH_DELAY_IN_MS` seconds ago.
```
raychu86: nit: This may have been missed 
```suggestion
            // Ensure that the previous certificate was created at least `MIN_BATCH_DELAY_IN_SECS` seconds ago.
```

raychu86 (dismissed):
Great cleanup and unification of disjoint usages! Code overall looks logically correct.

Would wait to merge until proper sync stress tests have been run and reported. @kpandl would appreciate any updates once you've had a change to run the tests

kpandl:
I've ran 2 clients syncing on mainnet.
Screenshot 2025-04-09 at 16 11 42

Yellow is staging, and green is this feature branch. So syncing generally works. For this feature branch, syncing was ~3% slower, so not much. Would need more runs to have significance, but since this PR mostly reorganizes, might be noise. LGTM.

Note that I haven't run the classic stress-test, since we are currently having issues with the automated test infra.

raychu86 (approved):

PR #3545, opened 2025-03-19
[Perf] Use the fastsync CDN address for syncing

ljedrz:


raychu86:
CI looks to be failing. 

@onetrickwolf Can we expose the latest.json object as well?

raychu86:
Looks like the `test_cdn_get` also fails due to an invalid URL. The previous URL was returning an "Access Denied" and somehow we are interpreting the bytes as a proper block height value. 

I believe this test can be removed since we no longer have a use for a URL of `"{TEST_BASE_URL}/mainnet/latest/height"`

onetrickwolf:
@raychu86 I believe `latest.json` should be available: https://blocks.provable.com/v0/mainnet/latest.json

I think the issue is this PR is doing provable **.org** instead of provable **.com** .

onetrickwolf:
@raychu86 and @ljedrz we have decided to change from `blocks.provable.com/v0` to `cdn.provable.com/v0/blocks`

For example these are the new URL patterns:

```
cdn.provable.com/v0/blocks/canary/0.50.blocks
cdn.provable.com/v0/blocks/testnet/0.50.blocks
cdn.provable.com/v0/blocks/mainnet/0.50.blocks
cdn.provable.com/v0/blocks/canary/latest.json
cdn.provable.com/v0/blocks/testnet/latest.json
cdn.provable.com/v0/blocks/mainnet/latest.json
```

If that causes any issues let me know happy to rework it 👍

cli/src/commands/start.rs
```diff
@@ -60,7 +60,7 @@ const DEVELOPMENT_MODE_RNG_SEED: u64 = 1234567890u64;
 const DEVELOPMENT_MODE_NUM_GENESIS_COMMITTEE_MEMBERS: u16 = 4;
 
 /// The CDN base url.
-pub(crate) const CDN_BASE_URL: &str = "https://blocks.aleo.org";
+pub(crate) const CDN_BASE_URL: &str = "https://blocks.provable.org";
```
onetrickwolf:
```suggestion
pub(crate) const CDN_BASE_URL: &str = "https://cdn.provable.com";
```

cli/src/commands/start.rs
```diff
@@ -284,9 +284,9 @@ impl Start {
                 },
                 // If no CDN URL is provided, determine the CDN URL based on the network ID.
                 None => match N::ID {
-                    MainnetV0::ID => Some(format!("{CDN_BASE_URL}/mainnet/v0")),
-                    TestnetV0::ID => Some(format!("{CDN_BASE_URL}/testnet/v0")),
-                    CanaryV0::ID => Some(format!("{CDN_BASE_URL}/canary/v0")),
+                    MainnetV0::ID => Some(format!("{CDN_BASE_URL}/v0/mainnet")),
+                    TestnetV0::ID => Some(format!("{CDN_BASE_URL}/v0/testnet")),
+                    CanaryV0::ID => Some(format!("{CDN_BASE_URL}/v0/canary")),
```
onetrickwolf:
```suggestion
                    MainnetV0::ID => Some(format!("{CDN_BASE_URL}/v0/blocks/mainnet")),
                    TestnetV0::ID => Some(format!("{CDN_BASE_URL}/v0/blocks/testnet")),
                    CanaryV0::ID => Some(format!("{CDN_BASE_URL}/v0/blocks/canary")),
```

node/cdn/src/blocks.rs
```diff
@@ -444,7 +444,7 @@ mod tests {
 
     type CurrentNetwork = MainnetV0;
 
-    const TEST_BASE_URL: &str = "https://blocks.aleo.org/mainnet/v0";
+    const TEST_BASE_URL: &str = "https://blocks.provable.org/v0/mainnet";
```
onetrickwolf:
```suggestion
    const TEST_BASE_URL: &str = "https://cdn.provable.com/v0/blocks/mainnet";
```

This will return a similar error now as the old CDN did so this might work. But also see @raychu86's suggested comment of probably just removing this test.

onetrickwolf (changes requested):

raychu86 (approved):

PR #3549, opened 2025-03-25
[doc] Document some code-level properties.

acoglio:
These are some preconditions, postconditions, and invariants in the syncing code.

These are currently *not* formally proved -- reviewers should double-check them. They are based on an examination of the code, with informal reasoning.


node/sync/src/block_sync.rs
```diff
@@ -86,6 +86,11 @@ impl BlockSyncMode {
 ///   the `request_timestamps` map remains unchanged.
 /// - When a response is removed/completed, the `requests` map and `request_timestamps` map also remove the entry for the request height.
 /// - When a request is timed out, the `requests`, `request_timestamps`, and `responses` map remove the entry for the request height.
+///
+/// Invariant: `requests` and `request_timestamps` always have the same keys.
```
kaimast, line 90: Why don't we have just one map for both then, then? 
It would be great if we could atomically update them instead of each having its own lock as well.

kaimast: Will put this on my TODO list.

kaimast (dismissed):
LGTM!

node/sync/src/block_sync.rs
acoglio: Yes, I would prefer a single map too, to build the invariant in. Documenting it was a first step — I'm hoping that reviewers can double-check the informal reasoning that it is indeed an invariant.

kaimast (approved):

kaimast:
I merged this commit into #3626.

PR #3557, opened 2025-03-26
Document --rotate-external-peers

vicsn:


kpandl (dismissed):
LGTM. I noticed the term bootstrap client isn't defined in the doc.
Might be a widely known term, or we could add something like: Outer clients can be bootstrap clients that serve as accessible entry points for new nodes joining the network with publicly known or static IPs.

raychu86 (dismissed):

raychu86:
Docs look good, but I agree with @kpandl's comment about potentially defining bootstrap peers. It's not exactly necessary, but I can see that the new line of documentation may confuse general users.

vicsn:
Thx for surfacing, please clarify the bootstrap distiction in the README @kpandl

kpandl (approved):

kpandl:
Done!

raychu86 (approved):

PR #3570, opened 2025-04-02
chore: add audit.toml to ignore known unmaintained crates

joske:
## Motivation
`cargo audit` complains that `ansi_term` and `paste` crates are unmaintained. 

For `paste`: As the rust community considers this 'done', we can safely ignore this warning.
For `ansi_term`: Crate was unmaintained for 4 years. Removed this dependency.

## Test Plan

no actual code changes


joske:
@niklaslong could you review? I can't set reviewers it seems.

niklaslong:
I think this is a reasonable approach to take, though we should be _very_ strict with the advisories we choose to include here. It might also be a good practice to periodically review this list. Could you also add comments to the file explaining why each advisory is ignored and perhaps link to relevant discussions?

joske:
Absolutely, we should think hard about adding anything to this list. There exists a fork of `paste` called `pastey` that is supposedly drop-in, but I'd be more weary to change to that than to just keep `paste`.

niklaslong:
> I'd be more weary to change to that than to just keep paste.

Agreed! 👍

kaimast:
Looks good to me! 

Should we create an issue for each unmaintained dependency, so we don't forget about it?

For `ansi_term` it looks like we would need to upgrade `tracing-test`, which might be good anyway because it should get rid of the dependency on `tracing-subscriber` 0.2 (snarkOS also depends on 0.3)

kaimast (dismissed):

vicsn:
> For ansi_term it looks like we would need to upgrade tracing-test, which might be good anyway because it should get rid of the dependency on tracing-subscriber 0.2 (snarkOS also depends on 0.3)

@joske We should indeed just try to get rid of ansi_term, its use seems to be limited. Then we can remove that from the ignore list.

Can you also add a CI job which runs cargo audit? Also for snarkVM?

> Should we create an issue for each unmaintained dependency, so we don't forget about it?

And yes we can make an issue to track the situation of the paste dependency...

joske:
@vicsn  (github) workflow added. It will fail on any warning.

joske:
Apparently `tokio` should be upgraded to 1.42.2. Should probably be another PR?

joske:
`ansi_term` dep removed.

Cargo.toml
```diff
@@ -184,13 +184,14 @@ incremental = true
...
 
 [profile.test]
-opt-level = 2
-lto = "thin"
+opt-level = 1
+lto = "off"
 incremental = true
 debug = true
-debug-assertions = true
+debug-assertions = false
```
vicsn: You can introduce profile changes in a separate PR with explanation if you think they're a good idea. In the past changing some of this slowed down CI, and I'm not sure we'd want to turn off debug assertions...

joske: oh this is a mistake! Good catch!

joske: removed

joske:
BTW, the audit workflow will fail until the cargo update PRs are merged (#3589 and the relevant one in snarkVM).

.integration/Cargo.toml
```diff
@@ -0,0 +1,37 @@
+[package]
+name = "snarkos-integration"
+version = "3.5.0"
+authors = [ "The Aleo Team <hello@aleo.org>" ]
+description = "A integration testing suite for a decentralized operating system"
```
niklaslong:
```suggestion
description = "An integration testing suite for a decentralized operating system"
```

.integration/Cargo.toml
```diff
@@ -0,0 +1,37 @@
+[package]
+name = "snarkos-integration"
```
niklaslong: Why do we need a new crate?

.github/workflows/audit.yml
```diff
@@ -0,0 +1,20 @@
+name: Cargo Audit
```
niklaslong: We might want to unify this job with the existing circle workflow instead?

.integration/Cargo.toml
joske: oh I did not see this is a new folder. Not sure where that came from.

.github/workflows/audit.yml
joske: Could do. Is that preferred?

kaimast: We use CircleCI for most/all CI, so it makes sense to move it there.

kaimast (dismissed):
Let's move the audit action to CircleCI, and then we can merge it.

kaimast:
I changed this PR to use CircleCI. It should be ready to merge, but I will not do it myself as I made changes.

vicsn (approved):

Issue #3546, opened 2025-03-19
[Bug] A test in the cli package fails when testing the entire workspace


kaimast:
## 🐛 Bug Report

Running individual tests works fine, but when testing the entire workspace, `test_parse_development_and_genesis` in `cli` fails with the following error:

```
thread 'commands::start::tests::test_parse_development_and_genesis' panicked at cli/src/commands/start.rs:921:82:
called `Result::unwrap()` on an `Err` value: Invalid block metadata: {"network":0,"round":0,"height":0,"cumulative_weight":"0","cumulative_proof_target":"0","coinbase_target":536870911,"proof_target":134217728,"last_coinbase_target":536870911,"last_coinbase_timestamp":1725462000,"timestamp":1725462000}
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

This is not a huge problem as CircleCI runs the tests individually, but might confuse new contributors. It looks like there is an inconsistency in how feature flags are set (or not set), which could cause other problems in the future.

## Steps to Reproduce

Run `cargo test --package snarkos-cli commands::start::tests::test_parse_development_and_genesis`, which should succeed, and then `cargo test --workspace commands::start::tests::test_parse_development_and_genesis`.

Both commands will run the same unit test, but the `--workspace` flag causes the test to break. Your output should look like the following:

```
~> cargo test --package snarkos-cli commands::start::tests::test_parse_development_and_genesis
    Finished `test` profile [optimized + debuginfo] target(s) in 0.41s
     Running unittests src/lib.rs (/Users/kaimast/dev/snarkOS/target/debug/deps/snarkos_cli-fe91258e0e9b0cb4)

running 1 test
test commands::start::tests::test_parse_development_and_genesis ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 30 filtered out; finished in 0.22s

~> cargo test --workspace commands::start::tests::test_parse_development_and_genesis
   Compiling snarkos v3.3.2 (/Users/kaimast/dev/snarkOS)
    Finished `test` profile [optimized + debuginfo] target(s) in 26.61s
     Running unittests snarkos/main.rs (/Users/kaimast/dev/snarkOS/target/debug/deps/snarkos-319e0b736a923806)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

[...]

PR #3572, opened 2025-04-02
Introduce test_network feature

kpandl:
## Motivation

For testing purposes, it would be practical to test the snarkVM consensus version (and proof targets) heights through a `--features test_network` compilation flag in snarkOS. This PR introduces such a flag.

It also updates the snarkVM ref to the version where https://github.com/ProvableHQ/snarkVM/pull/2693 is merged.

## Test Plan

Tested with an added CLI output of consensus height thresholds for testing purposes, and verified it works.


kaimast:
This looks fine to me.

There is one thing to consider, though. Should the `test` feature enable `test` for all crates in the workspace? `ledger-service`, `storage-service`, `router`, and `sync` all have such a feature. 
If not, we should leave a comment somewhere that explains what features/behavior it enables and why. Alternatively, we could name the feature something like `test-consenus-heights` instead. 

###  "Big Picture" Comments
* I dug through these feature flags a while ago when trying to debug #3546. It's a little confusing how the `test` and `test_targets` features are used in some, but not all, unit tests (usually by having the feature set in a dev-dependency).

* We should document all feature flags for the top-level crate/workspace. The Rust documentation points to [this README](https://github.com/rust-lang/regex/blob/1.4.2/src/lib.rs#L488-L583) as a good example of how to do it.

vicsn:
> We should document all feature flags for the top-level crate/workspace. The Rust documentation points to this README as a good example of how to do it.

Really good idea. Not sure if its overkill to use a tool to visualize the feature trees. This documentation can also document the discrepancy you noted regarding test and test_targets. CC Perhaps @joske wants to take a stab at this because he also has context on the addition of test_targets.

niklaslong:
> Not sure if its overkill to use a tool to visualize the feature trees.

You should be able to get quite far with `cargo tree`: https://doc.rust-lang.org/cargo/commands/cargo-tree.html, specifically the `-e` option used for searching through features, perhaps combined with `-i` for a reverse search.

vicsn:
> There is one thing to consider, though. Should the test feature enable test for all crates in the workspace? ledger-service, storage-service, router, and sync all have such a feature.

@kpandl want to give this a try, and see if we can still run a local devnet and land a transaction? No rush on this.

kpandl:
Thanks, I've added the `test` feature to `router`, `sync`, and through feature propagation of `bft` to `ledger-service` and `storage-service`. I've also updated the `router` crate to propagate the test feature to the submodule.

Ran a devnet and verified a transaction still lands when using the `test` feature (didn't test anything else).

kaimast (dismissed):
Looks good!

Cargo.toml
```diff
@@ -82,6 +82,12 @@ locktick = [
   "snarkos-node-sync/locktick",
   "snarkos-node-tcp/locktick"
 ]
+test = [
+  "snarkos-node-bft/test",
```
vicsn: Since we're importing snarkVM in this file, shouldn't we be specifying `"snarkvm/test"` here as well? If only for consistency. (Don't ask me why snarkVM is imported in two placed by snarkOS)

kpandl:
I've added `"snarkvm/test"` to the relevant crates that directly depend on snarkVM (specifically `node/bft/Cargo.toml` and `node/router/Cargo.toml`). I couldn't add it to the root Cargo.toml where I got this:

```
error: feature `test` includes `snarkvm/test`, but `snarkvm` is not a dependency
  --> Cargo.toml:85:8
   |
85 |   test = [
   |  ________^
86 | |   "snarkos-node-bft/test",
...  |
90 | |   "snarkvm/test"
91 | | ]
   | |_^
   |
error: failed to parse manifest at `/Users/kp/dev/snarkOS/Cargo.toml`

```

I've kept changes focused on the files already in the PR scope, though there are many more crates where snarkVM is added as a dependency.

Cargo.toml
kpandl: Addressed here: https://github.com/ProvableHQ/snarkOS/pull/3572/commits/e66c42ee4fcf34e2545503c3295fd143bda56086
I've added `"snarkvm/test"` to the relevant crates that directly depend on snarkVM (specifically `node/bft/Cargo.toml` and `node/router/Cargo.toml`). I couldn't add it to the root Cargo.toml where I got this:

```
error: feature `test` includes `snarkvm/test`, but `snarkvm` is not a dependency
  --> Cargo.toml:85:8
   |
85 |   test = [
   |  ________^
86 | |   "snarkos-node-bft/test",
...  |
90 | |   "snarkvm/test"
91 | | ]
   | |_^
   |
error: failed to parse manifest at `/Users/kp/dev/snarkOS/Cargo.toml`

```

I've kept changes focused on the files already in the PR scope, though there are many more crates where snarkVM is added as a dependency.

kaimast:
Seems like adding the snarkVM test feature broke something.

Does the test feature get enabled in unit tests? If it isn't, these changes shouldn't affect the tests, but it seems like they do.

vicsn:
> Seems like adding the snarkVM test feature broke something.
> Does the test feature get enabled in unit tests? If it isn't, these changes shouldn't affect the tests, but it seems like they do.

Indeed, it seems like for some subcrates, `features = [ "test" ]` is enabled in a `dev-dependencies` subsection. Note for example that router had an empty `test` feature which is now suddenly enabled with a lot more junk in there: https://github.com/ProvableHQ/snarkOS/pull/3572/files#diff-f5e82e097aa5005ad9c92fe29abb9de43cddfbf974a5dcafcdad98e4cbb7b715R20

I see a couple of directions to achieve the original goal: ensure we can run a snarkOS network where novel features are enabled on a low block height.
1. We introduce a new `test_consensus_versions` feature in snarkVM and snarkOS, which will use the test consensus version heights.
2. We introduce `test_targets` and `test_consensus_versions` in snarkVM, but use a single feature in snarkOS called `test_network` which will enable both in snarkVM.
3. We consolidate `test_targets` and `test_consensus_versions` into a single feature called `test`, but this requires untangling the morass of `git grep -C2 test | grep Cargo.toml`.

I vote for starting with option 2., and we can create an issue to clean up 3.

kpandl:
Thanks, addressed option 2 now. Tested it along with https://github.com/ProvableHQ/snarkVM/pull/2693 on a local devnet.
Note that CI will fail because it needs this snarkVM: https://github.com/ProvableHQ/snarkVM/pull/2693

I'd suggest I temporarily change the snarkVM ref here so that we can let CI run and see if it passes.

cli/src/commands/start.rs
```diff
@@ -523,8 +523,8 @@ impl Start {
 
         // Check if we are running with the lower coinbase and proof targets. This should only be
         // allowed in --dev mode.
-        if cfg!(feature = "test_targets") && self.dev.is_none() {
-            bail!("The 'test_targets' feature is enabled, but the '--dev' flag is not set");
+        if cfg!(feature = "test_network") && self.dev.is_none() {
```
vicsn, line 526: Can we also add a guard that we're not running mainnet? Because the feature only influences canary/testnet traits (we can question that decision later if it creates trouble).

Cargo.toml
```diff
@@ -82,7 +82,10 @@ locktick = [
   "snarkos-node-sync/locktick",
   "snarkos-node-tcp/locktick"
 ]
-test_targets = [ "snarkos-cli/test_targets" ]
+test_network = [
+  "snarkos-cli/test_network",
+  "snarkos-node/test_network"
```
vicsn: Why do we only have to activate these two subdirectories (cli and node)?

Given that we didn't activate `snarkos-node/test_targets` before, was that effectively some dead configuration code?

vicsn (approved):

vicsn (changes requested):
Lets wait with merging until snarkVM test_migration_heights is fixed

vicsn (dismissed):

Cargo.toml
vicsn: @kpandl this question still stands - is `snarkos-node/test_network` actually used? Should it be called from `cli/Cargo.toml` or not?

kpandl: I don't think `snarkos-node/test_network` is currently used - I've printed the consensus version heights and the coinbase target in `fn check_next_block` of snarkVM. Based on that, in the current state of the codebase (where we neither trigger `snarkos-node/test_network` in the root cargo, nor in the cli cargo), the `test_network` feature works.

vicsn: Should we remove it (and its children) from `Cargo.toml` then?

kpandl: Sure, addressed here: https://github.com/ProvableHQ/snarkOS/pull/3572/commits/491655f9a0657471551f0c971edef554c468093b
Verified it works by printing values in `fn check_next_block` of snarkVM.

vicsn (approved):

PR #3574, opened 2025-04-03
[Doc] Document "current committee" more precisely.

acoglio:


kaimast (approved):
LGTM!

I would prefer, though, if we just renamed the functions to `latest_committee` so that they have the same name as in snarkVM. 
"current" also implies (to me) that the committee is currently active, which is not always the case if I understand correctly.

acoglio:
I agree, I pushed a commit that does that.

vicsn (dismissed):

Issue #3559, opened 2025-03-27
Check --storage and --dev compatibility


niklaslong:
The `--storage` cli option (`--path` for `clean`) allows a user to specify a custom path for storage. While in dev mode, that path is hardcoded to the project directory. We could either allow custom paths to be passed while in dev mode or disable the use of both options at the same time.

joske:
@niklaslong Unless I'm overlooking something, it seems like `--path` or `--storage` already takes precedence? Consider this excerpt from `snarkos clean`:

```rust
impl Clean {
    /// Cleans the snarkOS node storage.
    pub fn parse(self) -> Result<String> {
        // Initialize the storage mode.
        let storage_mode = match self.path {
            Some(path) => StorageMode::Custom(path),
            None => match self.dev {
                Some(id) => StorageMode::Development(id),
                None => StorageMode::Production,
            },
        };
...
}
```

I've confirmed this with a unit test. Other commands seem to do similar things, see this bit from `snarkos start`:

```rust
        let storage_mode = match &self.storage {
            Some(path) => StorageMode::Custom(path.clone()),
            None => match self.dev {
                Some(id) => StorageMode::Development(id),
                None => StorageMode::Production,
            },
        };
```

It sets the `StorageMode` to `Custom` if `--storage` is passed, regardless of `--dev`.

niklaslong:
Yes, thanks for looking into this! I think this is perhaps the correct behaviour: custom paths when in dev may come in handy. Thoughts?

joske:
It seems to me it's fine like it is now.

niklaslong:
Cool, let's add some documentation and then we can close this.

joske:
Ok, where to add this documentation?

niklaslong:
I'd say in the cli options? Open to other suggestions if you think there's a better place.

joske:
@niklaslong Needs more than what I have in above PR?

joske:
@vicsn This can be closed now.

PR #3590, opened 2025-04-09
chore: clarify the storage/path option

joske:
## Motivation
#3559 Clarify --storage/--path

## Test Plan
no code changes

### DIFF
--- a/cli/src/commands/clean.rs
+++ b/cli/src/commands/clean.rs
@@ -30,7 +30,7 @@ pub struct Clean {
     /// Enables development mode, specify the unique ID of the local node to clean.
     #[clap(long)]
     pub dev: Option<u16>,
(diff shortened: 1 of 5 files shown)


cli/src/commands/clean.rs
```diff
@@ -30,7 +30,7 @@ pub struct Clean {
     /// Enables development mode, specify the unique ID of the local node to clean.
     #[clap(long)]
     pub dev: Option<u16>,
-    /// Specify the path to a directory containing the ledger
+    /// Specify the path to a directory containing the ledger. Overrides the default path.
```
niklaslong: I think it would be good to specify this also applies when the dev flag is set?

niklaslong (approved):

PR #3591, opened 2025-04-10
Improve documentation for consensus crate and remove (potentially dangerous) unwrap()

kaimast:
I found the role of the `snarkos-node-consensus` crate and its relation to the `snarkos-node-bft` crate hard to understand. This pull request adds more documentation to those two crates.

The PR also removes an `unwrap()` that used to be invoked in production code when calling `Consensus::primary_sender()`.
To do this, the functionality that was formerly in `Consensus::run` is now in its constructors, so that other code can only interact with the crate's public API after it is fully initialized, whereas before one could potentially not invoke `run()` and cause a panic.

Finally, the PR makes some functions, which were not used anywhere alse, as non-public or removes them entirely.


kaimast:
@acoglio This is (mostly) an improvement on documentation, so you might be a good person to review this.

node/bft/src/helpers/ready.rs
```diff
@@ -25,6 +25,7 @@ use snarkvm::{
 use indexmap::{IndexMap, IndexSet};
 use std::collections::{HashMap, VecDeque, hash_map::Entry::Vacant};
 
+/// Maintains a queue of verified ("ready") transactions.
```
acoglio:
```suggestion
/// Maintains a queue of verified ("ready") transmissions.
```

node/bft/src/primary.rs
```diff
@@ -1133,6 +1133,13 @@ impl<N: Network> Primary<N> {
 
 impl<N: Network> Primary<N> {
     /// Starts the primary handlers.
+    ///
+    /// For each receiver in the `primary_receiver` struct, there will be a dedicated tasks
```
```suggestion
    /// For each receiver in the `primary_receiver` struct, there will be a dedicated task
```

node/bft/src/primary.rs
```diff
@@ -1133,6 +1133,13 @@ impl<N: Network> Primary<N> {
...
+    /// that awaits new data and handles it accordingly.
+    /// Additionally, this spawns a tasks that periodically issues PrimaryPings and one that periodically
```
```suggestion
    /// Additionally, this spawns a task that periodically issues PrimaryPings and one that periodically
```

node/consensus/README.md
```diff
@@ -6,5 +6,5 @@
 
 The `snarkos-node-consensus` crate provides the consensus layer for the snarkOS node.
 
-It builds on top of the `snarkos-node-bft` crate, which provides an abstract implementationv of the AleoBFT.
-More concretely, this crate provides the communication logic that allows multiple AleoBFT nodes to interact with each other.
+The crate builds on top of the `snarkos-node-bft`, which implements AleoBFT.
+It manages a ratelimiter/mempool for incoming transmissions, and manages construction of blcoks from batches that have been confirmed by the BFT layer.
```
acoglio:
```suggestion
It manages a ratelimiter/mempool for incoming transmissions, and manages construction of blocks from batches that have been confirmed by the BFT layer.
```

acoglio (dismissed):

raychu86:
The documentation additions are great and the unification of `new` and `run` to get around the `.expect` call makes sense to me, but this brings up the question of consistency around the other modules.

A lot of other structs use a similar `new` and `run` interface (BFT, Primary, Sync, gateway, etc); I believe some even have an additional `initialize` function. There are likely ordering considerations that I have not thought about, but should these be unified for consistency?

acoglio:
From a (formal and informal) reasoning perspective, generally the more (ideally, all) "initialization" (in a broad sense) can be done in the constructor, the better, because we get the strongest invariants just after construction; otherwise, structs between construction and running/initialization may have weaker invariants.

This is not always possible (when there are circular dependencies of sort), and it should be balanced with other considerations, but it does have the advantage mentioned above.

kaimast:
> A lot of other structs use a similar new and run interface (BFT, Primary, Sync, gateway, etc); I believe some even have an additional initialize function. There are likely ordering considerations that I have not thought about, but should these be unified for consistency?

In some cases there are circular dependencies that make it hard to unify the two, but I can look into doing more of these in future PRs.

acoglio (approved):

vicsn (approved):

PR #3596, opened 2025-04-16
[Feature] Introduce rolling Message versions 

raychu86:
## Motivation

This PR updates the Message::Version used in the router to be based on a rolling migration. This allows for smoother behavior whenever there is a breaking migration that requires nodes to upgrade versions. The intention is to allow the following:

1. At block height `X`, the protocol upgrades to message version from `Y-1` to `Y`.
2. Client A upgrades and starts using message version `Y`.
3. Client B has not upgraded and still uses message version `Y-1`.
4. Until block `X`, they stay connected and can communicate.
5. After block `X`, Client A will reject messages from Client B.

This behavior will allow smoother upgrades by allowing client nodes to stay connected and process blocks up until the actual migration height rather than immediately after peers upgrade. 

Note that this does not touch Validator's `Event::Version` because we want those to stay fixed and in lock-step, at least until this change is fully tested.

## Test Plan

Basic validator + clients devnet has been run locally, but more extensive testing needs to be done to test general cases. A couple to consider are:

- Local network
  - [x] Client nodes + Prover nodes with this PR @kpandl 
- Canary network
  - [x] this PR snarkOS client @kpandl

### DIFF
--- a/node/router/messages/src/lib.rs
+++ b/node/router/messages/src/lib.rs
@@ -120,6 +120,36 @@ impl<N: Network> Message<N> {
         Self::VERSIONS.last().map(|(_, version)| *version).unwrap_or(0)
     }
 
+    /// Returns the lowest acceptable message version for the given block height.
+    pub fn lowest_accepted_message_version(current_block_height: u32) -> u32 {
+        // Fetch the latest message version.
+        let latest_message_version = Self::latest_message_version();
+
+        // Fetch the versions.
+        let versions = Self::VERSIONS;
+
+        // Determine the minimum accepted message version.
+        // Example scenario:
+        // - At block height `X`, the protocol upgrades to message version from `Y-1` to `Y`.
+        // - Client A upgrades and starts using message version `Y`.
+        // - Client B has not upgraded and still uses message version `Y-1`.
+        // - Until block `X`, they stay connected and can communicate.
+        // - After block `X`, Client A will reject messages from Client B.
+        N::CONSENSUS_VERSION(current_block_height).map_or(latest_message_version, |seek_version| {
+            // Search the consensus value for the specified version.
+            match versions.binary_search_by(|(version, _)| version.cmp(&seek_version)) {
+                // If a value was found for this consensus version, return it.
+                Ok(index) => versions[index].1,
+                // If the specified version was not found exactly, determine whether to return an appropriate value anyway.
+                Err(index) => {
+                    // Return the appropriate value belonging to the consensus version *lower* than the sought version.
+                    // If the constant is not yet in effect at this consensus version, use the earliest version.
+                    versions[index.saturating_sub(1)].1
+                }
+            }
+        })
+    }
+
     /// Returns the message name.
     #[inline]
     pub fn name(&self) -> Cow<'static, str> {
@@ -236,5 +266,66 @@ impl<N: Network> FromBytes for Message<N> {
         }
 
         Ok(message)
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use snarkvm::prelude::{CanaryV0, MainnetV0, TestnetV0};
+
+    /// Ensure that the message versions used at genesis is correct.
+    fn consensus_constants_at_genesis<N: Network>() {
+        let height = 0;
+        let consensus_version = Message::<N>::lowest_accepted_message_version(height);
+        assert_eq!(consensus_version as usize, Message::<N>::VERSIONS.first().unwrap().1 as usize);
+    }
+
+    /// Ensure that the consensus *versions* are unique and incrementing.
+    fn consensus_versions<N: Network>() {
+        let mut previous_version = Message::<N>::VERSIONS.first().unwrap().0;
+        for (version, _) in Message::<N>::VERSIONS.iter().skip(1) {
+            assert!(*version as usize > previous_version as usize);
+            previous_version = *version;
+        }
+    }
+
+    /// Ensure that *message versions* are unique and incrementing by 1.
+    fn consensus_constants_increasing_heights<N: Network>() {
+        let mut previous_message_version = Message::<N>::VERSIONS.first().unwrap().1;
+        for (_, message_version) in Message::<N>::VERSIONS.iter().skip(1) {
+            assert_eq!(*message_version, previous_message_version + 1);
+            previous_message_version = *message_version;
+        }
+    }
+
+    /// Ensure that `Message::VERSIONS` increases and is correctly defined.
+    /// See the constant declaration for an explanation why.
+    fn message_version_increasing<N: Network>() {
+        let mut previous_value = Message::<N>::VERSIONS.first().unwrap().1;
+        for (_, value) in Message::<N>::VERSIONS.iter().skip(1) {
+            assert!(*value >= previous_value);
+            previous_value = *value;
+        }
+    }
+
+    #[test]
+    #[allow(clippy::assertions_on_constants)]
+    fn test_consensus_constants() {
+        consensus_constants_at_genesis::<MainnetV0>();
+        consensus_constants_at_genesis::<TestnetV0>();
+        consensus_constants_at_genesis::<CanaryV0>();
+
+        consensus_versions::<MainnetV0>();
+        consensus_versions::<TestnetV0>();
+        consensus_versions::<CanaryV0>();
+
+        consensus_constants_increasing_heights::<MainnetV0>();
+        consensus_constants_increasing_heights::<TestnetV0>();
+        consensus_constants_increasing_heights::<CanaryV0>();
+
+        message_version_increasing::<MainnetV0>();
+        message_version_increasing::<TestnetV0>();
+        message_version_increasing::<CanaryV0>();
     }
 }
--- a/node/router/messages/src/lib.rs
+++ b/node/router/messages/src/lib.rs
@@ -112,10 +112,13 @@ impl<N: Network> From<DisconnectReason> for Message<N> {
 }
 
 impl<N: Network> Message<N> {
-    /// The version of the network protocol; it can be incremented in order to force users to update.
-    pub const VERSION: u32 = 17;
     /// The version of the network protocol; this can is incremented for breaking changes between migration versions.
-    pub const VERSIONS: [(ConsensusVersion, u32); 1] = [(ConsensusVersion::V5, 17)];
+    pub const VERSIONS: [(ConsensusVersion, u32); 2] = [(ConsensusVersion::V4, 16), (ConsensusVersion::V5, 17)];
+
+    /// Returns the latest message version.
+    pub fn latest_message_version() -> u32 {
+        Self::VERSIONS.last().map(|(_, version)| *version).unwrap_or(0)
+    }
 
     /// Returns the message name.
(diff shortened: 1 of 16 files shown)


acoglio:
This has been discussed for other kinds of upgrades, but a caveat is that a validator may take time to generate blocks from anchors (until it has enough votes), and so the current block height is somewhat "ephemeral" with respect to the current round number (to which messages and transactions are tied).

raychu86:
> This has been discussed for other kinds of upgrades, but a caveat is that a validator may take time to generate blocks from anchors (until it has enough votes), and so the current block height is somewhat "ephemeral" with respect to the current round number (to which messages and transactions are tied).

Note that this code does not change the behavior of validators connecting to each other - only to the surrounding clients. So the "ephemerality" of current block height should have no impact on block production

node/router/src/lib.rs
```diff
@@ -251,6 +256,53 @@ impl<N: Network> Router<N> {
...
+                N::CONSENSUS_VERSION(latest_block_height).map_or(latest_message_version, |seek_version| {
+                    // Search the consensus value for the specified version.
+                    match versions.binary_search_by(|(version, _)| version.cmp(&seek_version)) {
+                        // If a value was found for this consensus version, return it.
+                        Ok(index) => versions[index].1,
+                        Err(index) => {
+                            // If the specified version was not found exactly, determine whether to return an appropriate value anyway.
+                            match index {
+                                // This constant is not yet in effect at this consensus version.
+                                0 => Message::<N>::latest_message_version(),
```
vicsn: If I understand correctly, at block heights `0..N::CONSENSUS_HEIGHT(ConsensusVersion::V4` this will return 17 (due to `latest_message_version`, then at `ConsensusVersion::V4` jump to 16, then at `ConsensusVersion::V5` jump back to 17.

I think one way to fix this would be to use `ConsensusVersion::V1, 16` in the `VERSIONS` constant, so that it'll start at 16.

We should test with the test feature [once that PR is ready](https://github.com/ProvableHQ/snarkOS/pull/3572), so that we can observe the behaviour at low block heights. CC @kpandl

node/router/messages/src/lib.rs
```diff
@@ -111,8 +112,13 @@ impl<N: Network> From<DisconnectReason> for Message<N> {
 }
 
 impl<N: Network> Message<N> {
-    /// The version of the network protocol; it can be incremented in order to force users to update.
-    pub const VERSION: u32 = 17;
+    /// The version of the network protocol; this can is incremented for breaking changes between migration versions.
+    pub const VERSIONS: [(ConsensusVersion, u32); 2] = [(ConsensusVersion::V4, 16), (ConsensusVersion::V5, 17)];
```
vicsn, line 116: Cool! Can you add relevant tests on these constants similar to `test_consensus_constants` in snarkVM?

vicsn (changes requested):

node/router/src/lib.rs
```diff
@@ -251,6 +256,53 @@ impl<N: Network> Router<N> {
...
+        // - Until block `X`, they stay connected and can communicate.
+        // - After block `X`, Client A will reject messages from Client B.
+        let lowest_accepted_message_version = match self.node_type {
+            // Provers should always use the latest version.
+            NodeType::Prover => Message::<N>::latest_message_version(),
+            // Validators and clients accept messages from lower version based on the migration height.
+            NodeType::Validator | NodeType::Client => {
+                let latest_block_height = self.ledger.latest_block_height();
+                let versions = Message::<N>::VERSIONS;
+                // TODO (raychu86): Refactor `consensus_config_value` to support generic constant types.
```
vicsn: EDIT: I just tried using `$constant:path` instead of `$constant:ident`, which seems to work.

~It would be great if Ray or Kai can take one more 30 minute stab at this before we merge the PR, if its too cumbersome we can leave it.~

node/router/src/lib.rs
raychu86: Hmm the logic here is not exactly correct. 


The intention is this:
You will be using `17` for all heights. With this change, you will ALWAYS be using the latest version number at 17. However, you are able to accept connections from nodes using version 16 up until `N::CONSENSUS_HEIGHT(ConsensusVersion::V5)`

node/router/src/lib.rs
raychu86: Note logic slightly changed in [3eca0df](https://github.com/ProvableHQ/snarkOS/pull/3596/commits/3eca0dfa4df035e58f66ebc482c3bcc94dd3119a) which deviates from the macro.

node/router/messages/src/lib.rs
raychu86: Added in [3eca0df](https://github.com/ProvableHQ/snarkOS/pull/3596/commits/3eca0dfa4df035e58f66ebc482c3bcc94dd3119a)

node/router/src/lib.rs
raychu86: The new logic in [3eca0df](https://github.com/ProvableHQ/snarkOS/pull/3596/commits/3eca0dfa4df035e58f66ebc482c3bcc94dd3119a) should address this

node/router/messages/src/lib.rs
```diff
@@ -111,8 +112,43 @@ impl<N: Network> From<DisconnectReason> for Message<N> {
...
+        // - Client A upgrades and starts using message version `Y`.
+        // - Client B has not upgraded and still uses message version `Y-1`.
+        // - Until block `X`, they stay connected and can communicate.
+        // - After block `X`, Client A will reject messages from Client B.
+        N::CONSENSUS_VERSION(current_block_height).map_or(latest_message_version, |seek_version| {
+            // Search the consensus value for the specified version.
+            match versions.binary_search_by(|(version, _)| version.cmp(&seek_version)) {
+                // If a value was found for this consensus version, return it.
+                Ok(index) => versions[index].1,
+                // If the specified version was not found exactly, determine whether to return an appropriate value anyway.
```
vicsn:
```suggestion
                // If the specified version was not found exactly, return the appropriate value belonging to the consensus version *lower* than the sought version.
```

node/router/messages/src/lib.rs
```diff
@@ -111,8 +112,43 @@ impl<N: Network> From<DisconnectReason> for Message<N> {
...
+                Err(index) => {
+                    // Return the appropriate value belonging to the consensus version *lower* than the sought version.
+                    // If the constant is not yet in effect at this consensus version, use the earliest version.
```
vicsn:
```suggestion
                    // If the constant is not yet in effect at this consensus version, use the earliest version.
```

vicsn: Because I moved it up.

node/router/messages/src/lib.rs
```diff
@@ -232,3 +268,64 @@ impl<N: Network> FromBytes for Message<N> {
...
+
+    /// Ensure that `Message::VERSIONS` increases and is correctly defined.
+    /// See the constant declaration for an explanation why.
+    fn message_version_increasing<N: Network>() {
+        let mut previous_value = Message::<N>::VERSIONS.first().unwrap().1;
+        for (_, value) in Message::<N>::VERSIONS.iter().skip(1) {
+            assert!(*value >= previous_value);
+            previous_value = *value;
+        }
+    }
```
vicsn: This looks duplicate to the test above.

node/router/messages/src/lib.rs
```diff
@@ -120,6 +120,36 @@ impl<N: Network> Message<N> {
...
+        // Fetch the versions.
+        let versions = Self::VERSIONS;
+
+        // Determine the minimum accepted message version.
+        // Example scenario:
+        // - At block height `X`, the protocol upgrades to message version from `Y-1` to `Y`.
+        // - Client A upgrades and starts using message version `Y`.
+        // - Client B has not upgraded and still uses message version `Y-1`.
+        // - Until block `X`, they stay connected and can communicate.
+        // - After block `X`, Client A will reject messages from Client B.
```
vicsn: Better as a docblock for `lowest_accepted_message_version` I think.

vicsn (dismissed):
LGTM! Left some low priority nits.

I don't think we have to test with larger number of clients.

kpandl (dismissed):
LGTM, tested it and was able to trigger `OutdatedClientVersion` logs in certain, expected scenarios. Also connects successfully to canary.

For reference, this are the builds I used for testing:

1. this branch with the latest snarkVM (`f23bc3caca8806d99c047d58456105b3c2cfe418`), compiled with the `test` feature (commit `e8d89621ae8dc2d19976cf78b9cf474823d96291` [here](https://github.com/kpandl/snarkOS/commits/test-feat/rolling-message-version/))
2. this branch with the latest snarkVM (`f23bc3caca8806d99c047d58456105b3c2cfe418`), using a new message version at ConsensusHeight V6, compiled with the `test` feature (commit `c1605c295032c41860d8b5447f624959cc4b2901` [here](https://github.com/kpandl/snarkOS/commits/test-feat/rolling-message-version/))
3. this branch with the latest snarkVM (`f23bc3caca8806d99c047d58456105b3c2cfe418`), (commit `e8d89621ae8dc2d19976cf78b9cf474823d96291` [here](https://github.com/kpandl/snarkOS/commits/test-feat/rolling-message-version/), like build 1 but without the test feature)


This are the tests I ran:

Local devnets
- 4 vals + 1 client run build 2. Once the 5 nodes reached ConsensusHeight V6, turned on another client running build 1, it wouldn’t connect due to OutdatedClientVersion.
- Similar setup but the node running build 1 is a prover instead of a client, same result.

Canary
- Already had a local canary storage of some blocks on my machine that was past test height V6. Connected build 2 to canary. The client wouldn’t connect due to OutdatedClientVersion.
- After wiping the local storage, build 2 would sync up to block 10, when test height V2 was triggered, and perceived the next block to be invalid.
- Connected build 3 to canary and let it sync to height 200, worked fine.

niklaslong (dismissed):

node/router/messages/src/lib.rs
raychu86: Resolved with [80c308a](https://github.com/ProvableHQ/snarkOS/pull/3596/commits/80c308a6cd540d7e5d5f3b9a7daa3ff1add2d9d1)

node/router/messages/src/lib.rs
raychu86: Resolved with [80c308a](https://github.com/ProvableHQ/snarkOS/pull/3596/commits/80c308a6cd540d7e5d5f3b9a7daa3ff1add2d9d1)

node/router/messages/src/lib.rs
raychu86: Resolved with [80c308a](https://github.com/ProvableHQ/snarkOS/pull/3596/commits/80c308a6cd540d7e5d5f3b9a7daa3ff1add2d9d1)

node/router/messages/src/lib.rs
raychu86: Resolved with [80c308a](https://github.com/ProvableHQ/snarkOS/pull/3596/commits/80c308a6cd540d7e5d5f3b9a7daa3ff1add2d9d1)

vicsn (approved):

PR #3602, opened 2025-04-21
Simplify `Storage::rounds` map.

acoglio:
Since certificate ID and batch ID are always the same, we can simplify this map to avoid having two identical IDs, and having just one instead.

### DIFF
--- a/node/bft/src/helpers/storage.rs
+++ b/node/bft/src/helpers/storage.rs
@@ -86,13 +86,13 @@ pub struct StorageInner<N: Network> {
     /// The maximum number of rounds to keep in storage.
     max_gc_rounds: u64,
     /* Once per batch */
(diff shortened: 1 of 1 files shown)


raychu86 (approved):
LGTM, fairly straightforward.

Might be worth checking if any of our documentation (inside or outside snarkOS) that considers `batch_id` and `certificate_id` different or separate things.

kaimast:
> Might be worth checking if any of our documentation (inside or outside snarkOS) that considers batch_id and certificate_id different or separate things.

It looks like both `BatchCertificate::id` and `BatchCertificate::batch_id` call `self.batch_header.batch_id()`. We should probably remove or deprecate one of th two.

kaimast (approved):

PR #3603, opened 2025-04-21
[Feature] Make height metric available during CDN sync

kpandl:
## Motivation

Previously, metrics would show height 0 during CDN sync.
This PR fixes it - a node syncing through CDN will export the current sync height through the metrics.

## Test Plan

Run `snarkos start --client --nodisplay --rest 0.0.0.0:3030 --bft 0.0.0.0:5000 --node 0.0.0.0:4130 --verbosity 0 --metrics --network 0`. Once the node is syncing, open `http://localhost:9000/metrics`.

Under `snarkos_bft_height_total`, it should show the current syncing height. Previously, it showed only 0.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3706,6 +3706,7 @@ dependencies = [
  "reqwest 0.11.27",
  "serde",
  "serde_json",
+ "snarkos-node-metrics",
  "snarkvm",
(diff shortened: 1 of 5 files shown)


vicsn (approved):
Very nice

Issue #3361, opened 2024-07-12
[Bug] Return valid Json when executing programs 


nathanhopp:
## 🐛 Bug Report
Improve outputs when executing program commands  

## Steps to Reproduce

```
curl <host>/<network>/program/credits.aleo/mapping/bonded/<public_key> 
"{\n  validator: <public_key>,\n  microcredits: 21091223930u64\n}"
```
The above can't be parsed, and is not valid json. 
## Expected Behavior

It should return valid json, example: 
```
{
  "validator": "<public_key>",
  "microcredits": "21065418248u64"
}
```
Even better microcredits not having `u64`. 

## Your Environment
independent of env.

joske:
Can reproduce, even though it's returning `Ok(ErasedJson::pretty(mapping_value))`. Investigating.

PR #3606, opened 2025-04-22
fix: return proper JSON from REST API

joske:
## Motivation

Some endpoints return just a `Display` implementation for serializing of a `Value`. This leads to invalid JSON. See #3361. Note that this approach *does* preserve order.

Filing as draft to solicit feedback.

## Test Plan

Tested manually as described in the linked issue.


joske:
@vicsn @niklaslong I think this approach is acceptable, it only changes the behaviour for the REST call.

kaimast:
Closing this in favor of #3751.

PR #3612, opened 2025-04-28
fix: missed version update

joske:
## Motivation

snarkOS fails to build because one instance of the snarkOS version was forgotten.

## Test Plan

build now works


joske:
@vicsn

vicsn (approved):

PR #3613, opened 2025-04-28
fix: avoid rebuilding snarkOS binary every single time

joske:
## Motivation

When you issue `cargo run` or even `cargo test` with no code changes at all, `cargo` would still re-build the snarkOS binary (and this can take a long time) on every single invocation. The cause for this is `cargo:rerun-if-changed=.`, this line is not needed at all, and `cargo` will still notice when the code changes and rebuild as needed. With this line removed, you will no longer have a rebuild on every run. The `cargo:rerun-if-changed=` directive should only be used on extra things, like json schema's  or generated code.

## Test Plan

manually tested (and only build.rs changes)

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -51,9 +51,6 @@ fn check_file_licenses<P: AsRef<Path>>(path: P) {
             );
         }
     }
-
-    // Re-run upon any changes to the workspace.
-    println!("cargo:rerun-if-changed=.");
 }
 
 // The build script; it currently only checks the licenses.


vicsn (approved):

kaimast:
I think this should be merged, as it would speed up the development process noticeably. That function in the build script does not actually change any source file and should not cause a recompile.

However, `build.rs` invokes the `built` crate to update the build information, which is shown in the CLI for example. There might be a reason we want to recreate the binary with updated build information even if nothing has changed. Maybe @raychu86 know if these rebuilds are intentional? 

Also, with the fix, does a second run of `cargo test` still cause the code to recompile? If I understand correctly, the first run changes the build information and writes it to some source file, and on the second run cargo would detect the changed file and recompile.

niklaslong (approved):

PR #3614, opened 2025-04-28
[Doc] Document an invariant on current round.

acoglio:
Mention that, because of the invariant, a test somewhere else should never fail.

Other minor comment fixes and improvements.


kaimast (approved):
Looks good!

raychu86 (approved):

PR #3615, opened 2025-04-28
Change BFT Sync to process block responses as expected

kaimast:
## Motivation

Stress testing revealed an error in the recent changes to block synchronization. When validators fall behind they do not sync block properly. Meaning, the sync blocks like clients, and do not add the contained certificates to the BFT module.

This PR provides a fix for that by modifying `Sync::advance_with_sync_blocks` and adds more documentation to the `Sync` and `BlockSync` structs.

It also renames two functions to make their functionality clearer:
* `Sync::sync_storage_with_blocks` was changed to `Sync::try_advancing_block_synchronization` to make it obvious that this is the validator version of `BlockSync::try_advancing_block_synchronization`.
* `Sync::try_block_sync` was changed to `Sync::issue_block_requests` because it only create and sends them, but does not process any block responses.

Finally, the PR adds function `Sync::try_block_sync` was added that encapsulates a full iteration of a validator's block synchronization. This function is invoked manually in some unit tests (mostly in `Primary`'s tests). 

## Test Plan

We will run more stress tests before merging this into staging (thanks to @kpandl for doing most of the work here).

## Related PRs

#3543 introduced the bug that this PR fixes.

### DIFF
--- a/node/bft/src/sync/mod.rs
+++ b/node/bft/src/sync/mod.rs
@@ -121,9 +121,14 @@ impl<N: Network> Sync<N> {
         self.sync_storage_with_ledger_at_bootup().await
     }
 
-    /// Performs one iteration of the block synchronization.
+    /// Issues new requests for blocks, if needed.
+    ///
+    /// Additionally, this function removes obsolete and timed out block requests,
+    /// and disconnects/bans unresponsive peers.
+    ///
+    /// Responses to block requests will eventually be processed by `Self::try_advancing_block_synchronization`.
     #[inline]
-    pub async fn try_block_sync(&self) {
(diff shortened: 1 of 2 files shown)


raychu86:
Changes make sense, as we do not want the validator nodes to call the sync functions of the internal `block_sync`, but rather use higher level wrapper logic.

Couple questions:
1. Is renaming `sync_storage_with_blocks` really more clear? There is now two instances of `try_advancing_block_synchronization` and they can't be clearly identified between validator and client impls without reading the documentation directly. The name doesn't really provide context that it is for validators.
2. Is there a way for us to prevent future mistakes of calling the internal `block_sync` functions if we are on the validator level?

PR #3617, opened 2025-04-29
[Fix] Fix quorum check for round advancement.

acoglio:
When deciding whether to advance from round `R` to `R+1`, this must check whether a quorum of certificates exists at round `R`, not `R+1`.

This is just one of several ways in which a validator can advance its round, which is why this issue may have been "masked".

### DIFF
--- a/node/bft/src/primary.rs
+++ b/node/bft/src/primary.rs
@@ -1360,24 +1360,27 @@ impl<N: Network> Primary<N> {
                     continue;
                 }
                 // Attempt to increment to the next round.
-                let next_round = self_.current_round().saturating_add(1);
+                let current_round = self_.current_round();
+                let next_round = current_round.saturating_add(1);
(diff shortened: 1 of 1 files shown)


kaimast:
This looks fine. However, I think when finding potential bugs in the consensus layer, there should be an actual unit/integration test that fails before the fix. 

Have you thought about if/how this could be tested?

kaimast (approved):

kaimast:
I am approving the PR. We will think about how to improve testing for primary's state in the future.

kaimast:
@raychu86 do you want to take another look before I merge this?

raychu86 (approved):
Would like be able to test these properties more concretely, but this seems correct.

I believe the previous logic was safe due to the BFT properties, but didn't fully assist in maintaining the liveness because of the off-by-one.

PR #3619, opened 2025-04-30
Label tmux windows in devnet as validators or clients

kaimast:
## Overview
This PR labels the tmux windows spawned by `devnet.sh` more clearly. Each validator spawns in a window labeled `validator-i` (where i is its index) and each client in a window labeled `client-i`.

The change makes it a little easier to figure out which windows run validators, not clients. The labels are also identical to the files in which each node's log is stored.

Finally, the PR also makes log verbosity a constant, so you can change it (albeit only by modifying the file) to increase log output when you are debugging.

## Test Plan
`devnet.sh` is only used for local development and not in production. CI also has its own devnet script at `.circle/devnet_ci.sh`.
So, I ran the script locally and verified it works as expected, which should be sufficient for a change like this.

### DIFF
--- a/devnet.sh
+++ b/devnet.sh
@@ -20,13 +20,16 @@ build_binary=${build_binary:-y}
 read -p "Do you want to clear the existing ledger history? (y/n, default: n): " clear_ledger
 clear_ledger=${clear_ledger:-n}
 
+# Log verbosity is set to 1 (DEBUG) by default.
+verbosity=1
+
 if [[ $build_binary == "y" ]]; then
   # Ask the user if they want to enable validator telemetry
   read -p "Do you want to enable validator telemetry? (y/n, default: y): " enable_telemetry
   enable_telemetry=${enable_telemetry:-y}
 
   # Build command
-  build_cmd="cargo install --locked --path ."
+  build_cmd="cargo install --locked --path . --offline --debug"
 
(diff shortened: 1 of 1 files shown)


devnet.sh
```diff
@@ -20,13 +20,16 @@ build_binary=${build_binary:-y}
...
+verbosity=1
+
 if [[ $build_binary == "y" ]]; then
   # Ask the user if they want to enable validator telemetry
   read -p "Do you want to enable validator telemetry? (y/n, default: y): " enable_telemetry
   enable_telemetry=${enable_telemetry:-y}
 
   # Build command
-  build_cmd="cargo install --locked --path ."
+  build_cmd="cargo install --locked --path . --offline --debug"
```
vicsn: Can the offline option be an issue if you don't have any of the crates downloaded yet?

niklaslong: @kaimast I think we should remove both of these new options:
- running with `--debug` will make everything much less performant (especially as VM crypto is involved), not an ideal default. 
- running with `--offline` can cause issues if dependencies are not cached by the user.

kaimast: Good catch. I added them to speed up compilation and forgot to remove them.

niklaslong (approved):

PR #3626, opened 2025-05-07
Fix remaining synchronization bugs

kaimast:
This PR reverts the revert of PR #3543 and fixes two issues we detected during stress tests. 

The changes are structured as follows:

* [db40b35](https://github.com/kaimast/snarkOS/pull/3/commits/db40b35d8d8ef374ea5970878c76b106560fd442) reverts the revert (undos commit). These changes have already been reviewed for the reverted PR.
* [7f6762b](https://github.com/ProvableHQ/snarkOS/pull/3626/commits/7f6762b4bd4d1e951013afe3f0a2cc31f6b77c61) ensures that, for validators, BFT's sync module processes new blocks, instead of them being processed by the generic BlockSync module.
* [434a406](https://github.com/kaimast/snarkOS/pull/3/commits/434a406a97b40821e8e5d49704a6b6a543ea97e8) fixes another bug, that is unlikely to happen on mainnet but popped up during stress testing (see the description on "pending blocks" below).

## Pending Blocks
Currently, validators have two different sync modes. For older blocks (>=100 rounds ago) they sync like clients performing only some of the DAG checks, for more recent blocks (<100 rounds ago) validators check that there are enough votes for each new block's leader certificate in some subsequent block. This check happens in `snarkos_node_bft::sync::sync_storage_with_block`.

Checking for sufficient votes means that, in addition to blocks added to the ledger, there is also a set of "pending blocks" which cannot be added to the ledger yet because they did not reach the availability threshold yet.
The generic BlockSync module, however, is unaware of these pending blocks which can cause issues, as outlined below.

### Detected Issue and Proposed Fix
In the common case, there is only one pending block. However, during stress testing we triggered a case where a majority of validators restart after resetting their ledgers, and the most recent block did not receive sufficient votes. Here, there exists a block `i-1` and a block `i`, where block `i` does not contain sufficient votes for `i-1`.
Usually, a subsequent block, here w.l.o.g. block `i+1`, contains sufficient votes to add all its pending predecessors to the ledger.

The issue we found is that, in this scenario, BlockSync considers itself more than one block behind and, thus, not fully synced. The network then cannot make progress to create block i+1 because most validators are still syncing, but without `i+1` block `i-1` never gets added to the ledger.
The final commit in this PR considers pending blocks as part of the sync state to address this.

## Testing Plan
We already ran stresstests on this branch and did not detect any errors. However, I want to perform another run of stresstest to be on the safe side.

## Notes
There is a larger PR in the works to perform the checks on pending blocks on all nodes, not just validators, and move more of the DAG checks into snarkVM. However, I want to merge this version first, as the delta to the last PR is fairly small, and the larger PR requires more extensive testing and benchmarking.


node/bft/src/sync/mod.rs
```diff
@@ -102,16 +125,67 @@ impl<N: Network> Sync<N> {
...
     /// Starts the sync module.
+    ///
+    /// When this function returns sucessfully, the sync module will have spawned background tasks
+    /// that fetch blocks from other validators.
     pub async fn run(&self, sync_receiver: SyncReceiver<N>) -> Result<()> {
         info!("Starting the sync module...");
 
         // Start the block sync loop.
         let self_ = self.clone();
-        self.handles.lock().push(tokio::spawn(async move {
```
ljedrz, line 114: do we not want to keep track of this long-running task? it facilitates a cleaner shutdown

ljedrz: ah, it now goes through `self.spawn`, which does the same :+1:

node/bft/src/sync/mod.rs
```diff
@@ -435,14 +559,21 @@ impl<N: Network> Sync<N> {
     /// and its addition to the ledger is deferred until the check passes.
     /// Several blocks may be stored in `Sync::latest_block_responses`
     /// before they can be all checked and added to the ledger.
-    pub async fn sync_storage_with_block(&self, block: Block<N>) -> Result<()> {
+    async fn sync_storage_with_block(&self, block: Block<N>) -> Result<()> {
         // Acquire the sync lock.
         let _lock = self.sync_lock.lock().await;
         // Acquire the latest block responses lock.
         let mut latest_block_responses = self.latest_block_responses.lock().await;
```
ljedrz: perhaps we should move this assignment under the check below, in case of an early return?

ljedrz:
Nit: the link to the 2nd commit should be https://github.com/ProvableHQ/snarkOS/pull/3626/commits/7f6762b4bd4d1e951013afe3f0a2cc31f6b77c61.

ljedrz (approved):
LGTM :+1:

node/bft/src/sync/mod.rs
kaimast: Yes, exactly. Not sure why that line did not use self.spawn before, but it was probably just an oversight.

node/bft/src/helpers/dag.rs
```diff
@@ -113,19 +127,29 @@ impl<N: Network> DAG<N> {
         /* Updates */
 
         // Update the recently committed IDs.
-        self.recent_committed_ids.entry(certificate_round).or_default().insert(certificate_id);
+        let is_new = self.recent_committed_ids.entry(certificate_round).or_default().insert(certificate_id);
+        if !is_new {
+            //TODO (kaimast): return early here?
```
vicsn, line 132: Thank you!

Some open TODOs:
- PR message indicates "However, I want to perform another run of stresstest to be on the safe side.". Please indicate on a high level in the PR message which tests you ran.
- added @raychu86 for another independent strong review, maybe it would be good if you sit down together.
- @kaimast would you like to resolve the TODOs before or during review with Ray?

kaimast: My plan was to leave the TODOs for another PR but I can make those changes now if you prefer.

I already ran the stresstests but it would be good if someone else runs them to verify. However, now that it didn't land in 3.8 and we won't tag a new release for a while, I am not too worried about breaking things anymore.

node/bft/src/helpers/dag.rs
```diff
@@ -22,12 +22,24 @@ use snarkvm::{
 use indexmap::IndexSet;
 use std::collections::{BTreeMap, HashMap};
 
+/// Maintains an directed acyclic graph (DAG) of batches, from which we build a totally-ordered blockchain.
```
acoglio:
```suggestion
/// Maintains a directed acyclic graph (DAG) of certificates, from which we build a totally-ordered blockchain.
```

node/bft/src/primary.rs
```diff
@@ -1203,7 +1206,9 @@ impl<N: Network> Primary<N> {
         self.spawn(async move {
             while let Some((peer_ip, primary_certificate)) = rx_primary_ping.recv().await {
                 // If the primary is not synced, then do not process the primary ping.
-                if !self_.sync.is_synced() {
+                if self_.sync.is_synced() {
+                    trace!("Processig new primary ping from '{peer_ip}'");
```
```suggestion
                    trace!("Processing new primary ping from '{peer_ip}'");
```

node/bft/src/sync/mod.rs
```diff
@@ -333,31 +426,58 @@ impl<N: Network> Sync<N> {
...
+        // Figure out which height we are synchronized to.
+        // If there are queued block responses, this might be higher than the latest block in the ledger.
+        let sync_height = {
+            let responses = self.latest_block_responses.lock().await;
+            if let Some((height, _)) = responses.last_key_value() { *height } else { self.ledger.latest_block_height() }
+        };
+
         // Retrieve the next block height.
         // This variable is used to index blocks that are added to the ledger;
         // it is incremented as blocks as added.
```
acoglio:
```suggestion
        // it is incremented as blocks are added.
```

acoglio (dismissed):

node/bft/src/sync/mod.rs
kaimast: Done!

node/bft/src/bft.rs
```diff
@@ -1100,8 +1118,11 @@ mod tests {
         let storage = Storage::new(ledger.clone(), Arc::new(BFTMemoryService::new()), 10);
         // Initialize the account.
         let account = Account::new(rng)?;
-        // Initialize the BFT.
-        let bft = BFT::new(account.clone(), storage.clone(), ledger.clone(), None, &[], StorageMode::new_test(None))?;
+
+        // Set up the BFT logic.
+        let bft = initialize_bft(account.clone(), storage.clone(), ledger.clone())?;
+        assert!(bft.is_timer_expired()); // 0 + 5 < now()
```
vicsn: Are these comments needed?

vicsn (dismissed):

node/bft/src/bft.rs
kaimast: I removed them and added a comment that `bft.is_timer_expired` is always true when you first create an instance of `BFT`.

vicsn (approved):
Did not review the code in depth, but know the other reviewers did, and tests were run succesfully so approving. Amazing work folks!

PR #3628, opened 2025-05-09
[Feature] Prometheus Metrics for Telemetry

raychu86:
## Motivation

This PR adds Prometheus Metrics for the validator telemetry participation scores.

Additionally, this adjusts the scoring in two ways:
1. Calculate weighted at time of insertion rather than query.
2. Increase participation leniency for certificates to be every two rounds rather than every round.

### DIFF
--- a/node/bft/src/helpers/telemetry.rs
+++ b/node/bft/src/helpers/telemetry.rs
@@ -169,10 +169,8 @@ impl<N: Network> Telemetry<N> {
         let validator_signatures = self.validator_signatures.read();
         let validator_certificates = self.validator_certificates.read();
 
-        // Fetch the total number of certificates and rounds.
-        let num_certificate_rounds = tracked_certificates.len();
+        // Fetch the total number of certificates.
         let total_certificates = validator_certificates.values().map(|rounds| rounds.len()).sum::<usize>();
-        let tracked_rounds: Vec<_> = tracked_certificates.keys().copied().collect();
 
         // Calculate the signature participation scores for each validator.
         let signature_participation_scores: IndexMap<_, _> = cfg_iter!(validator_signatures)
@@ -185,6 +183,7 @@ impl<N: Network> Telemetry<N> {
 
         // Calculate the certificate participation scores for each validator.
         // This score is based on how many certificates the validator has included in every two rounds.
+        let tracked_rounds: Vec<_> = tracked_certificates.keys().skip_while(|r| *r % 2 == 0).copied().collect();
         let certificate_participation_scores: IndexMap<_, _> = cfg_iter!(validator_certificates)
             .map(|(address, certificate_rounds)| {
                 // Count the number of round pairs that are included in the certificate rounds.
@@ -192,7 +191,7 @@ impl<N: Network> Telemetry<N> {
                     .filter(|chunk| chunk.iter().any(|r| certificate_rounds.contains(r)))
                     .count();
                 // Calculate the number of round pairs.
-                let num_round_pairs = (num_certificate_rounds.saturating_add(1)).saturating_div(2);
+                let num_round_pairs = (tracked_rounds.len().saturating_add(1)).saturating_div(2);
                 // Calculate the score based on the number of certificate rounds the validator is a part of.
                 let score = num_included_round_pairs as f64 / num_round_pairs.max(1) as f64 * 100.0;
                 (*address, score as u16)
--- a/node/bft/src/helpers/telemetry.rs
+++ b/node/bft/src/helpers/telemetry.rs
@@ -36,11 +36,18 @@ use std::{collections::BTreeMap, sync::Arc};
 //  - Various stake weight considerations
 //  - The latest seen IP address of each validator (useful for debugging purposes)
 
+/// The participation scores for each validator.
+///     Certificate Score: The % of rounds the validator has a valid certificate
+///     Signature Score: The % of certificates the validator has a valid signature for
+///     Combined Score: The weighted score using the certificate and signature scores
+type ParticipationScores = (f64, f64, f64);
+
 /// Tracker for the participation metrics of validators.
 #[derive(Clone, Debug)]
 pub struct Telemetry<N: Network> {
     /// The certificates seen for each round
     /// A mapping of `round` to set of certificate IDs.
+    /// Note that this map is sorted to allow grouped iteration over rounds.
     tracked_certificates: Arc<RwLock<BTreeMap<u64, IndexSet<Field<N>>>>>,
 
     /// The total number of signatures seen for a validator, including for their own certificates.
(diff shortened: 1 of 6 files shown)


node/bft/src/helpers/telemetry.rs
```diff
@@ -51,8 +51,11 @@ pub struct Telemetry<N: Network> {
     /// A mapping of `address` to a list of rounds.
     validator_certificates: Arc<RwLock<IndexMap<Address<N>, IndexSet<u64>>>>,
 
-    /// The certificate and signature participation scores for each validator.
-    participation_scores: Arc<RwLock<IndexMap<Address<N>, (f64, f64)>>>,
+    /// The certificate, signature, and combined participation scores for each validator.
+    ///     Certificate Score: The % of rounds the validator has a valid certificate
+    ///     Signature Score: The % of certificates the validator has a valid signature for
+    ///     Combined Score: The weighted score using the certificate and signature scores
+    participation_scores: Arc<RwLock<IndexMap<Address<N>, (f64, f64, f64)>>>,
```
kaimast: Should we add a `ParticipationScores` struct that has these three fields? It might make the code easier to read.

node/bft/src/helpers/telemetry.rs
```diff
@@ -182,10 +182,18 @@ impl<N: Network> Telemetry<N> {
             .collect();
 
         // Calculate the certificate participation scores for each validator.
+        // This score is based on how many certificates the validator has included in every two rounds.
+        let tracked_rounds: Vec<_> = tracked_certificates.keys().skip_while(|r| *r % 2 == 0).copied().collect();
```
kaimast, line 190: Why do you use `skip_while` and not `filter` here?

Assuming I understand the [documentation](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.skip_while) correctly, this will give you all rounds from 2 onwards (or even from 0 onwards), not just the even rounds.

Side note: we really need to add a `is_certificate_round` function in the future.

node/bft/src/helpers/telemetry.rs
```diff
@@ -182,10 +182,18 @@ impl<N: Network> Telemetry<N> {
...
         let certificate_participation_scores: IndexMap<_, _> = cfg_iter!(validator_certificates)
             .map(|(address, certificate_rounds)| {
-                // Calculate a rough score for the validator based on the number of certificates seen.
-                let score = certificate_rounds.len() as f64 / num_certificate_rounds as f64 * 100.0;
+                // Count the number of round pairs that are included in the certificate rounds.
+                let num_included_round_pairs = cfg_chunks!(tracked_rounds, 2)
```
kaimast, line 194: Can this not create a lot of chunks? I am unclear how big `tracked_rounds` will get.

node/bft/src/helpers/telemetry.rs
```diff
@@ -182,10 +182,18 @@ impl<N: Network> Telemetry<N> {
...
+                    .filter(|chunk| chunk.iter().any(|r| certificate_rounds.contains(r)))
+                    .count();
+                // Calculate the number of round pairs.
+                let num_round_pairs = (tracked_rounds.len().saturating_add(1)).saturating_div(2);
+                // Calculate the score based on the number of certificate rounds the validator is a part of.
+                let score = num_included_round_pairs as f64 / num_round_pairs.max(1) as f64 * 100.0;
```
kaimast, line 200: Wouldn't this be faster and easier to read?
```suggestion
                let score = (num_included_round_pairs * 100) / (num_round_pairs.max(1) * 100);
```

node/bft/src/helpers/telemetry.rs
```diff
@@ -41,7 +41,7 @@ use std::sync::Arc;
 pub struct Telemetry<N: Network> {
     /// The certificates seen for each round
     /// A mapping of `round` to set of certificate IDs.
-    tracked_certificates: Arc<RwLock<IndexMap<u64, IndexSet<Field<N>>>>>,
+    tracked_certificates: Arc<RwLock<BTreeMap<u64, IndexSet<Field<N>>>>>,
```
kaimast, line 51: Can you add a comment why this needs to be sorted?

kaimast (changes requested):
I added a few comments and suggestions. But looks like a pretty straightforward change that we can hopefully merge soon.

node/bft/src/helpers/telemetry.rs
raychu86: No, the suggested logic would not work because `num_included_round_pairs` and `num_round_pairs` are of type usize.

An example is (2 * 100) / (3 * 100) = 0 rather than the 0.66 we want to represent it by

node/bft/src/helpers/telemetry.rs
raychu86: `tracked_rounds` should only ever have `gc_rounds` items, which is 100. This would create 50 chunks

node/bft/src/helpers/telemetry.rs
raychu86: `skip_while` is used because we just want to skip the initial odd round. I'm not sure filter could achieve that type of functionality.

raychu86: I am not accounting for complete correctness of the sudags, but rather just ingesting the block subdags. We can definitely add additional certificate round checks for robustness in the future.

node/bft/src/helpers/telemetry.rs
raychu86: Done - [11505f2](https://github.com/ProvableHQ/snarkOS/pull/3628/commits/11505f2fe3be89ca9603181ee2d6f3acb07a1ad1)

node/bft/src/helpers/telemetry.rs
raychu86: Done - [11505f2](https://github.com/ProvableHQ/snarkOS/pull/3628/commits/11505f2fe3be89ca9603181ee2d6f3acb07a1ad1)

node/bft/src/helpers/telemetry.rs
kaimast: Sorry, that was supposed to say `(num_included_round_pairs * 100) / num_round_pairs.max(1)`, but it probably does not make a huge difference.

node/bft/src/helpers/telemetry.rs
kaimast: Oh okay, so you are checking if a validator participated in either round of a block? I for some reason thought you are only counting commit rounds (probably because I confused certificate and commit round in my earlier comment).

raychu86: Yea that's right. It's not an exact bijection of block subdags - it's more of a best guess on participation by checking certificate existence in every 2 rounds.

node/bft/src/helpers/telemetry.rs
raychu86: I'd opt to keep it as is with floating points for more granularity, but i agree it's a bit wordy

kaimast (approved):

PR #3640, opened 2025-05-20
[Feature] Improve ergonomics of the CLI

kaimast:
## Motivation

There were a few things that made the CLI a little awkward to use for me. For example, the network type is passed an integer (network ID) and it is easy to forget which one is which. 
In some cases, the CLI would also panic on an invalid input, which is not a great user experience.

## Overview

There are four commits in this PR:
* [40d8ab5](https://github.com/ProvableHQ/snarkOS/pull/3640/commits/40d8ab5ce45e32880ba87cd281ea59ad8d988888) and [d3bcf3b](https://github.com/ProvableHQ/snarkOS/pull/3640/commits/d3bcf3b662fc3641e4e4d728929e1ba871bdb189) add a more detailed help text for the network ID. It will also only allow values in `0..=2`.

* [826217a](https://github.com/ProvableHQ/snarkOS/pull/3640/commits/826217ab5da335a94dd9fdf50dc687ea00898bf2) ensures that at most one node type can be specified and clarifies that `--client` is the default.

* [810f034](https://github.com/ProvableHQ/snarkOS/pull/3640/commits/810f0349442c8630aa592ae5c272c2bb2da01f22) ensures errors while parsing the node arguments do not cause a panic. This commit also changes how errors are printed. Please see below for more details.

### New Error Handling

Before the CLI simply printed `{error}` which only shows the top-level error, not the root cause (for this you would do `println!("{}", error.root_cause())` instead). So you would get something like this:
```
⚠️  Failed to parse node arguments
```

I decided to leave the top-level error as is and additionally print the rest of the error chain. Now it will show something like this:
```
⚠️  Failed to parse node arguments
     ↳ Missing the '--private-key' or '--private-key-file' argument
```

## Test Plan
Manually tested.

## Notes
We should consider picking a different emoji. The current one is a "warning sign", which is not a good fit for a (fatal) error.

### DIFF
--- a/cli/src/commands/clean.rs
+++ b/cli/src/commands/clean.rs
@@ -13,19 +13,19 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use snarkos_node::bft::helpers::proposal_cache_path;
-
 use aleo_std::StorageMode;
 use anyhow::{Result, bail};
 use clap::Parser;
 use colored::Colorize;
+use snarkos_node::bft::helpers::proposal_cache_path;
+use snarkvm::console::network::{CanaryV0, MainnetV0, Network};
 use std::path::PathBuf;
 
 /// Cleans the snarkOS node storage.
 #[derive(Debug, Parser)]
 pub struct Clean {
-    /// Specify the network to remove from storage.
-    #[clap(default_value = "0", long = "network")]
+    /// Specify the network to remove from storage (0 = mainnet, 1 = testnet, 2 = canary)
+    #[clap(default_value_t=MainnetV0::ID, long = "network", value_parser = clap::value_parser!(u16).range((MainnetV0::ID as i64)..=(CanaryV0::ID as i64)))]
     pub network: u16,
     /// Enables development mode, specify the unique ID of the local node to clean.
     #[clap(long)]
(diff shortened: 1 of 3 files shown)


ljedrz (approved):
LGTM :+1:

cli/src/commands/clean.rs
```diff
@@ -13,19 +13,19 @@
...
 // limitations under the License.
 
-use snarkos_node::bft::helpers::proposal_cache_path;
-
 use aleo_std::StorageMode;
 use anyhow::{Result, bail};
 use clap::Parser;
 use colored::Colorize;
+use snarkos_node::bft::helpers::proposal_cache_path;
+use snarkvm::console::network::{CanaryV0, MainnetV0, Network};
```
niklaslong: Nit: could you keep VM and OS deps separate from imported crates? 🙏

niklaslong (dismissed):

cli/src/commands/clean.rs
kaimast: Done!

kaimast:
I did another rebase to trigger CI.

niklaslong (approved):

PR #3644, opened 2025-05-21
[Build] Remove unused dependencies

kaimast:
## Overview
This PR removes unused dependencies from crates in snarkOS. I used `cargo machete` to determine which dependencies are unused. While unused code is removed at link-time, removing them outright reduces compile times slightly and decreases the size of `Cargo.toml` files. 

The second commit also adds a CI job that runs `cargo machete` to detect any unused dependencies on new commits to prevent future bloat. We can only merge the first commit if we decide to not automatically test for unused dependencies.

## Notes and "Future Work"
There are a few places (mostly the top-level snarkOS crate) where we add indirect dependencies as direct dependencies, solely to enable features on them.
Most of these superfluous dependencies are not needed. For example, `snarkos` depends on `snarkos-node` so it can enable the `locktick` feature on it. But that feature is also enabled by `snarkos-cli` which actually uses the `snarkos-node` dependency.
However, I did not want to add any more changes to this PR, so I simply set some dependencies to be ignored by `cargo machete` using the `package.metadata.cargo-machete.ignored` field in `Cargo.toml`.

There is also a feature on `snarkos-node` called `parallel` that does not seem to be used anywhere, but I did not want to remove it in this PR either.


node/Cargo.toml
```diff
@@ -172,3 +158,8 @@ features = [ "env-filter", "fmt" ]
 
 [dev-dependencies.rand_chacha]
 version = "0.3.0"
+
+[package.metadata.cargo-machete]
+ignored = [
+  "serde_json", # Need to set preverse_order feature
```
ljedrz:
```suggestion
  "serde_json", # Need to set preserve_order feature
```

ljedrz (approved):

kaimast:
Rebased to trigger CI, but I will probably need to merge/rebase again after #3666 is merged.

node/Cargo.toml
```diff
@@ -18,7 +18,7 @@ edition = "2021"
 
 [features]
 default = [ "parallel" ]
-parallel = [ "rayon" ]
+parallel = [ ] #TODO(kaimast): is this still used?
```
vicsn, line 21: FYI I found a while ago that although snarkVM has a very clean `serial` feature with consistent use of cfg_*, snarkOS does not at the moment

vicsn (dismissed):

kaimast:
Merged staging.

The only other change I had to do is lock the `base64ct` dependency to `1.7` as never versions require Rust 2024 (>=1.85) and we are still at Rust 1.83.

vicsn (approved):

PR #3647, opened 2025-05-21
fix(node/router): handle repated connects to same peer better

kaimast:
The heartbeat logic has no notion of pending connections, so it might call `Router::connect` multiple times for the same peer while the handshake is happening.

This change prevents frequent warnings being generated when a node attempts to connect to a peer, that it already connected or is currently connecting to, by not considering those cases an error.

In particular, it removes warnings like the following (they are still logged as debug messages):
```
WARN Dropping connection attempt to '127.0.0.1:4134' (already shaking hands as the initiator)
```

## Test Plan
I ran a local devnet to verify this works as expected.

Reviewers, let me know if you think this should also be stress tested. It really only changes the log output but is a networking change, nevertheless.

## Notes
 An alternative solution would be to add more state to the heartbeat logic. Currently, heartbeat has little state and, for example, discards the return value `Router::connect` instead of checking the result.

It seemed less invasive to make these changes in `Router` and keep `Heartbeat` unchanged.

### DIFF
--- a/node/router/src/lib.rs
+++ b/node/router/src/lib.rs
@@ -166,11 +166,18 @@ impl<N: Network> Router<N> {
 
 impl<N: Network> Router<N> {
     /// Attempts to connect to the given peer IP.
+    ///
+    /// Returns None if we are already connected to the peer or cannot connect.
+    /// Otherwise, it returns a handle to the tokio tasks that sets up the connection.
     pub fn connect(&self, peer_ip: SocketAddr) -> Option<JoinHandle<bool>> {
         // Return early if the attempt is against the protocol rules.
-        if let Err(forbidden_message) = self.check_connection_attempt(peer_ip) {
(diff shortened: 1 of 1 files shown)


ljedrz (approved):
LGTM :+1:

bendyarm (approved):
Looks good to me.
I didn't see any other callers to `check_connection_attempt`.
I don't think it needs stress testing separately; it is low-enough risk that it can be added to other changes that are stress-tested together, IMHO.

kaimast:
> I didn't see any other callers to `check_connection_attempt`.

Yeah. In hindsight, I could also just have merged the function into `connect`. 

> I don't think it needs stress testing separately; it is low-enough risk that it can be added to other changes that are stress-tested together, IMHO.

Good point. There should indeed be a few more stress tests run of staging until the next release.

PR #3649, opened 2025-05-22
[Build] Fix warning about yanked version of zip

kaimast:
`zip` 2.6.* was [yanked due to a breaking change](https://github.com/zip-rs/zip2/issues/337). This PR only updates `Cargo.lock` using the result of `cargo update zip`.
`zip` is not a direct dependency, but needed by the `self_update` crate, so we cannot update to 3.* or 4.* yet.


ljedrz (approved):

kaimast:
Closing in favor of #3662.

PR #3650, opened 2025-05-22
Fix or improve a few peering-related comments.

bendyarm:
in `gateway.rs` there was a vestigial reference to "router".
In `handshake.rs` fixed the incorrect logic in the comment.
In `start.rs` clarified that Client and Prover nodes ignore the `allow_external_peers` flag and always allow untrusted peers to connect.


niklaslong (approved):

node/router/src/handshake.rs
```diff
@@ -318,7 +318,7 @@ impl<N: Network> Router<N> {
         if self.is_connected(&peer_ip) {
             bail!("Dropping connection request from '{peer_ip}' (already connected)")
         }
-        // Only allow trusted peers to connect if allow_external_peers is set
```
kaimast, line 321: Thanks for catching this. I might have written this and obviously meant *untrusted* peers.

kaimast (approved):
Looks good!

PR #3661, opened 2025-05-27
Remove mainnet check for test_networks

meddle0x53:
## Motivation

As the stress tests are ran for a cluster in mainnet mode, but also need the feature `test_network` enabled, based on this conversation with @vicsn : https://github.com/ProvableHQ/stress-testing/pull/288#discussion_r2104512301 I'm proposing removing these checks, so we don't have to patch it for the tests.

## Test Plan

The current PR in the stress tests repository https://github.com/ProvableHQ/stress-testing/pull/288 for using the `test_network` feature works by doing this change with a manual patch. The tests pass (can be checked by running the upgrade test, for example from this PR). After this change I can remove the patch.

## Related PRs

https://github.com/ProvableHQ/snarkOS/pull/3572

### DIFF
--- a/cli/src/commands/start.rs
+++ b/cli/src/commands/start.rs
@@ -543,7 +543,7 @@ impl Start {
         if cfg!(feature = "test_network") && self.dev.is_none() {
             bail!("The 'test_network' feature is enabled, but the '--dev' flag is not set");
         }
-        
+
         // Parse the trusted peers to connect to.
         let mut trusted_peers = self.parse_trusted_peers()?;
         // Parse the trusted validators to connect to.
--- a/cli/src/commands/start.rs
+++ b/cli/src/commands/start.rs
@@ -520,7 +520,7 @@ impl Start {
     async fn parse_node<N: Network>(&mut self, shutdown: Arc<AtomicBool>) -> Result<Node<N>> {
         // Print the welcome.
         println!("{}", crate::helpers::welcome_message());
-        
+
         // Check if we are running with the lower coinbase and proof targets. This should only be
         // allowed in --dev mode and should not be allowed in mainnet mode.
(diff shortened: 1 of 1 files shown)


cli/src/commands/start.rs
```diff
@@ -521,15 +521,6 @@ impl Start {
         // Print the welcome.
         println!("{}", crate::helpers::welcome_message());
 
-        // Check if we are running with the lower coinbase and proof targets. This should only be
-        // allowed in --dev mode and should not be allowed in mainnet mode.
-        if cfg!(feature = "test_network") && self.dev.is_none() {
```
vicsn, line 543: We can and should keep the check in place that we can only use this in dev mode

vicsn (changes requested):

vicsn (dismissed):

cli/src/commands/start.rs
kaimast: Could we still `warn!` here?

vicsn: The original code is intact on the latest commit, we are bailing and exiting and I assume the error is reported in CLI. Do you want to warn on top for logging purposes?

kaimast: I mean instead of removing the if branch entirely, change it to use `warn!` instead of `bail!`

vicsn: More general, I like the idea of surfacing the feature flags and network. We already collect and print this for the --version flag, not sure if its a log line. Feel free to add in a separate PR.

kaimast:
I merged staging to trigger CI.

meddle0x53:
For some reason the `cargo fmt` didn't fix this and I now fixed it manually, strange.

raychu86 (approved):

Issue #3630, opened 2025-05-12
[Bug] snarkOS doesn't build with GCC 15


ljedrz:
## 🐛 Bug Report

Root issue: https://github.com/rust-rocksdb/rust-rocksdb/issues/991.

Workaround: `CXXFLAGS="$CXXFLAGS -include cstdint" cargo build --release`.

kaimast:
I have the same problem on Arch Linux. This also seems to happen with a more recent version of clang. My Linux machine is at clang 19.* while macOS is still at 17.*.

kaimast:
This is fixed upstream, but they didn't tag a release for it (yet). We are also not on the most recent version of the rocksdb crate, so not sure how easy upgrading will be. 

Maybe we can just add the CXXFLAGS to `.cargo/config.toml`? I assume macOS will eventually update clang and things will break there too.

ljedrz:
The upgrade is not an issue, I've already done it in https://github.com/ProvableHQ/snarkOS/pull/3671. As for including `CXXFLAGS` in `config.toml`, something like the following would work:
```
[env]
CXXFLAGS = "-include cstdint"
```

It's only a matter of whether we want to work around it right away; @vicsn?

vicsn:
IMO good to update the Cargo.toml if its permanent and useful - if it becomes superfluous or requires another change within a few weeks we can wait and ignore.

ljedrz:
Yeah, it's the latter.

PR #3671, opened 2025-06-03
[Feat] Database checkpoints

ljedrz:
This PR exposes the functionality introduced to snarkVM in https://github.com/ProvableHQ/snarkVM/pull/2742 via the REST API.

Creating a checkpoint at location `/home/aleo_ledger_checkpoints/1`:
```
curl -X POST http://localhost:3030/mainnet/db_backup?path=/home/aleo_ledger_checkpoints/1
```

Rolling back to (or, more precisely, switching to) the aforementioned checkpoint:
```
snarkos --storage=/home/aleo_ledger_checkpoints/1
```
Alternatively, the original ledger can be removed, and manually substituted with the checkpoint.

In addition to the [root issue](https://github.com/ProvableHQ/snarkVM/issues/2707):
- the checkpoint's location can have any name - it only needs to be a non-existent one
- (almost - depends on the hard link limits of the filesystem) any number of checkpoints can be created locally

Filing as a draft, pending community feedback wrt the scope.

### DIFF
--- a/.circleci/db_backup_ci.sh
+++ b/.circleci/db_backup_ci.sh
@@ -8,6 +8,8 @@ network_name="mainnet"
 # Stopping conditions
 checkpoint_height=3
 rollback_height=10
+num_checkpoints=0
+remaining_checkpoints=2
 
 # Use fixed JWT values in order to be able to create checkpoints
 jwt_secret="ZGJjaGVja3BvaW50dGVzdA=="
@@ -55,7 +57,8 @@ check_heights() {
 create_checkpoints() {
   for ((node_index = 0; node_index < $total_validators; node_index++)); do
     port=$((3030 + node_index))
-    result=$(curl -s -X "POST" -H "Authorization: Bearer ${jwt[node_index]}" "http://127.0.0.1:$port/$network_name/db_backup?path=/tmp/checkpoint$node_index" || echo "fail")
+    suffix="${node_index}_$1"
+    result=$(curl -s -X "POST" -H "Authorization: Bearer ${jwt[node_index]}" "http://127.0.0.1:$port/$network_name/db_backup?path=/tmp/checkpoint_$suffix" || echo "fail")
 
     # Track highest height for reporting
     if [ "$result" = "fail" ]; then
@@ -73,19 +76,26 @@ sleep 15
 
 # Check heights periodically with a timeout
 total_wait=0
-checkpoints_created=false
+checkpoint_created=false
 while [ $total_wait -lt 300 ]; do  # 5 minutes max
   # Apply short-circuiting
-  if [[ $checkpoints_created = true ]] || check_heights "$checkpoint_height"; then
-    if [[ $checkpoints_created = false ]]; then
+  if [[ $checkpoint_created = true ]] || check_heights "$checkpoint_height"; then
+    if [[ $checkpoint_created = false ]]; then
       # Create checkpoints at the specified height
-      create_checkpoints
-      checkpoints_created=true
+      create_checkpoints $num_checkpoints
+      checkpoint_created=true
+      checkpoint_height=$((checkpoint_height+2))
+      num_checkpoints=$((num_checkpoints+1))
+
+      echo "num_checkpoints: $num_checkpoints"
+      sleep 2
     fi
 
     # Wait until the specified rollback height is reached
     if check_heights "$rollback_height"; then
       echo "All nodes reached rollback height."
+
+      checkpoint_created=false
 
       # Gracefully shut down the validators
       for pid in "${PIDS[@]}"; do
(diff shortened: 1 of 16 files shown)


ljedrz:
note: the CI issues appear to have been caused by https://github.com/ProvableHQ/snarkOS/issues/3630

vicsn:
Can you revive and rebase: https://github.com/ProvableHQ/snarkOS/pull/3592

Can you write a short manual in the top PR message for how to backup (call endpoint) and recover (???)

.circleci/db_backup_ci.sh
```diff
@@ -0,0 +1,145 @@
...
+echo "Waiting 15 seconds for network to start up..."
+sleep 15
+
+# Check heights periodically with a timeout
+total_wait=0
+checkpoints_created=false
+while [ $total_wait -lt 300 ]; do  # 5 minutes max
+  # Apply short-circuiting
+  if [[ $checkpoints_created = true ]] || check_heights "$checkpoint_height"; then
+    if [[ $checkpoints_created = false ]]; then
```
vicsn: Can you test creating at least two checkpoints; recovering from genesis to the first, and from the first to the second?

.circleci/db_backup_ci.sh
```diff
@@ -22,7 +24,7 @@ declare -a PIDS
 
 # Start all validator nodes in the background
 for ((validator_index = 0; validator_index < $total_validators; validator_index++)); do
-  snarkos start --nodisplay --network $network_id --dev $validator_index --dev-num-validators $total_validators --validator --jwt-secret $jwt_secret --jwt-timestamp $jwt_ts &
+  ./target/release/snarkos start --nodisplay --network $network_id --dev $validator_index --dev-num-validators $total_validators --validator --jwt-secret $jwt_secret --jwt-timestamp $jwt_ts &
```
ljedrz: whoops, forgot to remove this; no worries, I can take care of it when I'm removing the temporary snarkVM branch

.circleci/db_backup_ci.sh
ljedrz: done - the test now goes through 2 checkpoints, and there's even a load from a 3rd one for good measure

vicsn: Sorry for the long back-and-forth, but I realise others will also comment on this (for good reason): could you remove this file and place them in the existing ci script appropriately? We're running that network anyway, we might as well check if the backup functionality stays unchanged.

If the existing CI script already restarts a node, you can make them restart from a backup. If not, you can just make them restart an extra time.

ljedrz: Do you mean `devnet_ci.sh`? I would recommend against merging them, for the following reasons:
- `devnet_ci` verifies that the basic functionality of maintaining consensus and producing blocks works for all networks, with additional metrics and full logs, and different validator/client counts
- `db_backup_ci` is specialized to verify that the db rollback functionality works in a single, minimal network, while _assuming_ regular runtime; it _could_ be extended to at the end wait until a specific block height is reached, but it would make it more complex
- I am no `bash` scripting expert, and wouldn't feel confident creating something this advanced, especially considering how essential the `devnet_ci` check is

Also, why do you believe it's desirable to merge them, other than to deduplicate some code? It's usually preferable to test different functionalities in separate tests, and this addition is neither especially long in runtime nor compute-intensive.

vicsn: Thank you, clear.

My intention was to let this be tested. But we have a huge test infra already, so I'm not keen to add yet another CI job. Let's leave it as is for now. Thank you.

vicsn:
This is ready for rebase and opening

ljedrz:
I stopped the `devnet` CI job, as it wasn't stopping on its own; let's see if it was a fluke.

Edit: yep, it was a fluke - there wasn't an issue locally either.

vicsn:
I'm wondering if we can get `.circleci/db_backup_ci.sh` in CI without having to spin up yet another CI job. Could we just run them one after the other? It shouldn't be an issue if state bleeds from one network to the other.
```
          name: "Run devnet test"
          timeout: 20m  # Allow 20 minutes total
          command: |
            echo "Starting db_backup_ci.sh"
             ./.circleci/db_backup_ci.sh << parameters >>
            echo "Starting devnet_ci.sh"
            ./.circleci/devnet_ci.sh << parameters.validators >> << parameters.clients >> << parameters.network_id >> << parameters.min_height >>
```

And to be clear, I think we should not migrate [to this other script](https://github.com/ProvableHQ/snarkOS/pull/3675) yet, we can evaluate that PR after this is through.

.circleci/db_backup_ci.sh
```diff
@@ -0,0 +1,166 @@
...
+    fi
+  done
+
+  echo "All nodes created a checkpoint"
+  return 0
+}
+
+# Wait for 15 seconds to let the network start
+echo "Waiting 15 seconds for network to start up..."
+sleep 15
```
vicsn, line 75: Does the script assume REST endpoints work after this sleep time?

Could we just use `curl` in a loop, which will both be faster and less error prone.

Similarly we should change the `Waiting 30 seconds for network to start up` in `devnet_ci.sh`

ljedrz:
> Could we just run them one after the other? It shouldn't be an issue if state bleeds from one network to the other.

Hmm bleeding state would be a problem, as the checkpoint script expects very specific block heights; it should be fine if `snarkos clean` is run first, though :thinking:.

That being said, database checkpoints are quite isolated from the other functionalities, so I wouldn't expect them to be fragile enough to require a test run after every PR.

vicsn:
> Hmm bleeding state would be a problem, as the checkpoint script expects very specific block heights; it should be fine if snarkos clean is run first, though 🤔.

Well I'm suggesting to run `db_backup_ci.sh` **first** before `devnet_ci.sh`. But doing `snarkos clean` at the beginning or end is also fine. Up to you.

ljedrz:
Seems to have worked just fine; a [tiny modification](https://github.com/ProvableHQ/snarkOS/pull/3671/commits/0394ea6cb1d3d1a00913652adf6263659c27b6c4), too.

vicsn (approved):

.circleci/db_backup_ci.sh
vicsn: We can tackle it later

ljedrz: Hmm, we would have to check all the nodes involved; also, the wait time, while crude, could reveal a legitimate slowdown here :thinking:.

PR #3672, opened 2025-06-03
[Feat] Optional constant JWT

ljedrz:
A revival of https://github.com/ProvableHQ/snarkOS/pull/3592. I've successfully tested it on a validator in a local devnet.

CC https://github.com/ProvableHQ/snarkOS/pull/3671

### DIFF
--- a/node/rest/src/helpers/auth.rs
+++ b/node/rest/src/helpers/auth.rs
@@ -96,8 +96,8 @@ pub async fn auth_middleware(request: Request<Body>, next: Next) -> Result<Respo
             }
         }
 
-        Err(_) => {
-            return Err(StatusCode::UNAUTHORIZED.into_response());
+        Err(err) => {
+            return Err((StatusCode::UNAUTHORIZED, err.to_string()).into_response());
         }
     }
 
--- a/node/rest/src/helpers/auth.rs
+++ b/node/rest/src/helpers/auth.rs
@@ -67,11 +67,6 @@ impl Claims {
         Self { sub: address.to_string(), iat: issued_at, exp: expiration }
     }
 
-    /// Returns true if the token is expired.
-    pub fn is_expired(&self) -> bool {
-        OffsetDateTime::now_utc().unix_timestamp() >= self.exp
-    }
-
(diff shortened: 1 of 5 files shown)


vicsn (commented):
Is there any way to sensibly unit test the behaviour of the endpoint? Do we want this? Noting that we do have a `node-rest` CI job which doesn't actually run any unit tests, because none are defined.

ljedrz:
I added a test to check that a JWT with consistent inputs results is constant.

vicsn:
Can you:
- confirm in the PR description that you tested that this new functionality works for a validator and client by running a local devnet?
- open an issue on AleoNet/welcome with the appropriate documentation needed for users on developer.aleo.org

ljedrz:
I've confirmed it to be constant with a validator in a devnet already, but I can also run some privileged requests.

Edit: done; I also made auth errors more informative, and enabled JWT for clients.

node/rest/src/helpers/auth.rs
```diff
@@ -80,16 +84,20 @@ pub async fn auth_middleware(request: Request<Body>, next: Next) -> Result<Respo
...
         Ok(decoded) => {
             let claims = decoded.claims;
             if claims.is_expired() {
                 return Err((StatusCode::UNAUTHORIZED, "Expired JSON Web Token".to_owned()).into_response());
             }
         }
 
-        Err(_) => {
-            return Err(StatusCode::UNAUTHORIZED.into_response());
+        Err(err) => {
```
vicsn: Hate to do one more nit, but I've seen it being good security practice not to return  internal errors to potential attackers. What about logging this locally instead of returning it?

vicsn (changes requested):

node/rest/src/helpers/auth.rs
ljedrz: That's reasonable; I'll make it into a `WARN`-tier log, perhaps, so that the server can have greater visibility in case of brute-force access attempts.

vicsn (approved):

PR #3674, opened 2025-06-06
[Fix] Handle duplicate TMUX session better and pass --metrics to first validator

kaimast:
After the recent change (#3619), `./devnet.sh` does not pass `--metrics` to the first validator anymore.

This PR fixes that, and also handles nested or duplicate TMUX sessions better.


bendyarm (dismissed):
LGTM

vicsn (commented):
Can we just enable metrics for all nodes? I think itll be useful for testing

kaimast:
I added a second commit that enables `--metrics` for all validators.

vicsn (approved):

PR #3675, opened 2025-06-08
[devnet] devnet-topo.sh is an alternative to devnet.sh that sets up a…

bendyarm:
Sets up a trusted peers topology that might better reflect mainnet.
Prevents validators from accepting external peer connections.
Otherwise mainly the same as `devnet.sh`.

### DIFF
--- a/devnet-topo.sh
+++ b/devnet-topo.sh
@@ -4,21 +4,23 @@
 
 # Like devnet.sh, but with these changes that stabilize the network topology:
 #
-# * Only asks for the number N of validators, not the number of clients.
-#   Makes the same number of clients as validators, to reflect
-#   the concept that each validator has has a dedicated core client.
-#   (Currently we are waiting for more information about how mainnet validators
-#   set up their trusted peers; this script can then be updated to match.)
+# * Asks for the number N of validators, not the number of clients.
+#   Asks how many core clients each validator should have.
+#   Makes the appropriate number of clients, to reflect
+#   the concept that each validator has 1 or 2 dedicated core clients.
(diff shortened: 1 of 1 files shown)


devnet-topo.sh
```diff
@@ -0,0 +1,192 @@
...
+# Loop through the list of validator indices and create a new window for each
+for validator_index in "${validator_indices[@]}"; do
+  # Generate a unique and incrementing log file name based on the validator index
+  name="validator-$validator_index"
+  log_file="$log_dir/$name.log"
+  window_index=$((validator_index + index_offset))
+  metrics_flag=""
+
+  if [ "$validator_index" -eq 0 ]; then
+    # Enable metrics only for the first validator.
```
kaimast: We can enable metrics for all validators, like we do in the devnet.sh script now.

devnet-topo.sh
```diff
@@ -0,0 +1,192 @@
...
+          if [ "$trusted_validators" = "" ]; then
+              trusted_validators="127.0.0.1:$((5000 + trusted_validator_index))"
+          else
+              trusted_validators="${trusted_validators},127.0.0.1:$((5000 + trusted_validator_index))"
+          fi
+      fi
+  done
+
+  # Create the list of peers (i.e., trusted peers) for the validator at the_validator_index
+  # It is the corresponding client now.
```
kaimast: Do you mean "it is only the corresponding client now"?

kaimast (dismissed):
Looks good! I ran the script locally and it seems to work as expected.

PR #3676, opened 2025-06-09
[CI] Add devnet test for syncing

kaimast:
This adds another test to CI that fetches ledger data from Google Cloud Storage, spawns a client from it, and then has other clients sync from the first client.

The goal is to have a quick check that loading state from storage works, and that basic syncing works and does not take unreasonable long. The test waits at most five minutes for the client to sync.

The test creates a chain of clients, so blocks to propagate through multiple clients to reach the final one. This ensures forwarding of blocks also works as expected.


acoglio (dismissed):
I'm not familiar with these scripts, but the PR looks reasonable to me.

kaimast:
There was a bug in the previous commit and CI failed. I pushed a fixed version. Maybe @bendyarm can review, as he is a little more familiar with the devnet script?

.circleci/config.yml
```diff
@@ -457,34 +488,89 @@ jobs:
...
 workflows:
   version: 2
 
   main-workflow:
     jobs:
-      - snarkos
+      # These do not run any tests
+      # - snarkos
+      # - node-sync-communication-service
+      # - node-rest
```
vicsn: A small test was added recently

.circleci/config.yml
```diff
@@ -184,260 +177,298 @@ commands:
...
     steps:
       - run: rustup toolchain install nightly-x86_64-unknown-linux-gnu
 
+  fetch_repository:
+    steps:
+      - attach_workspace:
+          at: /home/circleci
+
 jobs:
+  prepare-repository:
```
vicsn, line 307: What does this do? Why was it added?

.circleci/config.yml
```diff
@@ -55,29 +61,45 @@ commands:
             - target
           key: cargo-cache-{{ arch }}-{{ checksum "rust-version" }}-{{ checksum "Cargo.lock" }}
 
-  setup_environment:
-    description: "Setup testing environment"
+  prepare_environment:
```
vicsn: Are you bringing all these CircleCI refactors to snarkVM as well?

.circleci/config.yml
```diff
@@ -184,260 +177,298 @@ commands:
...
+    description: "Fetch the git repository and store it in the workspace"
+    steps:
+      - checkout
+      - run:
+          command: rm -rf .git
+      - persist_to_workspace:
+          root: /home/circleci
+          paths: project
+    
+  build-debug:
```
vicsn: What does this do? Why was it added?

vicsn (changes requested):
Couple nits and clarifications needed

vicsn:
@kaimast can this be closed?

kaimast:
Yeah we can close this.

PR #3677, opened 2025-06-10
[Test] Reduce sync test iterations to speed up CI

kaimast:
Sync tests currently take very long (over 15 minutes on CI). 

This PR makes two changes:
- ~Use release build on CI for sync tests~ (removed from the PR)
- Reduce the number of iterations from some of the tests (from 100,000 to 5,000).

For the latter, I wrote a helper that will ensure we always test the highest given height (usually 100,001) and randomly sampling another 4,999 values. This will still test the sync module with a similar rigor but much faster.

At a first glance, running `node-sync` now runs in less than 5 minutes. I observed this job to be the slowest on many post workflow runs, but sometime `bft`, `node-router-messages`, or the devnet test can be slower.

In the most recent run for this PR, `bft` took the longest (over 19 minutes). So, this change might not always reduce the overall execution time of CI, but will reduce cost.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -339,6 +339,7 @@ jobs:
     steps:
       - run_serial:
           workspace_member: node/sync
+          flags: --release
           cache_key: v3.3.1-rust-1.83.0-node-sync-cache
 
   node-sync-communication-service:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -339,7 +339,6 @@ jobs:
     steps:
       - run_serial:
           workspace_member: node/sync
-          flags: --release
           cache_key: v3.3.1-rust-1.83.0-node-sync-cache
 
   node-sync-communication-service:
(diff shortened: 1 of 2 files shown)


kaimast:
@raychu86 you wrote some of the tests, I think. Maybe you could chime in if this is a reasonable change?

.circleci/config.yml
```diff
@@ -339,6 +339,7 @@ jobs:
     steps:
       - run_serial:
           workspace_member: node/sync
+          flags: --release
```
vicsn: I'd remove to avoid the overhead of analysing whether this exhibits the same behaviour. We can tackle this some other time in a separate PR specific on updating stuff to release mode

vicsn (dismissed):
logic LGTM, minor nit

vicsn:
Can you also clarify in the PR message what the slowest CI job is before and after this PR

kaimast:
> Can you also clarify in the PR message what the slowest CI job is before and after this PR

sync sometimes was the slowest before, but it could also be bft or router-messages or the devnet test. In the most recent run for this PR, bft took the longest (over 19minutes).

vicsn (approved):

PR #3678, opened 2025-06-12
[Feature] Immediately send a Ping after generating new blocks

kaimast:
## Motivation
It can take many seconds for a client to detect that its transaction was confirmed, because block propagation is sometimes slow.
This needs to be fixed to improve the user experience.

## Overview
This PR introduces a `Ping` struct that keeps track of all ping messages (to be) sent to other peers. When a new block is received or created, nodes update the block locators in `Ping`, which immediately sends new ping messages.
This should reduce block propagation time, as the old approach introduces up to 20 seconds of delay for each network hop.

Every iteration, Ping's worker task waits on a `tokio::Notify` (similar to a condition variable), which gets notified whenever the block locators change. It also set a timeout to the earliest time we have to send a new ping message to a peer.
If there are a no peers, the task still wakes up every 20 seconds. The latter should not be needed, but I left it in place to make sure the task never gets stuck.

This change also moves the `Writer` implementation from `Validator`/`Client`/`Prover` to `Router`, which deduplicates some code and also makes the code more modular.

## Notes/Considerations
* This change means we now send a ping every block, so roughly every 4 seconds or so. This could cause issues, and we can consider changing it to not trigger a new ping message on every block. I also have ideas to make `BlockLocators` smaller so that Ping messages do not cause a noticeable overhead.
* I did not change the `Gateway`/`PrimaryPing` code. The `PrimaryPing` is really only used during initial sync. During normal operations, validators build blocks themselves and the pings do not affect latency.

## Result
Client block height is now identical to that of validators.
Screenshot 2025-06-22 at 6 14 34 PM

### DIFF
--- a/cli/src/helpers/logger.rs
+++ b/cli/src/helpers/logger.rs
@@ -55,12 +55,12 @@ pub fn initialize_logger<P: AsRef<Path>>(
     // Filter out undesirable logs. (unfortunately EnvFilter cannot be cloned)
     let [filter, filter2] = std::array::from_fn(|_| {
         let filter = EnvFilter::from_default_env()
-            .add_directive("mio=off".parse().unwrap())
-            .add_directive("tokio_util=off".parse().unwrap())
-            .add_directive("hyper=off".parse().unwrap())
-            .add_directive("reqwest=off".parse().unwrap())
-            .add_directive("want=off".parse().unwrap())
-            .add_directive("h2=off".parse().unwrap());
+            .add_directive("mio=warn".parse().unwrap())
+            .add_directive("tokio_util=warn".parse().unwrap())
+            .add_directive("hyper=warn".parse().unwrap())
+            .add_directive("reqwest=warn".parse().unwrap())
+            .add_directive("want=warn".parse().unwrap())
+            .add_directive("h2=warn".parse().unwrap());
 
         let filter = if verbosity >= 2 {
             filter.add_directive("snarkos_node_sync=trace".parse().unwrap())
@@ -91,7 +91,7 @@ pub fn initialize_logger<P: AsRef<Path>>(
         if verbosity >= 6 {
             filter.add_directive("snarkos_node_tcp=trace".parse().unwrap())
         } else {
-            filter.add_directive("snarkos_node_tcp=off".parse().unwrap())
+            filter.add_directive("snarkos_node_tcp=warn".parse().unwrap())
         }
     });
 
--- a/cli/src/helpers/logger.rs
+++ b/cli/src/helpers/logger.rs
@@ -54,13 +54,26 @@ pub fn initialize_logger<P: AsRef<Path>>(
 
     // Filter out undesirable logs. (unfortunately EnvFilter cannot be cloned)
     let [filter, filter2] = std::array::from_fn(|_| {
-        let filter = EnvFilter::from_default_env()
-            .add_directive("mio=warn".parse().unwrap())
-            .add_directive("tokio_util=warn".parse().unwrap())
-            .add_directive("hyper=warn".parse().unwrap())
-            .add_directive("reqwest=warn".parse().unwrap())
-            .add_directive("want=warn".parse().unwrap())
-            .add_directive("h2=warn".parse().unwrap());
+        let filter = EnvFilter::from_default_env();
+
+        // At high log levels, also show warnings of dependencies
+        let filter = if verbosity < 4 {
+            filter
+                .add_directive("mio=off".parse().unwrap())
+                .add_directive("tokio_util=off".parse().unwrap())
+                .add_directive("hyper=off".parse().unwrap())
+                .add_directive("reqwest=off".parse().unwrap())
+                .add_directive("want=off".parse().unwrap())
+                .add_directive("h2=off".parse().unwrap())
+        } else {
+            filter
+                .add_directive("mio=warn".parse().unwrap())
+                .add_directive("tokio_util=warn".parse().unwrap())
+                .add_directive("hyper=warn".parse().unwrap())
+                .add_directive("reqwest=warn".parse().unwrap())
+                .add_directive("want=warn".parse().unwrap())
+                .add_directive("h2=warn".parse().unwrap())
+        };
 
         let filter = if verbosity >= 2 {
             filter.add_directive("snarkos_node_sync=trace".parse().unwrap())
--- a/node/src/client/mod.rs
+++ b/node/src/client/mod.rs
@@ -28,10 +28,10 @@ use snarkos_node_router::{
     Routing,
     messages::{Message, NodeType, UnconfirmedSolution, UnconfirmedTransaction},
 };
-use snarkos_node_sync::{BLOCK_REQUEST_BATCH_DELAY, BlockSync};
+use snarkos_node_sync::{BLOCK_REQUEST_BATCH_DELAY, BlockSync, Ping};
 use snarkos_node_tcp::{
     P2P,
-    protocols::{Disconnect, Handshake, OnConnect, Reading, Writing},
+    protocols::{Disconnect, Handshake, OnConnect, Reading},
 };
 use snarkvm::{
     console::network::Network,
@@ -124,6 +124,8 @@ pub struct Client<N: Network, C: ConsensusStorage<N>> {
     handles: Arc<Mutex<Vec<JoinHandle<()>>>>,
     /// The shutdown signal.
     shutdown: Arc<AtomicBool>,
+    /// Keeps track of sending pings.
+    ping: Arc<Ping<N>>,
 }
 
 impl<N: Network, C: ConsensusStorage<N>> Client<N, C> {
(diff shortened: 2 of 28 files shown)


node/sync/src/ping.rs
```diff
@@ -0,0 +1,198 @@
...
+            tokio::spawn(async move {
+                Self::ping_task(&inner, &router, &notify).await;
+            });
+        }
+
+        Self { inner, router, notify }
+    }
+
+    /// Same as [`Self::new`] but for nodes that peers cannot sync from
+    /// such as provers.
```
vicsn, line 85: This surprises me, can you confirm this was the existing behaviour?

bendyarm (commented):
I did a quick review. This seems like a good idea.  I didn't have time to check all the details, though.   One thing I noticed is that, due to normal peer rotation, there might be too many messages `Attempted to send to a non-connected peer {peer_ip}` when attempting to send a recently-disconnected peer a ping.

acoglio (dismissed):

node/sync/src/ping.rs
```diff
@@ -0,0 +1,198 @@
...
+};
+use tokio::{sync::Notify, time::timeout};
+
+/// Internal state of the ping logic
+///
+/// Essentially, ping keeps an ordered map `next_ping` of time(rs) to peer IPs.
+/// When a new peer connects or a Pong message is received, an entry in next ping is created
+/// for when a peer should next be pinged.
+///
+/// TODO (kaimast): maybe keep track of the last ping too, to not trigger spam detection?
```
vicsn, line 35: Good one, though I assume typical honest behaviour will only result in at most2-3 messages in a burst, likely not enough for spam detection, but please double check

node/sync/src/ping.rs
```diff
@@ -0,0 +1,198 @@
...
+        inner.next_ping.insert(now + Self::MAX_PING_INTERVAL, peer_ip);
+
+        // self.notify.notify() is not needed as ping_task wakes up every MAX_PING_INTERVAL
+    }
+
+    /// Notify the ping logic that a new peer connected.
+    pub fn on_peer_connected(&self, peer_ip: SocketAddr) {
+        // Send the first ping.
+        let locators = self.inner.lock().block_locators.clone();
+        self.router.send_ping(peer_ip, locators);
```
vicsn: Should this attempt to reset any respective entry in `inner.next_ping`, i.e. set the max ping interval after pinging?

node/sync/src/ping.rs
```diff
@@ -0,0 +1,198 @@
...
+
+        loop {
+            // Do not hold the lock while waiting.
+            let sleep_time = {
+                let mut inner = inner.lock();
+                let now = Instant::now();
+
+                // Ping peers.
+                if new_block {
+                    Self::ping_all_peers(&mut inner, router);
```
vicsn, line 142: Should this attempt to reset any respective entry in `inner.next_ping`, i.e. set the max ping interval after pinging?

node/src/validator/mod.rs
```diff
@@ -72,6 +73,8 @@ pub struct Validator<N: Network, C: ConsensusStorage<N>> {
     handles: Arc<Mutex<Vec<JoinHandle<()>>>>,
     /// The shutdown signal.
     shutdown: Arc<AtomicBool>,
+    /// Keeps track of sending pings.
+    ping: Arc<Ping<N>>,
```
ljedrz, line 77: do the validators use these pings? it might be handled on the BFT side already

vicsn:
Please mention in the PR description how much this reduces syncing latency.

node/src/validator/mod.rs
kaimast: Yes, they ping clients using the same mechanism. `PrimaryPing` in BFT is only for talking to other validators.

node/sync/src/ping.rs
kaimast: The current implementation only schedules a new ping after receiving a pong. Should we change that?

If you put the peer back into `next_ping`, they will receive a ping as soon as the next block. I am a little worried that this might overburden slow clients as they might receive pings faster than they can respond.

node/sync/src/ping.rs
kaimast: If you look at the changed prover code, we remove the following lines:
```rust
        tokio::spawn(async move {
            // Sleep for the preset time before sending a `Ping` request.
            tokio::time::sleep(Duration::from_secs(Self::PING_SLEEP_IN_SECS)).await;
            // Check that the peer is still connected.
            if self_clone.router().is_connected(&peer_ip) {
                // Send a `Ping` message to the peer.
                self_clone.send_ping(peer_ip, None);
            }
        });
```

It set the argument for block locators to `None` when calling `send_ping`. This means it is not advertising any blocks to sync.

node/src/client/mod.rs
```diff
@@ -277,7 +284,14 @@ impl<N: Network, C: ConsensusStorage<N>> Client<N, C> {
         if block_requests.is_empty() && self.sync.has_pending_responses() {
             // Try to advance the ledger with the sync pool.
             trace!("No block requests to send, but there are still pending block responses.");
-            self.sync.try_advancing_block_synchronization();
+            let has_new_blocks = self.sync.try_advancing_block_synchronization();
```
vicsn, line 309: Is it worth / possible to call `try_block_sync` on receiving new block locators, instead of waiting a fixed 5 seconds?

node/sync/src/ping.rs
vicsn: > The current implementation only schedules a new ping after receiving a pong. Should we change that?

I just wanted to make sure that we avoid ping-spam. But indeed, if after every ping we assume to get a pong, and that pong will lead to resetting the 20 second default ping timer, that works!

>  I am a little worried that this might overburden slow clients as they might receive pings faster than they can respond.

Correct me if I'm wrong, my understanding is that the clients have the following processing to do:
- send a pong, this is light-weight
- block syncing, which is an independent process anyway. If however we'd [link block syncing](https://github.com/ProvableHQ/snarkOS/pull/3678#discussion_r2150811697), then we might need to avoid triggering sync while sync is still in progress

node/src/client/mod.rs
kaimast: Yes, that is another thing I was thinking of doing. We should trigger try_advancing_block_synchronization, not at a fixed interval but on specific events. It should just be the following
* New block locators were received
* All outstanding block requests were processed
* A peer disconnects (can affect outstanding block requests)

We can still keep the timer as a failsafe. I thought that might be better as a separate PR, but I can add some of it to this one.

kaimast:
Noticed some issues during stress testing. I will keep the PR a draft until everything looks good.

kaimast:
I added a few things to make debugging the issues I saw on testnet easier, such as actually printing logs that happen in TCP. If the PR is too complex now, I can also move the last few commits to a separate PR.

node/src/client/mod.rs
vicsn: Made an issue for it https://github.com/ProvableHQ/snarkOS/issues/3691

cli/src/helpers/logger.rs
```diff
@@ -55,12 +55,12 @@ pub fn initialize_logger<P: AsRef<Path>>(
     // Filter out undesirable logs. (unfortunately EnvFilter cannot be cloned)
     let [filter, filter2] = std::array::from_fn(|_| {
         let filter = EnvFilter::from_default_env()
-            .add_directive("mio=off".parse().unwrap())
-            .add_directive("tokio_util=off".parse().unwrap())
-            .add_directive("hyper=off".parse().unwrap())
-            .add_directive("reqwest=off".parse().unwrap())
-            .add_directive("want=off".parse().unwrap())
-            .add_directive("h2=off".parse().unwrap());
+            .add_directive("mio=warn".parse().unwrap())
```
vicsn: Would it make sense to only turn these into warns for higher verbosity levels? Is that possible?

kaimast: Yeah, I can do that. I still think we should at least keep it at error, because sometimes we don't log errors in snarkOS and it is unclear what happened.

node/src/client/mod.rs
```diff
@@ -480,7 +523,7 @@ impl<N: Network, C: ConsensusStorage<N>> NodeInterface<N> for Client<N, C> {
         self.shutdown.store(true, std::sync::atomic::Ordering::Release);
 
         // Abort the tasks.
-        trace!("Shutting down the validator...");
+        trace!("Stopping backgroudn tasks...");
```
vicsn: Please keep this logline unchanged, for better or worse our prerelease tests depend on it. Consider adding a comment above so we don't change it in the future.

kaimast: Should the message say `validator` for the client too? I would at least change that one.

cli/src/helpers/logger.rs
kaimast: I set it to off for anything lower than 4, and then warn for 4 and above.

node/src/client/mod.rs
vicsn: Good one, yes can you change it to `Shutting down the client`, thx!

vicsn (approved):

PR #3679, opened 2025-06-13
[Build] Force lower proptest version

kaimast:
`proptest` 1.6 seems to be the last version that uses `rand` 0.8. Upgrading to a newer version causes issues for me. 

This PR keeps it at 1.6 until we upgrade snarkOS to `rand` 0.9. I also ran `cargo update` to update the lock file.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -13,9 +13,9 @@ dependencies = [
 
 [[package]]
 name = "adler2"
-version = "2.0.0"
+version = "2.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "512761e0bb2578dd7380c6baaa0f4ce03e84f95e960231d1dec8bf4d7d6e2627"
+checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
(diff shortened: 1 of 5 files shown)


vicsn (changes requested):
Thank you, I think I noticed this in one of my compilations.

I have a slight preference for updating rand now already. In general if we experience dependency issues, we might as well try and upgrade as far as possible.

ljedrz:
On a related note, we will want to update `bincode` to `2.0` sometime soon as well.

kaimast:
> I have a slight preference for updating rand now already. In general if we experience dependency issues, we might as well try and upgrade as far as possible.

That would be great. Updating to rand might just be a longer process, because it is so widely used in snarkVM and touches some of the cryptography code if I understand correctly.

kaimast:
> On a related note, we will want to update `bincode` to `2.0` sometime soon as well.

I am happy to review a PR for that. I assume bincode 1.* will be maintained for the foreseeable future, though.

vicsn (approved):
Thx, I see the point, merging this small change so people don't run into the same issue you faced.

PR #3682, opened 2025-06-16
[Feature] Add sync_status REST endpoint

kaimast:
This adds a `<network>/sync_status` endpoint to the REST API. A call to it will return something like the following:
```json
~> curl http://localhost:3030/mainnet/sync_status
{
  "is_synced": false,
  "ledger_height": 2443,
  "sync_mode": "p2p",
  "cdn_height": 8167950,
}
```

`sync_mode` will switch to `p2p` once the initial sync is done. 

If the node is not connected to a CDN, the `cdn_height` field is omitted.
```json
~> curl http://localhost:3030/mainnet/sync_status
{
  "is_synced": false,
  "ledger_height": 2795,
  "sync_mode": "p2p",
  "p2p_height": 8638987
}
```

p2p height is only added to the response once a peer is connected

### DIFF
--- a/node/rest/src/routes.rs
+++ b/node/rest/src/routes.rs
@@ -42,14 +42,21 @@ pub(crate) struct Metadata {
     all: Option<bool>,
 }
 
-/// The return value for a `status` query.
+/// The return value for a `sync_status` query.
 #[skip_serializing_none]
 #[derive(Copy, Clone, Serialize)]
-struct NodeStatus<'a> {
+struct SyncStatus<'a> {
+    /// Is this node fully synced with the network?
     is_synced: bool,
+    /// The block height of this node.
     ledger_height: u32,
+    /// Which way are we sync'ing (either "cdn" or "p2p")
     sync_mode: &'a str,
+    /// The block height of the CDN (if connected to a CDN).
     cdn_height: Option<u32>,
+    /// The greatest known block height of a peer.
+    /// None, if no peers are connected yet.
+    p2p_height: Option<u32>,
 }
 
 impl<N: Network, C: ConsensusStorage<N>, R: Routing<N>> Rest<N, C, R> {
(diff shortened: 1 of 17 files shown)


vicsn:
Just had a chat with infra folks about how we can monitor CDN syncing. How about we return the following data:
```
{
  "is_synced": true
  "sync_mode": p2p/cdn/none,
  "cdn_height": 2, // 0 if we're not syncing from CDN currently.
  "ledger_height": 1
}
```

~Noting my DRY obsession, maybe we can:~
~- leave out the ledger height~
~- rename the endpoint to `sync_status`. In the future we can optionally add info about this node's view of block locators~

kaimast:
I updated the PR to include a change that will start the CDN connection before the REST server. This required some refactoring, so the change set is larger now, but allows querying the state of CDN sync.

Now, I wonder if we should also include the maximum peer height, so you can see how far along p2p sync is? 

Finally, the endpoint is simply called `status`, but something like `sync_status` might be more accurate?

vicsn:
> Now, I wonder if we should also include the maximum peer height, so you can see how far along p2p sync is?

Yes that would be a nice and symmetric alternative to `cdn_height`

> Finally, the endpoint is simply called status, but something like sync_status might be more accurate?

Indeed I prefer `sync_status`

node/rest/src/routes.rs
```diff
@@ -41,6 +42,16 @@ pub(crate) struct Metadata {
     all: Option<bool>,
 }
 
+/// The return value for a `status` query.
+#[skip_serializing_none]
```
vicsn, line 46: Very nice

node/rest/src/routes.rs
```diff
@@ -117,6 +128,26 @@ impl<N: Network, C: ConsensusStorage<N>, R: Routing<N>> Rest<N, C, R> {
         }
     }
 
+    // GET /<network>/status
+    pub(crate) async fn get_status(State(rest): State<Self>) -> Result<ErasedJson, RestError> {
+        // Get the CDN height (if we are connected to a CDN)
+        let cdn_height =
+            if let Some(cdn_sync) = &rest.cdn_sync { Some(cdn_sync.get_cdn_height().await?) } else { None };
```
vicsn: To prevent confusion, should we atomically determine both `cdn_height` and `sync_mode` based on `cdn_sync`? Using the `is_done()` call as you do below.

kaimast:
Should be ready to merge now.

vicsn (approved):

PR #3690, opened 2025-06-23
Add default scan endpoint

vicsn:
## Motivation

To make life easier.

## Test Plan

Manually tested.


kaimast (approved):
Looks good. Did it just use an empty string before if `--endpoint` wasn't set?

PR #3694, opened 2025-06-26
[Optimize] Reduce Network Distribution of Invalid Solutions

raychu86:
## Motivation

This PR reduces the potential number of solutions send around on the network. This is not needed from a safety perspective, because snarkVM has the final underlying check to abort solutions that exceed the ARC-0046 limit.

If a prover has already reached his solution limit according to ARC-0046, then there is no need to further verify and broadcast subsequent solutions from that prover. This will reduce the number of solutions that land in subdags (if we already know the solution will be aborted). The tradeoff is that there may be more solution error logs seen from the node operators due to this new condition.


vicsn (dismissed):
Very good idea

vicsn:
Fix conflict

vicsn (approved):

PR #3695, opened 2025-06-27
[Feature] Process transactions based on the priority fee

niklaslong:
This PR introduces a priority queue for transactions containing a nonzero priority fee. These are drained before the pre-existing ready queue by the worker when building batches. 

A few notes on the current design:
1. Priority fees for transactions already in the mempool cannot be updated.
2. There is no protection against starvation: if the priority pool stays full enough, transmissions from the ready queue will not be included in the batches.
3. Batch building is **not** an atomic operation, the queues can be updated in the background.

### DIFF
--- a/node/bft/src/worker.rs
+++ b/node/bft/src/worker.rs
@@ -24,7 +24,7 @@ use crate::{
 };
 use snarkos_node_bft_ledger_service::LedgerService;
 use snarkvm::{
-    console::{prelude::*, types::U64},
+    console::prelude::*,
     ledger::{
         block::Transaction,
         narwhal::{BatchHeader, Data, Transmission, TransmissionID},
@@ -238,7 +238,7 @@ impl<N: Network> Worker<N> {
         {
             let Ok(priority_fee) = transaction.priority_fee_amount() else { return };
 
-            if priority_fee != U64::zero() {
+            if !priority_fee.is_zero() {
                 self.ready_priority.write().insert(transmission_id, transmission, priority_fee);
                 return;
             }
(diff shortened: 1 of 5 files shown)


node/bft/src/worker.rs
```diff
@@ -58,6 +58,8 @@ pub struct Worker<N: Network> {
     proposed_batch: Arc<ProposedBatch<N>>,
     /// The ready queue.
     ready: Arc<RwLock<Ready<N>>>,
+    /// The priority queue.
+    ready_priority: Arc<RwLock<ReadyPriority<N>>>,
```
vicsn, line 62: Architecture question: why duplicate `Ready` instead of duplicating `transactions_queue`?

I see some downsides to duplicating `Ready`:
- Lots of extra lines of code
- Transactions in the `transactions_queue` (of size `1 << 10`) are processed without priority in mind, meaning when there are a lot of transactions, the priority fee will barely accelerate landing your transaction past the ready queues (each of size `100`).

A benefit of using `ready_priority` is that we can pop and push from it, prioritizing transactions with priority fees, but if we assume `ready` was populated with a good ordering in mind, and because it is so small anyway, it doesn't really matter.

niklaslong: Great point—I went straight to the BFT thinking the workers held the primary queues of the mempool... What processing is the first queue guarding against, I'm guessing verification? The expensive deserialisation happens when the message is first processed by the router. 

I _think_ it would be easy to move up the stack (as the fee is accessible), I'll draft that version as well.

Issue #3646, opened 2025-05-21
[Feature] Improve log message on refused connections


kaimast:
When peers refuse a connection (e.g., because they have reached their max connectivity) the following error is shown.

```
WARN Unable to connect to '127.0.0.1:4135' - '127.0.0.1:4135' disconnected before sending "Message::ChallengeResponse"
```

This message is much more opaque than it should be. Ideally, it would print something like `127.0.0.1:4135 refused connection because it already reached its maximum number of peers`.

To fix this, we need to change how the network layer handles disconnects, so it might make sense to tackle this and #3645 together.

ljedrz:
The maximum number of peers can be controlled at the low-level (`tcp` module) layer, and the high-level layer (the `Heartbeat` impl). While we couldn't provide extra context at the lower layer, we might be able to do so higher up, during the handshake (which is ultimately where this is being triggered, as the higher-level connection limit is lower).

This isn't related to how disconnects happen (at least not the "usual" ones, for the fully connected peers) - it's at the handshake level, where we could handle a response indicating the reason for connection refusal. I'll see if this could leverage the existing `DisconnectReason` object.

Edit: we actually already allow for this to happen, but I don't see the higher-level limit being used during the handshake; I'll need to dig deeper.

ljedrz:
Ah, I see now: the low-level maximum number of peers for both `Router` and the `Gateway` is inherited the from the higher layer:
- for `Router` it's `Heartbeat::MAXIMUM_NUMBER_OF_PEERS`
- for `Gateway` it's `Committee::<N>::max_committee_size`

Meaning the connection limit check happens as soon as possible, before even going to the handshake implementation. It was implemented that way for performance reasons - deferring this check to the handshake impl (by e.g. setting the low-level connection limit to a higher value) would be slower.

If we want to change this, my recommendation would be to set the lower-level connection limit to something only a little bit higher than the higher-level one (say, 1 or 2 connections more), and to add an additional limit check in the handshake impl, which would allow the node to send a `Disconnect` message/event to the peer who initiated the connection, leading to them seeing the reason.

Sounds good, or are we ok with the lower log clarity being rewarded with extra performance?

vicsn:
Thanks for the extra analysis.

Given that the handshake refusal may also happen for other reasons (specifically a node going offline, anything else?), I think perhaps it is sufficient for now to just change the log message to sth like "disconnected from X during handshake, possible reasons include them not being able to accept more peers or shutting down".

[...]

PR #3696, opened 2025-06-27
Remove debug ERRORs from the default verbosity

meddle0x53:
## Motivation

We've been testing snarkOs a lot and our tests are trying to detect problems with this new verbosity change we are getting a lot of ERROR logs of the type:

```
2025-06-27T08:51:53.757689Z ERROR tcp{name=“0”}: handshake with 18.237.141.71:4130 failed: ‘18.237.141.71:4130’ disconnected before sending “Message::ChallengeResponse”
```

So we talked with @vicsn to set the logging back to `off` when the verbosity is not `>= 6`.

## Test Plan

Can be seen by running `devnet.sh` without this fix and with it.

### DIFF
--- a/cli/src/helpers/logger.rs
+++ b/cli/src/helpers/logger.rs
@@ -104,7 +104,7 @@ pub fn initialize_logger<P: AsRef<Path>>(
         if verbosity >= 6 {
             filter.add_directive("snarkos_node_tcp=trace".parse().unwrap())
         } else {
-            filter.add_directive("snarkos_node_tcp=warn".parse().unwrap())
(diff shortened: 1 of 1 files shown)


vicsn (approved):

kaimast:
There are a few cases where we do not log refused connections in snarkOS proper, which is why I enabled them. I will try to tackle #3646 soon and improve disconnect / PeerRefresh log messages.

I think this warning is also an indicator that something about peering is going wrong. If a node continuously tries to connect to a peer and gets refused, either the setup is wrong or the heartbeat logic is too aggressive.

PR #3697, opened 2025-06-27
[Feature] process transactions based on the priority fee

niklaslong:
This PR introduces a priority queue for transactions containing a nonzero priority fee. This is drained before the pre-existing transactions queue when sending transactions to the BFT.

A few notes on the current design:

1. Priority fees for transactions already in the mempool cannot be updated.
2. There is no protection against starvation: if the priority pool stays full enough, transmissions from the standard queue will not be sent to the BFT.
3. Batch building is **not** atomic w.r.t. the memory pool. 

Supersedes #3695.


node/consensus/src/lib.rs
```diff
@@ -418,6 +516,8 @@ impl<N: Network> Consensus<N> {
             let capacity = Primary::<N>::MAX_TRANSMISSIONS_TOLERANCE.saturating_sub(num_unconfirmed_transmissions);
             // Acquire the lock on the transactions queue.
             let mut tx_queue = self.transactions_queue.lock();
+            // Acquire the lock on the priority queue.
+            let mut priority_queue = self.priority_queue.lock();
```
vicsn: Need to also use this to determine num_deployments, num_executions

node/consensus/src/lib.rs
```diff
@@ -89,6 +94,80 @@ impl<N: Network> Default for TransactionsQueue<N> {
...
+            self.executions.put(transaction_id, transaction);
+        }
+    }
+}
+
+struct PriorityQueueInner<N: Network> {
+    /// A counter to ensure fifo ordering for transmissions with the same fee.
+    counter: u64,
+    /// A map of transmissions ordered by fee and by fifo sequence.
+    transaction_ids: BTreeMap<(Reverse<U64<N>>, u64), N::TransactionID>,
```
vicsn: TransactionsQueue contains an LRUCache to prevent OOM.

So during insert we should limit the maximum size, and think about how to deal with excess transactions. 

Is it possible to pop the transaction with the lowest priority fee into the TransactionsQueue? That would be awesome, because then we wouldn't have to worry about whether our PriorityQueue has some optimal maximum capacity, it'll effectively be shared among the two queues.

node/consensus/src/lib.rs
```diff
@@ -89,6 +94,80 @@ impl<N: Network> Default for TransactionsQueue<N> {
...
+            self.deployments.put(transaction_id, transaction);
+        } else {
+            self.executions.put(transaction_id, transaction);
+        }
+    }
+}
+
+struct PriorityQueueInner<N: Network> {
+    /// A counter to ensure fifo ordering for transmissions with the same fee.
+    counter: u64,
```
vicsn: IMO we can get rid of the counter and not care about fifo in case priority fees are the same. To reduce complexity and improve maintainability.

This would only be relevant if the PriorityQueue is full for extended periods of time, which I think will be unlikely in the near term.

node/consensus/src/lib.rs
```diff
@@ -89,6 +94,80 @@ impl<N: Network> Default for TransactionsQueue<N> {
...
+    /// A map of transmissions ordered by fee and by fifo sequence.
+    transaction_ids: BTreeMap<(Reverse<U64<N>>, u64), N::TransactionID>,
+    /// A map of transmission IDs to transmissions.
+    transactions: HashMap<N::TransactionID, Transaction<N>>,
```
vicsn: Is it possible to use ((priority_fee, txid) transaction) directly in the BTreeMap?

Then we only need a single container and we can get rid of this wrapper.

I'm assuming the difference in retrieval speed is negligible (walking down the tree v.s. hashing), its just a bit simpler.

node/consensus/src/lib.rs
```diff
@@ -389,15 +489,13 @@ impl<N: Network> Consensus<N> {
...
             }
-            // Add the transaction to the memory pool.
-            trace!("Received unconfirmed transaction '{}' in the queue", fmt_id(transaction_id));
-            if transaction.is_deploy() {
-                if self.transactions_queue.lock().deployments.put(transaction_id, transaction).is_some() {
-                    bail!("Transaction '{}' exists in the memory pool", fmt_id(transaction_id));
-                }
-            } else if self.transactions_queue.lock().executions.put(transaction_id, transaction).is_some() {
+            // Check that the transaction is not in the mempool.
+            if self.contains_transaction(&transaction_id) {
```
vicsn, line 381: Any chance we can get rid of contains, by letting insert return a boolean if it was already present? Similar to the pre-existing behaviour for TransactionsQueue? That way we only have to lock once.

vicsn (changes requested):
Very nice! Do we have any existing unit tests for transactions_queue which we can extend for the priority_queue? If not, I'm happy to run a test on devnet.

node/consensus/src/lib.rs
niklaslong: I did consider this but I wanted to minimise the potential for MEV with tx id crafting and maximise fairness. If that's not a concern, happy to remove it.

node/consensus/src/lib.rs
niklaslong: Retrieval would be O(log(n)) with just the `BTreeMap` vs O(1) with the `HashMap`, which is relevant for inclusion checks at the moment. Also, the keys in the tree need to implement `Ord` and `TransactionID` currently does not afaik?

node/consensus/src/lib.rs
niklaslong: The behaviour of the `TransactionsQueue` was to return an `Option` on insert: `Some` if an existing item was updated. It would also bail in that case, which makes me think the implementation contained a bug? 

If we don't keep the contains call, I'd suggest we return a bool on insert and bail if it returns `false` i.e a similar API to `HashSet` with no overwritten value.

node/consensus/src/lib.rs
niklaslong: > happy to remove it.

Ah but I don't think can entirely remove it and use a `BTreeSet` as `TransactionID` doesn't implement `Ord`. We need a tiebreaker for the same fee value.

node/consensus/src/lib.rs
niklaslong: We could but I'm hesitant to mix the two queues for simplicity's sake, I do see your point though 🤔

node/consensus/src/lib.rs
niklaslong: fd4317e

node/consensus/src/lib.rs
vicsn: > The behaviour of the TransactionsQueue was to return an Option on insert: Some if an existing item was updated. It would also bail in that case, which makes me think the implementation contained a bug?

Bug is a strong word (but unfortunately we only have a vocabulary of size 2: "bug" and "feature"). :)

But yes we should not have `bail`'d, just returning `Ok(())` would be been fine!

node/consensus/src/lib.rs
vicsn: We could serialize into a string, but yeah at that point the performance comparison becomes unclear, we can leave it as is, thx for the pushback

node/consensus/src/lib.rs
vicsn: We could serialize into a string, and do we really need a tiebreaker? But we can leave it as is, thx for the pushback

node/consensus/src/lib.rs
niklaslong: I see! Should I update the implementation in this PR to allow overwriting transactions in the mempool (I removed it thinking it was not desired)?

vicsn: > I see! Should I update the implementation in this PR to allow overwriting transactions in the mempool (I removed it thinking it was not desired)?

It's probably fine, but you can still avoid it by 1. taking the lock 2. calling `contains_key` 3. if it doesn't contain, then insert it.

A related nit: executions are more common than deployments, so if you want to short_circuit, try executions first, or just use `is_deploy` like below: https://github.com/ProvableHQ/snarkOS/pull/3697/files#diff-8b822371aae41d2275688c666d4c5d5d304e67aa422628ee42c405b5e8d7f572R162

niklaslong: > 1. taking the lock 2. calling contains_key 3. if it doesn't contain, then insert it.

Yep, this is the current approach (though I could reuse the lock if we don't want to move the inclusion check to the top of the method). I've also pushed 59e6738 to check for executions before deployments.

vicsn: As far as I can tell, at the time of writing, the code is trying to acquire a lock twice: once for `contains` and once for `insert`, which is suboptimal.

Can we change `priority_queue` to be a `RwLock`? This wasn't possible for the `LRUCache` but it should be possible for this one.

node/consensus/src/lib.rs
niklaslong: Still considering a few options but I'll probably go with:
1. One global cap for both queues (i.e. a memory pool capacity).
2. Grow any queue as needed but evict only from the low priority queue. 
3. If only the priority queue is full when the cap is reached, reject transactions until the node has room to process new ones. We could also consider evicting the lowest priority fee in the queue to let new transactions with higher fees through—likely not an immediate concern.

vicsn: Sounds good!

niklaslong: I've pushed a new version of the transaction queue that does what is described above. I still need to add tests but a local devnet seems to run successfully (though that won't exercise the limits or the queue ordering).

node/consensus/src/lib.rs
niklaslong: [31677fd](https://github.com/ProvableHQ/snarkOS/pull/3697/commits/31677fdf2258ee79f6ecc5d3b67272fce8c6ee07)

niklaslong: Note: I've kept it simple with one `RwLock` over the collection of executions and deployments. It would be easy enough to decouple this in future if need be.

node/consensus/src/transactions_queue.rs
```diff
@@ -0,0 +1,188 @@
...
+    pub fn len(&self) -> usize {
+        self.queue.len().saturating_add(self.priority_queue.len())
+    }
+
+    fn contains(&self, transaction_id: &N::TransactionID) -> bool {
+        self.queue.contains(transaction_id) || self.priority_queue.transactions.contains_key(transaction_id)
+    }
+
+    fn insert(&mut self, transaction_id: N::TransactionID, transaction: Transaction<N>) -> Result<()> {
+        let priority_fee = transaction.priority_fee_amount()?;
```
vicsn: Cleanup nit: you're only using `let has_priority_fee = transaction.priority_fee_amount()?.is_zero()` below.

vicsn (commented):
Main logic LGTM!

Can you add tests still? https://github.com/ProvableHQ/snarkOS/pull/2977/files#diff-8b822371aae41d2275688c666d4c5d5d304e67aa422628ee42c405b5e8d7f572R485

node/consensus/src/transactions_queue.rs
```diff
@@ -0,0 +1,342 @@
...
+impl<N: Network> TransactionsQueue<N> {
+    pub fn contains(&self, transaction_id: &N::TransactionID) -> bool {
+        self.executions.contains(transaction_id) || self.deployments.contains(transaction_id)
+    }
+
+    pub fn insert(
+        &mut self,
+        transaction_id: N::TransactionID,
+        transaction: Transaction<N>,
+        priority_fee: U64<N>,
```
niklaslong, line 51: Note: I now include the priority fee as an input to `insert` to make testing easier. Transaction sampling helpers in VM don't currently support specifying a fee's value so when that eventually gets included we can fully encapsulate the fee retrieval from the transaction in the queue logic.

vicsn (approved):
LGTM! We can merge next week after canary 4.0.0 has branched off

node/consensus/src/transactions_queue.rs
```diff
@@ -0,0 +1,354 @@
...
+    }
+
+    pub fn pop(&mut self) -> Option<(N::TransactionID, Transaction<N>)> {
+        self.priority_queue.pop().or_else(|| self.queue.pop_lru())
+    }
+}
+
+struct PriorityQueue<N: Network> {
+    /// A counter to ensure fifo ordering for transmissions with the same fee.
+    counter: u64,
```
raychu86, line 140: Would using a unix timestamp be easier to manage than a rolling counter?

node/consensus/src/transactions_queue.rs
```diff
@@ -0,0 +1,354 @@
...
+        if self.transaction_ids.is_empty() {
+            return;
+        }
+
+        // If the lowest fee in the collection is higher than the new fee, no-op.
+        //
+        // SAFETY: the empty check guarantees an item will be returned
+        let ((Reverse(lowest_fee), _), _) = self.transaction_ids.last_key_value().expect("item must be present");
+        if lowest_fee > &fee {
+            return;
```
raychu86, line 181: In this case why is it not an error? You technically tried to add a TX into the priority but it didn't happen.

node/consensus/src/transactions_queue.rs
```diff
@@ -0,0 +1,354 @@
...
+            self.transaction_ids.insert((Reverse(fee), self.counter), transaction_id);
+            // Increment the counter.
+            self.counter += 1;
+        }
+    }
+
+    fn compare_insert(&mut self, transaction_id: N::TransactionID, transaction: Transaction<N>, fee: U64<N>) {
+        // Make sure the collection isn't empty.
+        if self.transaction_ids.is_empty() {
+            return;
```
raychu86, line 173: Same as the other case. Why not error here? The TX was not actually added to the queue.

raychu86:
Changes look great!

Couple questions:

1. Starvation is quite easy to pull off by an attacker who fills the priority queue with `priority_fee`'s of only 1 microcredit and prevent transactions with no `priority_fee` to ever be included in a batch?.

In theory we could use a weighted algorithm based on time to equate how long the transactions were waiting and how much (if any) priority fee was included. This might be left for future work depending on scope.

2. Are we not trying to pack the batch with as close to the batch propose limit as possible? I recall we had iterated on a solution where we used an iterator and took the top txes (based on priority fee) to fill. And if the tx `execution_cost` was too big, we moved on to the next item in the list.

vicsn:
> Starvation is quite easy to pull off by an attacker who fills the priority queue with priority_fee's of only 1 microcredit and prevent transactions with no priority_fee to ever be included in a batch?.

Isn't this issue already present, i.e. our mempool simply has a limited size.

> Are we not trying to pack the batch with as close to the batch propose limit as possible? I recall we had iterated on a solution where we used an iterator and took the top txes (based on priority fee) to fill. And if the tx execution_cost was too big, we moved on to the next item in the list.

Thx for the reminder. I think we may want to cache the `execution_cost` in the `Ready` queue for this, and I'd leave this as an issue to open after this PR lands.

node/consensus/src/transactions_queue.rs
niklaslong: I chose a logical clock because a unix timestamp only has a precision of 1 second—likely not granular enough to guarantee ordering when there are a lot of inbound transactions?

node/consensus/src/transactions_queue.rs
niklaslong: This check is unnecessary with the current logic, if `compare_insert` is called, it is assumed the priority queue is already full and a swap needs to occur. I'd be happy to add an error but perhaps a debug assertion or a `WARN` log would be more suitable as it isn't a public API?

node/consensus/src/transactions_queue.rs
niklaslong: Indeed, though I don't think this is an error case but expected behaviour for a memory pool that is full (and designed to guard against DoS) but we could perhaps emit a `WARN` log?

niklaslong:
> Starvation

This is a race to the bottom (or top, depending on how you look at it): one can always outbid another if the notion of "priority" exists in the mempool. Granted, the floor is set by us, though paying vs non-paying seems reasonable to me? 

> Are we not trying to pack the batch with as close to the batch propose limit as possible

This PR is only meant to introduce priority fee ordering in the initial enqueueing of transactions into the workers. Batch packing will require a little more thought down in the ready queue, though it should be possible with minimal interference in the current logic.

node/consensus/src/transactions_queue.rs
raychu86: I believe our previous logic would error if the transaction failed to insert. Can you confirm if that is correct?

I think a warn log could be useful, but also we should be explicit regarding what happened to the TX. Without the error, I imagine the TX will be less "trackable".

niklaslong: I think we didn't error on the cache call itself, i.e. we were using [`LruCache::put`](https://docs.rs/lru/latest/lru/struct.LruCache.html#method.put) which just evicts the oldest value when an insert is performed at capacity.

vicsn (approved):
Very nice! @niklaslong can you still open a new issue regarding smarter packing?

niklaslong:
Aye, it's on my list!

PR #3698, opened 2025-06-27
Limit number of outstanding block requests

kaimast:
So far, the only limit on outstanding block requests is how often `prepare_block_requests` is called. 

This PR changes `BlockSync` to always have at most `MAX_BLOCK_REQUESTS` (set to 50 right now) outstanding requests. This should lighten the load on the network and increase sync performance.


vicsn (approved):
Looks solid

vicsn:
40 validator devnet passes with flying colors.

PR #3699, opened 2025-06-28
[Chore] Upgrade to rust 1.88

kaimast:
This PR changes the Rust version to 1.88 and changes most crates to use the 2024 edition of Rust. I am marking this as draft, because I will rebase it after the 4.0.0 postrelease merge.

There are some open tasks
* Some crates use `std::env::set_var` which is unsafe in 2024 
* I needed to change quite a few instances of `rng.gen()` to `rng.r#gen()` because `gen` is a reserved keyword now. This is just temporary until we upgrade the `rand` crate.

The PR also cleans up the CircleCI file slightly to make future upgrades easier.


kaimast:
Looks like we have to wait until CircleCI releases an image for Rust 1.88 anyway.

ljedrz (approved):
Not a CircleCI config expert, but the Rust bits look fine to me :+1:.

Cargo.toml
```diff
@@ -167,6 +167,7 @@ locktick = [
   "snarkos-node-sync/locktick",
   "snarkos-node-tcp/locktick"
 ]
+serial = [ "snarkos-cli/serial" ]
```
vicsn, line 170: Should we just let this also enable `serial` in snarkVM (for easy use in flamegraphing, debugging, and anything else really). Do we have any usecase for `serial` in snarkOS but not in snarkVM? As a developer, we can always adjust it, its just about what we'd like to have as default.

CC @niklaslong , this also applies to other services depending on snarkVM

vicsn (approved):
Thank you!!!

Left one comment for future thinking

PR #3704, opened 2025-07-01
[Fix] Update sync height in bft::Sync correctly

kaimast:
I noticed two potential issues in `snarkos_node_bft::Sync`.

First, when performing the long-range (non-BFT) sync, it would not update the sync height and, thus, block sync would try to sync older blocks that we already received.

Second, there seems to be a case where it would switch to BFT sync even if the node is still behind more than GC blocks. This could only happen if you received a block right after a non-BFT sync, so it might have never been triggered in production.

The PR aims to fix both of these issues. It is very short and just changes the behavior of `Sync::try_advancing_block_synchronization` slightly, so hopefully it can get merged quickly.

### DIFF
--- a/node/rest/Cargo.toml
+++ b/node/rest/Cargo.toml
@@ -19,14 +19,13 @@ edition = "2021"
 [features]
 default = [ "parallel" ]
 parallel = [ "rayon" ]
-history = [ "snarkvm-synthesizer/history" ]
+history = [ "snarkvm/history" ]
 telemetry = [ "snarkos-node-consensus/telemetry" ]
 cuda = [ "snarkvm/cuda", "snarkos-node-consensus/cuda", "snarkos-node-router/cuda" ]
 locktick = [
   "dep:locktick",
   "snarkos-node-consensus/locktick",
   "snarkos-node-router/locktick",
-  "snarkvm-synthesizer/locktick"
 ]
 
 [dependencies.anyhow]
@@ -81,14 +80,6 @@ workspace = true
 [dependencies.snarkos-node-router]
 workspace = true
 
-[dependencies.snarkvm-synthesizer]
-#path = "../../../snarkVM/synthesizer"
-git = "https://github.com/ProvableHQ/snarkVM.git"
-rev = "c194f3e"
-#version = "=3.8.0"
-default-features = false
-optional = true
-
 [dependencies.rand]
 version = "0.8"
 
@@ -117,8 +108,3 @@ version = "0.1"
 
 [dev-dependencies.base64]
 workspace = true
-
-[package.metadata.cargo-machete]
-ignored = [
-  "snarkvm-synthesizer", # Needed for locktick feature
-]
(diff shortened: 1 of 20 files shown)


kaimast:
This is based on top of #3707, so that one should be merged first.

kaimast:
Did a final rebase now that #3707 is merged. Hopefully we can merge this soon.

node/rest/Cargo.toml
```diff
@@ -19,14 +19,13 @@ edition = "2021"
 [features]
 default = [ "parallel" ]
 parallel = [ "rayon" ]
-history = [ "snarkvm-synthesizer/history" ]
+history = [ "snarkvm/history" ]
```
vicsn, line 22: Did you compile with history feature to make sure it still works as expected? (I think checking if it creates a new db directory with `history-` prefix and it grows in size is sufficient, IIRC).

UPDATE: I built and ran, can confirm it still works.

node/sync/src/block_sync.rs
```diff
@@ -270,12 +277,44 @@ impl<N: Network> BlockSync<N> {
...
+                        warn!("Failed to send block request to peer '{sync_ip}': no such peer");
+                        false
+                    }
+                }
+            });
+
+            tasks.push(task);
+        }
+
+        // Wait for all sends to finish at the same time.
```
vicsn, line 306: I recall from old benchmarks that we started processing blocks while new requests were still coming in. Is this behaviour changed in this PR or was it already synchronous before?

node/sync/src/block_sync.rs
```diff
@@ -874,22 +907,35 @@ impl<N: Network> BlockSync<N> {
     fn construct_requests(
         &self,
         sync_peers: &IndexMap<SocketAddr, BlockLocators<N>>,
-        current_height: u32,
+        sync_height: u32,
```
vicsn, line 910: Should we rename this variable in the caller(s) as well?

node/sync/src/block_sync.rs
```diff
@@ -874,22 +907,35 @@ impl<N: Network> BlockSync<N> {
...
-        max_requests: usize,
+        max_blocks_to_request: u32,
     ) -> Vec<(u32, PrepareSyncRequest<N>)> {
+        // Compute the start height for the block requests.
+        let start_height = {
+            let requests = self.requests.read();
+            let mut start_height = sync_height + 1;
+
+            loop {
+                if requests.contains_key(&start_height) {
```
vicsn, line 920: Should we only skip requests which have been fulfilled? Does this relate to the redundancy factor?

vicsn (commented):
Some noob questions

node/sync/src/block_sync.rs
vicsn: This is not critical so resolving for now.

node/sync/src/block_sync.rs
kaimast: If we don't check it here it will fail on `check_next_block_request` further down. I added this for clarity and also reduce the number of log messages.

The redundancy factor is still 1 in production, so you send the request to exactly one peer.
Maybe we should change it in the future to have another timeout where after like 30 seconds you send to another peer so make progress faster?

vicsn: Thx for explaining! https://github.com/ProvableHQ/snarkOS/issues/3709

node/sync/src/block_sync.rs
vicsn: Speeding up block sync is on this quarter's agenda anyway, so I guess we'll find out what are the best first-principles ways to make it faster.

vicsn (approved):

PR #3708, opened 2025-07-04
[Feat] Support Migration for Record Versioning Upgrade

raychu86:
## Motivation

This PR adds support for the new Record Versioning and Inclusion updates. In addition, migration logic is added to support the transition between the consensus versions that activates the new features.

Two new rules are added:
1. Do not include deployments in new proposals if you are behind `ConsenusVersion::V8`.
2. Do not sign proposals that have deployments if you are behind `ConsenusVersion::V8`. 

These two rules can be removed after the migration occurs on each network

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -46,7 +46,7 @@ default-features = false
 [workspace.dependencies.snarkvm] # If this is updated, the rev in `node/rest/Cargo.toml` must be updated as well.
 #path = "../snarkVM"
 git = "https://github.com/ProvableHQ/snarkVM.git"
-rev = "c194f3e"
+rev = "4642f9b"
 #version = "=3.8.0"
 default-features = false
 #features = [ "circuit", "console", "rocks" ]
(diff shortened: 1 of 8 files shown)


Cargo.toml
```diff
@@ -43,10 +43,10 @@ members = [
 version = "1.0.1"
 default-features = false
 
-[workspace.dependencies.snarkvm] # If this is updated, the rev in `node/rest/Cargo.toml` must be updated as well.
+[workspace.dependencies.snarkvm]
 #path = "../snarkVM"
 git = "https://github.com/ProvableHQ/snarkVM.git"
-rev = "c194f3e"
+rev = "4642f9b"
```
vicsn: Can this be updated so CI can run using the latest snarkVM?

vicsn (dismissed):

Cargo.toml
raychu86: Updated in [c48e07f](https://github.com/ProvableHQ/snarkOS/pull/3708/commits/c48e07f61d33cbefc63f1ec4a185d56062ab3964)

vicsn (approved):

Issue #3710, opened 2025-07-05
[Bug] /sync_status output and logging output incorrect


vicsn:
## 🐛 Bug Report

snarkOS commit `420b4fa4af2f22ecc899c03b97cc3fbed4d2a2da`. Full command: `./target/release/snarkos start --client --rest 0.0.0.0:3030 --nodisplay --verbosity 6 --cdn https://cdn.provable.com/v0/blocks/mainnet`

Example output from a fresh client, `is_synced` is clearly incorrect:

```
$ curl localhost:3030/mainnet/sync_status
{
  "is_synced": true,
  "ledger_height": 718,
  "sync_mode": "p2p",
  "cdn_height": 8920250,
  "outstanding_block_requests": 0
}
```

Moreover, `sync_mode` is actually also incorrect. We are by default starting to sync from CDN.

And what's worse, it seems all normal logging is not visible, perhaps due to https://github.com/ProvableHQ/snarkOS/pull/3707/
The only line I see printed is:
```
2025-07-05T14:54:31.824420Z DEBUG tcp{name="0"}: snarkos_node_tcp::tcp: The node is ready
```
If I change all `info/debug/trace` to `println` then I start seeing CDN sync is in progress, e.g.:
```
Synced up to block 43672 of 8925449 - 0% complete
```

PR #3711, opened 2025-07-07
[Fix] Set up log filters correctly

kaimast:
I recently changed log filter to not use environment variables anymore (because `std::env::set_var` is unsafe in Rust 2024). For this to work correctly, the log filter cannot be prefixed with `RUST_LOG=` anymore. This caused the issue reported in #3710.

This PR also fixes the sync mode being reported incorrectly and makes `tracing` a workspace dependency. The former was caused by a missing `!`. The latter is just to avoid different versions of tracing causing logging issues, and adding workspace gradually instead of in one large commit, seems easier to manage.

Finally, I added a test to the devnet_ci script that tests for nodes creating any log messages. This is a good sanity check to have and does not add a noticeable overhead to testing.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -195,7 +195,6 @@ commands:
           name: "Run devnet test"
           timeout: 20m  # Allow 20 minutes total
           command: |
-            chmod +x .circleci/devnet_ci.sh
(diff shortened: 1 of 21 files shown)


vicsn (approved):
LGTM, logs show up again for me.

I'm still seeing `  "is_synced": true,` even if we're not based on `cdn_height`.

PR #3713, opened 2025-07-08
[Refactor] Tighten up high-level connection/peer handling

### LINKED ISSUE #3705: [Bug] The Resolver can get unaligned with the higher-level address maps
There is currently the possibility of the TCP address resolver becoming unaligned with the higher-level address maps, which is most likely caused by the modularity of the network stack (which results in the `Resolver` needing to be updated before the connection is finalized). While it doesn't happen too often, and we have a simple workaround for it, we should ideally not have to rely on it.

ljedrz:
We currently have several high-level connection/peer-related collections, which provides us with a good overview when reading the objects, but isn't without potential issues: it involves multiple lock operations whenever a peer is added or removed, which can easily lead to race conditions.

This PR is an attempt to get rid of these issues altogether by collapsing several collections of simple items in favor of a single collection of more complex ones. The basic summary of the changes is this:
```diff
# InnerRouter<N: Network> {
#    (...)
-    /// The map of connected peer IPs to their peer handlers.
-    connected_peers: RwLock<HashMap<SocketAddr, Peer<N>>>,
-    /// The set of handshaking peers. While `Tcp` already recognizes the connecting IP addresses
-    /// and prevents duplicate outbound connection attempts to the same IP address, it is unable to
-    /// prevent simultaneous "two-way" connections between two peers (i.e. both nodes simultaneously
-    /// attempt to connect to each other). This set is used to prevent this from happening.
-    connecting_peers: Mutex<HashMap<SocketAddr, Option<Peer<N>>>>,
-    /// The set of candidate peer IPs.
-    candidate_peers: RwLock<HashSet<SocketAddr>>,
-    /// The set of restricted peer IPs.
-    restricted_peers: RwLock<HashMap<SocketAddr, Instant>>,
+    /// The collection of both candidate and connected peers.
+    peer_pool: RwLock<HashMap<SocketAddr, Peer<N>>>,
```

A short summary of the other changes involved:
- the `Peer` becomes an `enum` with 3 possible variants: `Candidate`, `Connecting`, and `Connected`
- the `Peer` is no longer `Clone` (to avoid any possibility of unintentional duplication)
- new concrete types, `CandidatePeer` and `ConnectedPeer`, are isolated for added type safety
- the completion of the high-level network handshake and the high-level disconnect now only involve accessing a single collection
- the `Resolver` is only ever modified indirectly, through the `Peer`
- the `Resolver` no longer needs to map from the listener address to the connected address (the `peer_pool` can do it)

I tried to limit the scope of the logic changes to minimum, and confirmed that they are working by running all the tests locally, plus the devnet script.

Fixes https://github.com/ProvableHQ/snarkOS/issues/3705.


vicsn:
~Triggered prerelease tests against this.~ 

EDIT: Nevermind, still adding more clients to prerelease tests so it more accurately reflects the load.

vicsn:
Connections on a 40 validator 80 client network look healthy. The visible drops to 0 connections are due to expected restarts. 
Screenshot 2025-07-21 at 15 26 55

Can you evaluate if the logs look good to you too @ljedrz ?
https://drive.google.com/file/d/1WZcRemR1vUWhWkeLEsdsl-Pp3c_HK1tl/view?usp=sharing

node/router/src/handshake.rs
```diff
@@ -123,15 +123,21 @@ impl<N: Network> Router<N> {
 
         // Perform the handshake; we pass on a mutable reference to peer_ip in case the process is broken at any point in time.
         let handshake_result = if peer_side == ConnectionSide::Responder {
-            self.handshake_inner_initiator(peer_addr, &mut peer_ip, stream, genesis_header, restrictions_id).await
+            self.handshake_inner_initiator(peer_addr, stream, genesis_header, restrictions_id).await
         } else {
             self.handshake_inner_responder(peer_addr, &mut peer_ip, stream, genesis_header, restrictions_id).await
         };
 
         if let Some(ip) = peer_ip {
```
kaimast: Also change this to `listener_addr`?

I love that we are finally renaming this one, but there are a lot of other places in snarkos where we use the term `peer_ip`. So we might want to document the change somewhere, as we most likely won't change it for the entire codebase in the short term.

node/router/src/heartbeat.rs
```diff
@@ -217,16 +220,16 @@ pub trait Heartbeat<N: Network>: Outbound<N> {
             let peers_to_disconnect = self
                 .get_removable_peers()
                 .into_iter()
-                .filter(|peer| !peer.is_prover()) // remove provers as those are handled separately
-                .map(|p| p.ip())
+                .filter(|peer| peer.node_type != NodeType::Prover) // remove provers as those are handled separately
```
kaimast: I think you can still write 
```rust
!peer.node_type.is_prover()
```

If you find your version to be better style, we should consider removing those helper functions from `NodeType`.

node/router/src/heartbeat.rs
```diff
@@ -314,7 +317,7 @@ pub trait Heartbeat<N: Network>: Outbound<N> {
             .iter()
             .filter_map(|peer_ip| {
                 // If the peer is not connected, attempt to connect to it.
```
kaimast: Change the comment to something like 
```suggestion
                // If the peer is not already connecting or connected, attempt to connect to it.
```

node/router/src/helpers/peer.rs
```diff
@@ -13,108 +13,127 @@
...
+    Connecting,
+    // A fully connected (post-handshake) peer.
+    Connected(ConnectedPeer<N>),
+}
+
+/// A candidate peer.
+#[derive(Clone)]
+pub struct CandidatePeer {
+    pub listener_addr: SocketAddr,
+    pub restricted: Option<Instant>,
```
kaimast, line 39: I am not sure what `restricted` is for, even after looking at `CandidatePeer::restrict`. Can you add a doc comment here?

node/router/src/lib.rs
```diff
@@ -342,47 +306,51 @@ impl<N: Network> Router<N> {
...
+
+    /// Returns `true` if the node is connecting to the given peer IP.
+    pub fn is_connecting(&self, ip: &SocketAddr) -> bool {
+        self.peer_pool.read().get(ip).map(|peer| peer.is_connecting()).unwrap_or(false)
     }
 
     /// Returns `true` if the node is connected to the given peer IP.
     pub fn is_connected(&self, ip: &SocketAddr) -> bool {
-        self.connected_peers.read().contains_key(ip)
+        self.peer_pool.read().get(ip).map(|peer| peer.is_connected()).unwrap_or(false)
```
kaimast: Can you use `is_some_and` here?

node/router/src/lib.rs
```diff
@@ -342,47 +306,51 @@ impl<N: Network> Router<N> {
...
 
     /// Returns `true` if the given IP is restricted.
     pub fn is_restricted(&self, ip: &SocketAddr) -> bool {
-        self.restricted_peers
-            .read()
-            .get(ip)
-            .map(|time| time.elapsed().as_secs() < Self::RADIO_SILENCE_IN_SECS)
-            .unwrap_or(false)
+        if let Some(Peer::Candidate(peer)) = self.peer_pool.read().get(ip) {
+            peer.restricted.map(|ts| ts.elapsed().as_secs() < Self::RADIO_SILENCE_IN_SECS).unwrap_or(false)
```
kaimast: It might be a good time to change `Self::RADIO_SILENCE_IN_SECS` to something like

```rust
const MAX_RADIO_SILENCE: Duration = Duration::from_secs(150);
```

node/router/src/lib.rs
```diff
@@ -397,73 +365,105 @@ impl<N: Network> Router<N> {
...
-        self.connected_peers.read().values().cloned().collect()
+    pub fn get_connected_peers(&self) -> Vec<ConnectedPeer<N>> {
+        self.peer_pool
+            .read()
+            .values()
+            .filter_map(|peer| if let Peer::Connected(p) = peer { Some(p.clone()) } else { None })
+            .collect()
     }
 
     /// Returns the list of connected peers.
```
kaimast, line 414: Can we clarify here which address is returned?

node/router/src/lib.rs
```diff
@@ -99,17 +98,8 @@ pub struct InnerRouter<N: Network> {
...
-    /// and prevents duplicate outbound connection attempts to the same IP address, it is unable to
-    /// prevent simultaneous "two-way" connections between two peers (i.e. both nodes simultaneously
-    /// attempt to connect to each other). This set is used to prevent this from happening.
-    connecting_peers: Mutex<HashMap<SocketAddr, Option<Peer<N>>>>,
-    /// The set of candidate peer IPs.
-    candidate_peers: RwLock<HashSet<SocketAddr>>,
-    /// The set of restricted peer IPs.
-    restricted_peers: RwLock<HashMap<SocketAddr, Instant>>,
+    /// The collection of both candidate and connected peers.
+    peer_pool: RwLock<HashMap<SocketAddr, Peer<N>>>,
```
kaimast, line 102: We write lock this whenever we update `last_seen` for any peer, which happens fairly frequently.  This could cause lock contention on nodes with many peers.

It might make sense to use [`DashMap`](https://docs.rs/dashmap/latest/dashmap/struct.DashMap.html) here. Alternatively, we could add a `Mutex` or `RwLock` around every peer entry, so we only have to write-lock the `peer_pool` when adding/removing an entry.

node/router/src/helpers/peer.rs
```diff
@@ -13,108 +13,127 @@
...
-    pub const fn is_prover(&self) -> bool {
-        self.node_type.is_prover()
-    }
-
-    /// Returns `true` if the peer is a client.
-    pub const fn is_client(&self) -> bool {
-        self.node_type.is_client()
-    }
+    /// Demote a peer to candidate status, marking it as disconnected.
+    pub fn downgrade_to_candidate(&mut self, listener_addr: SocketAddr, restrict: bool) {
```
kaimast, line 100: This is great! Can we add unit tests for downgrading peers, or do they exist already somewhere?

node/router/src/lib.rs
```diff
@@ -224,49 +205,32 @@ impl<N: Network> Router<N> {
...
         if self.is_connected(&peer_ip) {
             debug!("Dropping connection attempt to '{peer_ip}' (already connected)");
             return Ok(true);
         }
         // Ensure the peer is not restricted.
         if self.is_restricted(&peer_ip) {
-            bail!("Dropping connection attempt to '{peer_ip}' (restricted)")
-        }
-        // Ensure the node is not already connecting to this peer.
-        match self.connecting_peers.lock().entry(peer_ip) {
```
kaimast, line 237: I think I added this to prevent certain race conditions. Would it be possible to hold the lock to the peer pool, check the connection attempt, and then insert if successful? 

If I understand correctly, you mark a peer as connecting later when initiating the handshake, but you could already mark it as connecting here.

kaimast (changes requested):
These changes are exciting as they improve the code quality and maintainability a lot.
I left a few comments, and we can hopefully get this merged soon :)

node/router/src/handshake.rs
ljedrz: I changed it in several places, yes, but wanted to limit the scope of changes to prevent bitrot; my plan is to change the rest as a follow-up. This file should be safe to update, though.

node/router/src/lib.rs
ljedrz: This is a good point, though this setup is roughly the same as the current one, since we don't connect as often as we update connected peers; I think we should remember this, and it will be trivial to adjust if we ever detect contention there (e.g. with the `locktick` feature).

node/router/src/heartbeat.rs
ljedrz: While the method is trivial, I don't mind keeping it; it might have been a leftover from an earlier iteration, happy to bring it back here and in a few more spots.

node/router/src/helpers/peer.rs
ljedrz: Added a comment; also, ultimately I'd vote for removing it in favor of the ban feature, as the conditions and the "cooldown" involved are very similar.

node/router/src/helpers/peer.rs
ljedrz: Indirectly yes, as we have tests to detect the number of connected peers. It's not really testable in a truly isolated unit fashion, as it involves the resolver too.

node/router/src/lib.rs
ljedrz: I would love to be able to do it, but it would be a pretty serious attack vector if we optimistically assumed that the peer wouldn't take their time to finish connecting.

As for marking an _outgoing_ connection attempt as "connecting", this is primarily done by the lower-level layer (`tcp`, which also has a notion of "partial" connections), which is sufficient to avoid duplicate attempts. Of course it is trickier in case of inbound connections, but we do it as soon as possible (in `ensure_peer_is_allowed`).

node/router/src/lib.rs
ljedrz: I'm not opinionated about the name here (we rarely use it), but I'd suggest leaving it for a follow-up, when the notion of a "restricted peer" (which utilizes it) is cleared up (as suggested [here](https://github.com/ProvableHQ/snarkOS/pull/3713#discussion_r2224854708)).

ljedrz:
@vicsn The logs look ok to me :+1:.

ljedrz:
The CI failure is unrelated:
```
    Scanning Cargo.lock for vulnerabilities (603 crate dependencies)
Crate:     curve25519-dalek
Version:   4.2.0
Warning:   yanked
Dependency tree:
curve25519-dalek 4.2.0
└── ed25519-dalek 2.2.0
    └── zipsign-api 0.1.5
        └── self_update 0.42.0
            └── snarkos-cli 3.8.0
                └── snarkos 4.0.0

error: 1 denied warning found!
```

kaimast:
I had the audit issue too. `cargo update` should downgrade that package and fix it.

node/router/src/lib.rs
kaimast: Sounds good

node/router/src/helpers/peer.rs
kaimast: Is it hard to set up a resolver state? If not, it would still be great to have dedicated tests for changing peer state.
If it is too complicated, the E2E tests are sufficient.

node/router/src/lib.rs
kaimast: Sounds fine. On second thought, it is less invasive to reduce the frequency of calling `update_last_seen`. `Instant` usually has a granularity of 1s, so calling it on every inbound message is a little excessive. 

But that can also be done in another PR.

node/router/src/helpers/peer.rs
ljedrz: yeah, it requires the entire `Router`, which in turn requires the ledger, etc; that being said, the peer state changes are quite trivial

kaimast (approved):
This should be good to merge now 🥳

node/router/src/lib.rs
kaimast: Oh, I didn't see the logic in `ensure_peer_is_allowed`. Thanks for clarifying!

PR #3714, opened 2025-07-08
Add "--store-format" option to snarkos develop

meddle0x53:
## Motivation

We want to be able to store transactions in string/JSON format so we can broadcast them via the API easily when testing SnarkOS.

Basically with this option we can do:

```
snarkos developer execute --private-key <PK> --query http://localhost:3030 credits.aleo transfer_public <addr> 1000u64 --network 1  --store tx.json --store-format string
```

And then later we can load this transaction and broadcast it with:

```
curl -X POST http://localhost:3030/testnet/transaction/broadcast   -H "Content-Type: application/json"   --data-binary @tx.json
```

The old functionality of storing transactions in bytes format is still kept and if `--store-format` is not passed it will behave like before, so this is just  new logic in addition to the old one.

## Test Plan

Tested VS testnet locally and also tested with `./devnet.sh`

### DIFF
--- a/cli/src/commands/developer/execute.rs
+++ b/cli/src/commands/developer/execute.rs
@@ -13,6 +13,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use crate::commands::StoreFormat;
 use super::Developer;
 use snarkvm::{
     console::network::{CanaryV0, MainnetV0, Network, TestnetV0},
@@ -35,7 +36,6 @@ use aleo_std::StorageMode;
 use anyhow::{Result, anyhow, bail};
 use clap::Parser;
 use colored::Colorize;
-use crate::commands::StoreFormat;
 use std::{path::PathBuf, str::FromStr};
 use zeroize::Zeroize;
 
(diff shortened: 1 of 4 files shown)


cli/src/commands/developer/execute.rs
```diff
@@ -35,6 +35,7 @@ use aleo_std::StorageMode;
 use anyhow::{Result, anyhow, bail};
 use clap::Parser;
 use colored::Colorize;
+use crate::commands::StoreFormat;
```
vicsn: Can you move this up? We list inside and outside dependencies in separate blocks.

vicsn (dismissed):
One nit

cli/src/commands/developer/execute.rs
meddle0x53: Yeah, I was wondering where to put it and added it in alphabetical order, but will move it.

meddle0x53: Updated.

vicsn (approved):

PR #3723, opened 2025-07-10
Add static query to snarkos developer execute

meddle0x53:
## Motivation

After a sync with @vicsn we are moving to using `snarkos developer execute` instead of the cannon for testing. One of the steps is to pregenerate transactions using the genesis state root and on the latest consensus version.

This can be done by passing them as the `--query` parameter to the command. Here is an example:

```
snarkos developer execute \
  --query '{"state_root": "sr1287g435kqmlu6w8fzpp9dyggfwxyyaeymg89p3nzefq86xxst5pq9eht34", "height": 249}' \
  --private-key <PK> \
  credits.aleo transfer_public aleo1rhgdu77hgyqd3xjj8ucu3jj9r2krwz6mnzyd80gncr5fxcwlh5rsvzp9px 1000u64 \
  --network 1  \
  --store file.txt \
  --store-format string
```

This now works and these transaction can be sent later with:

```
curl http://localhost:3030/testnet/transaction/broadcast -X POST -H "Content-Type: application/json" -d @file.txt
```

We will be able to pre-generate many transfer_public (and not only) transactions that we can send later while running the tests.

## Test Plan

Used `devnet.sh`, got a state root (`curl localhost:3030/testnet/stateRoot/latest`) and a height for the test, ran the new command, broadcasted the transaction and checked that it was available : DONE.

Wrote a small unit test for the new static query parsing logic. Ran all the CLI unit tests.


vicsn:
@niklaslong following up on an earlier conversation, I'd be keen to hear if you think it makes sense to make this a supported part of the snarkVM library somewhere, with e.g. a state path map as well.

EDIT: Ah I see SDK also has such an object already: https://github.com/ProvableHQ/sdk/blob/mainnet/wasm/src/programs/offline_query.rs#L33

cli/src/helpers/static_query.rs
```diff
@@ -0,0 +1,93 @@
...
+    prelude::query::QueryTrait,
+};
+
+use anyhow::{Result, anyhow};
+use async_trait::async_trait;
+use serde::Deserialize;
+use std::str::FromStr;
+
+#[derive(Clone, Debug)]
+pub struct StaticQuery<N: Network> {
```
vicsn: Can you move this into `snarkVM/ledger/query/src/static_query.rs` ?

Can you make it:
```
#[derive(Clone)]
pub struct StaticQuery<N: Network> {
    block_height: u32,
    state_root: N::StateRoot,
    state_paths: HashMap<Field<N>, StatePath<N>>,
}

impl<N: Network> StaticQuery<N> {
    pub fn new(
        block_height: u32,
        state_root: N::StateRoot,
        state_paths: HashMap<Field<N>, StatePath<N>>,
    ) -> Self {
        Self {
            block_height,
            state_root,
            state_paths,
        }
    }
}

#[async_trait(?Send)]
impl<N: Network> QueryTrait<N> for StaticQuery<N> {
    /// Returns the current state root.
    fn current_state_root(&self) -> Result<N::StateRoot> {
        Ok(self.state_root)
    }

    /// Returns the current state root.
    async fn current_state_root_async(&self) -> Result<N::StateRoot> {
        unimplemented!();
    }

    /// Returns a state path for the given `commitment`.
    fn get_state_path_for_commitment(&self, commitment: &Field<N>) -> Result<StatePath<N>> {
        match self.state_paths.get(commitment) {
            Some(state_path) => Ok(state_path.clone()),
            None => bail!("could not find state path for commitment '{commitment}'"),
        }
    }

    /// Returns a state path for the given `commitment`.
    async fn get_state_path_for_commitment_async(
        &self,
        _commitment: &Field<N>,
    ) -> Result<StatePath<N>> {
        unimplemented!();
    }

    /// Returns the current block height
    fn current_block_height(&self) -> Result<u32> {
        Ok(self.block_height)
    }

    /// Returns the current block height
    async fn current_block_height_async(&self) -> Result<u32> {
        unimplemented!();
    }
}
```

kaimast:
Running `cargo +nightly fmt` and `cargo update` should fix the failing CI.

meddle0x53:
Yeah, I'm not fixing it as it should actually be reworked following the merge of this one : https://github.com/ProvableHQ/snarkVM/pull/2806

But am using it to run tests, building from my commits to it here.

cli/src/commands/developer/deploy.rs
```diff
@@ -113,7 +108,10 @@ impl Deploy {
     /// Construct and process the deployment transaction.
     fn construct_deployment<N: Network, A: Aleo<Network = N, BaseField = N::Field>>(&self) -> Result<String> {
         // Specify the query
-        let query = Query::<N, BlockMemory<N>>::from(&self.query);
+        let query: Box<dyn QueryTrait<N>> = match self.query.parse::<StaticQuery<N>>() {
```
vicsn: I see I had an oversight in the snarkVM PR: can we make `Static` a member of the enum `Query`?

In the `impl<N: Network, B: BlockStorage<N>> From<&String> for Query<N, B> {`, we can parse the string, and determine whether it is supposed to be a `Static` or a `REST`

meddle0x53: The snarkVM PR has been merged...

vicsn (approved):
LGTM!

PR #3727, opened 2025-07-14
Simplify BlockSync

kaimast:
### Removes the dedicated responses map
This part simplifies the BlockSync code a lot because we do not need to update two maps when removing requests.
One limitation is that `has_pending_responses` is now slower, but it is only called once every sync iteration.

### Merges `check_block_response` and `insert_block_response`
By merging these two functions, we avoid potential race conditions as, because, before, each locked the requests map independently.


node/sync/src/block_sync.rs
```diff
@@ -378,12 +373,16 @@ impl<N: Network> BlockSync<N> {
     /// Inserts a new block response from the given peer IP.
     ///
     /// Returns an error if the block was malformed, or we already received a different block for this height.
+    /// This functiona also removes all block requests from the given peer IP on failure.
```
```suggestion
    /// This function also removes all block requests from the given peer IP on failure.
```
nit

node/bft/src/sync/mod.rs
```diff
@@ -316,7 +298,10 @@ impl<N: Network> Sync<N> {
             return false;
         }
 
-        self.issue_block_requests().await;
+        // Prepare the block requests, if any.
+        // In the process, we update the state of `is_block_synced` for the sync module.
+        let batch = self.block_sync.prepare_block_requests();
+        self.send_block_requests(batch).await;
```
ljedrz: perhaps this could be conditional based on the requests being non-empty (in case it's possible for there to be none) for added clarity?

node/src/client/mod.rs
```diff
@@ -348,18 +338,24 @@ impl<N: Network, C: ConsensusStorage<N>> Client<N, C> {
...
-                tokio::time::sleep(BLOCK_REQUEST_BATCH_DELAY).await;
+    async fn send_block_requests(
+        &self,
+        block_requests: Vec<(u32, PrepareSyncRequest<N>)>,
+        sync_peers: IndexMap<SocketAddr, BlockLocators<N>>,
+    ) {
+        // Issues the block requests in batches.
+        for requests in block_requests.chunks(DataBlocks::<N>::MAXIMUM_NUMBER_OF_BLOCKS as usize) {
+            if !self.sync.send_block_requests(self, &sync_peers, requests).await {
+                // Stop if we fail to process a batch of requests.
```
ljedrz, line 356: perhaps we should issue a relevant `DEBUG` log here?

vicsn (approved):

PR #3732, opened 2025-07-14
Remove uses of cow_to_cloned and cow_to_copied.

mikebenfield:
These macros were removed in snarkVM, and we can just do *x or x.into_owned() instead.

### DIFF
--- a/node/cdn/src/blocks.rs
+++ b/node/cdn/src/blocks.rs
@@ -24,7 +24,7 @@ use snarkvm::prelude::{
     Network,
     Serialize,
     block::Block,
-    store::{ConsensusStorage, cow_to_copied},
+    store::ConsensusStorage,
 };
 
 use anyhow::{Result, anyhow, bail};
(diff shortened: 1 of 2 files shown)


node/cdn/src/blocks.rs
```diff
@@ -138,7 +138,7 @@ impl CdnBlockSync {
                 debug!("Synced the ledger up to block {completed_height}");
 
                 // Retrieve the latest height, according to the ledger.
-                let node_height = cow_to_copied!(ledger.vm().block_store().heights().max().unwrap_or_default());
+                let node_height = ledger.vm().block_store().heights().max().unwrap_or_default().into_owned();
```
ljedrz:
```suggestion
                let node_height = *ledger.vm().block_store().heights().max().unwrap_or_default();
```
should be fine as long it's just the usual `u32`

kaimast (approved):
Looks good. I triggered CI again.

PR #3733, opened 2025-07-14
Fix client sync off by one

vicsn:
## Motivation

One more syncing off by one which had a very small chance of triggering:
```
quest to construct. Height for the next block request is 10000, but minimum common block locator ancestor is only 10000 (sync_height=9999 greatest_pee│
```

## Test Plan

- [ ] Deployed to devnet succesfully


vicsn:
Superseded by https://github.com/ProvableHQ/snarkOS/pull/3734

PR #3734, opened 2025-07-14
Fix/simplify sync

kaimast:


node/sync/src/block_sync/helpers.rs
```diff
@@ -0,0 +1,65 @@
...
+    ret
+}
+
+#[cfg(test)]
+mod tests {
+    use super::rangify_heights;
+
+    #[test]
+    fn test_rangify_heights() {
+        let heights = &[0, 1, 3, 4, 5, 6, 7, 10, 11, 12, 20, 22, 23, 24, 25, 27, 28];
```
vicsn, line 102: Can we add more edge cases? Empty set, duplicates if possible?

node/bft/src/sync/mod.rs
```diff
@@ -515,12 +502,55 @@ impl<N: Network> Sync<N> {
...
+                info!("Syncing the BFT to block {}...", block.height());
+                // Sync the storage with the block.
+                match self.sync_storage_with_block(block).await {
+                    Ok(_) => {
+                        // Update the current height if sync succeeds.
+                        current_height = next_height;
+                    }
+                    Err(err) => {
+                        // Mark the current height as processed in block_sync.
+                        self.block_sync.remove_block_response(next_height);
```
vicsn, line 535: Where do we remove outstanding block responses in the happy Ok path? Do we need to?

vicsn (approved):

PR #3739, opened 2025-07-17
 Ureq default features to enable tls

vicsn:
## Motivation

rusttls is part of ureq's default features.
But because we're since recently enabling json, we need to manually enable default-features

## Test Plan

Tested snarkos developer execute on macOS.

## Related PRs

https://github.com/ProvableHQ/snarkVM/pull/2817

CI issues:

- [x] Output record must be Version 0 before Consensus V8 - caused by some test innapropriately sampling record version 1 in combination with https://github.com/ProvableHQ/snarkVM/pull/2800

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3312,6 +3312,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "rustls-pemfile"
+version = "2.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "dce314e5fee3f39953d46bb63bb8a46d40c2f8fb7cc5a3b6cab2bde9721d6e50"
+dependencies = [
+ "rustls-pki-types",
+]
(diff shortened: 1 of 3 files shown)


raychu86:
Seems to work correctly. Will approve after the snarkVM sister PR lands

kaimast:
Your intention is to enable gzip compression too (which is the other default feature), right?

Generally, I find it nicer to explicitly enable the features you want instead of `default-features = true`, because the set of default features might change in the future.
If you keep it that way maybe add a comment that explains why it is enabled, so  its not changed in the future by accident.

raychu86 (approved):

vicsn:
> Generally, I find it nicer to explicitly enable the features you want instead of default-features = true, because the set of default features might change in the future.

Good point, I'll do this after we get this release out of the door.

PR #3751, opened 2025-07-23
[Fix] Return proper JSON from REST APIs

kaimast:
Replaces PR #3606 and fixes issue #3361.

### Reproducing the Bug
On `staging`, run test network using `./devnet.sh`. When you query the bonded validator mapping, it should give you the incorrectly formatted output.

```
~> curl http://localhost:3030/testnet/program/credits.aleo/mapping/bonded/$VALIDATOR_ADDR
"{\n  validator: $VALIDATOR_ADDR ,\n  microcredits: 187500058346732u64\n}"
```
(here $VALIDATOR_ADDR is a bonded address, that can be retrieved from the logs)

### Testing
On the PR branch (`fix/value-json`) run another test network and execute the above command again. Now you will get an output with the correct formatting.

```
~> curl http://localhost:3030/testnet/program/credits.aleo/mapping/bonded/$VALIDATOR_ADDR
{
  "validator": "$VALIDATOR_ADDR",
  "microcredits": "187500297811718u64"
}
```

### DIFF
--- a/node/rest/src/helpers/json.rs
+++ b/node/rest/src/helpers/json.rs
@@ -1,0 +1,56 @@
+// Copyright (c) 2019-2025 Provable Inc.
+// This file is part of the snarkOS library.
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at:
+
+// http://www.apache.org/licenses/LICENSE-2.0
+
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use snarkvm::prelude::{Network, Plaintext, Value};
+
+use anyhow::Result;
+
+/// Convert a `Value` to JSON.
+/// `Value` implements Serialize by just calling `to_string()` on the value, which leads to Display
+/// output which (sometimes) looks like JSON but isn't.
+/// We produce actual JSON by calling `serde_json::to_value` on the inner object.
+pub(crate) fn value_to_json<N: Network>(mapping_value: Option<&Value<N>>) -> Result<serde_json::Value> {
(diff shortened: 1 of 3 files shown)


raychu86:
Would the `value_to_json` and `mapping_to_json` functions be more useful in snarkVM?

kaimast:
> Would the `value_to_json` and `mapping_to_json` functions be more useful in snarkVM?

Yeah, that would make sense if you think there are other use cases for it. Maybe we should even have something like `Value::into_json()`.

Initially, I was also wondering if it makes more sense to just fix the serialization of `Value`, but I wasn't sure if that would break something else, and the previous PR was already close to finished.

vicsn:
Thank you for picking this up. @kaimast which other endpoints exhibit this issue?

Based on Kai's answer, @HarukaMa @miazn @iamalwaysuncomfortable are you ok if we change the serialization in a backwards-incompatible way? Or will this break existing tooling to an unbearable degree?

HarukaMa:
I don't think I'm relying on this specific api behavior so I'm not affected.

However, I think the old format is potentially useful if people are manually executing transactions via `snarkos developer execute`, as the old output format is directly usable as inputs (especially records, not sure if it's applicable here). There would be no other issues if the execute cli can also consume valid json.

kaimast:
> However, I think the old format is potentially useful if people are manually executing transactions via `snarkos developer execute`, as the old output format is directly usable as inputs (especially records, not sure if it's applicable here). There would be no other issues if the execute cli can also consume valid json.

Thank you for pointing this out. I will test that `developer execute` still works as expected and fix it, if needed.

kaimast:
@vicsn this issue is tagged for Q4, but I am not sure what we decided on with regard to changing the API behavior. I think it should change, but we could add a flag to force the old behavior. 

Also, we haven't decided if the value-to-JSON functionality should reside in snarkVM or remain in snarkOS (as implemented by this PR). The former would require more discussion, as we are trying to avoid API breakage in snarkVM.

vicsn:
This is not priority at the moment but adding this gem here for shared context: https://github.com/ProvableHQ/snarkVM/pull/2404

PR #3752, opened 2025-07-24
Tightened deserialization bounds

Antonio95:
This PR mainly adds few maximum checks when deserialising structures. This required, among other things, moving the constant `MAX_VALIDATORS_TO_SEND` (it has been re-exported in the place where it was originally defined in order to make the change breaking with minimal code modifications).

A couple of other tiny improvements in deserialisation functions have been made.


vicsn:
We might want to use a different PR title

Antonio95:
> We might want to use a different PR title

Apologies, I think I tried to copy the other title from the snarkVM for consistency and instead the link address got copied. Fixed.

node/bft/events/src/lib.rs
```diff
@@ -82,6 +82,8 @@ use serde::{Deserialize, Serialize};
 pub use std::io::{self, Result as IoResult};
 use std::{borrow::Cow, net::SocketAddr};
 
+pub use validators_response::MAX_VALIDATORS_TO_SEND;
```
raychu86: nit: Move this to the above `pub use validators_response::{...}`

node/bft/src/gateway.rs
```diff
@@ -101,7 +101,7 @@ const RESTRICTED_INTERVAL: i64 = (MAX_CONNECTION_ATTEMPTS as u64 * MAX_BATCH_DEL
 /// The minimum number of validators to maintain a connection to.
 const MIN_CONNECTED_VALIDATORS: usize = 175;
 /// The maximum number of validators to send in a validators response event.
-const MAX_VALIDATORS_TO_SEND: usize = 200;
+use snarkos_node_bft_events::MAX_VALIDATORS_TO_SEND;
```
raychu86: nit: Move this up

node/bft/events/src/validators_response.rs
```diff
@@ -15,6 +15,9 @@
 
 use super::*;
 
+/// The maximum number of validators to send in a validators response event.
+pub const MAX_VALIDATORS_TO_SEND: usize = 200;
```
raychu86: The use of `Committee::<N>::max_committee_size()` or `N::LATEST_MAX_CERTIFICATES()` may be more appropriate, because that is how we are bounding the maximum number of validators in

Antonio95: Thank you for pointing me to those methods. I have switched the bound check to `N::LATEST_MAX_CERTIFICATES()`, which in addition means I have been able to undo the moving of `MAX_VALIDATORS_TO_SEND` from this PR's first commit - this also addresses your two comments preceding this one.

node/bft/events/src/validators_response.rs
```diff
@@ -45,6 +45,15 @@ impl<N: Network> FromBytes for ValidatorsResponse<N> {
     fn read_le<R: Read>(mut reader: R) -> IoResult<Self> {
         // Read the number of validators.
         let num_validators = u16::read_le(&mut reader)?;
+
+        // Ensure the number of validators is within bounds
+        if num_validators > N::LATEST_MAX_CERTIFICATES().unwrap() {
```
kaimast: We usually avoid calling `unwrap` in production code, unless we know that this is always `Some`.

kaimast (changes requested):
Looks good. I am just a little unsure whether the unwraps are safe in this code.

node/bft/events/src/validators_response.rs
raychu86: Unfortunately, it looks like we do use these `unwraps` in `snarkVM`, but they would be better handled here with a `.map_err`

Antonio95: I actually wrote `map_err` first, then realised `unwrap()` was being used with that function in other places and switched to the latter. I'm happy to change all `N::LATEST_MAX_CERTIFICATES().unwrap()` to suitable errors (there are 6 instances, or 3 if one doesn't count closely related instances) or leave things as they are. @kaimast @raychu86 ?

vicsn: You can remove the unwraps. Its silly, but it will prevent future readers from worrying.

Antonio95: Done! There are three instances of `N::LATEST_MAX_CERTIFICATES()` in functions which don't return a `Result`-like type, so I have left them as they were. In the cases where it was possible to return an error, the `?` operator couldn't be used directly because of differing `Restult`-like types, so I used `map_err` (as suggested above).

raychu86 (dismissed):

kaimast (approved):

kaimast (dismissed):
I just noticed some tests fail in `snarkos-node-bft-events` due to these checks. Could you take a look?

Antonio95:
You're completely right, @kaimast , thanks for catching that.

**What was happening**:  [This test](https://github.com/ProvableHQ/snarkOS/blob/staging/node/bft/events/src/validators_response.rs#L93) was using [this function](https://github.com/ProvableHQ/snarkOS/blob/staging/node/bft/events/src/validators_response.rs#L82) to generate a response with 50 validators, but `MAX_CERTIFICATES` is set to 40 for `ConsensusVersion::V9` on `MainnetV0` in test mode (cf. [here](https://github.com/ProvableHQ/snarkVM/blob/fc4951e24d1c2bbae71104b4bfce29d13a3c6289/console/network/src/mainnet_v0.rs#L177)). Because the limit was not tested when deserialising `ValidatorsResponse` before, the issue wasn't caught, but introducing the bound checks caused deserialisation to fail.

**The fix**: I've made it so that [the test generates](https://github.com/ProvableHQ/snarkOS/blob/938f63109d938c4d7a24a19042652fa55e434899/node/bft/events/src/validators_response.rs#L97) not 50 validators, but `min(CurrentNetwork::LATEST_MAX_CERTIFICATES().unwrap() + 1, 50)`. The test now passes. In addition, I have added [a separate test](https://github.com/ProvableHQ/snarkOS/blob/938f63109d938c4d7a24a19042652fa55e434899/node/bft/events/src/validators_response.rs#L119) to ensure the bound is being checked correctly. It manually generates `LATEST_MAX_CERTIFICATES + 1` and checks deserialisation results in the expected "Too many" error.

**Two questions**:
 1. Is it normal that, in the preexisting test, deserialisation still works with `CurrentNetwork::LATEST_MAX_CERTIFICATES().unwrap() + 1` (but fails with `CurrentNetwork::LATEST_MAX_CERTIFICATES().unwrap() + 2`)? I suspect this is a result of how `proptest/strategy/boxed` work, which I am not familiar with, but I would like confirmation on this. Perhaps `proptest` use one of the generated addresses as the "requester address" and only the rest are put into the `ValidatorsResponse` by the `proptest` machinery? In any case, the newly introduced test indeed shows that the deserialisation error is triggered exactly when the number of validators is `MAX_CERTIFICATES().unwrap() + 1` or more (note that in that test, the `ValidatorsResponse` object is generated manually and not through the `strategy/boxed/proptest` machinery).
 2. Is `mod prop_tests` the right place for [the newly introduced test](https://github.com/ProvableHQ/snarkOS/blob/938f63109d938c4d7a24a19042652fa55e434899/node/bft/events/src/validators_response.rs#L119), which doesn't use any `proptest` machinery? I'm happy to move it elsewhere, e. g. to a new `mod tests {...}` in the same file.

raychu86 (approved):

PR #3753, opened 2025-07-24
Allow for transaction checking on broadcast

vicsn:
This PR adds support for synchronous checking of transactions at the following endpoint, by specifying a query parameter:
```
// POST /<network>/transaction/broadcast?check_transaction={true}
```

## Motivation

It is furiatingly hard to know why transactions failed.
See discussion in: https://github.com/ProvableHQ/snarkOS/pull/3046#issuecomment-3099196286

CC @HarukaMa (can't add you as a reviewer so tagging you here)

## Test Plan

- [x] Confirm spam protection on a local devnet using k6, whilst it passes just fine using curl

### DIFF
--- a/node/rest/src/lib.rs
+++ b/node/rest/src/lib.rs
@@ -54,7 +54,10 @@ use axum_extra::response::ErasedJson;
 use locktick::parking_lot::Mutex;
 #[cfg(not(feature = "locktick"))]
 use parking_lot::Mutex;
-use std::{net::SocketAddr, sync::Arc};
+use std::{
+    net::SocketAddr,
+    sync::{Arc, atomic::AtomicUsize},
+};
 use tokio::{net::TcpListener, task::JoinHandle};
 use tower_governor::{GovernorLayer, governor::GovernorConfigBuilder};
 use tower_http::{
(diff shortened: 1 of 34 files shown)


node/rest/src/lib.rs
```diff
@@ -80,6 +83,10 @@ pub struct Rest<N: Network, C: ConsensusStorage<N>, R: Routing<N>> {
     handles: Arc<Mutex<Vec<JoinHandle<()>>>>,
     /// A reference to BlockSync,
     block_sync: Arc<BlockSync<N>>,
+    /// The number of ongoing deploy transaction verifications via REST.
+    num_verifying_deploys: Arc<AtomicUsize>,
```
niklaslong, line 87: This is perfectly reasonable for this use case. Another option that is slightly heavier but fairer and more ergonomic, should that be desired, is a [`Semaphore`](https://docs.rs/tokio/latest/tokio/sync/struct.Semaphore.html).

node/rest/src/routes.rs
```diff
@@ -552,6 +560,40 @@ impl<N: Network, C: ConsensusStorage<N>, R: Routing<N>> Rest<N, C, R> {
...
+            // Try to acquire a slot.
+            let prev = counter.fetch_add(1, Ordering::Relaxed);
+            if prev >= limit {
+                counter.fetch_sub(1, Ordering::Relaxed);
+                return Err(RestError(err_msg.to_string()));
+            }
+            // Perform the check.
+            let res = rest
+                .ledger
+                .check_transaction_basic(&tx, None, &mut rand::thread_rng())
```
niklaslong, line 589: Sweet, this spawns a blocking task internally.

niklaslong (dismissed):
LGTM overall, left one suggestion and double-checked a blocking task is spawned for the check 🚀

niklaslong (approved):

PR #3754, opened 2025-07-25
[Refactor] More high-level network cleanups

ljedrz:
This PR is a direct follow-up to https://github.com/ProvableHQ/snarkOS/pull/3713, continuing with a refactoring sweep of the higher-level network layer around the `Router`.

A summary of the changes:
- remove the notion of a restricted peer in favor of the banning feature, which is targeting the same issue (connection spam), but is faster to trigger and has a wider reach
- `RADIO_SILENCE_IN_SECS` is renamed to `MAX_RADIO_SILENCE_SECS`
- the collection of trusted peers is removed in favor of enhancing the `Peer` object with a `trusted` flag
- several `heartbeat` and `outbound` methods are refactored for brevity and improved performance
- a few peer-related `Router` methods are removed
- peers can no longer have their node type or version updated without reconnecting (this may have been considered, but has never been possible - the node must be restarted in order for either to change)

Note: I currently only removed the updates to the `RESTRICTED` metric; should the metric be removed altogether, or perhaps show the number of banned peers instead?

### DIFF
--- a/node/router/src/handshake.rs
+++ b/node/router/src/handshake.rs
@@ -154,10 +154,10 @@ impl<N: Network> Router<N> {
     ) -> io::Result<ChallengeRequest<N>> {
         match self.peer_pool.write().entry(peer_addr) {
             Entry::Vacant(entry) => {
-                entry.insert(Peer::Connecting);
+                entry.insert(Peer::new_connecting(false));
             }
             Entry::Occupied(mut entry) if matches!(entry.get(), Peer::Candidate(_)) => {
-                entry.insert(Peer::Connecting);
+                entry.insert(Peer::new_connecting(entry.get().is_trusted()));
             }
             Entry::Occupied(_) => {
                 return Err(error(format!("Duplicate connection attempt with '{peer_addr}'")));
@@ -309,24 +309,24 @@ impl<N: Network> Router<N> {
 
     /// Ensure the peer is allowed to connect.
     fn ensure_peer_is_allowed(&self, listener_addr: SocketAddr) -> Result<()> {
-        // Ensure the peer IP is not this node.
-        if self.is_local_ip(&listener_addr) {
-            bail!("Dropping connection request from '{listener_addr}' (attempted to self-connect)")
-        }
-        // Ensure either the peer is trusted or `allow_external_peers` is true.
-        if !self.allow_external_peers() && !self.is_trusted(&listener_addr) {
-            bail!("Dropping connection request from '{listener_addr}' (untrusted)")
-        }
-        // Ensure the node is not already connecting to this peer.
         match self.peer_pool.write().entry(listener_addr) {
             Entry::Vacant(entry) => {
-                entry.insert(Peer::Connecting);
+                // Ensure that it's not a self-connect attempt.
+                if self.is_local_ip(&listener_addr) {
+                    bail!("Dropping connection request from '{listener_addr}' (attempted to self-connect)")
+                }
+                // Unknown peers are untrusted, so check if `allow_external_peers` is true.
+                if !self.allow_external_peers() {
+                    bail!("Dropping connection request from '{listener_addr}' (untrusted)")
+                }
+
+                entry.insert(Peer::new_connecting(false));
             }
             Entry::Occupied(mut entry) => match entry.get_mut() {
                 peer @ Peer::Candidate(_) => {
-                    let _ = mem::replace(peer, Peer::Connecting);
+                    let _ = mem::replace(peer, Peer::new_connecting(peer.is_trusted()));
                 }
-                Peer::Connecting => {
+                Peer::Connecting(_) => {
                     bail!("Dropping connection request from '{listener_addr}' (already connecting)");
                 }
                 Peer::Connected(_) => {
--- a/node/router/src/handshake.rs
+++ b/node/router/src/handshake.rs
@@ -309,17 +309,17 @@ impl<N: Network> Router<N> {
 
     /// Ensure the peer is allowed to connect.
     fn ensure_peer_is_allowed(&self, listener_addr: SocketAddr) -> Result<()> {
+        // Ensure that it's not a self-connect attempt.
+        if self.is_local_ip(&listener_addr) {
+            bail!("Dropping connection request from '{listener_addr}' (attempted to self-connect)")
+        }
+        // Unknown peers are untrusted, so check if `allow_external_peers` is true.
+        if !self.allow_external_peers() && !self.is_trusted(&listener_addr) {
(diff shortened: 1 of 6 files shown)


ljedrz:
One test has failed, looking into it :detective:.

All good now :ok_hand:.

node/router/src/lib.rs
```diff
@@ -121,7 +119,7 @@ impl<N: Network> Router<N> {
     const MAX_CONNECTION_ATTEMPTS: usize = 10;
     /// The duration in seconds after which a connected peer is considered inactive or
     /// disconnected if no message has been received in the meantime.
-    const RADIO_SILENCE_IN_SECS: u64 = 150; // 2.5 minutes
+    const MAX_RADIO_SILENCE_SECS: u64 = 150; // 2.5 minutes
```
kaimast: Should we make the type of this `Duration`?

node/router/src/lib.rs
```diff
@@ -215,10 +218,6 @@ impl<N: Network> Router<N> {
             debug!("Dropping connection attempt to '{peer_ip}' (already connected)");
             return Ok(true);
         }
-        // Ensure the peer is not restricted.
-        if self.is_restricted(&peer_ip) {
```
kaimast, line 222: Has restricted peers been removed? I am actually not sure what it was used for.

node/router/src/lib.rs
```diff
@@ -329,33 +328,9 @@ impl<N: Network> Router<N> {
...
-    pub fn is_restricted(&self, ip: &SocketAddr) -> bool {
-        if let Some(Peer::Candidate(peer)) = self.peer_pool.read().get(ip) {
-            peer.restricted.is_some_and(|ts| ts.elapsed().as_secs() < Self::RADIO_SILENCE_IN_SECS)
-        } else {
-            false
-        }
-    }
-
     /// Returns `true` if the given IP is trusted.
     pub fn is_trusted(&self, ip: &SocketAddr) -> bool {
```
kaimast: Change this to `listener_addr` too?

node/router/src/lib.rs
```diff
@@ -368,46 +343,31 @@ impl<N: Network> Router<N> {
...
-            .filter(|peer| matches!(peer, Peer::Candidate(peer) if peer.restricted.is_some()))
-            .count()
-    }
-
     /// Returns the connected peer given the peer IP, if it exists.
     pub fn get_connected_peer(&self, ip: &SocketAddr) -> Option<ConnectedPeer<N>> {
         if let Some(Peer::Connected(peer)) = self.peer_pool.read().get(ip) { Some(peer.clone()) } else { None }
     }
 
     /// Returns the connected peers.
```
kaimast:
```suggestion
    /// Returns all connected peers.
```

node/router/src/lib.rs
```diff
@@ -368,46 +343,31 @@ impl<N: Network> Router<N> {
...
     pub fn get_connected_peers(&self) -> Vec<ConnectedPeer<N>> {
+        self.filter_connected_peers(|_| true)
+    }
+
+    pub fn filter_connected_peers<P: FnMut(&ConnectedPeer<N>) -> bool>(
```
kaimast, line 370:
```suggestion
    /// Returns all connected peers that satisify the given predicate.
    pub fn filter_connected_peers<P: FnMut(&ConnectedPeer<N>) -> bool>(
```

node/router/src/lib.rs
```diff
@@ -368,46 +343,31 @@ impl<N: Network> Router<N> {
...
+        &self,
+        mut predicate: P,
+    ) -> Vec<ConnectedPeer<N>> {
         self.peer_pool
             .read()
             .values()
-            .filter_map(|peer| if let Peer::Connected(p) = peer { Some(p.clone()) } else { None })
+            .filter_map(|peer| if let Peer::Connected(peer) = peer { Some(peer) } else { None })
+            .filter(|peer| predicate(peer))
```
kaimast: Would it make sense to merge these filters? Let chains are stable now. 
```suggestion
            .filter_map(|peer| if let Peer::Connected(peer) = peer && predicate(peer) { Some(peer) } else { None })
```

kaimast (changes requested):
Just a few minor comments. Looks great!

node/router/src/lib.rs
ljedrz: sure, why not; ever since they can be created as `const`, it's fine

node/router/src/lib.rs
ljedrz: yes; it was basically used for the same thing as bans (too many connection attempts), it was just the older rendition that has become redundant

node/router/src/lib.rs
ljedrz: oh yeah, good point

node/router/src/lib.rs
```diff
@@ -121,7 +120,7 @@ impl<N: Network> Router<N> {
     const MAX_CONNECTION_ATTEMPTS: usize = 10;
     /// The duration in seconds after which a connected peer is considered inactive or
     /// disconnected if no message has been received in the meantime.
-    const RADIO_SILENCE_IN_SECS: u64 = 150; // 2.5 minutes
+    const MAX_RADIO_SILENCE_SECS: Duration = Duration::from_secs(150); // 2.5 minutes
```
kaimast: Does it make sense to still have the `_SECS` suffix as Duration is not tied to a specific unit? Could be `INTERVAL` or `DURATION`.

node/router/src/handshake.rs
```diff
@@ -309,32 +309,24 @@ impl<N: Network> Router<N> {
...
-                    if peer
-                        .restricted
-                        .map(|ts| ts.elapsed().as_secs() < Self::RADIO_SILENCE_IN_SECS)
-                        .unwrap_or(false) =>
-                {
-                    bail!("Dropping connection request from '{listener_addr}' (restricted)");
-                }
                 peer @ Peer::Candidate(_) => {
-                    let _ = mem::replace(peer, Peer::Connecting);
+                    let _ = mem::replace(peer, Peer::new_connecting(peer.is_trusted()));
```
kaimast, line 327: I'm curious if you could just write
```suggestion
                    *peer = Peer::new_connecting(peer.is_trusted());
```

kaimast (approved):

kaimast:
Looks great now. I had one minor bit but we can also merge as it.

node/router/src/handshake.rs
ljedrz: it _is_ in fact possible, but the merge was faster :smiley:; in any case, both are equivalent, so it can be done as a future cleanup

PR #3755, opened 2025-07-25
added test_network to default feature list

Roee-87:
## Motivation

This PR includes the `test_network` to the list of default features that are automatically included in the default installation of SnarkOS.  
Previously, users would have to specify the `test_network` as an added feature flag during installation.  This would be easy to miss (and was missed) by some teams who routinely use local test networks for development.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -145,7 +145,7 @@ name = "snarkos"
 path = "snarkos/main.rs"
 
 [features]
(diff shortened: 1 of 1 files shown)


vicsn:
This PR would enable `test_network` and all its test effects in production, which we don't want.

Do you want to give this a try? https://github.com/ProvableHQ/snarkOS/pull/3757

PR #3756, opened 2025-07-25
Add minimal transaction submission tests in devnet_ci.sh

### LINKED ISSUE #3701: [Feature] Land a deployment and execution in the CI devnet
## 🚀 Feature

If we're going through the trouble of running a devnet in CI and syncing anyway, we should also land different types of transactions. This would have caught https://github.com/ProvableHQ/snarkOS/pull/3700 early.

vicsn:
## Motivation

Closes https://github.com/ProvableHQ/snarkOS/issues/3701

The hardcoded timers are bound to break someday if we ever change CI hardware, but we can cross that bridge when we get there.

### DIFF
--- a/.circleci/devnet_ci.sh
+++ b/.circleci/devnet_ci.sh
@@ -1,4 +1,6 @@
 #!/bin/bash
+set -eo pipefail # error on any command failure
+set -x # print commands before executing them for easier debugging
 
 # Set parameters directly
 total_validators=$1
@@ -32,14 +34,29 @@ esac
 echo "Using network: $network_name (ID: $network_id)"
 
 # Create log directory
-log_dir=".logs-$(date +"%Y%m%d%H%M%S")"
+log_dir=${PWD}".logs-$(date +"%Y%m%d%H%M%S")"
 mkdir -p "$log_dir"
+chmod 755 "$log_dir"
 
 # Array to store PIDs of all processes
 declare -a PIDS
 
+# Define a cleanup function to kill all processes on exit
+cleanup() {
+  echo "🚨 Cleaning up ${#PIDS[@]} process(es)…"
+  for pid in "${PIDS[@]}"; do
+    if kill -0 "$pid" 2>/dev/null; then
+      kill -9 "$pid" 2>/dev/null || true
+    fi
+  done
+}
+trap cleanup EXIT
+trap 'echo "⛔️ Error in $BASH_SOURCE at line $LINENO: \"$BASH_COMMAND\" failed (exit $?)"' ERR
+
 # Start all validator nodes in the background
 for ((validator_index = 0; validator_index < $total_validators; validator_index++)); do
+  snarkos clean --dev $validator_index
+
   log_file="$log_dir/validator-$validator_index.log"
   if [ "$validator_index" -eq 0 ]; then
     snarkos start --nodisplay --network $network_id --dev $validator_index --allow-external-peers --dev-num-validators $total_validators --validator --logfile $log_file --metrics &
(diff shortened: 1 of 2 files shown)


.circleci/devnet_ci.sh
```diff
@@ -94,42 +114,124 @@ check_heights() {
...
-sleep 30
+function main:
+    input r0 as u32.public;
+    input r1 as u32.private;
+    add r0 r1 into r2;
+    output r2 as u32.private;
+
+constructor:
+    assert.eq true true;
+""" > program/main.aleo
```
kaimast, line 190: You could maybe just store those files in .circleci? But this is fine too.

.circleci/devnet_ci.sh
```diff
@@ -94,42 +114,124 @@ check_heights() {
...
+  echo "❌ Test failed! Expected 4 records, but found $num_records: $scan_result"
+  exit 1
+else
+  echo "✅ Scan returned 4 records correctly: $scan_result"
+fi
 
 # Check heights periodically with a timeout
 total_wait=0
-while [ $total_wait -lt 900 ]; do  # 15 minutes max
+while [ $total_wait -lt 600 ]; do  # 10 minutes max
```
kaimast, line 232: I increased this before because tests would take slightly longer and fail sometimes. Not sure if it's a good idea to change it back?

.circleci/devnet_ci.sh
```diff
@@ -94,42 +114,124 @@ check_heights() {
...
   if check_heights; then
     echo "🎉 Test passed! All nodes reached minimum height."
     
-    # Cleanup: kill all processes
```
kaimast, line 128: I assume you're removing this because the container will shut down anyways?

.circleci/devnet_ci.sh
```diff
@@ -94,42 +114,124 @@ check_heights() {
...
+# Wait for the deployment to be processed.
+sleep 10
+# Execute a function in the deployed program.
+execute_result=$(snarkos developer execute --private-key APrivateKey1zkp8CZNn3yeCseEtxuVPbDCwSyhGW6yZKUYKfgXmcpoGPWH --network $network_id --query http://localhost:3030 --broadcast http://localhost:3030/$network_name/transaction/broadcast test_program.aleo main 1u32 1u32)
+# Wait for the execution to be processed.
+sleep 10
+# Fail if the execution transaction does not exist.
+tx=$(echo "$execute_result" | tail -n 1)
+found=$(curl -s -o /dev/null -w "%{http_code}" "http://localhost:3030/$network_name/transaction/$tx")
+if [[ "$found" -lt 200 || "$found" -ge 300 ]]; then
```
kaimast, line 213: Is this checking the HTTP status or what is the number for? A comment would be good

.circleci/devnet_ci.sh
```diff
@@ -32,14 +34,29 @@ esac
...
-log_dir=".logs-$(date +"%Y%m%d%H%M%S")"
+log_dir=${PWD}".logs-$(date +"%Y%m%d%H%M%S")"
 mkdir -p "$log_dir"
+chmod 755 "$log_dir"
 
 # Array to store PIDs of all processes
 declare -a PIDS
 
+# Define a cleanup function to kill all processes on exit
+cleanup() {
```
kaimast: Nit: you can write 'function cleanup()' for better legibility

kaimast (changes requested):

kaimast:
Looks mostly good! I was planning to add something like this to the CLI PR but let's merge this first :)

vicsn (commented):
TODO: increase min_height to 75, check CI doesn't take too long

.circleci/devnet_ci.sh
vicsn: There is a new timeout above for `consensus_version_stable` which takes like 20 blocks at the moment, which may also take a few minutes

.circleci/devnet_ci.sh
vicsn: Moved it into `trap cleanup EXIT`

.circleci/devnet_ci.sh
vicsn: bd4c86e72

.circleci/devnet_ci.sh
vicsn: 9319c29a4

kaimast (approved):

PR #3757, opened 2025-07-25
Default use test_network in devnet.sh

vicsn:
## Motivation

To make it easier for testers to use the newest consensus logic.


kaimast (approved):
LGTM

Issue #3758, opened 2025-07-26
[Bug] `snarkos developer execute` doesn't work under consensus V8


HarukaMa:
## 🐛 Bug Report

The cli command only inserts the programs used once, and those programs are added as edition 0 to the VM stack because it's the first program initialization with an empty VM. However after consensus V8 it's not possible to execute programs in edition 0, thus all executions will fail locally.

This applies to any program besides `credits.aleo`.

## Steps to Reproduce

Execute a program other than `credits.aleo` with `snarkos developer execute`.

## Expected Behavior

The command should deal with the edition restriction.

## Your Environment

- v4.0.1

vicsn:
Thank you for reporting!

snarkVM `v4.1.0` has exposed a new utility which should be used. Process now exposes `pub fn add_programs_with_editions` which allows for exposing Process with the latest Program editions.

kaimast:
If possible, could you test if the problem is fixed for you on [this branch](https://github.com/ProvableHQ/snarkOS/pull/3759)? Note, that the argument handling for `snarkos dev execute` is slightly different on the branch.

The potential fix is in [commit f83b8e6](https://github.com/ProvableHQ/snarkOS/pull/3759/commits/f83b8e64ab20ef06b69cde58adc91cb4599fef5c), which changed the CLI to set the edition based on the network's consensus version.

HarukaMa:
Sorry for the late report. I can confirm that branch works for executing, but it seems like the help text need to be updated:

```
  -b, --broadcast
          The endpoint used to broadcast the generated transaction
```

raychu86:
> Sorry for the late report. I can confirm that branch works for executing, but it seems like the help text need to be updated:

Updated here - https://github.com/ProvableHQ/snarkOS/pull/3759/commits/ce908e26ffb697b2318dedcc4b4301c5101d2bf5

raychu86:
Closing with https://github.com/ProvableHQ/snarkOS/pull/3759

HarukaMa:
Hi, I just tried to execute something on v4.3.0 tag, and I'm getting this again:

```
2025-11-03T19:04:34.840808Z DEBUG snarkos_cli::commands::developer::execute: At block height 12843303 and consensus V10
⚠️ VM failed to execute transaction locally
     ↳ Cannot execute token_registry.aleo on edition 0
```

vicsn:
@HarukaMa I cannot reproduce. 

[...]

PR #3748, opened 2025-07-21
[Chore] Make all snarkos crates workspace dependencies

kaimast:
The PR also makes some commonly used third-party crates workspace dependencies, and updates `lru` to `0.16`.

### `extern crate`
One noticeable change is that I needed to explicitly rename the metrics dependency for the crates to use it so you see the following sometimes.
```rust
#[cfg(feature = "metrics")]
extern crate snarkos_node_metrics as metrics;
```

I could also have renamed the workspace dependency, but this way all local workspace dependencies are prefixed with `snarkos-`.

PR #3638, opened 2025-05-18
[Feature] Program Upgradability.

d0cd:
This PR introduces updates needed to support program upgradability. Concretely:
- updates to the deployment cost to account for constructors.
- updates to the REST API to account for multiple program editions.
- updates to `snarkos developer` to support upgrades.

## Testing with `snarkos developer`

@reviewers open question on where this should be documented.

1. Start up a `testnet` devnet using `devnet.sh`. Be sure to use the `test_network` feature.
2. Set up a test program. In a new directory, create:
```
// program.json
{
  "program": "test_program.aleo",
  "version": "0.1.0",
  "description": "",
  "license": "",
  "dependencies": null,
  "editions": {}
}
```
```
// main.aleo
program test_program.aleo;

function main:
    input r0 as u32.public;
    input r1 as u32.private;
    add r0 r1 into r2;
    output r2 as u32.private;

constructor:
    assert.eq true true;  
```
3. Deploy the initial program.
```
snarkos developer deploy --private-key APrivateKey1zkp8CZNn3yeCseEtxuVPbDCwSyhGW6yZKUYKfgXmcpoGPWH --network 1 --priority-fee 0 --broadcast http://localhost:3030/testnet/transaction/broadcast --query http://localhost:3030 test_program.aleo
```
Note, your `snarkos` will need to have the `test_network` feature enabled.
4. Check the edition.
```
curl http://localhost:3030/testnet/program/test_program.aleo/latest_edition
```
5. Update the edition in `program.json` for `test_program.aleo` to 1.
```
{
  "program": "test_program.aleo",
  "version": "0.1.0",
  "description": "",
  "license": "",
  "dependencies": null,
  "editions": {
    "test_program.aleo": 1
  }
}
```
6. Deploy again.
```
snarkos developer deploy --private-key APrivateKey1zkp8CZNn3yeCseEtxuVPbDCwSyhGW6yZKUYKfgXmcpoGPWH --network 1 --priority-fee 0 --broadcast http://localhost:3030/testnet/transaction/broadcast --query http://localhost:3030 test_program.aleo
```
7. Check the edition.
```
curl http://localhost:3030/testnet/program/test_program.aleo/latest_edition
```

PR #3759, opened 2025-07-26
[Feature] CLI developer fixes

kaimast:
Depends on [snarkVM#2838](https://github.com/ProvableHQ/snarkVM/pull/2838).

This PR includes an update to `ureq` and a set of improvements to the command line interface. #3748 should be merged before this one.

## ureq update
The `ureq` API changed slightly so I added helpers functions to `snarkos-cli` to get the behavior we want. In particular, the existing code does not only want the HTTP status code but also the response body when an error occurs.
There is now a `http_get` and a `http_get_json` function that perform result/error conversion to achieve this behavior.

## CLI rework
I changed some of the `snarkos developer` had a multitude of bugs. `deploy` and `execute` did not work as expected.
The PR not only fixes those commands, but also makes them align more with the ones in Leo's CLI. One difference is that `--broadcast` does not take an argument anymore, and `--query` was renamed to `--enpdoint`. 
Endpoint is now also passed as a `http::Uri` and handles missing protocol schemes better.

Finally, I changed the `parse` function for each command to `execute`, as there is already `parse_from` method provided by `clap`. `execute` is also a more fitting term as the command will run, not just be parsed.

### New CLI features
While trying to reproduce #3578 I added a set of new features that make it easier to use the `developer` command and simplify or CI scripts.

* `snarkos dev` is now an alias for `snarkos developer`.
* `snarkos dev execute` now has a `--wait` flag that blocks until the issued transaction is confirmed (or rejected).
* Transactions can now be issued with `--dev-key i` where i is the index of a devnet validator.
* `snarkos dev query transaction` allows getting transaction information similar to `leo query transaction`.
* Execution always works now, even with newer consensus version (fixing #3758).
* Priority fees are set to 0 by default and the `--priority-fee` flag can be omitted.
* The protocols scheme ("http" or "https") can be omitted for the endpoint, and will default to HTTP.
* Logging can be enabled (useful for debugging) by setting `--verbosity` like you would for `snarkos start`. When the flag it is not set, logging will be disabled entirely and not set to `INFO` like with `snarkos start`.

## `sccache` update
I updated sccache in this branch to 0.10, like we did for snarkOS, because I saw some issues with CI. Not sure if that was what fixed it or if it was just the usual flakiness, but it seemed a good time to do the update.


vicsn:
Would you mind tackling this one too along with it? https://github.com/ProvableHQ/snarkOS/issues/3758#issuecomment-3128186120

cli/src/commands/developer/execute.rs
```diff
@@ -96,13 +104,7 @@ impl Drop for Execute {
...
 impl Execute {
     /// Executes an Aleo program function with the provided inputs.
-    #[allow(clippy::format_in_format_args)]
-    pub fn parse(self) -> Result<String> {
-        // Ensure that the user has specified an action.
-        if !self.dry_run && self.broadcast.is_none() && self.store.is_none() {
-            bail!("❌ Please specify one of the following actions: --broadcast, --dry-run, --store");
-        }
-
+    pub fn execute(self) -> Result<String> {
```
raychu86: Please keep all the `parse` calls as is, unless there are actual breaking changes.

`execute` is also quite a loaded term that that clashes with our existing terms.

kaimast: Alright. I will change it back and add a doc comment to better describe what this does.

cli/src/commands/developer/scan.rs
```diff
@@ -32,30 +35,43 @@ use std::{
...
-    #[clap(short, long)]
+    #[clap(short, long, group = "key", conflicts_with = "dev_key")]
     private_key: Option<String>,
 
     /// The view key used to scan for records.
-    #[clap(short, long)]
+    #[clap(short, long, group = "key")]
     view_key: Option<String>,
 
+    /// Use a developer validator key tok generate the deployment.
```
raychu86:
```suggestion
    /// Use a developer validator key to generate the deployment.
```

cli/src/commands/account/sign.rs
```diff
@@ -0,0 +1,78 @@
...
+
+    /// Specify the account private key of the node
+    #[clap(long)]
+    pub(super) private_key: Option<String>,
+
+    /// Specify the path to a file containing the account private key of the node
+    #[clap(long, conflicts_with = "private_key")]
+    pub(super) private_key_file: Option<String>,
+
+    /// Use a developer validator key tok generate the deployment.
```
raychu86:
```suggestion
    /// Use a developer validator key to generate the deployment.
```

cli/src/commands/account/sign.rs
```diff
@@ -0,0 +1,78 @@
...
+    #[clap(long, group = "key")]
+    pub(super) dev_key: Option<u16>,
```
raychu86, line 40: The `dev_key` should conflict with `private_key` and `private_key` file since it won't be clear which key should be used

raychu86: Ah you have the new `key` groups that should resolve this. However, it's not implemented here for this command.

cli/src/commands/developer/decrypt.rs
```diff
@@ -14,50 +14,37 @@
...
-            MainnetV0::ID => Self::decrypt_ciphertext::<MainnetV0>(&self.ciphertext, &self.view_key),
-            TestnetV0::ID => Self::decrypt_ciphertext::<TestnetV0>(&self.ciphertext, &self.view_key),
-            CanaryV0::ID => Self::decrypt_ciphertext::<CanaryV0>(&self.ciphertext, &self.view_key),
-            unknown_id => bail!("Unknown network ID ({unknown_id})"),
-        }
-    }
-
-    /// Decrypts the ciphertext record with provided the view key.
-    fn decrypt_ciphertext<N: Network>(ciphertext: &str, view_key: &str) -> Result<String> {
+    pub fn parse<N: Network>(self) -> Result<String> {
```
raychu86, line 42: How come this particular command does not have `--network` that uses `network_id_parser` like the other commands? 

How is the network here selected?

cli/src/commands/developer/scan.rs
```diff
@@ -32,30 +35,43 @@ use std::{
...
 
 const MAX_BLOCK_RANGE: u32 = 50;
 
 /// Scan the snarkOS node for records.
-#[derive(Debug, Parser, Zeroize)]
+#[derive(Debug, Parser)]
+#[clap(
+    // The user needs to set view_key, private_key, or dev_key,
+    // but they cannot set private_key and dev_key.
+    group(clap::ArgGroup::new("key").required(true).multiple(true))
```
raychu86: It might be simpler to just disallow multiple uses. For instance, why allow using `private_key` AND `view_key`?

cli/src/commands/developer/query.rs
```diff
@@ -0,0 +1,56 @@
...
+
+#[derive(Debug, Parser)]
+pub enum QueryCommand {
+    Transaction(QueryTransaction),
+}
+
+#[derive(Debug, Parser)]
+pub struct Query {
+    /// The endpoint to scan blocks from.
+    #[clap(long, default_value = "https://api.explorer.provable.com/v1", global = true)]
```
raychu86: We seem to use `https://api.explorer.provable.com/v1` directly for each of these `--endpoint` inputs. Maybe we can define it as a global constant somewhere to reference?

cli/src/commands/developer/scan.rs
```diff
@@ -175,20 +192,16 @@ impl Scan {
...
-    fn parse_cdn<N: Network>() -> Result<String> {
-        match N::ID {
-            MainnetV0::ID => Ok(format!("{CDN_BASE_URL}/mainnet")),
-            TestnetV0::ID => Ok(format!("{CDN_BASE_URL}/testnet")),
-            CanaryV0::ID => Ok(format!("{CDN_BASE_URL}/canary")),
-            _ => bail!("Unknown network ID ({})", N::ID),
-        }
+    fn parse_cdn<N: Network>() -> Result<Uri> {
+        // This should always succeed as the base URL is hardcoded.
+        Uri::try_from(format!("{CDN_BASE_URL}/{}", N::SHORT_NAME)).with_context(|| "Unexpected error")
```
raychu86, line 174: Nit: Maybe a more descriptive error message?

cli/src/commands/developer/scan.rs
kaimast: Yeah, I was wondering about that too...

cli/src/commands/developer/decrypt.rs
kaimast: The `network` argument was moved to `developers/mod.rs` (using [`global` Option](https://docs.rs/clap/latest/clap/struct.Arg.html#method.global)) as all developer commands use it.
The PR had some left network arguments in the subcommands, but it is all cleaned up now.

Cargo.toml
```diff
@@ -46,7 +46,8 @@ default-features = false
 [workspace.dependencies.snarkvm]
 #path = "../snarkVM"
 git = "https://github.com/ProvableHQ/snarkVM.git"
-rev = "fc4951e"
+#rev = "fc4951e"
+branch = "feat/query-uri"
```
vicsn: Note to self: only merge once this is updated

cli/src/commands/developer/deploy.rs
```diff
@@ -34,51 +37,66 @@ use snarkvm::{
...
     priority_fee: u64,
     /// The record to spend the fee from.
     #[clap(short, long)]
     record: Option<String>,
-    /// The endpoint used to broadcast the generated transaction.
-    #[clap(short, long, conflicts_with = "dry_run")]
-    broadcast: Option<String>,
+    /// Set the transaction to be broadcasted to the given endpoint.
+    #[clap(short, long, group = "mode")]
+    broadcast: bool,
```
vicsn: This will break usage for anyone relying on this flag today. The cleaner interface is not worth the many hours of developer time which this will cause. Please revert all backwards-incompatible changes of this PR.

vicsn (changes requested):

cli/src/commands/developer/execute.rs
```diff
@@ -33,56 +37,69 @@ use snarkvm::{
...
+    #[clap(long, default_value_t = 0)]
+    priority_fee: u64,
     /// The record to spend the fee from.
     #[clap(short, long)]
     record: Option<String>,
     /// The endpoint used to broadcast the generated transaction.
-    #[clap(short, long, conflicts_with = "dry_run")]
-    broadcast: Option<String>,
+    #[clap(short, long, group = "mode")]
+    broadcast: bool,
```
vicsn: Moreover, having a separate query and broadcast endpoint is a privacy feature. Because you can query private state (records) from one node, and broadcast the transaction to another.

cli/src/commands/developer/execute.rs
```diff
@@ -33,56 +37,69 @@ use snarkvm::{
...
+    #[clap(long, group = "mode")]
     store: Option<String>,
     /// If --store is specified, the format in which the transaction should be stored : string or
     /// bytes, by default : bytes.
-    #[clap(long, value_enum, default_value_t = StoreFormat::Bytes)]
+    #[clap(long, value_enum, default_value_t = StoreFormat::Bytes, requires="store")]
     store_format: StoreFormat,
-    /// Specify the path to a directory containing the ledger. Overrides the default path (also for
-    /// dev).
+    /// Wait for the transaction to be accepted by the network. Requires --broadcast.
```
vicsn, line 100: Why did you add this? Our existing shell scripts work just fine. `snarkos developer` is just meant for very lightweight **internal** testing, Leo already exposes a much better waiting API for developers.

kaimast: I found it really useful to test https://github.com/ProvableHQ/snarkOS/pull/3638 and also reproduce the other bug fixes by this PR. Otherwise, I would have to issue the transaction and then query multiple times to see if it went through, or write a custom script.

cli/src/commands/developer/deploy.rs
kaimast: Added backwards compatibility in [4ae48c9](https://github.com/ProvableHQ/snarkOS/pull/3759/commits/4ae48c9a01538133a4c8e695fe722381a7167352)

cli/src/commands/developer/deploy.rs
```diff
@@ -34,51 +37,73 @@ use snarkvm::{
...
+#[command(
+    group(clap::ArgGroup::new("mode").required(true).multiple(false)),
+    group(clap::ArgGroup::new("key").required(true).multiple(false))
+)]
 pub struct Deploy {
     /// The name of the program to deploy.
     program_id: String,
-    /// Specify the network to create a deployment for.
-    #[clap(default_value = "0", long = "network")]
-    pub network: u16,
```
vicsn, line 51: Can you revive `network` so this PR is backwards compatible? Please review the argument changes one more time so it is backwards compatible. Can you also review the PR description one more time, and write it more like a changelog, so I can copy-paste it into the release blog post and changelog?

cli/src/commands/developer/deploy.rs
```diff
@@ -34,51 +37,73 @@ use snarkvm::{
...
-    #[clap(short, long)]
-    private_key: String,
-    /// The endpoint to query node state from.
-    #[clap(short, long)]
-    query: String,
+    #[clap(short = 'p', long, group = "key", value_parser=NonEmptyStringValueParser::default())]
+    private_key: Option<String>,
+    /// Use a developer validator key tok generate the deployment.
+    #[clap(long, group = "key")]
+    dev_key: Option<u16>,
```
vicsn, line 67: This is also not backwards compatible, our stress testing suite, and maybe that of other parties, would have to waste time refactoring this. It is a simple change, but it will cause many hours of work in the ecosystem for no good reason.

cli/src/commands/developer/execute.rs
```diff
@@ -33,56 +37,76 @@ use snarkvm::{
...
+    #[clap(value_parser=NonEmptyStringValueParser::default())]
     program_id: String,
     /// The function name.
+    #[clap(value_parser=NonEmptyStringValueParser::default())]
     function: String,
     /// The function inputs.
     inputs: Vec<String>,
-    /// Specify the network to create an execution for.
-    #[clap(default_value = "0", long = "network")]
-    pub network: u16,
```
vicsn, line 53: Keep this, even if it's ignored.

cli/src/commands/developer/deploy.rs
vicsn: Nevermind I'm mixing up the deploy and start commands, this is fine I think

cli/src/commands/developer/deploy.rs
vicsn: Nevermind I see it still exists and is moved up a level.

cli/src/commands/developer/execute.rs
vicsn: Nevermind I see it still exists and is moved up a level.

cli/src/commands/developer/query.rs
```diff
@@ -0,0 +1,54 @@
...
+use clap::Parser;
+use ureq::http::Uri;
+
+#[derive(Debug, Parser)]
+pub enum QueryCommand {
+    Transaction(QueryTransaction),
+}
+
+#[derive(Debug, Parser)]
+pub struct Query {
```
raychu86: We should remove this for now. In the meantime `leo` has all this functionality already in their CLI. 

In the future, we can adjust snarkVM `Query` or some base level repo (`aleo-std`) to define all the query endpoints for use in the higher level reps like `snarkVM`, `snarkOS`, and `leo`.

cli/src/commands/developer/execute.rs
```diff
@@ -160,50 +160,60 @@ impl Execute {
             let vm = VM::from(store)?;
 
             if !is_static_query {
+                let height = query.current_block_height()?;
+                let version = N::CONSENSUS_VERSION(height)?;
+                debug!("At block height {height} and consensus {version:?}");
+
+                // Ensure the correct edition is used if on newer consensus versions.
+                let edition = if version < ConsensusVersion::V8 { 0 } else { 1 };
```
raychu86: Query for the latest edition in order to select, rather than using the consensus version

kaimast: Done in [ea11958](https://github.com/ProvableHQ/snarkOS/pull/3759/commits/ea11958eb0656906324a99df4f74a20fa69b5d8c).

cli/src/commands/developer/query.rs
kaimast: Done in [4482bb9](https://github.com/ProvableHQ/snarkOS/pull/3759/commits/4482bb95b49d7bfca49b9c279447fc029dbbf5ae).

raychu86 (dismissed):

raychu86 (approved):

PR #3760, opened 2025-07-27
Fix build issue

vicsn:
## Motivation

```
error[E0515]: cannot return value referencing temporary value
   --> node/consensus/src/lib.rs:270:9
    |
270 |         self.transactions_queue.read().transactions().map(|(id, tx)| (id, Data::Object(tx)))
    |         ------------------------------^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |         |
    |         returns a value referencing data owned by the current function
    |         temporary value created here
    |
note: this call may capture more lifetimes than intended, because Rust 2024 has adjusted the `impl Trait` lifetime capture rules
   --> node/consensus/src/lib.rs:270:9
    |
270 |         self.transactions_queue.read().transactions().map(|(id, tx)| (id, Data::Object(tx)))
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    = help: use `.collect()` to allocate the iterator
help: use the precise capturing `use<...>` syntax to make the captures explicit
   --> node/consensus/src/transactions_queue.rs:60:91
    |
60  |     pub fn transactions(&self) -> impl Iterator<Item = (N::TransactionID, Transaction<N>)> + use<N> {
    |                                                                                            ++++++++

```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -275,7 +275,7 @@ jobs:
 
   node-consensus:
     executor: rust-docker
-    resource_class: << pipeline.parameters.medium >>
+    resource_class: << pipeline.parameters.large >>
     steps:
       - run_serial:
           workspace_member: node/consensus
--- a/node/consensus/src/transactions_queue.rs
+++ b/node/consensus/src/transactions_queue.rs
@@ -57,7 +57,7 @@ impl<N: Network> TransactionsQueue<N> {
         }
     }
 
-    pub fn transactions(&self) -> impl Iterator<Item = (N::TransactionID, Transaction<N>)> {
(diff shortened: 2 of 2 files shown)


kaimast:
This is odd. It looks like the test process for `snarkos-node-consensus` was killed. This usually happens when running out of memory, but, for this job, it never goes past 75%.

ljedrz (commented):
LGTM as long as the CI issue is fixed (not sure if the handful of test adjustments could impact this).

niklaslong (approved):

PR #3762, opened 2025-07-28
By default don't generate txs in devnet

vicsn:
## Motivation

Because this eats up resources and most people don't use it. This is a remnant from early snarkOS where this was needed for advancing.

CC @Roee-87


meddle0x53 (approved):
LGTM

PR #3767, opened 2025-07-30
[Feature] Show list of peers in terminal UI

kaimast:
This PR makes the "Overview" page in the terminal UI (`display`) usable. It now shows the currently connected peers and information about the latest block.

### Screenshot
﻿Screenshot of Overview Page

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -4,3 +4,6 @@ rustflags = [
     # See here: https://github.com/rust-lang/rust/issues/71515
     "-C", "link-arg=-fuse-ld=lld",
 ]
+
+[env]
+CXXFLAGS = "-include cstdint" #TODO: remove once RocksDB 0.24 is r
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -6,4 +6,4 @@ rustflags = [
 ]
 
 [env]
-CXXFLAGS = "-include cstdint" #TODO: remove once RocksDB 0.24 is r
+CXXFLAGS = "-include cstdint" #TODO: remove once RocksDB 0.24 is released
--- a/cli/src/commands/start.rs
+++ b/cli/src/commands/start.rs
@@ -634,8 +634,6 @@ impl Start {
             false => self.rest.or_else(|| Some("0.0.0.0:3030".parse().unwrap())),
         };
 
-        // If the display is not enabled, render the welcome message.
-        if self.nodisplay {
             // Print the Aleo address.
             println!("👛 Your Aleo address is {}.\n", account.address().to_string().bold());
             // Print the node type and network.
@@ -669,7 +667,6 @@ impl Start {
                     }
                 }
             }
-        }
 
         // If the node is a validator, check if the open files limit is lower than recommended.
         #[cfg(target_family = "unix")]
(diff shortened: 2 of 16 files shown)


display/src/pages/overview.rs
```diff
@@ -13,40 +13,94 @@
...
+            .map(|peer| {
+                let state = if peer.is_candidate() {
+                    "candidate"
+                } else if peer.is_connecting() {
+                    "connecting"
+                } else if peer.is_connected() {
+                    "connected"
+                } else {
+                    unreachable!("Unknown peer state");
+                }.to_string();
```
raychu86: Why not a `match peer` here for the types directly?

cli/src/commands/start.rs
```diff
@@ -710,7 +707,12 @@ impl Start {
         match node_type {
             NodeType::Validator => Node::new_validator(node_ip, self.bft, rest_ip, self.rest_rps, account, &trusted_peers, &trusted_validators, genesis, cdn, storage_mode, self.allow_external_peers, dev_txs, self.dev, shutdown.clone()).await,
             NodeType::Prover => Node::new_prover(node_ip, account, &trusted_peers, genesis, self.dev, shutdown.clone()).await,
-            NodeType::Client => Node::new_client(node_ip, rest_ip, self.rest_rps, account, &trusted_peers, genesis, cdn, storage_mode, self.rotate_external_peers, self.dev, shutdown).await,
+            NodeType::Client => {
+                if !self.nodisplay && !self.nocdn {
```
raychu86: Can you clarify why this is this only for Client and not Validator/Prover? 

Also, is the CDN syncing status still not being displayed?

kaimast: To show the CDN status, we would need to initialize the display earlier, similar to how it works with `snarkos-node-rest`. That would be a slightly bigger refactor and did not seem worth it to me right now.

Provers do not sync from CDN if I understand correctly. However, validators do, so let me change this.

display/src/pages/overview.rs
kaimast: Good point. I changed it.

kaimast:
I added a commit ([dbec188](https://github.com/ProvableHQ/snarkOS/pull/3767/commits/dbec188a3e965f4ab86be6b9b8da7086bdf2fc4c)) to fix building RocksDB on newer compilers. We do the same in snarkVM, so this is hopefully not too big of a change to tag onto this PR.

kaimast:
@raychu86 do you have time to take another look at this?
If time permits, run it locally as well and see if it looks okay.

.cargo/config.toml
```diff
@@ -4,3 +4,6 @@ rustflags = [
     # See here: https://github.com/rust-lang/rust/issues/71515
     "-C", "link-arg=-fuse-ld=lld",
 ]
+
+[env]
+CXXFLAGS = "-include cstdint" #TODO: remove once RocksDB 0.24 is r
```
raychu86: Comment is cut off here

raychu86 (dismissed):
This is great! The display is more clear than before

vicsn (approved):
Rubber stamp

PR #3768, opened 2025-07-30
feat: Add support for both ip and hostname for peers

### LINKED ISSUE #378: [Feature] Add DNS resolver for node connections
## 🚀 Feature

A DNS resolver can be added to snarkOS to facilitate node connections.

## Motivation

Node connections are purely reliant on IP addresses, which is much more difficult to reason about than a human readable URL. A simple DNS resolver will greatly aid in manual connection to nodes and delineation between bootnodes.

## Implementation

A DNS resolver such as `ns-dns-tokio` could be used to translate URLs into IPs that the node can connect to.

vicsn:
Closes https://github.com/ProvableHQ/snarkOS/issues/378

It is convinent to be able to specify both IPs and hostnames as peers. Especially, when you are building ephemeral testnets and want to specify peers as

```
./snarkos start --peers "validator0:4130,client0:4130"
```

Where you can populate the `/etc/hosts` file with subnet hostnames.

## Related PRs

Supersedes: https://github.com/ProvableHQ/snarkOS/pull/3432

### DIFF
--- a/cli/src/commands/start.rs
+++ b/cli/src/commands/start.rs
@@ -49,7 +49,7 @@ use rand::{Rng, SeedableRng};
 use rand_chacha::ChaChaRng;
 use serde::{Deserialize, Serialize};
 use std::{
-    net::{Ipv4Addr, SocketAddr, SocketAddrV4},
+    net::{Ipv4Addr, SocketAddr, SocketAddrV4, ToSocketAddrs},
     path::PathBuf,
     sync::{Arc, atomic::AtomicBool},
 };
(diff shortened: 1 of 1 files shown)


cli/src/commands/start.rs
```diff
@@ -313,20 +313,24 @@ impl Start {
...
+                .split(',')
+                .flat_map(|ip_or_hostname| {
+                    let trimmed = ip_or_hostname.trim();
+                    match trimmed.to_socket_addrs() {
+                        Ok(mut ip_iter) => {
+                            // A hostname might resolve to multiple IP addresses. We will use only the first one,
+                            // assuming this aligns with the user's expectations.
+                            ip_iter.next()
+                        }
+                        Err(e) => panic!("The hostname or IP supplied to --peers ('{trimmed}') is malformed: {e}"),
```
kaimast: We shouldn't panic here, should we?

cli/src/commands/start.rs
```diff
@@ -313,20 +313,24 @@ impl Start {
...
+            true => Ok(vec![]),
+            false => Ok(peers
+                .split(',')
+                .flat_map(|ip_or_hostname| {
+                    let trimmed = ip_or_hostname.trim();
+                    match trimmed.to_socket_addrs() {
+                        Ok(mut ip_iter) => {
+                            // A hostname might resolve to multiple IP addresses. We will use only the first one,
+                            // assuming this aligns with the user's expectations.
+                            ip_iter.next()
```
kaimast: If the socket address does not resolve, does this just fail silently? Might be good to print a warning if `ip_iter.next()` returns `None`.

kaimast (changes requested):
Two nits.

cli/src/commands/start.rs
vicsn: Good point, I'd argue it's not a nit :) 003bae5b4

cli/src/commands/start.rs
vicsn: Good point, I'd argue it's not a nit :) 003bae5b4

kaimast (approved):
Looks good now :)

PR #3769, opened 2025-07-31
WIP: Add TCP traffic metrics

vicsn:
## Motivation

These were and will continue to be instrumental in improving snarkOS node performance.

## Test Plan

- [ ] Should run a full devnet to assess performance regressions.


node/metrics/src/names.rs
```diff
@@ -44,8 +44,15 @@ pub(super) const GAUGE_NAMES: [&str; 26] = [
...
 ];
 
-pub(super) const HISTOGRAM_NAMES: [&str; 3] =
-    [bft::COMMIT_ROUNDS_LATENCY, consensus::CERTIFICATE_COMMIT_LATENCY, consensus::BLOCK_LATENCY];
+pub(super) const HISTOGRAM_NAMES: [&str; 7] = [
+    bft::COMMIT_ROUNDS_LATENCY,
+    consensus::CERTIFICATE_COMMIT_LATENCY,
+    consensus::BLOCK_LATENCY,
+    tcp::TCP_GATEWAY_EVENTS_OUTBOUND,
+    tcp::TCP_GATEWAY_EVENTS_INBOUND,
```
ljedrz, line 52:
```suggestion
    tcp::TCP_GATEWAY_BYTES_OUTBOUND,
    tcp::TCP_GATEWAY_BYTES_INBOUND,
```
IMO "events" would be more apt for the number of sent/received messages (which we might also want to track; they could even be used to detect DoS attempts)

node/metrics/src/names.rs
```diff
@@ -44,8 +44,15 @@ pub(super) const GAUGE_NAMES: [&str; 26] = [
...
+    tcp::TCP_ROUTER_MESSAGES_OUTBOUND,
+    tcp::TCP_ROUTER_MESSAGES_INBOUND,
```
ljedrz, line 54:
```suggestion
    tcp::TCP_ROUTER_BYTES_OUTBOUND,
    tcp::TCP_ROUTER_BYTES_INBOUND,
```

ljedrz (changes requested):
Actually, today I realized we keep track of those metrics for the node [already](https://github.com/ProvableHQ/snarkOS/blob/staging/node/tcp/src/tcp.rs#L108), it's just a matter of exposing it.

Issue #3779, opened 2025-08-09
[Proposal] Sync benchmark in CI


vicsn:
## 💥 Proposal

We should have some kind of representative benchmark in CI in order to assess the impact of changes such as https://github.com/ProvableHQ/snarkOS/pull/3681 . 

Because syncing consists of many components (disk I/O, compute, networking), no benchmark will be perfect, so this data should be cross-checked against the client sync performance from canary/testnet/mainnet logs.

Based on Kai's preliminary research, I propose the following: 

## setup
In a snarkVM unit test using rocksdb storage, generate a devnet chain with 60+ blocks with full subdags. That is, each block takes 2 rounds, each round contains MAX_CERTIFICATES number of certificates with 50%-100% MAX_TRANSMISSIONS worth of transfer_public executions. The resulting rocksdb storage folder can be zipped and uploaded to GCP storage.

Note that: 
- if we can't find the location of the rocksdb storage, perhaps we can use the rocksdb checkpoint mechanism to store the db to a known path. 
- this other PR also depends on GCP, it is likely we will want to keep private-write-public-read GCP storage in our CI workflows. https://github.com/ProvableHQ/snarkOS/pull/3740 
- we may want to update this rocksdb file every time we increase e.g. max validators or max transmissions 
- mainnet only has ~250 deployments, so we dont need to care about syncing deployments currently.
- syncing canary/testnet/mainnet state directly would not offer a good representative sample because often blocks are empty, especially early in the chain
- @meddle0x53 should be able to provide a link to pregenerated transactions you can download from S3, so you don't have to spend hours waiting for the generation to complete.

## benchmark
Just like the snarkVM benchmarks (https://github.com/ProvableHQ/snarkVM/blob/staging/.github%2Fworkflows%2Fbenchmarks.yml), run a snarkOS benchmark after every merge. 

The benchmark will: 
1. Download and unzip the rocksdb storage 
2. Start a single dev client node using the rocksdb storage 
3. Start a second client node which syncs from the first client node. 
4. Report how long the sync took. One dumb approach would be to poll the block height every tenth of a second.

vicsn:
https://github.com/ProvableHQ/snarkOS/pull/3783 has completed most of this, except for using transactions because they would lengthen the CI workflow

PR #3783, opened 2025-08-11
[CI] Sync benchmarks

kaimast:
This PR adds a benchmark workflow using Github Actions (similar to the one in snarkVM). It measures sync performance. The benchmarks only run for commits on staging and when manually triggered. 

## Graphs
Results for benchmark runs can be found on the [Dashboard](https://provablehq.github.io/snarkOS/dev/bench/). The dashboard will only contain runs from this PR's branch and staging. See a screenshot of the current graph for p2p sync below.

Picture of the p2p sync graph

The graph also contains additional information for each run, such as the git commit used to create the snapshot, in the tooltip.
Picture of the tooltip with extra data

## Workflow Summaries
Alternatively, you can go to the workflow summary to see results. That is useful when manually triggering runs of the benchmarks. See [here](https://github.com/ProvableHQ/snarkOS/actions/runs/17249923552/attempts/1#summary-48949336576) for an example of such a summary.

### DIFF
--- a/.ci/bench_bft_sync.sh
+++ b/.ci/bench_bft_sync.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 network_id=1
-min_height=240
+min_height=250
 
 # The number of validators that are syncing
 num_nodes=1
@@ -16,6 +16,10 @@ poll_interval=1 # Check block heights every second
 
 network_name=$(get_network_name $network_id)
 echo "Using network: $network_name (ID: $network_id)"
+
+info.txt > snapshot_info
+echo "Snapshot_info:"
+echo ${snapshot_info}
 
 # Create log directory
 log_dir=".logs-$(date +"%Y%m%d%H%M%S")"
@@ -75,7 +79,7 @@ while (( total_wait < max_wait )); do
     echo "🎉 Test passed!. Waited $total_wait for $min_height blocks. Throughput was $throughput blocks/s."
 
     # Append data to results file.
-    printf "{ \"name\": \"bft-sync\", \"unit\": \"blocks/s\", \"value\": %.3f, \"extra\": \"total_wait=%is\" },\n" \
+    printf "{ \"name\": \"bft-sync\", \"unit\": \"blocks/s\", \"value\": %.3f, \"extra\": \"total_wait=%is, target_height=${min_height}, ${snapshot_info}\" },\n" \
        "$throughput" "$total_wait" | tee -a results.json
     exit 0
   fi
--- a/.ci/bench_bft_sync.sh
+++ b/.ci/bench_bft_sync.sh
@@ -1,0 +1,99 @@
+#!/bin/bash
+
+network_id=1
+min_height=240
+
+# The number of validators that are syncing
+num_nodes=1
+
+# Adjust this to show more/less log messages
+log_filter="warn,snarkos_node_sync=debug"
+
+max_wait=1800 # Wait for up to 30 minutes
+poll_interval=1 # Check block heights every second
+
+. ./.ci/utils.sh
+
+network_name=$(get_network_name $network_id)
+echo "Using network: $network_name (ID: $network_id)"
+
+# Create log directory
+log_dir=".logs-$(date +"%Y%m%d%H%M%S")"
+mkdir -p "$log_dir"
+
+# Array to store PIDs of all processes
+declare -a PIDS
+
+# Define a trap handler that cleans up all processes on exit.
+function exit_handler() {
+  shutdown "${PIDS[@]}"
+}
+trap exit_handler EXIT
+
+# Define a trap handler that prints a message when an error occurs 
+trap 'echo "⛔️ Error in $BASH_SOURCE at line $LINENO: \"$BASH_COMMAND\" failed (exit $?)"' ERR
+
+# Shared flags betwen all nodes
+common_flags=" --nodisplay --network $network_id --nocdn --dev-num-validators=40 \
+  --no-dev-txs --log-filter=$log_filter"
+
+# The client that has the ledger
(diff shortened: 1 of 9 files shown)


.ci/bench_bft_sync.sh
```diff
@@ -0,0 +1,100 @@
+#!/bin/bash
+
+network_id=1
+min_height=200
```
vicsn: Just an idea, if the bft sync takes so much longer than client sync with the current setup, we could lower this to a height of 60 (which still surpassing the GC limit if we have strictly 2-round blocks). The smaller the benchmark, the lower the chance for noise building up in the measurement.

kaimast: If it is okay, I will just tackle BFT in another PR. It would be good to just get this merged now and make more incremental changes to it in the future.

.ci/bench_cdn_sync.sh
```diff
@@ -0,0 +1,73 @@
+#!/bin/bash
+
+####################################################
+# Measures a client syncing 1000 blocks from the CDN
```
vicsn, line 4: 1000 or 250 as mentioned in min_height? Same question for the other tests

.github/workflows/benchmarks.yml
```diff
@@ -0,0 +1,97 @@
...
+            echo "deb [signed-by=/usr/share/keyrings/cloud.google.gpg] https://packages.cloud.google.com/apt cloud-sdk main" | sudo tee -a /etc/apt/sources.list.d/google-cloud-sdk.list
+            sudo apt-get update
+            sudo apt-get install google-cloud-cli -y
+            echo $GCLOUD_SERVICE_KEY > ${HOME}/gcloud-key.json
+            gcloud auth activate-service-account --key-file=${HOME}/gcloud-key.json
+
+      - name: "Fetch Ledger Data"
+        run: |
+            gcloud config set project protocol-development-sandbox
+            gcloud storage cp gs://ci_testdata/sync-ledger-val40-250.zip ledger.zip
```
vicsn, line 46: Can we add the commit hash and consensus version to the filename? Can you add a comment/link here about how this file was generated? This should be reproducible.

.github/workflows/benchmarks.yml
```diff
@@ -0,0 +1,97 @@
...
+          alert-threshold: '150%'
+          comment-on-alert: true
+          summary-always: true
+          fail-on-alert: true
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          alert-comment-cc-users: '@kaimast'
+
+      - name: Push benchmark result
+        if: github.branch_name == 'staging'
+        run: git push 'https://kaimast:${{ secrets.GITHUB_TOKEN }}@github.com/ProvableHQ/snarkOS.git' gh-pages:gh-pagesu
```
vicsn: Why is this block needed and why does it need your account? I don't see this on the snarkVM counterpart: https://github.com/ProvableHQ/snarkVM/blob/staging/.github/workflows/benchmarks.yml

vicsn (changes requested):
Very nice! Some very minor points before we can merge

.github/workflows/benchmarks.yml
kaimast: I would like it to only push the results for staging. I enabled workflow dispatch so you can trigger it on any branch, but the graph should only record stuff that has been merged. They discuss this in the workflow README [here](https://github.com/benchmark-action/github-action-benchmark?tab=readme-ov-file#run-only-on-your-branches).
For other runs, you can still see the data on the summary page, e.g., [here](https://github.com/ProvableHQ/snarkOS/actions/runs/17182497823?pr=3783) for the most recent run.

Let me try to change it so it does not have my username. Also, I should add more documentation to this before we merge it, as per your previous point.

.github/workflows/benchmarks.yml
kaimast: Adding the hash to the filename is a good idea, but that would mean we have to change the benchmarks.yml file every time we update the zip file.
Alternatively, I could add a text file inside the zip that has the commit hash and print it when the workflow runs. There is an "extra_data" field in the results that I can add the commit hash to. What do you think?

I will also make a script to generate the ledger and add it to the `.ci` folder. This PR already moves a lot of code to `.ci/utils.sh` so it will be fairly short.

vicsn: > Adding the hash to the filename is a good idea, but that would mean we have to change the benchmarks.yml file every time we update the zip file.

I think we will need to do this, because we might have two different PRs, one on ConsensusVersion::V10 and the next on V11, and we want to be able to run benchmarks on both concurrently. 

Though as long as the transaction generation logic doesn't change, we can keep using the same file with the same `{commit_hash}_{consensus_version}` pre/postfix.

.ci/generate_ledger.sh
```diff
@@ -0,0 +1,93 @@
+#!/bin/bash
```
vicsn, line 1: I though we discussed populating the ledger with pregenerated transactions: https://github.com/ProvableHQ/snarkOS/issues/3779

Are we not doing this anymore? Do you think its too much work for the first iteration or was it unintentional or something else?

kaimast: Right now it takes over 20 minutes to sync 250 mostly empty blocks. Maybe we can tackle that first and then increase the number of blocks and transactions? Otherwise, it will be expensive to run the benchmarks.

vicsn: Thank you for explaining the rationale. I think we don't necessarily have to fill the entire DAG with transactions, and we could also sync less blocks, but happy to merge the PR so we can deliver.

vicsn (approved):
The graphs looks amazing 🚀 Remaining comments can be tackled in the future but this looks like a solid base.

PR #3786, opened 2025-08-12
CI: migrate workflows to checkout v5

rejected-l:
GitHub-hosted runners now use Node 24, so actions/checkout@v5 is required. Minimum runner version v2.327.1. Workflows only updated—no functional changes.

See: https://github.com/actions/checkout/releases/tag/v5.0.0

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -15,7 +15,7 @@ jobs:
       contents: write
     steps:
       - name: Checkout
-        uses: actions/checkout@v4
+        uses: actions/checkout@v5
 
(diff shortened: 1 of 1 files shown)


kaimast:
This looks fine.

Could you point to where it says checkout v4 does not work anymore? I don't see any warnings or errors about it on our workflow runs.

rejected-l:
@kaimast You're right - v4 still works. v5 is recommended for better Node.js 24 compatibility and future-proofing, but not strictly required yet

ljedrz (approved):

PR #3789, opened 2025-08-13
Handle errors in the REST API better

kaimast:
This PR uses #3759 as its base. I will change this PR's base to `staging` after the other PR is merged.

* The REST API now returns the error chain as JSON
* It will (try to) set the HTTP status code correctly
* CLI can now parse the returned error
* This PR also adds a few new command line flags (to help remove noise from the devnet CI logs)
     - `--nobanner` hides the Aleo banner at startup
     - `--noupdater` does not check for updates

### DIFF
--- a/cli/src/commands/start.rs
+++ b/cli/src/commands/start.rs
@@ -179,7 +179,7 @@ pub struct Start {
     /// Write log message to stdout instead of showing a terminal UI.
     ///
     /// This is useful, for example, for running a node as a service instead of in the foreground or to pipe its output into a file.
-    #[clap(long, verbatim_doc_comment, hide = true)]
+    #[clap(long, verbatim_doc_comment)]
     pub nodisplay: bool,
 
     /// Do not show the Aleo banner and information about the node on startup.
--- a/cli/src/commands/start.rs
+++ b/cli/src/commands/start.rs
@@ -179,8 +179,12 @@ pub struct Start {
     /// Write log message to stdout instead of showing a terminal UI.
     ///
     /// This is useful, for example, for running a node as a service instead of in the foreground or to pipe its output into a file.
-    #[clap(long, verbatim_doc_comment)]
+    #[clap(long, verbatim_doc_comment, hide = true)]
     pub nodisplay: bool,
+
+    /// Do not show the Aleo banner and information about the node on startup.
+    #[clap(long, hide = true)]
+    pub nobanner: bool,
 
     /// Specify the log verbosity of the node.
     /// [options: 0 (lowest log level) to 6 (highest level)]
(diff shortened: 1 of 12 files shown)


damons:
There are projects that rely on response codes as they are.  Changing the responses is a breaking change.   The general practice for APIs is to have a version in the URL.  Example:  `/v2/*`.    See https://github.com/ProvableHQ/snarkOS/pull/3796 for additional suggestions.

cli/src/commands/start.rs
```diff
@@ -182,6 +182,10 @@ pub struct Start {
     #[clap(long, verbatim_doc_comment)]
     pub nodisplay: bool,
 
+    /// Do not show the Aleo banner and information about the node on startup.
+    #[clap(long)]
+    pub nobanner: bool,
```
raychu86, line 187: Can you hide these docs?

`#[clap(long, hide = true)]`

kaimast: Done

cli/src/commands/start.rs
```diff
@@ -179,9 +179,13 @@ pub struct Start {
     /// Write log message to stdout instead of showing a terminal UI.
     ///
     /// This is useful, for example, for running a node as a service instead of in the foreground or to pipe its output into a file.
-    #[clap(long, verbatim_doc_comment)]
+    #[clap(long, verbatim_doc_comment, hide = true)]
```
raychu86:
```suggestion
    #[clap(long, verbatim_doc_comment)]
```

raychu86 (dismissed):

kaimast:
Commit [ecbb623](https://github.com/ProvableHQ/snarkOS/pull/3789/commits/ecbb623b43d042d1731a9bfaa3d490b48350a206) moved the new behavior to `/v2` using the approach outlined in #3796.

node/rest/src/lib.rs
```diff
@@ -261,9 +257,21 @@ impl<N: Network, C: ConsensusStorage<N>, R: Routing<N>> Rest<N, C, R> {
...
-        let v1_router = self.build_routes(rest_rps).layer(middleware::map_response(v1_error_middleware));
-        let v2_router = axum::Router::new().nest("/v2", self.build_routes(rest_rps));
-        let router = v1_router.merge(v2_router);
+        // Add the v1 API as default and under "/v1".
+        let default_router = axum::Router::new().nest(
+            &format!("/{}", N::SHORT_NAME),
+            self.build_routes(rest_rps).layer(middleware::map_response(v1_error_middleware)),
+        );
+        let v1_router = axum::Router::new().nest(
+            &format!("/v1/{}", N::SHORT_NAME),
```
vicsn, line 267: Introducing /v1/ does not seem backwards compatible. Please correct me if I'm wrong, or if its not possible to do without.

kaimast: It adds the old as API to `/` and `/v1`. In the future we may want to change the default, so if people want to keep using the `/v1` API they can specify it explicitly.

vicsn (approved):

PR #3796, opened 2025-08-20
Feature/add v2 api routes with error handling

damons:
## Motivation

The current REST API returns 500 status code for all errors/unsuccessful calls, even for calls that are successful but aren't functionally sufficient for intended results (i.e., blocks not found, but the query was appropriate).  This effectively inhibits clients from knowing the nature of the error/failure.  Sometimes those errors are not really errors.  The REST API should return meaningful status codes.

For example, block range requests for ranges that the client node does not yet contain currently returns a 500 instead of a 404.  Also, a 500 is returned for when the node is currently syncing instead of a service unavailable status code.  In each of these cases, if callers had better status codes they could choose alternate paths to fulfill the requirements of their application needs.

The current version of the API does not include a version indicator in the URL (i.e., `/v1/*`).  Any change to the response codes is a breaking change for all clients/projects who have built on the existing API.  A new version of the API needs to be implemented using a coherent version scheme that is industry standard while at the same time keeping the old API in place, allowing gradual migration for dependent projects.   This PR proposes a `/v2/*` implementation with enhanced HTTP StatusCode implementations.  This would allow the API to change over time with a new version being released for breaking changes.

This PR only addresses the following two concerns:

1.  The addition of a `/v2/*` API to the existing API which will still be available for use.
2.  The addition of the appropriate `StatusCode::` for all `/v2` API calls.

## Test Plan

Tests were added to `rest/src/lib.rs` to show routes continue to return `StatusCode::INTERNAL_SERVER_ERROR` for `Response<Body>`.    Also tests were added to verify routes return the new appropriate status codes for the `/v2` api.

Additional perf and CI tests should be added, I'm assuming.  Looking for feedback here and any help ushering this in as appropriate for the project.

## Related PRs

This PR also addresses status codes:  https://github.com/ProvableHQ/snarkOS/pull/3789

That PR also includes a lot of other changes that are unrelated to just the status codes.  It does not include the addition of a `/v2` API while also allowing the existing functionality to continue to exist, effectively allowing a sane migration path without breaking dependent projects and infrastructure.

I propose that we make the changes proposed here first.  This will pave the way for the changes in the other PR with a potential `/v3` possible.   Also, the other PR introduces more JSON into responses in clients that are already overloaded with abundant JSON overhead.  Simply returning a status code does not add bloat while at the same time enables developers of applications interacting with snarkOS clients dramatically enhanced/meaningful responses.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -288,7 +288,7 @@ dependencies = [
  "serde_urlencoded",
  "sync_wrapper",
  "tokio",
- "tower",
+ "tower 0.5.2",
  "tower-layer",
  "tower-service",
  "tracing",
@@ -333,7 +333,7 @@ dependencies = [
  "rustversion",
  "serde",
  "serde_json",
- "tower",
+ "tower 0.5.2",
  "tower-layer",
  "tower-service",
  "typed-json",
@@ -3215,7 +3215,7 @@ dependencies = [
  "tokio",
  "tokio-native-tls",
  "tokio-rustls",
- "tower",
+ "tower 0.5.2",
  "tower-http",
  "tower-service",
  "url",
@@ -4025,6 +4025,7 @@ dependencies = [
  "snarkvm",
  "time",
  "tokio",
+ "tower 0.4.13",
  "tower-http",
  "tower_governor",
  "tracing",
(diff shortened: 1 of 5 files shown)


vicsn (commented):
Thank you for the contribution!

We'll be prioritizing https://github.com/ProvableHQ/snarkOS/pull/3789 because it was opened a week earlier, it has an E2E test and returns more verbose errors, which at this stage seems more useful. But you are 100% right we should build in backwards compatibility, it'll be added along with your new unit tests and you'll be attributed!

vicsn:
Closing now that https://github.com/ProvableHQ/snarkOS/pull/3789 is merged, thank you again for raising this! I would appreciate if you can test that it's backwards compatible for what you had in mind.

PR #3802, opened 2025-08-23
[CI] Sync variance benchmark

kaimast:
## New Features
* `BlockSync` now measures sync speed, which is reported in the `sync/status` endpoint and shown in the TUI.
* Measures changes in sync speed. This will help to reduce sync burstiness in the future.
* Enables measuring BFT sync speeds.
* Adds a simple benchmark for the REST API (fetching random blocks). The idea is that improving the speed for the block endpoints will also help sync performance.

## Fixes/Improvements
* Throughput calculation is now based on system time (using `$SECONDS`) for higher accuracy. `sleep` might sleep longer than the given time, and time spent calling `curl` did not count towards the total time with the previous approach.
* Do not count initial connection handshake as sync time. These currently take up to 30 seconds and distort results.
* Make the scripts runnable on macOS (so we can test them easier)

### DIFF
--- a/.ci/utils.sh
+++ b/.ci/utils.sh
@@ -62,7 +62,7 @@ function check_heights() {
     return 0
   else
     if (( elapsed > 0 && ((elapsed % 60) == 0) )); then
-      elapsed_mins=$(elapsed/60)
+      elapsed_mins=$((elapsed / 60))
       echo "⏳ WAITING: Not all nodes reached minimum height of $min_height (highest so far: $highest_height, elapsed: $elapsed_mins minutes))"
     fi
 
--- a/.ci/utils.sh
+++ b/.ci/utils.sh
@@ -124,6 +124,9 @@ function stop_nodes() {
       kill -9 "$pid" 2>/dev/null || true
     fi
   done
+
+  # block until all nodes have shut down
+  wait
 }
 
 # Succeeds if all nodes are available.
--- a/.ci/utils.sh
+++ b/.ci/utils.sh
@@ -10,6 +10,16 @@ declare -a PIDS
 # Flag is set true once a node process stopped
 node_stopped=false
 
+# Tasksets to pin processes to specfic CPUs.
+# This is a no-op on MacOS.
+if [[ "$(uname)" == "Darwin" ]]; then
+  TASKSET1=""
+  TASKSET2=""
+else
+  TASKSET1="taskset -c 0,1"
+  TASKSET2="taskset -c 2,3"
+fi
+
 # Handler for a child process exiting
 function child_exit_handler() {
   # only set to true if this was indeed a node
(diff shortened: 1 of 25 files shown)


.ci/utils.sh
```diff
@@ -10,6 +10,16 @@ declare -a PIDS
 # Flag is set true once a node process stopped
 node_stopped=false
 
+# Tasksets to pin processes to specfic CPUs.
```
vicsn, line 13: Very fancy

vicsn (dismissed):

vicsn:
The [dashboard](https://provablehq.github.io/snarkOS/dev/bench/) notes that p2p-sync is ~0.2 blocks/second and bft-sync is ~50 blocks/seconds. Are we accidentally inverting one of the two metrics?

kaimast:
> The [dashboard](https://provablehq.github.io/snarkOS/dev/bench/) notes that p2p-sync is ~0.2 blocks/second and bft-sync is ~50 blocks/seconds. Are we accidentally inverting one of the two metrics?

This should be resolved with the latest changes.

vicsn:
The nodes are not syncing in the CDN benchmark, there may be a new bug.

node/sync/src/block_sync.rs
```diff
@@ -861,8 +884,13 @@ impl<N: Network> BlockSync<N> {
...
         // Remove the request entry for the given height.
         if let Some(e) = self.requests.write().remove(&height) {
-            trace!("Block request for height {height} was completed in {}ms", e.timestamp.elapsed().as_millis());
+            trace!(
+                "Block request for height {height} was completed in {}ms (sync speed is {})",
+                e.timestamp.elapsed().as_millis(),
+                self.get_sync_speed(),
+            );
         }
+        self.metrics.count_request_completed();
```
raychu86: This might also count `sync_speed` for requests that have failed.

I believe `remove_block_response` is called by the validator when it fails to execute `sync_storage_with_block` or `sync_ledger_with_block_without_bft`.

kaimast: Good point. I addressed it in [7a38b9e](https://github.com/ProvableHQ/snarkOS/pull/3802/commits/7a38b9eb20694174d453dc92a157bf306c97fbc9).

kaimast:
The PR is ready to be merged now (for real). Here is a summary of what was fixed in the latest commits:

* **[6b0f156](https://github.com/ProvableHQ/snarkOS/pull/3802/commits/6b0f1569ce9fd3606bef6b9c0db81ad715748360)**: CDN benchmark checked the height of the wrong node. For all other ones, it checks the height of the second node, but CDN only has one node, so checking the height of the second will never succeed. 
* **[7a38b9e](https://github.com/ProvableHQ/snarkOS/pull/3802/commits/7a38b9eb20694174d453dc92a157bf306c97fbc9):** Sync speed was counted incorrectly. Nodes now count them when advancing blocks. 
* **[e0e6273](https://github.com/ProvableHQ/snarkOS/pull/3802/commits/e0e6273531b9377a5e7fdbd84a02d74e6852df5a):** The rest benchmarks didn't get logged with the correct names. It stored both as `rest-` instead of `rest-block-height` and `rest-get-block`.
* **[9d9099d](https://github.com/ProvableHQ/snarkOS/pull/3802/commits/9d9099dc0ef1a0cac2abe50def3d8251573bb82d):** One small typo.

vicsn (approved):
LGTM!

PR #3808, opened 2025-08-26
[hotfix] Use correct cost determination for transactions after V10

vicsn:
## Motivation

There was a bug in the adaptation to https://github.com/ProvableHQ/snarkVM/pull/2890

As noted in that issue:
- for deployments we care about the constructor limit: https://github.com/ProvableHQ/snarkVM/pull/2890/files#diff-7d50c9d8aeb6ed044ebd7996abf9afcccfc58a6277737f85f78a15152f4a02e5R341
  - We also care about the individual function finalize costs, but AFAIK those may be expensive to calculate here as we have to construct a Stack, so we defer it until Stack creation. 
- for executions we care about the finalize cost limit: https://github.com/ProvableHQ/snarkVM/pull/2890/files#diff-897e9a9f59d4fc522348f00c936a1fd07ee9c654858dc846a658d6a02f05810aR420

All of this logic is still unnecessarily hard to track, as noted in this issue to clean up in the future: https://github.com/provableHQ/snarkVM/issues/2892

## Test Plan

- [x] deployments and transactions have a higher success rate in automated devnet


raychu86 (approved):

meddle0x53 (approved):
Could be good to see some message in the logs when this happens?

PR #3810, opened 2025-08-28
[Perf] Block Synchronization Pipeline

kaimast:
The `BlockSync` code in its current from prefers sending more requests over advancing block sync. I am not sure if there was a technical reason for this at some point, but it causes nodes to generate many requests before making any progress.

One issue with this approach is that it causes sync speed to contain spikes. Essentially, the node will issue hundreds of requests and not make any progress until all of them are sent successfully. Then, while processing responses, it will not issue any more requests. 
Another problem is that processing block responses is computationally expensive, causing a significant delay until a new batch of requests can be sent.

This PR changes the block sync loging to a more pipelined approach, where sending block requests and processing block responses can make progress semi-independently. This is achieved by having two dedicated tasks, each with their own `Notify` object to be woken up.
The separation of work allows sending more block requests *while* still waiting on previous responses to be processed in a blocking task. Similarly, block responses can be processed while still in the progress of sending a block request batch.

## Overview
The code currently adds the two tasks in `node/sync/src/block_sync.rs` and `node/bft/src/sync/mod.rs`. A future PR will unify more of the block sync logic between clients and validators. Below is an outline of what each of the two tasks do.

### "Outgoing" Task
* Generates new block requests if the node has not reached its limit yet
* Gets woken up on when the node receives new block locators or the ledger has advanced
* Re-issues timed out block requests (block requests can only time out if they have not received a response yet)

### "Incoming" Task
* Processes block responses and advances the ledger.
* Gets woken up when a pending block request received a response.
* Removes completed block requests (and wakes up the outgoing task)

## Testing
This branch requires new stresstests due to the change in sync logic.


vicsn:
CC https://github.com/ProvableHQ/snarkOS/pull/3410 just in case its useful

kaimast:
I just rebased this on most recent staging, and it would be great to get this merged soon. Maybe @raychu86 can take a look? 

The PR is fairly small, but it does change the sync logic noticeably. So a second, or third, set of eyes to check there are no obvious issues with the change would be great.

node/bft/src/sync/mod.rs
```diff
@@ -280,58 +305,64 @@ impl<N: Network> Sync<N> {
...
         }
+
+        // Send the block requests to peers.
+        self.send_block_requests(requests, sync_peers).await;
+    }
+
+    /// Test-only method to manually trigger block synchronization.
+    /// This combines both request generation and response processing for testing purposes.
+    #[cfg(test)]
+    pub(crate) async fn try_block_sync(&self) {
```
raychu86: nit
```suggestion
    pub(crate) async fn testing_only_try_block_sync_testing_only(&self) {
```

raychu86:
Conceptually seems like a great optimization, to parallelize block requesting and block processing. Logic seems fairly straightforward as well.

Couple questions:
1. In the current logic, we send a `peer_notify` only after we finish processing all of our received blocks. I realize that there are other instances of `peer_notify` like peer disconnects or inbound pings; however, couldn't we start requesting blocks that we haven't received yet?

2. Do you have benches on how much this improves sync speeds?

node/bft/src/sync/mod.rs
kaimast: Done! I also changed `test_update_peer_locators` to use the same naming scheme.

node/bft/src/sync/mod.rs
```diff
@@ -166,27 +169,49 @@ impl<N: Network> Sync<N> {
...
+                if !sleep_time.is_zero() {
+                    tokio::time::sleep(sleep_time).await;
                 }
+
+                // Wait for peer updates or timeout
+                let _ = tokio::time::timeout(Self::SYNC_INTERVAL, self_.block_sync.wait_for_peer_update()).await;
+
+                // Issue block requests to peers.
+                self_.try_issuing_block_requests().await;
+                last_update = now;
```
vicsn: Given that we already slept on line 183, would it make sense to move `last_update = now;` to line 185? And if not, to just use the previous simple approach of always sleeping `SYNC_INTERVAL` amount.

Same applies to the loop below

node/sync/src/block_sync.rs
```diff
@@ -158,8 +158,11 @@ pub struct BlockSync<N: Network> {
     /// The lock used to ensure that [`Self::advance_with_sync_blocks()`] is called by one task at a time.
     advance_with_sync_blocks_lock: TMutex<()>,
 
-    /// Gets notified when there was an update to the locators, a peer disconnected, or we received a new block response.
-    notify: Notify,
+    /// Gets notified when there was an update to the locators or a peer disconnected.
```
vicsn, line 165: Do I read correctly that we issue new block requests if a peer disconnected? Why?

kaimast: When a peer disconnects, it removes all block requests for that peer. In that case, you might need to reissue requests.

Notifying the outgoing task does not necessarily issue new requests, though. It will simply wake up and check if it can issue new requests.

kaimast:
The results from the benchmark look promising (see below). About a 5x speed up for P2P and BFT sync!

The REST API numbers also look at lot better, but that is likely due to the changes in the REST benchmark from #3828. The benchmark for merging that PR failed, due to snapshot issues fixed in this PR.

### Results
| Benchmark suite | Current: 50aa0027ed73326acdfefef30ed12569f8b238e8 | Previous: 83805f68a59640c0861218be289d2a77012165b9 | Ratio |
|-|-|-|-|
| `rest-get-block` | `2.68524245510956` ops/s | `0.65` ops/s | `0.24` |
| `rest-block-height` | `7863.4395405039495` ops/s | `185.18` ops/s | `0.023549491166830062` |
| `p2p-sync` | `1.3` blocks/s | `0.25` blocks/s | `0.19` |
| `p2p-sync-speed-variance` | `1.240252` blocks^2/s^2 | `0.202773` blocks^2/s^2 | `0.16` |
| `bft-sync` | `1.16` blocks/s | `0.2` blocks/s | `0.17` |
| `cdn-sync` | `1.23` blocks/s | `0.43` blocks/s | `0.35` |

This comment was automatically generated by [workflow](https://github.com/ProvableHQ/snarkOS/actions?query=workflow%3ARun%20snarkOS%20Benchmarks) using [github-action-benchmark](https://github.com/marketplace/actions/continuous-benchmark).

node/bft/src/sync/mod.rs
kaimast: The idea was to base the sleep on how long it took to issue the last set of requests to set some sort of lower bound to how often requests can be sent. 

However, I realized we already have a rate limiting mechanism in `send_block_requests` and there is also a long timeout for `wait_for_peer_updates`. So I ended up removing the sleeps entirely and also added more documentation (in commit [0010be3](https://github.com/ProvableHQ/snarkOS/pull/3810/commits/0010be3c30b00ab61be83c058de7a3fe993f22bb)). 

I ran a mainnet node with the changed code and did not see any issue due to frequent requests. My MacBook was able to sync at a steady 20+ blocks/s.

kaimast:
> In the current logic, we send a peer_notify only after we finish processing all of our received blocks. I realize that there are other instances of peer_notify like peer disconnects or inbound pings; however, couldn't we start requesting blocks that we haven't received yet?

I added a mechanism that wakes up the sending task whenever a block request is fully processed (in `BlockSync::remove_block_response`). This should produce the expected behavior that a node issues new requests whenever the number of in-flight requests is less than the maximum.

> Do you have benches on how much this improves sync speeds?

I added a comment above with some benchmark results that show significantly improved sync performance, albeit most of it comes from the faster `get_block` PR we recently merged. 

A noticeable improvement is that syncing now is much more steady instead of having frequent bursts. However, I need to think of how to benchmark this correctly. The sync speed variance benchmark does not capture this well, which might be due to the benchmark setup being much more CPU-limited than my laptop or the VMs I am using.

vicsn:
Triggered stress-tests on branch `prerelease-v4.3.11` - this failed, some validators halted.

vicsn (approved):
Tests passed under prerelease-v4.3.20

PR #3819, opened 2025-09-01
Use a process with loaded imports for deploymeny cost determination

vicsn:
## Motivation

Deployment of programs with imports was recently broken, because the deployment_cost function assumes the process contains all of the imports. This PR ensures we pass in a Process with the loaded imports.

## Test Plan

Manually tested success of nested deployments. In the future we'll have regression tests in CI on all known types of programs.

## Related PRs

https://github.com/ProvableHQ/snarkVM/pull/2910

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4130,7 +4130,7 @@ dependencies = [
 [[package]]
 name = "snarkvm"
 version = "4.1.0"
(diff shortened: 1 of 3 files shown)


ljedrz (commented):
LGTM, but needs a snarkVM update before merging.

ljedrz (approved):

Issue #3580, opened 2025-04-06
[Feature] Automate validator & client ip discovery


vicsn:
## 🚀 Feature

Currently, validators share their own and their core client ip addresses out of band and manually set them as a flag to `snarkos start` via a centralized document (trusted for liveness not safety).

This process has issues:
- liveness risk due to using a centralized process
- the manual update process is error-prone and incurs high latency

The main use-cases to keep in mind:
- how does a new node access the network?
  - Clients already gossip via the bootstrap peers
  - Validator IP addresses should also be gossipped via the bootstrap peers
    - bootstrap clients currently only run a `Router` object. To enable validator connections, we should consider letting them run a `Gateway` object, or creating a special class of `bootstrap validator`, or perhaps there are other approaches.
    - to avoid attacks validators should only believe an IP address **update** when having observed it from at least a quorum's worth of other validators. However, there are open questions around what constitutes a valid validator due to Aleo's support for dynamic committees.
      - validators which are out of sync are somewhat out of luck and may need to be optimistically lenient in order to be able to catch up
      - validators which are at tip should know what the correct committee is. They could in theory determine this by learning the committee's respective block heights.
- when all nodes restart, how can they connect to each other again?
  - The simple answer: a peer cache stored to disk.
- when somehow a quorum's worth of validators are forced to update their IP address at the same time
  - Gossip via the bootstrap peer should be sufficient.
- Are there any other usecases we should keep in mind?

## Implementation
- [x] implement router peer cache and get rid of manual client ip address tracking.
- [x] implement validator peer cache and get rid of manual client ip address tracking.

## Deferred tasks
We can worry about optimizing that core clients connect to each other later. Noting we will probably keep the core client concept for a while.

PR #3820, opened 2025-09-03
[Feat] Persistent peer cache for clients

ljedrz:
This PR introduces a simple, persistent peer cache, which saves high-quality peers on shutdown, and introduces them to the peer pool as candidate peers on startup. The scoring metrics are basically "free", as we already keep track of them.

This was tested locally as follows:
1. 4 `--dev` validators with `--allow-external-peers`, plus one `--dev` client, were started
2. the client was promptly shut down
3. the client had the addresses of all the validators saved at the expected location
4. one of the aforementioned addresses was manually changed
5. the client was restarted
6. the client's logs were inspected for the changed address (it was there)

### DIFF
--- a/node/router/src/lib.rs
+++ b/node/router/src/lib.rs
@@ -52,6 +52,7 @@ use snarkos_node_tcp::{Config, ConnectionSide, P2P, Tcp, is_bogon_ip, is_unspeci
 
 use snarkvm::prelude::{Address, Network, PrivateKey, ViewKey};
 
+use aleo_std::{StorageMode, aleo_ledger_dir};
 use anyhow::{Result, bail};
 #[cfg(feature = "locktick")]
 use locktick::parking_lot::{Mutex, RwLock};
@@ -61,7 +62,9 @@ use parking_lot::{Mutex, RwLock};
 use std::net::IpAddr;
 use std::{
     collections::{HashMap, HashSet},
+    fs,
     future::Future,
+    io::Write,
     net::SocketAddr,
     ops::Deref,
     str::FromStr,
@@ -72,6 +75,9 @@ use tokio::task::JoinHandle;
 
 /// The default port used by the router.
 pub const DEFAULT_NODE_PORT: u16 = 4130;
+
+/// The name of the file containing cached peers.
+const PEER_CACHE_FILENAME: &str = "cached_router_peers";
 
 /// The router keeps track of connected and connecting peers.
 /// The actual network communication happens in Inbound/Outbound,
@@ -108,6 +114,8 @@ pub struct InnerRouter<N: Network> {
     rotate_external_peers: bool,
     /// If the flag is set, the node will engage in P2P gossip to request more peers.
     allow_external_peers: bool,
+    /// The storage mode.
+    storage_mode: StorageMode,
     /// The boolean flag for the development mode.
     is_dev: bool,
 }
@@ -138,16 +146,27 @@ impl<N: Network> Router<N> {
         max_peers: u16,
         rotate_external_peers: bool,
         allow_external_peers: bool,
+        storage_mode: StorageMode,
         is_dev: bool,
     ) -> Result<Self> {
         // Initialize the TCP stack.
         let tcp = Tcp::new(Config::new(node_ip, max_peers));
 
-        let trusted_peers = trusted_peers
+        let mut initial_peers = trusted_peers
             .iter()
             .copied()
             .map(|addr| (addr, Peer::new_candidate(addr, true)))
             .collect::<HashMap<_, _>>();
+
+        let mut peer_cache_path = aleo_ledger_dir(N::ID, &storage_mode);
+        peer_cache_path.push(PEER_CACHE_FILENAME);
+        if let Ok(cached_peers_str) = fs::read_to_string(&peer_cache_path) {
+            for peer_addr_str in cached_peers_str.lines() {
+                if let Ok(addr) = SocketAddr::from_str(peer_addr_str) {
+                    initial_peers.insert(addr, Peer::new_candidate(addr, false));
+                }
+            }
+        }
 
         // Initialize the router.
         Ok(Self(Arc::new(InnerRouter {
@@ -157,10 +176,11 @@ impl<N: Network> Router<N> {
             ledger,
             cache: Default::default(),
             resolver: Default::default(),
-            peer_pool: RwLock::new(trusted_peers),
+            peer_pool: RwLock::new(initial_peers),
             handles: Default::default(),
             rotate_external_peers,
             allow_external_peers,
+            storage_mode,
             is_dev,
         })))
     }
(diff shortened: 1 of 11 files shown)


node/router/src/lib.rs
```diff
@@ -138,17 +146,28 @@ impl<N: Network> Router<N> {
...
             .copied()
             .map(|addr| (addr, Peer::new_candidate(addr, true)))
             .collect::<HashMap<_, _>>();
 
+        let mut peer_cache_path = aleo_ledger_dir(N::ID, &storage_mode);
+        peer_cache_path.push(PEER_CACHE_FILENAME);
+        if let Ok(cached_peers_str) = fs::read_to_string(&peer_cache_path) {
+            for peer_addr_str in cached_peers_str.lines() {
+                if let Ok(addr) = SocketAddr::from_str(peer_addr_str) {
+                    initial_peers.insert(addr, Peer::new_candidate(addr, false));
```
vicsn, line 166: Lazy question from mobile review: what does false do?

node/router/src/lib.rs
```diff
@@ -528,9 +548,51 @@ impl<N: Network> Router<N> {
...
+            let peer_ip = peer.listener_addr().ip();
+            if let Some(peer_stats) = known_peers.get(&peer_ip) {
+                (peer_stats.sent().0 as f64) > avg_sent_msgs * 0.75
+                    && (peer_stats.received().0 as f64) > avg_recv_msgs * 0.75
+                    && (peer_stats.failures() as f64) <= avg_failures
+            } else {
+                false
+            }
+        });
+        peers.truncate(MAX_PEERS_TO_SEND);
```
vicsn, line 584: I think it would be more robust to always try and retain MAX_PEERS_TO_SEND, ordered by their score. Failures being the most important I guess? But up to you how to order.

Then we don't have to make up magic numbers in the decision and don't have to worry about edge cases. Its cheap to store 256 lines, and we can try to connect just to the best initial ones at startup.

node/router/src/lib.rs
```diff
@@ -528,9 +548,51 @@ impl<N: Network> Router<N> {
...
+
+        // Dump the connected peers to a file.
+        let mut path = aleo_ledger_dir(N::ID, &self.storage_mode);
+        path.push(PEER_CACHE_FILENAME);
+        let mut file = fs::File::create(path)?;
```
vicsn, line 589: Will this succeed if the file already exists? (I think it should, think about multiple restarts)

Will we overwrite it with the new peers? (I think it should, otherwise the file has tons of old peers)

vicsn (changes requested):

vicsn:
Can you also report in the PR description how you manually tested on a local devnet?

node/router/src/lib.rs
ljedrz: it means that the peer is not trusted; we don't create peers often, but if you'd like, I could introduce something like
```
enum PeerTrustLevel {
    Trusted,
    Untrusted,
}
```
for maximum clarity

node/router/src/lib.rs
ljedrz: Yes, those were the goals:
> This function will create a file if it does not exist, and will truncate it if it does.

node/router/src/lib.rs
ljedrz: Sounds good; I can sort by the number of failures (ascending), as the other metrics aren't as obvious - a relatively low number of sent/received files might mean the node is not spamming, or that it doesn't "apply itself" enough, and there's no way of telling. We could introduce a more advanced statistical approach, but it would probably be an overkill.

vicsn (approved):

kaimast:
One side effect of this change is that sharing a ledger snapshot with another node now includes the cache. Maybe we should document this somewhere?

For example, I ran the sync benchmarks just now and saw these warnings (the benchmark runs with only two nodes, but the snapshot was created with 40)
```
2025-09-27T16:45:51.096469Z  WARN snarkos_node_router: Unable to connect to '127.0.0.1:4157' - already connecting to node at 127.0.0.1:4157
2025-09-27T16:45:51.096497Z  WARN snarkos_node_router: Unable to connect to '127.0.0.1:4154' - I/O error: Connection refused (os error 61)
2025-09-27T16:45:51.096501Z  WARN snarkos_node_router: Unable to connect to '127.0.0.1:4155' - I/O error: Connection refused (os error 61)
2025-09-27T16:45:51.096530Z  WARN snarkos_node_router: Unable to connect to '127.0.0.1:4168' - I/O error: Connection refused (os error 61)
2025-09-27T16:45:51.096736Z  WARN snarkos_node_router: Unable to connect to '127.0.0.1:4148' - I/O error: Connection refused (os error 61)
2025-09-27T16:45:51.096744Z  WARN snarkos_node_router: Unable to connect to '127.0.0.1:4157' - I/O error: Connection refused (os error 61)
2025-09-27T16:45:51.096637Z  WARN snarkos_node_router: Unable to connect to '127.0.0.1:4161' - I/O error: Connection refused (os error 61)
2025-09-27T16:45:51.096763Z  WARN snarkos_node_router: Unable to connect to '127.0.0.1:4140' - I/O error: Connection refused (os error 61)
2025-09-27T16:45:51.096753Z  WARN snarkos_node_router: Unable to connect to '127.0.0.1:4141' - I/O error: Connection refused (os error 61)
2025-09-27T16:45:51.096708Z  WARN snarkos_node_router: Unable to connect to '127.0.0.1:4132' - I/O error: Connection refused (os error 61)
2025-09-27T16:45:51.096833Z  WARN snarkos_node_router: Unable to connect to '127.0.0.1:4146' - I/O error: Connection refused (os error 61)
2025-09-27T16:45:51.096843Z  WARN snarkos_node_router: Unable to connect to '127.0.0.1:4150' - I/O error: Connection refused (os error 61)
2025-09-27T16:45:51.096843Z  WARN snarkos_node_router: Unable to connect to '127.0.0.1:4147' - I/O error: Connection refused (os error 61)
2025-09-27T16:45:51.096779Z  WARN snarkos_node_router: Unable to connect to '127.0.0.1:4136' - I/O error: Connection refused (os error 61)
2025-09-27T16:45:51.096847Z  WARN snarkos_node_router: Unable to connect to '127.0.0.1:4138' - I/O error: Connection refused (os error 61)
2025-09-27T16:45:51.096856Z  WARN snarkos_node_router: Unable to connect to '127.0.0.1:4135' - I/O error: Connection refused (os error 61)
2025-09-27T16:45:51.096825Z  WARN snarkos_node_router: Unable to connect to '127.0.0.1:4163' - I/O error: Connection refused (os error 61)
```

Alternatively, we could change the warnings to be more descriptive. Failing to connect to a candidate peer that is not trusted or a bootstrap node is not necessarily an indication that something went wrong. The candidate peer might just not be online anymore or accept new connections.

vicsn:
> Maybe we should document this somewhere?

It will be documented in the upcoming release notes. 

> For example, I ran the sync benchmarks just now and saw these warnings (the benchmark runs with only two nodes, but the snapshot was created with 40)

Awesome, the cache works! And yes the log level can be downgraded to debug.

PR #3829, opened 2025-09-06
Ensure txs are broadcast either if not in sync, or when processed

vicsn:
## Motivation

Previous behaviour before https://github.com/ProvableHQ/snarkOS/pull/3825 meant transactions were not broadcast when not synced, a very common occurance and bad UX.

This PR ensures txs are broadcast either if not in sync, or when processed.

Open question: should being out of sync return a 503 `service_unavailable` or a 202 `accepted`?

## Test Plan

non-trivial broadcast success rates in automated stress tests should suffice.

### DIFF
--- a/node/rest/src/routes.rs
+++ b/node/rest/src/routes.rs
@@ -629,9 +629,7 @@ impl<N: Network, C: ConsensusStorage<N>, R: Routing<N>> Rest<N, C, R> {
             }
             Err(other_rejection) => return Err(other_rejection.into()),
         };
-
-        // Determine transaction type.
-        let is_exec = tx.is_execute();
+        let tx_id = tx.id();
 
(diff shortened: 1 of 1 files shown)


kaimast:
I am not sure how easy it is to parse the returned error in the current setting. A node might return `service unavailable` for unrelated reasons, and matching against the error message string is a little cumbersome.

It might be better to return 200 in either case and a JSON document in the body that contains not only the transaction ID, but also information about whether it was successfully checked and/or broadcasted.
That would break the API, but we could use the recent bump to v2 (which has not reached mainnet AFAIK) to make the transition easier, or add a flag to enable this behavior.

vicsn:
Closing in favour of: https://github.com/ProvableHQ/snarkOS/pull/3832

PR #3831, opened 2025-09-08
[CI] Tests for unconfirmed endpoint

kaimast:
This is a low-cost test that will prevent regressions in the future. I was not able to add this before #3830 was merged.


kaimast:
There seems to be another issue with `cargo audit`. This time it is for `tracing-subscriber` and I don't see an easy way to resolve it. 

All other tests pass.

vicsn (dismissed):
Can you change the base to staging? Thx!

kaimast:
> Can you change the base to staging? Thx!

I added the same commits to #3828, so I will just close this PR in favor of the former.

PR #3837, opened 2025-09-09
[Fix] Don't panic if peer upgrade fails

kaimast:
There seem to some cases where a peer's gets downgraded to candidate, while the handshake is still in progress. In that case, a node should not panic but simply throw an error message.

(Includes changes from #3835)

### DIFF
--- a/node/router/src/helpers/peer.rs
+++ b/node/router/src/helpers/peer.rs
@@ -16,6 +16,7 @@
 use crate::{NodeType, Router, messages::ChallengeRequest};
 use snarkvm::prelude::{Address, Network};
 
+use anyhow::{Result, ensure};
 use std::{net::SocketAddr, time::Instant};
 
 /// A peer of any connection status.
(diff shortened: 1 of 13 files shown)


node/router/src/helpers/peer.rs
```diff
@@ -82,9 +83,14 @@ impl<N: Network> Peer<N> {
...
-    pub fn upgrade_to_connected(&mut self, connected_addr: SocketAddr, cr: &ChallengeRequest<N>, router: Router<N>) {
+    pub fn upgrade_to_connected(
+        &mut self,
+        connected_addr: SocketAddr,
+        cr: &ChallengeRequest<N>,
+        router: Router<N>,
+    ) -> Result<()> {
         // Logic check: this can only happen during the handshake.
-        assert!(matches!(self, Self::Connecting(_)));
+        ensure!(matches!(self, Self::Connecting(_)), "Peer is not in `connecting` state");
```
raychu86, line 93: It seems to me that this is the only change relevant to the PR title. 

Are the log changes related/necessary?

kaimast:
Closing this in favor of #3842.

node/router/src/helpers/peer.rs
kaimast: I ended up adding a lot of logging to this PR while we were debugging testnet issues. Sorry for the confusion.

Issue #3797, opened 2025-08-20
[Proposal] Unify underlying CLI capabilities


raychu86:
## 💥 Proposal

There should be a unified crate or utility where the `snarkOS`, `leo`, and sdk CLIs can pull shared functionality from.

For example, both `snarkOS` and `leo` CLIs have the following capabilities:
- account creation
- execute
- deploy

They are all different in implementation and are maintained separately. In addition, it would be nice for `snarkOS` to have the `leo query` capabilities as well, but it should not need to fetch it from `leo`.

Unifying all these impls will reduce code duplication and make maintaining the codebases easier. An example could be updating `Query` in `snarkVM` to support these and propagated up to `leo` and `snarkOS`.

raychu86:
@d0cd @kaimast ^From our CLI discussions in https://github.com/ProvableHQ/snarkOS/pull/3759#discussion_r2289244686

kaimast:
In #3789 I am also adding serialization of `anyhow::Error` to the API (see [here](https://github.com/ProvableHQ/snarkOS/pull/3789/files#diff-fbe1fdc35fe9fe6a93da50c81e90f7f528a1a35dce4883a53942592a63cc54edR86)). A unified crate should include this functionality as well. It could rely on functionality introduced in [snarkVM PR #2813](https://github.com/ProvableHQ/snarkVM/pull/2813).

We also need to set up a process to keep the endpoints, client library, and API documentation in sync. There are some crates that allow generating an OpenAPI spec from `axum` code, e.g., [utopia](https://docs.rs/utoipa/latest/utoipa). In theory, this would be great as it can automatically create documentation and Rust client code for the endpoints, but it might not work for our specific case.

vicsn:
I'll note that we should have a small design document or more specific scoped specification in this issue for this before we move forward with any implementation (as we should have had for compact blocks and other features too ;) )

PR #3841, opened 2025-09-12
[Feat] Implement `snarkos account import`

### LINKED ISSUE #3776: [Feature] snarkOS account [import]
## 🚀 Feature

A `snarkOS account import` feature would allow for better usability for deriving view keys and addresses from private keys. We should also have an import from view key to address as well.

## Motivation

There is currently no `import` feature on the `snarkOS account` CLI. We can match the functionality in the `leo` CLI, which already has an import.

raychu86:
## Motivation

This PR implements `snarkos account import` to allow derivation of view key and address from a given private key.

Closes https://github.com/ProvableHQ/snarkOS/issues/3776

### DIFF
--- a/cli/src/commands/account.rs
+++ b/cli/src/commands/account.rs
@@ -568,6 +568,94 @@ mod tests {
     }
 
     #[test]
+    fn test_import() -> Result<()> {
+        for _ in 0..3 {
+            let account = Account::Import { network: 0, private_key: None, discreet: false, save_to_file: None };
+            account.parse().with_context(|| "Account import failed")?;
+        }
+
+        let mut expected = format!(
+            " {:>12}  {}\n",
+            "Private Key".cyan().bold(),
+            "APrivateKey1zkp2n22c19hNdGF8wuEoQcuiyuWbquY6up4CtG5DYKqPX2X"
+        );
+        expected += &format!(
+            " {:>12}  {}\n",
+            "View Key".cyan().bold(),
+            "AViewKey1pNxZHn79XVJ4D2WG5Vn2YWsAzf5wzAs3dAuQtUAmUFF7"
+        );
+        expected += &format!(
+            " {:>12}  {}",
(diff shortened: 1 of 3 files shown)


cli/src/commands/account.rs
```diff
@@ -68,6 +68,21 @@ pub enum Account {
         #[clap(long)]
         save_to_file: Option<String>,
     },
+    /// Derive an Aleo account from a private key
+    Import {
+        /// Account private key
+        private_key: Option<String>,
```
kaimast, line 74: I assume this mirrors the implementation in leo, but it would be good to add a group here that requires `private_key` or `discreet` to be set.

cli/src/commands/account.rs
```diff
@@ -68,6 +68,21 @@ pub enum Account {
...
+        /// Specify the network to create an execution for.
+        /// [options: 0 = mainnet, 1 = testnet, 2 = canary]
+        #[clap(long, default_value_t=MainnetV0::ID, long, value_parser = network_id_parser())]
+        network: u16,
+        /// Print sensitive information (such as the private key) discreetly in an alternate screen
+        #[clap(long)]
+        discreet: bool,
+        /// Specify the path to a file where to save the account in addition to printing it
+        #[clap(long)]
+        save_to_file: Option<String>,
```
kaimast, line 84: You could also add a `conflicts_with="discreet"` here.

raychu86: Addressed with [2fbc9ad](https://github.com/ProvableHQ/snarkOS/pull/3841/commits/2fbc9add260035dac4809e2ae9ed478a839daf83)

cli/src/commands/account.rs
raychu86: I'd like to leave that to https://github.com/ProvableHQ/snarkOS/issues/3797 for now unless you feel that we should add it here.

vicsn (approved):

Issue #3844, opened 2025-09-15
[Feature] Sort peers by height


vicsn:
## 🚀 Feature

We need a few more pieced in order to tackle: https://github.com/ProvableHQ/snarkOS/issues/3580

Clients should make sure that they're communicating with the most advanced peers in order to always sync with the network.

Could we do something like the following, without breaking too many abstractions:
1. We track a `last_height_seen` metric for each peer we connected to.
2. Whenever we save to disk, attempt to order the peers not just by reputation but also by the optional last seen height.
3. Whenever we submit a `PeerResponse` in `fn peer_request`, attempt to order the peers by reputation and by the optional last seen height. Whenever we receive a `PeerResponse`, insert in the `peer_pool` sorted by optional `last_height_seen`, so that when we fetch `candidate_peers` we choose:
  a. `num_deficient - 1` peers which have the highest last seen height (assuming honest reporting)
  b. 1 peer randomly from the rest of the set (in order to protect against dishonest reporting)

PR #3848, opened 2025-09-16
[Feat] Track peer height for cache purposes

ljedrz:
This PR extends the `ConnectedPeer` object with a new member representing its last known block height, as informed by its `BlockLocators`, and uses it to sort the entries in the persistent peer cache, and those sent in response to `PeerRequest` messages.

For now I've omitted the last part of the proposed implementation, as I think it was intended to be related to the `PeerRequest` messages we _receive_, not provide. That being said, this is something that could be considered a hardening feature separate from these changes, and as such, it could be introduced in another PR.

CC https://github.com/ProvableHQ/snarkOS/issues/3844

### DIFF
--- a/node/router/src/inbound.rs
+++ b/node/router/src/inbound.rs
@@ -196,7 +196,7 @@ pub trait Inbound<N: Network>: Reading + Outbound<N> {
                 }
                 // If the peer is a prover, ensure there are no block locators.
                 else if message.node_type.is_prover() && message.block_locators.is_some() {
-                    bail!("Peer '{peer_ip}' is a prover or client, but block locators were provided");
+                    bail!("Peer '{peer_ip}' is a prover, but block locators were provided");
                 }
 
                 // Process the ping message.
--- a/node/router/src/inbound.rs
+++ b/node/router/src/inbound.rs
@@ -36,7 +36,7 @@ use snarkvm::prelude::{
 
 use anyhow::{Result, anyhow, bail};
 use snarkos_node_tcp::is_bogon_ip;
-use std::net::SocketAddr;
+use std::{cmp, net::SocketAddr};
 use tokio::task::spawn_blocking;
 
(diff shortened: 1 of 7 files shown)


node/router/src/inbound.rs
```diff
@@ -314,21 +314,34 @@ pub trait Inbound<N: Network>: Reading + Outbound<N> {
...
-                .filter(|ip| *ip != peer_ip && self.router().is_valid_peer_ip(ip))
-                .take(MAX_PEERS_TO_SEND)
-                .collect(),
-        };
+        });
+        // Get the low-level peer stats.
+        let known_peers = self.tcp().known_peers().snapshot();
+
+        // Sort the prospect peers.
+        peers.sort_unstable_by_key(|peer| {
```
vicsn, line 331: Given that we have two exact replica's, shall we extract into a `sort_peers_by_quality` function?

vicsn (approved):
LGTM, minor cleanup idea.

> For now I've omitted the last part of the proposed implementation, as I think it was intended to be related to the PeerRequest messages we receive, not provide. 

I think the item is worded correctly, note that it says `PeerResponse`, but indeed you can tackle this in a follow-up PR.

ljedrz:
> I think the item is worded correctly, note that it says PeerResponse, but indeed you can tackle this in a follow-up PR.

Ah, so it was just lumped together; I just saw `peer_request` in tandem with `num_deficient`, and errored out.

node/router/src/inbound.rs
ljedrz: I thought about it, but unfortunately, they are different, as one operates on `Peer`s, and the other on `ConnectedPeer`s. They are very similar at first glance, but the types differ - note that one uses a method, and the other directly references the `listening_addr` or `last_height_seen`.

vicsn: Then leave it

PR #3860, opened 2025-09-17
[Feat] Filter out lower-height peers from PeerResponse messages

### LINKED ISSUE #3844: [Feature] Sort peers by height

ljedrz:
A follow-up to https://github.com/ProvableHQ/snarkOS/pull/3848, completing #3844.

This PR enhances the `PeerResponse` messages with optional peer heights, which are then used to filter out candidate peers which are lower than equal to our node in terms of their current block height.

Edit: the most recent iteration adds a `last_height_seen` field to the `CandidatePeer`, and updates it whenever we receive a `PeerResponse`; we later use it in order to make sure that at least half of the peers we attempt to connect to (in case we have too few of them) are at a greater height than us.

Closes https://github.com/ProvableHQ/snarkOS/issues/3844


node/router/messages/src/peer_response.rs
```diff
@@ -39,20 +39,56 @@ impl ToBytes for PeerResponse {
...
         }
         Ok(())
     }
 }
 
 impl FromBytes for PeerResponse {
     fn read_le<R: io::Read>(mut reader: R) -> io::Result<Self> {
-        let count = u8::read_le(&mut reader)?;
+        // Read the peer count if their heights aren't present; otherwise, interpret this value
+        // as the message version. It is a workaround for a currently missing version value.
```
vicsn, line 64: I think it's worth also documenting the worst case scenario. If I understand correctly: for nodes which don't upgrade, and which have just a single peer, that peer may not be able to deserialize the empty `PeerResponse`. It would be nice to have a testcase for the success and failure cases, perhaps at either of these places: 
- https://github.com/ProvableHQ/snarkOS/blob/mainnet/node/router/messages/src/helpers/codec.rs#L123
- https://github.com/ProvableHQ/snarkOS/tree/mainnet/node/router/messages/src/lib.rs#L336

node/router/messages/src/peer_response.rs
```diff
@@ -39,20 +39,56 @@ impl ToBytes for PeerResponse {
             return Err(io::Error::new(io::ErrorKind::InvalidInput, format!("Too many peers: {}", self.peers.len())));
         }
 
+        // A version indicator; we don't expect empty peer responses, so a zero value can serve
+        // as an indicator that this message is to be processed differently. The version value
+        // can be changed to a 2 in the future, once everyone expects it there.
+        0u8.write_le(&mut writer)?;
```
raychu86, line 45: This will require either a synchronous restart or a dedicated rollout based on a `Message::Version`.

raychu86: Ah, I didn't read the comment. We would then need to enforce that we never send 0 as the peer length

node/router/src/inbound.rs
```diff
@@ -340,25 +340,42 @@ pub trait Inbound<N: Network>: Reading + Outbound<N> {
...
-    fn peer_response(&self, _peer_ip: SocketAddr, peers: &[SocketAddr]) -> bool {
+    fn peer_response(&self, _peer_ip: SocketAddr, peers: &[(SocketAddr, Option<u32>)]) -> bool {
         // Check if the number of peers received is less than MAX_PEERS_TO_SEND.
         if peers.len() > MAX_PEERS_TO_SEND {
             return false;
         }
-        // Filter out invalid addresses.
+        // Prepare a height filtering closure.
+        let own_height = self.router().ledger.latest_block_height();
+        let height_check = |peer_height: Option<u32>| -> bool {
```
raychu86: Do we use `MINIMUM_NUMBER_OF_PEERS` for anything other than pruning peers? 

If so, this may prevent nodes that are sufficiently caught up (close to tip) from finding peers to connect to.

node/router/messages/src/peer_response.rs
vicsn: It's not ideal, but I think this will not cause major issues: https://github.com/ProvableHQ/snarkOS/pull/3860#discussion_r2355764518

node/router/messages/src/peer_response.rs
ljedrz: done in https://github.com/ProvableHQ/snarkOS/pull/3860/commits/da7452e63a9ec240d6b791f879dbf69118b4f3a7

node/router/src/inbound.rs
vicsn: > Do we use MINIMUM_NUMBER_OF_PEERS for anything other than pruning peers?

I don't understand the question

> If so, this may prevent nodes that are sufficiently caught up (close to tip) from finding peers to connect to.

I think it is likely fine. A node which is at tip will likely have plenty of active peers and peers in its peer cache (which we now also persist to disk). A node which is not at tip will keep collecting more peers. A node without any peers will default to the bootstrap peers, and re-enter the network.

But more important than peer collection, really, is prioritizing connection to peers with higher observed heights in the heartbeat. @ljedrz could you find a simple approach to this still?

vicsn: @ljedrz you mentioned the complexity is annoying. But maybe whenever we fetch peers, and also periodically, we can re-order the peer pool by observed height... Or we keep two separate peer_pools: one where we have an observation above our height, and one for the rest.

vicsn: After we feel confident, I'm also running a 120 client devnet to test the peering behaviour

ljedrz: @vicsn if the goal is to further increase the likelihood of being connected to high-quality peers, the next thing we could do that is IMO harder to "game" is dropping peers with the lowest/unknown height when we rotate them during the heartbeat

vicsn: > @vicsn if the goal is to further increase the likelihood of being connected to high-quality peers, the next thing we could do that is IMO harder to "game" is dropping peers with the lowest/unknown height when we rotate them during the heartbeat.

The downside of this is that it might negatively impact the experience of peers which are behind, without significantly improving our connectivity to higher peers. Random selection from the 10.000-sized `peer_pool` does not guarantee getting higher peers.

So I think we should attempt to connect to **some** peers which (ideally recently) were announced to be above us. With two separate `peer_pool`s: one where we have an observation above our height, and one for the rest. To make this more robust, we should not simply discard peers when `peer_pool.contains_key(peer_ip)`, but re-insert appropriately based on whether the newly observed height is above or below us.

- If all peers are honest, this will ensure peers try to stay connected to some peers which are above us.
- If some peers are dishonest, their influence is only marginal because we only rotate **some** and we only select based on whether they are above us, not based on how far above us.

vicsn (dismissed):
LGTM! Looking forward to the other reviews, after which we can start testing to observe sparsely connected client connectivity

ljedrz:
Rebased.

@vicsn If possible, I'd prefer https://github.com/ProvableHQ/snarkOS/pull/3905 to be merged first, as this PR will be a lot easier to rebase.

node/router/src/inbound.rs
```diff
@@ -341,25 +341,25 @@ pub trait Inbound<N: Network>: Reading + Outbound<N> {
...
 
     /// Handles a `PeerResponse` message.
-    fn peer_response(&self, _peer_ip: SocketAddr, peers: &[SocketAddr]) -> bool {
+    fn peer_response(&self, _peer_ip: SocketAddr, peers: &[(SocketAddr, Option<u32>)]) -> bool {
         // Check if the number of peers received is less than MAX_PEERS_TO_SEND.
         if peers.len() > MAX_PEERS_TO_SEND {
             return false;
         }
-        // Filter out invalid addresses.
+        // Filter out invalid addresses and peers with a lower known block height.
```
kaimast: Where does the filtering of lower block height happen? I might be missing something.

ljedrz: No, you're right, it must have been some earlier change that got superseded.

node/router/src/lib.rs
```diff
@@ -393,13 +405,19 @@ pub trait PeerPoolHandling<N: Network>: P2P {
...
-    fn candidate_peers(&self) -> Vec<SocketAddr> {
+    fn get_candidate_peers(&self) -> Vec<CandidatePeer> {
         let banned_ips = self.tcp().banned_peers().get_banned_ips();
         self.peer_pool()
             .read()
             .iter()
             .filter_map(|(addr, peer)| {
-                (matches!(peer, Peer::Candidate(_)) && !banned_ips.contains(&addr.ip())).then_some(*addr)
+                if let Peer::Candidate(peer) = peer
+                    && !banned_ips.contains(&addr.ip())
```
ljedrz: actually, this is no longer needed after https://github.com/ProvableHQ/snarkOS/pull/3905; I'll remove it shortly

node/router/src/lib.rs
```diff
@@ -212,15 +212,16 @@ pub trait PeerPoolHandling<N: Network>: P2P {
...
+    fn insert_candidate_peers(&self, mut listener_addrs: Vec<(SocketAddr, Option<u32>)>) {
         // Hold a write guard from now on, so as not to accidentally slash multiple times
         // based on multiple batches of candidate peers, and to not overwrite any entries.
         let mut peer_pool = self.peer_pool().write();
 
         // Perform filtering to ensure candidate validity.
-        listener_addrs.retain(|&addr| {
-            !peer_pool.contains_key(&addr)
+        listener_addrs.retain(|&(addr, _)| {
+            peer_pool.get(&addr).map(|peer| peer.is_candidate()).unwrap_or(true)
```
ljedrz: note for reviewers: we now _do_ want to potentially update candidates; we should drop such entries without a corresponding height, though

node/router/src/lib.rs
ljedrz: I'll update the calculation soon - updates shouldn't be counted

node/router/messages/src/peer_response.rs
vicsn: In more detail:
- new peers receiving an empty PeerResponse from old peers, or old peers receiving a PeerResponse from new peers, should gracefully fail to parse the message without causing any ip banning
- new peers can still receive non-empty PeerResponses from old peers.

vicsn (approved):
LGTM!

PR #3863, opened 2025-09-18
[Tweak] Disallow marking bootstrap peers as trusted

### LINKED ISSUE #3632: [Bug] Do not allow bootstrap peers to be trusted peers
## 🐛 Bug Report

There appears not to be any code that prevents a peer from being both a bootstrap peer and a trusted peer.  If it is both, then there is suboptimal code that can cause the peer to be disconnected and then immediately reconnected (possibly with a gap of being able to use it). This can happen if two bootstrap peers happen to be connected, in which case handle_bootstrap_peers(), (which is only called from heartbeat()), will disconnect a random one of them.  If the disconnected one is a trusted peer, the next function called from heartbeat() is handle_trusted_peers(), which will initiate a reconnection attempt.  We should fix this.

ljedrz:
Closes https://github.com/ProvableHQ/snarkOS/issues/3632.

I tested it locally.

### DIFF
--- a/node/router/src/lib.rs
+++ b/node/router/src/lib.rs
@@ -153,12 +153,14 @@ impl<N: Network> Router<N> {
         // Initialize the TCP stack.
         let tcp = Tcp::new(Config::new(node_ip, max_peers));
 
+        // Add the trusted peers to the peer pool
         let mut initial_peers = trusted_peers
             .iter()
             .copied()
             .map(|addr| (addr, Peer::new_candidate(addr, true)))
             .collect::<HashMap<_, _>>();
 
+        // Load additional peers from the peer cache (if present).
         let mut peer_cache_path = aleo_ledger_dir(N::ID, &storage_mode);
         peer_cache_path.push(PEER_CACHE_FILENAME);
         if let Ok(cached_peers_str) = fs::read_to_string(&peer_cache_path) {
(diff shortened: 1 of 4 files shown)


node/router/src/lib.rs
```diff
@@ -183,7 +185,25 @@ impl<N: Network> Router<N> {
...
             is_dev,
-        })))
+        }));
+
+        // Ensure there are no bootstrappers among the trusted peers.
+        {
+            let bootstrap_peers = router.bootstrap_peers();
+            let mut peer_pool = router.peer_pool.write();
+            let initial_peer_count = peer_pool.len();
+            peer_pool.retain(|addr, peer| !(peer.is_trusted() && bootstrap_peers.contains(addr)));
```
vicsn: This check should be in start.rs so its for both validators and clients, and so it doesnt include peers stored to disk added to the peer_pool, I think

ljedrz: That would require `bootstrap_peers` to become a free function (it is currently a method on `Router`); I can change it if it's acceptable.

vicsn: Yes that's ok!

ljedrz: Done in https://github.com/ProvableHQ/snarkOS/pull/3863/commits/967977b2e910babeb95d80d0a632598786ec0663.

vicsn (approved):

PR #3876, opened 2025-09-22
[CI] Measure speed of retrieving the most recent block

kaimast:
The most recent block should be cached at some layer in snarkOS. This PR adds a benchmark to see if fetching it is actually fast.


kaimast:
There's a new issue with `cargo audit` failing unrelated to this PR...

vicsn (approved):

kaimast:
Benchmark succeeded. Looks like the most recent blocks can be retrieved about 500 times faster than others!

PR #3877, opened 2025-09-23
[Feat] Allow trusted validator addresses to include hostnames

ljedrz:
Technically more of a refactoring than a feature - it just extends the hostname resolution that we've been doing to `--peers` to `--validators`.


vicsn (changes requested):
Lpgic LGTM. Can you rebase on and target canary?

vicsn (approved):

Issue #3872, opened 2025-09-19
[Proposal] Consolidate the Resolvers


ljedrz:
We currently have 2 different `Resolver` objects - one for the `Router`, and one for the `BFT`. We could instead just have one, as there is an overlap between them.

Issue #3869, opened 2025-09-18
[Feature] Ensure gateway can persist peer cache


vicsn:
## 🚀 Feature

The internals may look a bit different from the client, in that e.g. the same validator can't have two ip addresses

vicsn:
CC https://github.com/ProvableHQ/snarkOS/issues/3580

PR #3879, opened 2025-09-23
[Refactor] Align peer handling between Router and Gateway

### LINKED ISSUE #3873: [Proposal] Introduce a peer pool for the validators
This was recently done for clients in https://github.com/ProvableHQ/snarkOS/pull/3713; the validators would also benefit from such a change.

ljedrz:
The `Router` and the `Gateway` have very similar functionalities, so we can benefit from aligning and deduplicating some code between them. This PR was mostly inspired by https://github.com/ProvableHQ/snarkOS/issues/3869 (which would otherwise require duplicated code), but other than that, it will allow both objects to benefit from similar bugfixes and optimizations (e.g. those from https://github.com/ProvableHQ/snarkOS/pull/3540 and https://github.com/ProvableHQ/snarkOS/pull/3713, which had only targeted the `Router`, but apply to the `Gateway` as well).

The commits may be read individually (the code compiles after each one) or together. The short list of changes is as follows:
- decouple the `Peer` from the `Router`
- simplify peer handling in the `Gateway` by leveraging the `Peer`
- introduce a `PeerPoolHandling` trait which contains methods applicable to both `Router` and `Gateway`

This integration can be pushed further (CC https://github.com/ProvableHQ/snarkOS/issues/3872), but I wanted to limit the scope of changes for a better reviewing experience. This PR also doesn't contain https://github.com/ProvableHQ/snarkOS/issues/3869 just yet, but it will now be trivial to follow up with it.

Closes https://github.com/ProvableHQ/snarkOS/issues/3873.

### DIFF
--- a/node/bft/src/gateway.rs
+++ b/node/bft/src/gateway.rs
@@ -42,7 +42,7 @@ use snarkos_node_bft_events::{
     ValidatorsResponse,
 };
 use snarkos_node_bft_ledger_service::LedgerService;
-use snarkos_node_router::{ConnectedPeer, NodeType, Peer};
+use snarkos_node_router::{NodeType, Peer, PeerPoolHandling};
 use snarkos_node_sync::{MAX_BLOCKS_BEHIND, communication_service::CommunicationService};
 use snarkos_node_tcp::{
     Config,
@@ -74,7 +74,7 @@ use rand::seq::{IteratorRandom, SliceRandom};
 #[cfg(not(any(test)))]
 use std::net::IpAddr;
 use std::{
-    collections::{HashMap, HashSet, hash_map::Entry},
+    collections::{HashMap, HashSet},
     future::Future,
     io,
     net::{Ipv4Addr, SocketAddr, SocketAddrV4},
@@ -148,6 +148,12 @@ pub struct Gateway<N: Network> {
     handles: Arc<Mutex<Vec<JoinHandle<()>>>>,
     /// The development mode.
     dev: Option<u16>,
+}
+
+impl<N: Network> PeerPoolHandling<N> for Gateway<N> {
+    fn peer_pool(&self) -> &RwLock<HashMap<SocketAddr, Peer<N>>> {
+        &self.peer_pool
+    }
 }
 
 impl<N: Network> Gateway<N> {
@@ -360,25 +366,15 @@ impl<N: Network> Gateway<N> {
         // Retrieve the peer IP of the given address.
         match self.resolver.read().get_peer_ip_for_address(address) {
             // Determine if the peer IP is connected.
-            Some(peer_ip) => self.is_connected_ip(peer_ip),
+            Some(peer_ip) => self.is_connected(peer_ip),
             None => false,
         }
-    }
-
-    /// Returns `true` if the node is connected to the given peer IP.
-    pub fn is_connected_ip(&self, ip: SocketAddr) -> bool {
-        self.peer_pool.read().get(&ip).is_some_and(|peer| peer.is_connected())
-    }
-
-    /// Returns `true` if the node is connecting to the given peer IP.
-    pub fn is_connecting_ip(&self, ip: SocketAddr) -> bool {
-        self.peer_pool.read().get(&ip).is_some_and(|peer| peer.is_connecting())
     }
 
     /// Returns `true` if the given peer IP is an authorized validator.
     pub fn is_authorized_validator_ip(&self, ip: SocketAddr) -> bool {
         // If the peer IP is in the trusted validators, return early.
-        if self.trusted_validators().contains(&ip) {
+        if self.trusted_peers().contains(&ip) {
             return true;
         }
         // Retrieve the Aleo address of the peer IP.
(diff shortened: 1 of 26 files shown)


node/bft/tests/gateway_e2e.rs
```diff
@@ -78,79 +79,10 @@ async fn handshake_responder_side_timeout() {
...
     // Check the test peer hasn't been added to the gateway's connected peers.
-    assert!(gateway.connected_peers().read().is_empty());
+    assert!(gateway.connected_peers().is_empty());
     assert_eq!(gateway.tcp().num_connected(), 0);
 }
 
-// The test peer connects to the gateway and sends an unexpected event.
-// The gateway's handshake should be interrupted and the peer should be
-// disconnected.
-macro_rules! handshake_responder_side_unexpected_event {
```
ljedrz, line 88: note: these tests no longer apply with the handshake "tightening" (like the one in https://github.com/ProvableHQ/snarkOS/pull/3713); other than that, they check for a relatively trivial feature (rejecting unexpected handshake messages)

node/router/src/lib.rs
```diff
@@ -80,6 +81,206 @@ pub const DEFAULT_NODE_PORT: u16 = 4130;
...
+        self.peer_pool()
+            .read()
+            .iter()
+            .filter_map(
+                |(addr, peer)| if let Peer::Candidate(peer) = peer { peer.trusted.then_some(*addr) } else { None },
+            )
+            .collect()
+    }
+
+    // Save the best peers to disk.
```
kaimast: Make this a doc comment. Also, maybe explain what "best" means here.

node/router/src/lib.rs
```diff
@@ -80,6 +81,206 @@ pub const DEFAULT_NODE_PORT: u16 = 4130;
 /// The name of the file containing cached peers.
 const PEER_CACHE_FILENAME: &str = "cached_router_peers";
 
+pub trait PeerPoolHandling<N: Network>: P2P {
+    fn peer_pool(&self) -> &RwLock<HashMap<SocketAddr, Peer<N>>>;
```
kaimast, line 85: Does this mean an external function can now call this and change the set of peers directly? Should we allow that?

node/bft/src/gateway.rs
```diff
@@ -914,13 +891,14 @@ impl<N: Network> Gateway<N> {
     }
 
     /// Disconnects from the given peer IP, if the peer is connected.
```
kaimast: You should document here what the returned boolean means.

node/router/src/lib.rs
```diff
@@ -81,6 +81,206 @@ pub const DEFAULT_NODE_PORT: u16 = 4130;
...
+        peers.sort_unstable_by_key(|peer| {
+            if let Some(peer_stats) = known_peers.get(&peer.listener_addr().ip()) {
+                // Prioritize greatest height, then lowest failure count.
+                (cmp::Reverse(peer.last_height_seen()), peer_stats.failures())
+            } else {
+                // Unreachable; use an else-compatible dummy.
+                (cmp::Reverse(peer.last_height_seen()), 0)
+            }
+        });
+        peers.truncate(MAX_PEERS_TO_SEND);
```
vicsn, line 230: I finally see why consolidating the Gateway and Router also has it's downsides, because there are still some different assumptions. Most notably, that we'll usually want to connect to all of our validators, a number which is not tightly coupled to `MAX_PEERS_TO_SEND`.

Also, we don't want `PEER_CACHE_FILENAME` to be the same for both peer sets.

I guess both variables can simply be passed into this function. Feel free to tackle in the follow up PR.

vicsn (dismissed):

kaimast (dismissed):
The code looks good to me.

I am mostly wondering if we want to export the trait publicly, as it adds a method that allows to side-step Router/Gatways logic and directly modify the peer pool.
Right now, there are not many (any?) users of snarkOS as a library, but would like to minimize the amount of functions that are added to the public API of snarkOS.

Maybe it is better to introduce a `PeerPool` struct that wraps around the map of peers and `Gateway`/`Router` can use to implement their functionality? 

Also, what is the difference in functionality between Gateway and Router after we are done with all the cleanups? Could we get to a point where there is only one struct to manage p2p and BFT peers?

node/router/src/lib.rs
ljedrz: snarkOS is not a library, so I wouldn't consider it a concern - there is no potential for accidental misuse, and we don't have control over "deliberate misuse"

ljedrz:
> I am mostly wondering if we want to export the trait publicly, as it adds a method that allows to side-step Router/Gatways logic and directly modify the peer pool.

The only intended users of this trait are objects carrying a specific peer pool, and since only the `Router` and `Gateway` contain it, only they can implement it. There are likely several other traits that could be misused in the way you're describing, but it couldn't be done by accident. What is more, the `peer_pool` under the `Gateway` shouldn't have an individual `Arc` over itself (only the `Gateway` needs to be shallowly clonable), but I decided to do it in a small, dedicated follow-up. This will further ensure that misuse is impossible.

> Right now, there are not many (any?) users of snarkOS as a library, but would like to minimize the amount of functions that are added to the public API of snarkOS.

To my knowledge there are none, and thus we haven't maintained API stability for snarkOS (otherwise we'd routinely run into such concerns).

> Also, what is the difference in functionality between Gateway and Router after we are done with all the cleanups?

Specific handshaking and peering logic, some caches, probably some other small details. Eventually having just one of these objects is likely possible, but it's not the goal of this PR.

vicsn (approved):

PR #3887, opened 2025-09-25
[Feat] Introduce a persistent validator cache

### LINKED ISSUE #3869: [Feature] Ensure gateway can persist peer cache

ljedrz:
This PR introduces a persistent validator cache, similarly to https://github.com/ProvableHQ/snarkOS/pull/3820.

As a drive-by, there is also a reduction of `Arc`s within the `Gateway`, which will reduce the number of atomic operations on each clone and drop by 7. This also makes it harder to misuse the new `PeerPoolHandling` trait by making separating the peer pool from the `Gateway` impossible (CC https://github.com/ProvableHQ/snarkOS/pull/3879).

Note: the associated issue states that
```
the same validator can't have two ip addresses
```

I'm assuming this means that we can't have two connected `Gateway` `Peer`s that have the same Aleo address. This, however, would be a more general change, as it would also affect the way the peer lists should be handled.

I've tested it locally on new validators and after a restart - both `cached_router_peers` and `cached_gateway_peers` files are created and processed, and the addresses are the ones I expected. Manually altering them confirmed that the cache would work for other (non-`dev`) addresses too.

Closes https://github.com/ProvableHQ/snarkOS/issues/3869.

### DIFF
--- a/node/bft/src/gateway.rs
+++ b/node/bft/src/gateway.rs
@@ -24,6 +24,7 @@ use crate::{
     helpers::{Cache, PrimarySender, Resolver, Storage, SyncSender, WorkerSender, assign_to_worker},
     spawn_blocking,
 };
+use aleo_std::StorageMode;
 use snarkos_account::Account;
 use snarkos_node_bft_events::{
     BlockRequest,
@@ -110,6 +111,9 @@ const CONNECTION_ATTEMPTS_SINCE_SECS: i64 = 10;
 /// The amount of time an IP address is prohibited from connecting.
 const IP_BAN_TIME_IN_SECS: u64 = 300;
 
+/// The name of the file containing cached validators.
+const VALIDATOR_CACHE_FILENAME: &str = "cached_gateway_peers";
+
 /// Part of the Gateway API that deals with networking.
 /// This is a separate trait to allow for easier testing/mocking.
 #[async_trait]
@@ -146,6 +150,8 @@ pub struct Gateway<N: Network> {
     sync_sender: Arc<OnceCell<SyncSender<N>>>,
     /// The spawned handles.
     handles: Arc<Mutex<Vec<JoinHandle<()>>>>,
+    /// The storage mode.
+    storage_mode: StorageMode,
     /// The development mode.
     dev: Option<u16>,
 }
@@ -164,6 +170,7 @@ impl<N: Network> Gateway<N> {
         ledger: Arc<dyn LedgerService<N>>,
         ip: Option<SocketAddr>,
         trusted_validators: &[SocketAddr],
+        storage_mode: StorageMode,
         dev: Option<u16>,
     ) -> Result<Self> {
         // Initialize the gateway IP.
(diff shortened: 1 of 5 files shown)


node/router/src/lib.rs
```diff
@@ -208,9 +208,30 @@ pub trait PeerPoolHandling<N: Network>: P2P {
...
+        let peers = if let Ok(cached_peers_str) = fs::read_to_string(&peer_cache_path) {
+            let mut cached_peers = Vec::new();
+            for peer_addr_str in cached_peers_str.lines() {
+                if let Ok(addr) = SocketAddr::from_str(peer_addr_str) {
+                    cached_peers.push(addr);
+                }
+            }
+            cached_peers
+        } else {
+            Vec::new()
```
vicsn: Can we add a warn log here? So we can potentially surface e.g. filesystem permission errors?

node/router/src/lib.rs
```diff
@@ -208,9 +208,30 @@ pub trait PeerPoolHandling<N: Network>: P2P {
...
+    fn load_cached_peers(storage_mode: &StorageMode, filename: &str) -> Result<Vec<SocketAddr>> {
+        let mut peer_cache_path = aleo_ledger_dir(N::ID, storage_mode);
+        peer_cache_path.push(filename);
+
+        let peers = if let Ok(cached_peers_str) = fs::read_to_string(&peer_cache_path) {
+            let mut cached_peers = Vec::new();
+            for peer_addr_str in cached_peers_str.lines() {
+                if let Ok(addr) = SocketAddr::from_str(peer_addr_str) {
+                    cached_peers.push(addr);
+                }
```
vicsn, line 225: Can we add a warn log here if it failed? It shouldn't

node/bft/src/gateway.rs
```diff
@@ -175,30 +192,37 @@ impl<N: Network> Gateway<N> {
...
-        Ok(Self {
+        Ok(Self(Arc::new(InnerGateway {
             account,
             storage,
             ledger,
             tcp,
             cache: Default::default(),
             resolver: Default::default(),
-            peer_pool: Arc::new(RwLock::new(initial_peers)),
+            peer_pool: RwLock::new(initial_peers),
```
vicsn, line 214: If I see correctly, the heartbeat currently ignores the untrusted peers. Perhaps a clean-ish way to fix this is to connect to them in handle_min_connected_validators. And while we're add it, we can fix that MIN_CONNECTED_VALIDATORS should actually be N::LATEST_MAX_CERTIFICATES().unwrap()

ljedrz: > If I see correctly, the heartbeat currently ignores the untrusted peers. Perhaps a clean-ish way to fix this is to connect to them in handle_min_connected_validators.

Just to confirm; do you mean that we should keep proactively connecting to untrusted validators as well (in addition to what `handle_trusted_validators` is doing) if we're under `MIN_CONNECTED_VALIDATORS`?

> MIN_CONNECTED_VALIDATORS should actually be N::LATEST_MAX_CERTIFICATES().unwrap()

Should that be a new associated function for the `Gateway` or for the `Network`?

vicsn: > Just to confirm; do you mean that we should keep proactively connecting to untrusted validators as well (in addition to what handle_trusted_validators is doing) if we're under MIN_CONNECTED_VALIDATORS?

Yes!

> Should that be a new associated function for the Gateway or for the Network?

N::LATEST_MAX_CERTIFICATES().unwrap() already exists, does that answer the question?

vicsn:
Can you confirm the tests you ran in the PR deacription

ljedrz:
@vicsn done; I applied more or less the same treatment as with the `Router` peer cache.

vicsn (approved):

Issue #3888, opened 2025-09-26
[Feature] Harden validator peering


vicsn:
## 🚀 Feature

It seems possible for validators to accidentally connect to other validators multiple times. This does not have to imply malicious behaviour, maybe a validator moves from one IP address to another or is reachable via multiple IP addresses.

The Gateway should track which validators it is already connected to, and ensure connections from/to existing validators don't happen. One exception here is that we should prioritize **trusted** peer connections over **untrusted** peer connections.

We should also think of a way for the gateway peer cache not to grow full with stale entries. I guess we should:
- remove banned peers from our in-memory peer pool if we don't do so already
- put a liberal cap on the size of the peer pool for the gateway like we do in router of MAXIMUM_CANDIDATE_PEERS so we avoid a DoS attack on that front

ljedrz:
> remove banned peers from our in-memory peer pool if we don't do so already

We disallow connection attempts from banned peers, but we don't currently prohibit ourselves from connecting to them. Should I add this to the checks done whenever we call `connect`? Or do we want them completely erased from the list of candidates for as long as the ban lasts (which would indirectly lead to the same result)?

ljedrz:
> put a liberal cap on the size of the peer pool for the gateway like we do in router of MAXIMUM_CANDIDATE_PEERS so we avoid a DoS attack on that front

Sounds like a `PeerPoolHandling`-controlled limit; however, note that the candidate peers can change into connecting/connected peers (and the other way round), so I'm assuming this limit would instead apply to the entirety of the peer pool, i.e. to all kinds of peers? It's pretty unlikely we'd get saturated with useless candidates (without the possibility of getting new connected ones), but this could also be some heartbeat-based cleanup routine.

That being said, this could also be limited to candidate peers, but then we might reach a state where a freshly disconnected peer gets removed from the pool (because it would change into a candidate, with them being at their limit), unless we apply some LRU-ish logic.

vicsn:
> Or do we want them completely erased from the list of candidates for as long as the ban lasts (which would indirectly lead to the same result)?

I feel like this would be the cleanest solution. As it leaves optionality in case the node wants to manually initiate a connection one way or another to banned peers anyway.

> That being said, this could also be limited to candidate peers, but then we might reach a state where a freshly disconnected peer gets removed from the pool (because it would change into a candidate, with them being at their limit), unless we apply some LRU-ish logic.

[...]

PR #3889, opened 2025-09-26
[Fix] Disallow multiple conns per validator

ljedrz:
This PR targets https://github.com/ProvableHQ/snarkOS/issues/3888.

It appears that we are already guarding the validators against connecting to those with an already-connected `Address` during the handshake (both sides, via `Gateway::verify_challenge_request`), but we're doing a bad job at it, as we're assuming that the multiple connections can only be attributed to a single listener address. This is trivially solved with https://github.com/ProvableHQ/snarkOS/commit/9258e342e6e97446087a81abcc91d421f54aaa88.

Next, I took a closer look at banning, and realized that the ones triggered by delayed block sync responses currently partially do not work, as the address used to disconnect is the listener address, which is the wrong one to use in case of connections we had accepted (as opposed to initiated). This is fixed with https://github.com/ProvableHQ/snarkOS/commit/ff2672c510dd98e858ff4106c4f82889cc18a1e7. A side bonus of this change is that banning peers and high-level disconnects have been moved to the `PeerPoolHandling` trait, deduplicating some code.

In addition, the `Gateway`'s `Resolver` [was shrunk](https://github.com/ProvableHQ/snarkOS/pull/3889/commits/378ea824a76af1020d809641640c6e621bff4c8c), as the peer pool is able to resolve listener addresses to connected addresses and Aleo addresses. It's a pure cleanup.

The linked issue requires further changes, but since the diff has grown quite a bit already, I decided to cut it short here and leave the rest to a follow-up.

### DIFF
--- a/node/bft/src/gateway.rs
+++ b/node/bft/src/gateway.rs
@@ -311,17 +311,6 @@ impl<N: Network> CommunicationService for Gateway<N> {
     async fn send(&self, peer_ip: SocketAddr, message: Self::Message) -> Option<oneshot::Receiver<io::Result<()>>> {
         Transport::send(self, peer_ip, message).await
     }
-
-    fn ban_peer(&self, peer_ip: SocketAddr) {
-        trace!("Banning peer {peer_ip} for timing out on block requests");
-
-        let tcp = self.tcp().clone();
-        tcp.banned_peers().update_ip_ban(peer_ip.ip());
-
-        tokio::spawn(async move {
-            tcp.disconnect(peer_ip).await;
-        });
-    }
 }
 
 impl<N: Network> Gateway<N> {
@@ -906,20 +895,6 @@ impl<N: Network> Gateway<N> {
                 Ok(true)
             }
         }
-    }
-
-    /// Disconnects from the given peer IP, if the peer is connected. The returned boolean
-    /// indicates whether the peer was actually disconnected from, or if this was a noop.
-    pub fn disconnect(&self, peer_ip: SocketAddr) -> JoinHandle<bool> {
-        let gateway = self.clone();
-        tokio::spawn(async move {
-            if let Some(peer) = gateway.get_connected_peer(peer_ip) {
-                let connected_addr = peer.connected_addr;
-                gateway.tcp.disconnect(connected_addr).await
-            } else {
-                false
-            }
-        })
     }
 
     /// Initialize a new instance of the heartbeat.
--- a/node/bft/src/gateway.rs
+++ b/node/bft/src/gateway.rs
@@ -400,7 +400,7 @@ impl<N: Network> Gateway<N> {
             return true;
         }
         // Retrieve the Aleo address of the peer IP.
-        match self.resolver.read().get_address(ip) {
+        match self.resolve_to_aleo_addr(ip) {
             // Determine if the peer IP is an authorized validator.
             Some(address) => self.is_authorized_validator_address(address),
             None => false,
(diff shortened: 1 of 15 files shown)


node/router/src/lib.rs
```diff
@@ -666,6 +684,31 @@ impl<N: Network> Router<N> {
     }
 }
 
+#[async_trait]
+impl<N: Network> CommunicationService for Router<N> {
```
ljedrz, line 688: note: this trait was moved here from the `Client`, as the `Router` is sufficient to handle communications

node/sync/communication-service/src/lib.rs
```diff
@@ -75,9 +67,5 @@ pub mod test_helpers {
         ) -> Option<oneshot::Receiver<io::Result<()>>> {
             None
         }
-
-        fn ban_peer(&self, peer_ip: SocketAddr) {
```
ljedrz, line 79: note: the related tests are shrunk, as they were insufficient to verify this functionality in a TCP setting, where the listener address differs from the connected address when we're on the receiving side of the connection

node/sync/src/block_sync.rs
```diff
@@ -936,12 +936,12 @@ impl<N: Network> BlockSync<N> {
...
     /// This removes the corresponding block responses and returns the set of peers/addresses that timed out.
-    /// It will ask the communication service to ban any timed-out peers.
+    /// It will ask the peer pool handling service to ban any timed-out peers.
     ///
     /// Finally, it will return a set of new of block requests that replaced the timed-out requests (if needed).
-    pub fn handle_block_request_timeouts<C: CommunicationService>(
+    pub fn handle_block_request_timeouts<P: PeerPoolHandling<N>>(
         &self,
-        communication: &C,
+        peer_pool_handler: Option<&P>,
```
ljedrz: this is a simple workaround to disable banning in tests without code duplication

node/bft/src/helpers/resolver.rs
```diff
@@ -19,12 +19,8 @@ use std::{collections::HashMap, net::SocketAddr};
 
 #[derive(Debug)]
 pub struct Resolver<N: Network> {
-    /// The map of the listener address to (ambiguous) peer address.
```
vicsn, line 22: Perhaps worth adding a doc-comment to Resolver, noting the reverse maps live effectively in the peer pool

node/bft/src/gateway.rs
```diff
@@ -385,12 +385,8 @@ impl<N: Network> Gateway<N> {
...
     /// Returns `true` if the node is connected to the given Aleo address.
     pub fn is_connected_address(&self, address: Address<N>) -> bool {
-        // Retrieve the peer IP of the given address.
-        match self.resolver.read().get_peer_ip_for_address(address) {
-            // Determine if the peer IP is connected.
-            Some(peer_ip) => self.is_connected(peer_ip),
-            None => false,
-        }
+        // The resolver only contains data on connected peers.
+        self.resolver.read().get_peer_ip_for_address(address).is_some()
```
vicsn, line 378: My (potentially wrong) understanding is that the Resolver contains Connected peers. If so, isn't your change equivalent, except perhaps that it's a bit more stringent as we were about to disconnect from this peer but it's not cleaned up from the resolver yet?

node/sync/communication-service/src/lib.rs
vicsn: Can we revive the tests somehow, to make them properly test the two scenario's? Happy to discuss design before implementation.

node/bft/src/gateway.rs
ljedrz: The `Resolver` does only contain connected peers, but with this change, we are able to reject validators who have a different IP, but the same Aleo `Address`, which could happen when some validator wanted to e.g. launch multiple instances of snarkOS. As it stands now, we would not prohibit it.

node/sync/communication-service/src/lib.rs
ljedrz: test revival done in https://github.com/ProvableHQ/snarkOS/pull/3889/commits/55a23928394c196070ffcde5acbe4895ee0b113f; a separate banning test will follow in another PR

node/sync/src/block_sync.rs
ljedrz: no longer needed

vicsn (approved):

PR #3890, opened 2025-09-26
Store the generated JWT secret to a file

vicsn:
## Motivation

Validators should be able to figure out what their node's JWT is (for calling the important `/db_backup` endpoint) without having to comb through the node's stdout output.

## Test Plan

Ran a node locally and the file is created as expected. Restarted the node and the file is overwritten.


vicsn:
Closing in favour of the more comprehensive: https://github.com/ProvableHQ/snarkOS/pull/3891

PR #3891, opened 2025-09-26
Create and document native backup utility

vicsn:
## Motivation

Validators should be able to create fast local backups either:
- using a JWT, which they can read from a file instead of stdout
- explicitly passing in --nojwt, not having to authenticate if they already have a solid firewall and don't want to deal with the complexity of JWT creation and backup

To make this possible, this PR:

- [c61a1de](https://github.com/ProvableHQ/snarkOS/pull/3891/commits/c61a1de26c45cb46b2dc77e025ba1811821c2f6d) Moves scripts into their own folder, with the exception of `build_ubuntu.sh` which is likely relied upon in automated test and production systems
- [4572958](https://github.com/ProvableHQ/snarkOS/pull/3891/commits/4572958c20c7eef06760620987c5147d2e1fafcc) adds a totally AI generated **minimal** backup script. Node operators may diverge if they want, this is just an example.
- [b1db9e2](https://github.com/ProvableHQ/snarkOS/pull/3891/commits/b1db9e2dcc897999665c0532f86e3140c0e01732) introduce --nojwt
- [d934f54](https://github.com/ProvableHQ/snarkOS/pull/3891/commits/d934f54d66684992d1d3b40c5ec5d02b87a916e5) stores the generated JWT to a file after startup. I tried recreating it from bash using the secret and timestamp but gave up after 2 hours.

## Test Plan

Confirmed with local runs that the backup script creates backups as expected, and that the `jwt_secret*.txt` file is created as expected. We also have `db_backup_ci.sh` in CI. Next, this will be tested by a select few validators on canary, before moving on to testnet and mainnet.


cli/src/commands/start.rs
```diff
@@ -634,21 +641,25 @@ impl Start {
...
+                            }
+                            Some(jwt_bytes)
+                        } else {
+                            None
+                        };
+
+                        if let Ok(jwt_token) = snarkos_node_rest::Claims::new(account.address(), jwt_secret, self.jwt_timestamp).to_jwt_string() {
+                            println!("🔑 Your one-time JWT token is {}\n", jwt_token.dimmed());
+                            // Store the JWT secret to a file.
+                            std::fs::write(format!("jwt_secret_{}.txt", account.address()), jwt_token)?;
```
ljedrz: hmm perhaps we should write it to the ledger folder, so that it's at a predictable location? or at least notify the user where it is being stored, though that could quickly get lost among the other logs

scripts/backup.sh
```diff
@@ -0,0 +1,53 @@
+#!/bin/sh
+
+set -eu
+
+# The network of the node which you're making a backup for.
+NETWORK="mainnet"
+# The directory in which you want backups to be stored.
+BASE_DIR="${HOME}/snarkOS/aleo_ledger_checkpoints/${NETWORK}/"
+# The node REST endpoint to trigger backup creation from.
+ENDPOINT="http://localhost:3030/${NETWORK}/db_backup"
```
ljedrz: I'd make the port configurable individually, but I guess the user could specify the entire `ENDPOINT` in that case too

ljedrz (dismissed):
LGTM, left one or two suggestions

scripts/backup.sh
```diff
@@ -0,0 +1,45 @@
+#!/bin/sh
+
+set -eu
+
+NETWORK="mainnet"
+BASE_DIR="${HOME}/snarkOS/aleo_ledger_checkpoints/${NETWORK}/"
+ENDPOINT="http://localhost:3030/${NETWORK}/db_backup"
+JWT="eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJzdWIiOiJhbGVvMXJoZ2R1NzdoZ3lxZDN4amo4dWN1M2pqOXIya3J3ejZtbnp5ZDgwZ25jcjVmeGN3bGg1cnN2enA5cHgiLCJpYXQiOjE3NDkxMTYzNDUsImV4cCI6MjA2NDQ3NjM0NX0.LiqFGiQds3OGHGJ5K3xi359g-uTQBZCrAskGj9UWAbM"
+
+post_backup() {
```
kaimast: I would rename this to `create_backup` or add a comment `# Asks the node to create a new backup (via HTTP POST)`.

At a first glance, I thought "post" meant "after" here, and this was some kind of cleanup function.

scripts/backup.sh
```diff
@@ -0,0 +1,45 @@
...
+    slot="$1"
+    slot_path="${BASE_DIR}/${slot}"
+    rm -rf -- "$slot_path" 2>/dev/null || :
+    curl -fsS -X POST -H "Authorization: Bearer ${JWT}" "${ENDPOINT}?path=${slot_path}" >/dev/null 2>&1 || :
```
kaimast: Is it a good idea to discard the return value here? IMHO it should print something when it fails or successfully creates a backup.

scripts/backup.sh
```diff
@@ -0,0 +1,45 @@
...
+
+NETWORK="mainnet"
+BASE_DIR="${HOME}/snarkOS/aleo_ledger_checkpoints/${NETWORK}/"
+ENDPOINT="http://localhost:3030/${NETWORK}/db_backup"
+JWT="eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJzdWIiOiJhbGVvMXJoZ2R1NzdoZ3lxZDN4amo4dWN1M2pqOXIya3J3ejZtbnp5ZDgwZ25jcjVmeGN3bGg1cnN2enA5cHgiLCJpYXQiOjE3NDkxMTYzNDUsImV4cCI6MjA2NDQ3NjM0NX0.LiqFGiQds3OGHGJ5K3xi359g-uTQBZCrAskGj9UWAbM"
+
+post_backup() {
+    slot="$1"
+    slot_path="${BASE_DIR}/${slot}"
+    rm -rf -- "$slot_path" 2>/dev/null || :
```
kaimast: Similar to the comment about the line below: is it safe to discard the result here? If the old snapshot cannot be removed, should we still attempt to create a new one?

kaimast (dismissed):
Looks correct. I left some code quality comments.

raychu86 (dismissed):

cli/src/commands/start.rs
vicsn: @copilot fix this please. You can look for other examples of `storage_mode` to see how to build the path.

vicsn: 9d68713d5

scripts/backup.sh
vicsn: be226d9c5

scripts/backup.sh
vicsn: Fixed! https://github.com/ProvableHQ/snarkOS/commit/be226d9c59a3d1aa9f979a5e8fcc1767f1e5d817

scripts/backup.sh
vicsn: The script is just an example and likely not something directly usable in production, but good point, especially for testing it should not discard errors: https://github.com/ProvableHQ/snarkOS/commit/be226d9c59a3d1aa9f979a5e8fcc1767f1e5d817

scripts/backup.sh
vicsn: https://github.com/ProvableHQ/snarkOS/commit/be226d9c59a3d1aa9f979a5e8fcc1767f1e5d817

cli/src/commands/start.rs
```diff
@@ -618,6 +625,45 @@ impl Start {
...
+                None => StorageMode::Production,
+            },
+        };
+
+        // Helper function to obtain the path to the JWT secret.
+        let jwt_secret_path = |network: u16, storage_mode: &StorageMode, address: &Address<N>| {
+            let mut jwt_secret_path = aleo_std::aleo_ledger_dir(network, storage_mode);
+            jwt_secret_path.push(format!("jwt_secret_{address}.txt"));
+            jwt_secret_path
+        };
```
ljedrz: nit: it seems that this path is constructed just once, so a closure isn't really needed; it's not a blocker ofc

ljedrz (dismissed):

cli/src/commands/start.rs
vicsn: Yeah perhaps I overoptimized here, for clarity I wanted to make sure it was only ever instantiated in the two code paths where it is actually used.

ljedrz (approved):

PR #3894, opened 2025-09-29
[Logs] Reduce the log level when failing to connect to non-trusted/bootstrap peers

ljedrz:
CC https://github.com/ProvableHQ/snarkOS/pull/3820, where this was [suggested](https://github.com/ProvableHQ/snarkOS/pull/3820#issuecomment-3341914955).

This involves some code deduplication that facilitates this change for all types of peers (validators and non-validators), picking the best implementations between the `Gateway` and `Router`.

### DIFF
--- a/node/bft/src/gateway.rs
+++ b/node/bft/src/gateway.rs
@@ -51,8 +51,6 @@ use snarkos_node_tcp::{
     ConnectionSide,
     P2P,
     Tcp,
-    is_bogon_ip,
-    is_unspecified_or_broadcast_ip,
     protocols::{Disconnect, Handshake, OnConnect, Reading, Writing},
 };
 use snarkvm::{
@@ -165,6 +163,8 @@ pub struct InnerGateway<N: Network> {
 }
 
 impl<N: Network> PeerPoolHandling<N> for Gateway<N> {
+    const OWNER: &str = CONTEXT;
+
     fn peer_pool(&self) -> &RwLock<HashMap<SocketAddr, Peer<N>>> {
         &self.peer_pool
     }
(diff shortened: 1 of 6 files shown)


node/bft/src/gateway.rs
```diff
@@ -443,59 +425,11 @@ impl<N: Network> Gateway<N> {
...
-                }
-            }
-        }))
-    }
-
-    /// Ensure we are allowed to connect to the given peer.
-    fn check_connection_attempt(&self, peer_ip: SocketAddr) -> Result<()> {
-        // Ensure the peer IP is not this node.
-        if self.is_local_ip(peer_ip) {
-            bail!("{CONTEXT} Dropping connection attempt to '{peer_ip}' (attempted to self-connect)")
```
kaimast, line 467: We're losing this CONTEXT information, right? I think it is fine, but not sure if some people rely on it. 

If it is needed, we could add something like  `PeerPoolHandling::pool_type` or `PeerPoolHandling::pool_name` to the trait.

node/router/src/lib.rs
```diff
@@ -84,6 +84,104 @@ const PEER_CACHE_FILENAME: &str = "cached_router_peers";
...
+        // Determine whether the peer is trusted or a bootstrap node in order to decide
+        // how problematic any potential connection issues are.
+        let is_trusted_or_bootstrap =
+            self.is_trusted(listener_addr) || bootstrap_peers::<N>(false).contains(&listener_addr);
+
+        let tcp = self.tcp().clone();
+        Some(tokio::spawn(async move {
+            debug!("Connecting to {listener_addr}...");
+            // Attempt to connect to the peer.
+            match tcp.connect(listener_addr).await {
```
kaimast, line 163: Do we keep candidate peers in this case?

kaimast (dismissed):
LGTM. I left two questions/suggestions, but nothing preventing it from getting merged. 

(Excited about all the deduplication!)

node/bft/src/gateway.rs
ljedrz: ah, I must have used the version without `CONTEXT`, as one of the objects didn't utilize it; I don't think it is being used in this particular scenario, but I can re-add it later on; distinguishing between the implementors is also a good idea

node/router/src/lib.rs
ljedrz: if a connection is successful, then even if a peer had been unknown until then, it is retained as a future candidate after a disconnect

node/bft/src/gateway.rs
vicsn: Yes the logs should emit whether they're from router or gateway. Happy to approve if that is the case currently, otherwise I'll wait

ljedrz: `CONTEXT` restored in https://github.com/ProvableHQ/snarkOS/pull/3894/commits/555d0304442324569aeb22a123b4860e57d49848

ljedrz: well, almost - I need to check why it's not found here

ljedrz: fixed in https://github.com/ProvableHQ/snarkOS/pull/3894/commits/85cd265eae216593ea2ea6b77a75749910cf5988

vicsn (dismissed):
LGTM, please merge in staging

ljedrz:
Squashed and rebased as agreed offline (`git merge` was having a hard time with these changes).

vicsn (approved):

PR #3901, opened 2025-09-30
[Tests] Add IP-banning test cases

### LINKED ISSUE #3895: [Test] Add tests for peer bans
While this feature is relatively straightforward, we are currently missing test cases that would ensure that it doesn't break.

This was found while working on https://github.com/ProvableHQ/snarkOS/pull/3889, and would ideally build on that PR, which aligns banning implementations between the `Gateway` and the `Router`.

ljedrz:
This PR adds 2 straightforward tests that ensure that IP-banning works for both possible sides of the connection.

Closes https://github.com/ProvableHQ/snarkOS/issues/3895.

### DIFF
--- a/node/bft/src/gateway.rs
+++ b/node/bft/src/gateway.rs
@@ -70,8 +70,6 @@ use locktick::parking_lot::{Mutex, RwLock};
 #[cfg(not(feature = "locktick"))]
 use parking_lot::{Mutex, RwLock};
 use rand::seq::{IteratorRandom, SliceRandom};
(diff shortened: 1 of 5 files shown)


ljedrz:
Btw, these tests can also be extended with checks that banned peers can't be connected to, if we're ok with applying it in test conditions (it's currently feature-gated in handshake code, but I don't see a good reason for it - it could easily be moved to `PeerPoolHandling::check_connection_attempt`).

vicsn (approved):

PR #3900, opened 2025-09-30
[Fix] Dedup validator heartbeat conns & don't downgrade peers on connection duplicates

ljedrz:
This is a twofold fix to a handshake-related logic assertion hit in a `debug` build:
1. https://github.com/ProvableHQ/snarkOS/pull/3900/commits/1881d68711c158401f0ddf4de43c1a1b129c1186 - deduplicates outbound connection attempts to trusted validators (also removes redundant connection checks that are done in `PeerPoolHandling::connect`); this reduces the likelihood of the issue
2. https://github.com/ProvableHQ/snarkOS/pull/3900/commits/85bd920d63a76d078499b257c1919d1965a39672 - the root fix; it causes duplicate handshake attempts to not register as errors, which allows us to not downgrade a peer's status when they occur at the "wrong" time (it can happen when we have concurrent bilateral handshake attempts with a single peer - a latter one can currently invalidate a former one, downgrading the peer to `CandidatePeer` status instead of allowing that handshake to conclude)

Fixes https://github.com/ProvableHQ/snarkOS/issues/3897

PR #3902, opened 2025-10-01
[Feature] Improve (dis-)connect handling between peers

### LINKED ISSUE #3893: [Bug] Remove/reduce "already connecting to node" warnings
## 🐛 Bug Report

Frequently, when starting two nodes that connect to each other, I get warnings like the following.
```
2025-09-27T16:53:48.694791Z  WARN snarkos_node_router: Unable to connect to '127.0.0.1:4130' - already connecting to node at 127.0.0.1:4130
```

With the recent peering changes, this now also generates the following errors which are misleading.
```
2025-09-27T16:53:48.704736Z ERROR tcp{name="0"}: snarkos_node_tcp::protocols::handshake: handshake with 127.0.0.1:52182 failed: Dropping connection request from '127.0.0.1:4131' (untrusted)
2025-09-27T16:53:48.704800Z ERROR tcp{name="0"}: snarkos_node_tcp::protocols::handshake: handshake with 127.0.0.1:4130 failed: the peer disconnected before sending "Message::ChallengeResponse", likely due to peer saturation or shutdown
2025-09-27T16:53:48.704888Z ERROR tcp{name="0"}: snarkos_node_tcp::tcp: Unable to initiate a connection with 127.0.0.1:4130: the peer disconnected before sending "Message::ChallengeResponse", likely due to peer saturation or shutdown
```

### Expected Behavior
Attempting to connect to a peer that is already connecting to us should not create a warning or error of any kind.

kaimast:
Fixes #3893. The PR is quite large, below is a high-level overview of the changes.
 
### Better Connection Handling
The main purpose of this PR is to propagate errors during handshakes "up", so the error is logged at the appropriate location. It also enables filtering out benign errors such as "Already connecting". 

This is achieved by using a concrete `ConnectionError` type instead of `anyhow::Error`. Previously, the code also used `Option` or booleans to indicate a duplicate connection, which did not make the code very readable. To replace those, the new disconnect reasons, mentioned below, are very useful.

### New Disconnect Reasons
The PR adds a four new disconnect reasons "already connecting", "already connected", "connecting to myself", "no untrusted external peers allowed". This avoids those confusing `peer disconnected before sending `Message::ChallengeResponse` messages.

We previously added support for unknown disconnect reasons (see #4018), so this can be added without causing problems.

### Significant Reduction of Connection-Related Warnings
Nodes still generated a significant number of warnings even with these changes. One culprit was that the heartbeat logic sometimes reattempted failed connections too frequently. The PR now adds a mandatory 10-second cooldown period between connection attempts.
In the future, we should consider increasing the cooldowns with every failed attempt.

Additionally, I also expanded the notion of a grace period at startup, where nodes do not log errors, such as "no connected validators". Now, nodes do not generate warnings in the first minute about not being fully connected.
We already did this for the stake-related checks, but not for other router/gateway errors. 

To test these changes, the PR adds a new check to `devnet_ci.sh` that limits the number of warnings a node is allowed to generate. For now, I limited them to 10 per node.
There are some connections related errors at startup that we cannot remove easily; nodes all start concurrently and then fail to connect to each other at the first attempt.


ljedrz:
Some of the goals of this PR are achieved with https://github.com/ProvableHQ/snarkOS/pull/3900 (which is also a bugfix). In addition, I can see that some of the changes here are conflicting with the aforementioned PR, which deduplicates some of the related code.

That being said, the idea to introduce a new, concrete error type is solid, and so is the `Result` change in the `Handshake` protocol. Since I've been tinkering with peering quite a lot lately, my recommendation would be as follows:
1. close this PR for now, as the code it is applied to can still be compacted, making the change surface smaller
2. simplify/deduplicate the network-related code further (we still have plenty of duplicate functionalities between the `Gateway` and the `Router`)
3. update and extend the connectivity/peering tests to make sure they cover everything we need (it's been a long time since their introduction)
4. reconsider these changes and apply them in small chunks, without any unrelated adjustments

vicsn:
Agreed with all of the above except:

> simplify/deduplicate the network-related code further (we still have plenty of duplicate functionalities between the Gateway and the Router)

Please no refactors unless agreed upon and resolving an existing issue.

ljedrz:
> Please no refactors unless agreed upon and resolving an existing issue.

Fair enough; I was mostly thinking about the prospect of a new node type (`BootstrapClient`), and some of the existing issues (like https://github.com/ProvableHQ/snarkOS/issues/3888), which are likely to naturally lead to such a result.

.ci/utils.sh
```diff
@@ -82,14 +82,18 @@ function check_heights() {
...
   local log_dir=$1
   local total_validators=$2
   local total_clients=$3
   
   local all_reached=true
   local highest_height=0
 
+  # The maximum number of warnings allow in each node's log file.
+  # Nodes may create some warnings at startup because they cannot connect to each other yet.
+  local max_warnings=10
```
ljedrz: we may want this to be an input param (same with errors, btw) in order for jobs like https://github.com/ProvableHQ/snarkOS/pull/4069 (where warnings are a certainty) to pass

node/network/src/peering.rs
```diff
@@ -39,6 +36,24 @@ use std::{
 use tokio::task;
 use tracing::*;
 
+/// Errors that can occur when adding a peer to the pool.
+#[derive(Debug, Clone, thiserror::Error)]
+pub enum AddPeerError {
```
ljedrz: IMO this extra error is an overkill, since we'll only even encounter it in the context of a `ConnectError`

node/network/src/peering.rs
```diff
@@ -83,77 +98,62 @@ pub trait PeerPoolHandling<N: Network>: P2P {
...
         // Ensure the peer IP is not this node.
         if self.is_local_ip(listener_addr) {
-            bail!("{} Dropping connection attempt to '{listener_addr}' (attempted to self-connect)", Self::OWNER);
+            return Err(ConnectError::SelfConnect { address: listener_addr });
         }
-        // Ensure the node does not surpass the maximum number of peer connections.
-        if self.number_of_connected_peers() >= self.max_connected_peers() {
-            bail!("{} Dropping connection attempt to '{listener_addr}' (maximum peers reached)", Self::OWNER);
+        // Ensure the peer IP is not banned.
+        if self.is_ip_banned(listener_addr.ip()) {
```
ljedrz: why was this reordered? it's not a big difference, just wondering about the rationale; it was originally checked after connected/connecting because those 2 are a lot more likely to result in an early return

node/network/src/peering.rs
```diff
@@ -83,77 +98,62 @@ pub trait PeerPoolHandling<N: Network>: P2P {
...
-            debug!("{} Dropping connection attempt to '{listener_addr}' (already connecting)", Self::OWNER);
-            return Ok(true);
+            return Err(ConnectError::AlreadyConnecting { address: listener_addr });
         }
-        // If the IP is already banned, reject the attempt.
-        if self.is_ip_banned(listener_addr.ip()) {
-            bail!("{} Rejected a connection attempt to a banned IP '{}'", Self::OWNER, listener_addr.ip());
+        // Ensure the node does not surpass the maximum number of peer connections.
+        if self.number_of_connected_peers() >= self.max_connected_peers() {
+            return Err(ConnectError::MaximumConnectionsReached { limit: self.max_connected_peers() as u16 });
```
ljedrz: back when I wrote this code I think I just moved it from the original location, but I'm pretty sure that this is already guarded at the lower level; then again, it _would_ be faster to return early here in case of outgoing connections, so we might as well keep it

node/network/src/peering.rs
```diff
@@ -83,77 +98,62 @@ pub trait PeerPoolHandling<N: Network>: P2P {
...
-            Err(error) => {
-                warn!("{} {error}", Self::OWNER);
-                return None;
-            }
-        }
+        self.check_connection_attempt(listener_addr)?;
 
-        // Determine whether the peer is trusted or a bootstrap node in order to decide
-        // how problematic any potential connection issues are.
-        let is_trusted_or_bootstrap =
```
ljedrz, line 139: I remember this was a specific requirement in order to pick the right log level; just making a note on it in case this is still expected

node/router/messages/src/helpers/disconnect.rs
```diff
@@ -135,6 +154,34 @@ impl std::fmt::Display for DisconnectReason {
...
+    }
+}
+
+impl DisconnectReason {
+    pub fn into_connect_error(self, address: SocketAddr) -> ConnectError {
+        match self {
+            DisconnectReason::NoExternalPeersAllowed => ConnectError::NoExternalPeersAllowed,
+            DisconnectReason::SelfConnect => ConnectError::SelfConnect { address },
+            DisconnectReason::AlreadyConnected => ConnectError::AlreadyConnected { address },
+            DisconnectReason::AlreadyConnecting => ConnectError::NoExternalPeersAllowed,
```
ljedrz:
```suggestion
            DisconnectReason::AlreadyConnecting => ConnectError::AlreadyConnecting { address },
```

node/router/src/handshake.rs
```diff
@@ -285,21 +278,24 @@ impl<N: Network> Router<N> {
         *listener_addr = Some(SocketAddr::new(peer_addr.ip(), peer_request.listener_port));
         let listener_addr = listener_addr.unwrap();
 
+        // Knowing the peers listening address, ensure we are not already connecting/connected to it.
+        if self.is_connecting_or_connected(listener_addr) {
```
ljedrz: this is already handled by `add_connecting_peer` at https://github.com/ProvableHQ/snarkOS/pull/3902/files#diff-bc670e8d24a2c82c7eb983102dc07a5fe94a223f37abceb4a276ddd3dc64d7c3R293

node/router/src/heartbeat.rs
```diff
@@ -54,16 +63,16 @@ pub trait Heartbeat<N: Network>: Outbound<N> {
     const IP_BAN_TIME_IN_SECS: u64 = 300;
 
     /// Handles the heartbeat request.
-    async fn heartbeat(&self) {
+    async fn heartbeat(&self, uptime: Duration) {
```
ljedrz: the uptime of the node can actually already be obtained from `Tcp` > `Stats` > `timestamp`; a new method exposing it would be practical, too

node/tcp/src/tcp.rs
```diff
@@ -74,15 +75,46 @@ pub enum ConnectError {
     AlreadyConnecting { address: SocketAddr },
     #[error("already connected to node at {address:?}")]
     AlreadyConnected { address: SocketAddr },
+    #[error("already connected to the given aleo address")]
+    AlreadyConnectedToAleoAddress,
     #[error("attempt to self-connect (at address {address:?}")]
     SelfConnect { address: SocketAddr },
-    #[error("I/O error: {0}")]
+    #[error("no external peers allowed")]
+    NoExternalPeersAllowed,
```
ljedrz: I would recommend against including business logic errors in the low level network errors

node/router/src/heartbeat.rs
```diff
@@ -299,30 +334,59 @@ pub trait Heartbeat<N: Network>: Outbound<N> {
...
+                    && last_connection_attempt.elapsed() < Self::MINIMUM_TIME_BETWEEN_CONNECTION_ATTEMPTS
+                {
+                    return None;
+                }
+
+                let peer_type = if peer.trusted { "trusted peer" } else { "peer" };
+
+                // Do not attempt to reconnect too frequently.
+                // TODO (kaimast): Consider increasing the minimum time based on the number of failed attempts.
+                if let Some(last_connection_attempt) = peer.last_connection_attempt
```
ljedrz, line 347: this check is duplicated

node/bft/src/gateway.rs
```diff
@@ -1530,31 +1570,38 @@ impl<N: Network> Gateway<N> {
...
-            return Err(error("Skipping request to connect to self".to_string()));
+            return Err(ConnectError::SelfConnect { address: peer_addr });
         }
 
         // Obtain the peer's listening address.
         *peer_ip = Some(SocketAddr::new(peer_addr.ip(), peer_request.listener_port));
         let peer_ip = peer_ip.unwrap();
 
+        // Knowing the peers listening address, ensure we are not already connecting/connected to it.
+        if self.is_connecting_or_connected(peer_addr) {
```
ljedrz: already handled by `add_connecting_peer` at https://github.com/ProvableHQ/snarkOS/pull/3902/files#diff-8cf4dc35dbc765800ac3ba1ae72c96036548f24681f36c88ad606b84ef31d252R1592

ljedrz (commented):
Left some comments and suggestions.

node/network/src/peering.rs
kaimast: Good point. I will revert it to the old order.

.ci/utils.sh
kaimast: I made it configurable in `devnet_ci.sh`. This might be useful for testing in the future as well.

node/network/src/peering.rs
kaimast: Removed!

node/router/src/heartbeat.rs
kaimast: I added `Tcp::uptime`.

node/router/src/heartbeat.rs
kaimast: Removed.

node/tcp/src/tcp.rs
kaimast: I introduced an `ApplicationError` variant that takes anything that implements `Display` as its payload. It is still a little ugly as Router and Gateway have two different versions of `DisconnectReason` and `peering` can also generate a custom error.

node/network/src/peering.rs
kaimast: The error is now propagated to `heartbeat` and printed there as a warning, so I don't think this is needed anymore.
Let me know if you think otherwise.

kaimast:
> Left some comments and suggestions.

Thanks! I think I addressed everything. 

Note: Sometimes the devnet test fails due to an error in syncing. This is something I am fixing in another PR and, ideally, should not hold up this one.

node/network/src/peering.rs
```diff
@@ -39,6 +36,23 @@ use std::{
 use tokio::task;
 use tracing::*;
 
+/// Application-level errors generated by the peering module.
+/// This never returend directly, but only as they payload for a `ConnectError`.
```
ljedrz:
```suggestion
/// This is never returned directly, only as the payload for a `ConnectError`.
```

node/network/src/peering.rs
```diff
@@ -83,77 +97,59 @@ pub trait PeerPoolHandling<N: Network>: P2P {
...
         }
-        Ok(false)
+
+        Ok(())
     }
 
     /// Attempts to connect to the given peer's listener address.
     ///
-    /// Returns None if we are already connected to the peer or cannot connect.
+    /// Returns an earlier error, if, for example, we are already connected to the peer.
```
ljedrz, line 132:
```suggestion
    /// Returns an earlier error if, for example, we are already connected to the peer.
```

node/network/src/peering.rs
```diff
@@ -83,77 +97,59 @@ pub trait PeerPoolHandling<N: Network>: P2P {
...
     /// Otherwise, it returns a handle to the tokio tasks that sets up the connection.
-    fn connect(&self, listener_addr: SocketAddr) -> Option<task::JoinHandle<bool>> {
+    ///
+    /// # Concurrency
+    /// Only one task may call this function for a given listener address at a time.
```
ljedrz, line 136: do we enforce this? I don't think we do, so the current wording might be imprecise ("may" instead of "should")

ljedrz (dismissed):
LGTM with some final nits :ok_hand:.

node/network/src/lib.rs
```diff
@@ -44,8 +44,8 @@ pub fn bootstrap_peers<N: Network>(is_dev: bool) -> Vec<SocketAddr> {
         // Development testing contains optional bootstrap peers loaded from the environment.
         match std::env::var("TEST_BOOTSTRAP_PEERS") {
             Ok(peers) => peers.split(',').map(|peer| SocketAddr::from_str(peer).unwrap()).collect(),
-            Err(err) => {
-                warn!("Failed to load bootstrap peers from environment: {err}");
+            Err(_err) => {
+                // error may include the innocent 'environment variable not found'
```
vicsn: I recall we had a change somewhere to distinguish the not found case

node/tcp/src/tcp.rs
```diff
@@ -76,13 +80,54 @@ pub enum ConnectError {
...
+    BannedIp { ip: IpAddr },
+    // Socket errors, such as "connection refused".
+    #[error(transparent)]
     IoError(std::io::Error),
+    // An application-specific reason to reject the connection or abort the handshake.
+    // For snarkOS, this is either a `DisconnectReason` or a `PeeringError`, which do not fully implement `std::error::Error`.
+    #[error("{0}")]
+    ApplicationError(Box<dyn ApplicationError>),
+    /// An unexpected error at the application layer and certain deserialization errors.
+    /// TODO(kaimast): (some of) these should be treated with higher severity, as they indicate a bug or corrupted state,
```
vicsn, line 93: Do `warn` this category of errors in this PR?

What happened to them before this PR?

node/network/src/lib.rs
kaimast: It is in https://github.com/ProvableHQ/snarkOS/pull/4050. I thought it would be merged at that point, but I can add the commit to this branch as well.

node/network/src/peering.rs
kaimast: The heartbeat logic operates in a closed loop and waits for all connection attempts to complete before retrying. In this setup connection attempts can happen concurrently (if heartbeat selects multiple peers to connect to), but not to the same address.

I want to use "may"/"shall" here to make it clear that it is not how the API is intended to be used. To support this properly, we would need to write lock the peer pool before checking the connection.

node/network/src/lib.rs
kaimast: I will push the commit to this branch as well.

node/tcp/src/tcp.rs
kaimast: Before, these errors would not be returned as a `ConnectError`, Instead it would print warning in `connect` and return `None` [here](https://github.com/ProvableHQ/snarkOS/pull/3902/changes/BASE..ded363b916846ba71e14c9d6bb394cb352cd8d89#diff-53a9b3557e08dc171a0fa4df2120a4df691d0cb695bf3f4c4bae7ac415f8611dL128). 

We log a warning for all failed connection attempts. With this change the warning happens in the heartbeat logic.

vicsn (dismissed):
Pending Lukasz' approval!

ljedrz (dismissed):
LGTM

kaimast:
I resolved a conflict with `staging` and it dismissed the reviews again. Please re-approve.

vicsn (approved):

ljedrz (approved):

Issue #3904, opened 2025-10-01
[Feature] Limit the size of KnownPeers


ljedrz:
This is something I've noticed while working on https://github.com/ProvableHQ/snarkOS/issues/3888, which targets the high-level list of peers. We also have a lower-level list, which may also grow large given enough time.

PR #3905, opened 2025-10-01
[Feat] Limit peer pool size, remove banned peers from the pool

### LINKED ISSUE #3872: [Proposal] Consolidate the Resolvers

ljedrz:
This PR is larger than I expected, so I decided to also create an [issue](https://github.com/ProvableHQ/snarkOS/issues/3904) for the remaining bits.

The changes are as follows:
- https://github.com/ProvableHQ/snarkOS/commit/650bcd8ed55bab870a8bf9bde1a056bf24e169ce - a prerequisite for aligned peer removal, and a useful change on its own; it just replaces the `Router`'s `Resolver` with the one from the `Gateway`, using it for both entities
- https://github.com/ProvableHQ/snarkOS/commit/0d553b3884cb4f5d505d7a930730c040f92a40ba - makes peer removal distinct from its status downgrade, and aligns it between the `Gateway` and the `Router`, moving all auxiliary cleanups to the `Disconnect` trait which already handles those
- https://github.com/ProvableHQ/snarkOS/commit/edabb73ce7ba92a21e868fc96e87bf9c5fc3d6d9 - causes bans to prune the related peers from the pool
- https://github.com/ProvableHQ/snarkOS/commit/2a58ca2bfcf6c3cfdbcc4a3f4b66fa474d4bacb5 - disallows outbound connections to IP-banned peers; it's already done in the handshake, but this makes it more performant
- https://github.com/ProvableHQ/snarkOS/commit/52f0934b5bcee3a3086d5f8464ea8efa98f8b8de - removes any banned IPs from inbound peer lists
- https://github.com/ProvableHQ/snarkOS/commit/dfd553c7d7e2b5f22b5a19d94d04f84f09b623af - aligns the way candidate peers are inserted in both the `Gateway` and the `Router`, and makes it so that validator peer lists go through the peer pool before any connection attempts are made to them (in order for the heartbeat to fully control that process)
- https://github.com/ProvableHQ/snarkOS/commit/07763e54f2e25ecfc002832ac07ccbb8779eeeca - applies candidate peer slashing when new entries are about to breach the peer pool size limit

Open points: 
- the exact values for the new limits
- this approach doesn't guard us from breaching the peer pool size limit when we receive a direct connection from an unknown peer; if we want to do so, I propose to do it as a follow-up, as it may also be a larger change than one might expect (if it is to be performant and idiomatic)
- there is a minor difference between how the `Gateway` and the `Router` filtered `dev` addresses (`!is_local_ip` vs `!is_bogon_ip`); I left the latter for now, but this may need to be distinguished based on `Self::OWNER` if we see any test/devnet issues

Closes https://github.com/ProvableHQ/snarkOS/issues/3872.
CC https://github.com/ProvableHQ/snarkOS/issues/3888.


node/router/src/lib.rs
```diff
@@ -183,6 +199,79 @@ pub trait PeerPoolHandling<N: Network>: P2P {
         }
     }
 
+    /// Downgrades a connected peer to candidate status.
+    fn downgrade_peer_to_candidate(&self, listener_addr: SocketAddr) {
+        if let Some(peer) = self.peer_pool().write().get_mut(&listener_addr) {
+            if let Peer::Connected(peer) = peer {
+                self.resolver().write().remove_peer(peer.connected_addr, peer.aleo_addr);
```
vicsn: Is the router peer's aleo_addr verified with a signature verification? If not then we would risk someone spoofing the validator address

node/bft/src/gateway.rs
```diff
@@ -161,11 +161,21 @@ pub struct InnerGateway<N: Network> {
 }
 
 impl<N: Network> PeerPoolHandling<N> for Gateway<N> {
+    const MAXIMUM_POOL_SIZE: usize = 200;
     const OWNER: &str = CONTEXT;
+    const PEER_SLASHING_COUNT: usize = 40;
```
vicsn:
```suggestion
    const PEER_SLASHING_COUNT: usize = 20;
```

node/router/src/inbound.rs
```diff
@@ -349,20 +349,17 @@ pub trait Inbound<N: Network>: Reading + Outbound<N> {
     }
 
     /// Handles a `PeerResponse` message.
-    fn peer_response(&self, _peer_ip: SocketAddr, peers: &[SocketAddr]) -> bool {
-        // Check if the number of peers received is less than MAX_PEERS_TO_SEND.
-        if peers.len() > MAX_PEERS_TO_SEND {
+    fn peer_response(&self, _peer_ip: SocketAddr, peers: Vec<SocketAddr>) -> bool {
+        // Check if the number of peers received is less than MAX_PEERS_TO_SEND or empty.
+        if peers.len() > MAX_PEERS_TO_SEND || peers.is_empty() {
```
vicsn: Could an empty peer list be valid behaviour? I would remove the change as it does not feel like durable/important.

node/router/src/lib.rs
```diff
@@ -83,8 +83,20 @@ const PEER_CACHE_FILENAME: &str = "cached_router_peers";
 pub trait PeerPoolHandling<N: Network>: P2P {
     const OWNER: &str;
 
+    /// The maximum number of peers permitted to be stored in the peer pool.
+    const MAXIMUM_POOL_SIZE: usize = 1_000;
```
vicsn: I would not set any defaults so we have uniform reading experience for gateway v.s. router.

An individual peer may send us 256 ip addresses, so let's stick to 10k so we don't go over the limit so quickly.

node/router/src/lib.rs
```diff
@@ -183,6 +199,79 @@ pub trait PeerPoolHandling<N: Network>: P2P {
...
+            });
+        }
+
+        // If we've managed to filter out every entry, there's nothing to do.
+        if listener_addrs.is_empty() {
+            return;
+        }
+
+        // If we're about to exceed the peer pool size limit, apply candidate slashing.
+        if self.number_of_peers() + listener_addrs.len() >= Self::MAXIMUM_POOL_SIZE && Self::PEER_SLASHING_COUNT != 0 {
```
vicsn, line 234: If we're close to the limit, and we receive multiple PeerResponses, might we trigger this logic multiple times, potentially slashing down to 0?

Granted, the snapshot is unlikely to be very different, so we'll likely slash the same peers.

Holding a lock should not be too much of an issue because this function is just a background utility. In fact, doing the slashing serially might be more kind to CPU.

node/bft/src/gateway.rs
```diff
@@ -752,45 +741,24 @@ impl<N: Network> Gateway<N> {
...
+                    .into_iter()
+                    .filter_map(|(listener_addr, aleo_addr)| {
+                        (self.account.address() != aleo_addr
+                            && !self.is_connected_address(aleo_addr)
+                            && self.is_authorized_validator_address(aleo_addr))
+                        .then_some(listener_addr)
+                    })
+                    .collect::<Vec<_>>();
+                if !valid_addrs.is_empty() {
+                    self.insert_candidate_peers(valid_addrs);
```
kaimast, line 755: Great change!

node/bft/src/gateway.rs
```diff
@@ -1150,13 +1118,23 @@ impl<N: Network> Disconnect for Gateway<N> {
...
     async fn handle_disconnect(&self, peer_addr: SocketAddr) {
         if let Some(peer_ip) = self.resolve_to_listener(&peer_addr) {
-            self.remove_connected_peer(peer_ip);
-
+            self.downgrade_peer_to_candidate(peer_ip);
+            // Remove the peer from the sync module. Except for some tests, there is always a sync sender.
+            if let Some(sync_sender) = self.sync_sender.get() {
+                let tx_block_sync_remove_peer_ = sync_sender.tx_block_sync_remove_peer.clone();
+                tokio::spawn(async move {
+                    if let Err(e) = tx_block_sync_remove_peer_.send(peer_ip).await {
```
kaimast, line 1123: This will only block in the (unlikely) case the channel is full. Does it make sense to spawn a dedicate task for it?

node/bft/src/gateway.rs
```diff
@@ -1150,13 +1118,23 @@ impl<N: Network> Disconnect for Gateway<N> {
...
+                        warn!("Unable to remove '{peer_ip}' from the sync module - {e}");
```
kaimast, line 1124: This error indicates the channel was closed and not whether the sync module successfully removed the peer (part of the reason why I was hoping we can merge #3858 eventually).

Can this happen during normal operation or only during failure/shutdown? If it is the latter, the severity should be bumped to `error`. Also, maybe change the message to something like "Unable to notify sync module about peer removal".

node/router/src/helpers/resolver.rs
```diff
@@ -13,30 +13,87 @@
...
 use std::{collections::HashMap, net::SocketAddr};
 
-/// The `Resolver` provides the means to map the connected address (used in the lower-level
-/// `tcp` module internals, and provided by the OS) to the listener address (used as the
-/// unique peer identifier in the higher-level functions).
-#[derive(Debug, Default)]
-pub(crate) struct Resolver {
-    /// The map of the connected peer address to the correponding listener address.
+/// The resolver contains some reverse maps for peers which are not available
+/// by default to the implementors of PeerPoolHandling (who already contain
```
kaimast: nit: 
```suggestion
/// by default to the implementors of PeerPoolHandling (which already contains
```

node/router/src/helpers/resolver.rs
```diff
@@ -13,30 +13,87 @@
...
+
 use std::{collections::HashMap, net::SocketAddr};
 
-/// The `Resolver` provides the means to map the connected address (used in the lower-level
-/// `tcp` module internals, and provided by the OS) to the listener address (used as the
-/// unique peer identifier in the higher-level functions).
-#[derive(Debug, Default)]
-pub(crate) struct Resolver {
-    /// The map of the connected peer address to the correponding listener address.
+/// The resolver contains some reverse maps for peers which are not available
```
kaimast: `some` seems too vague. Change it to "additional" or remove it.

node/router/src/helpers/resolver.rs
```diff
@@ -13,30 +13,87 @@
...
+/// by default to the implementors of PeerPoolHandling (who already contain
+/// maps from the peer's listening address to their various components).
+#[derive(Debug)]
+pub struct Resolver<N: Network> {
+    /// The map of the (ambiguous) peer address to listener address.
     to_listener: HashMap<SocketAddr, SocketAddr>,
+    /// A map of `address` to `peer IP`.
```
kaimast: Could you clarify which IP/address this is? I assume the listener one? It would be good to use the same terminology for both doc comments.

node/router/src/helpers/resolver.rs
```diff
@@ -13,30 +13,87 @@
...
+impl<N: Network> Resolver<N> {
+    /// Returns the listener address for the given (ambiguous) peer address, if it exists.
+    pub fn get_listener(&self, peer_addr: SocketAddr) -> Option<SocketAddr> {
+        self.to_listener.get(&peer_addr).copied()
     }
 
-    /// Inserts a new mapping of a connected address to the corresponding listener address.
-    pub fn insert_peer(&mut self, listener_addr: SocketAddr, connected_addr: SocketAddr) {
-        self.to_listener.insert(connected_addr, listener_addr);
+    /// Returns the peer IP for the given address.
```
kaimast:
```suggestion
    /// Returns the peer IP for the given Aleo address.
```
To reduce the `SocketAddr`/`Address` confusion a little.

Also, the same comment as above. If this is the listening address, we should change it to 
```suggestion
    /// Returns the listening address for the peer with the given Aleo address.
```
(This also uses "address" twice, but it is a little clearer)

node/router/src/helpers/resolver.rs
```diff
@@ -13,30 +13,87 @@
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use snarkvm::{prelude::Rng, utilities::TestRng};
+
+    type CurrentNetwork = snarkvm::prelude::MainnetV0;
+
+    #[test]
+    fn test_resolver() {
```
kaimast, line 90: It would be good to leave top-level comment here that explains what the test does, so the next person knows what to do if/when it breaks.
Alternatively, you could leave comments at each assert block.

node/router/src/handshake.rs
```diff
@@ -134,7 +134,7 @@ impl<N: Network> Router<N> {
         if let Some(addr) = listener_addr {
             if let Ok(ref cr) = handshake_result {
                 if let Some(peer) = self.peer_pool.write().get_mut(&addr) {
-                    self.resolver.write().insert_peer(peer.listener_addr(), peer_addr);
+                    self.resolver.write().insert_peer(peer.listener_addr(), peer_addr, cr.address);
                     peer.upgrade_to_connected(peer_addr, cr.listener_port, cr.address, cr.node_type, cr.version);
                 }
```
kaimast, line 152: Unrelated to your changes, but I recently loooked at this section as well: Is it not a problem when the peer is not in the peer pool? Does it make sense to silently fail here, or should we at least log?

node/router/src/lib.rs
```diff
@@ -183,6 +199,79 @@ pub trait PeerPoolHandling<N: Network>: P2P {
...
+
+        // If we've managed to truncate to 0, exit.
+        if listener_addrs.is_empty() {
+            return;
+        }
+
+        // Insert new candidate peers.
+        let mut peer_pool = self.peer_pool().write();
+        for addr in listener_addrs {
+            peer_pool.insert(addr, Peer::new_candidate(addr, false));
```
kaimast, line 271: Is this safe? Couldn't it overwrite existing entries?

node/router/src/lib.rs
```diff
@@ -422,15 +516,19 @@ pub trait PeerPoolHandling<N: Network>: P2P {
...
 
     /// Temporarily IP-ban and disconnect from the peer with the given listener address and an
-    /// optional reason for the ban.
+    /// optional reason for the ban. This also removes the peer from the candidate pool.
     fn ip_ban_peer(&self, listener_addr: SocketAddr, reason: Option<&str>) {
         let ip = listener_addr.ip();
         debug!("IP-banning {ip}{}", reason.map(|r| format!(" reason: {r}")).unwrap_or_default());
 
+        // Insert/update the low-level IP ban list.
         let tcp = self.tcp().clone();
```
kaimast: Is this clone needed?

kaimast (dismissed):
Looks great to me. I left multiple small comments that we can hopefully address quickly.

node/bft/src/gateway.rs
ljedrz: That's a valid point, though it's beyond the scope of this PR - this code is only moved. It would be a great idea to soon do a fresh pass with `tokio console` to ensure that we're not spawning tasks too much etc.

node/bft/src/gateway.rs
ljedrz: Again, a valid concern, but outside of this PR's scope; feel free to open a related issue if you feel like this should be adjusted. I think this can only happen during shutdown, which aborts all registered task handles (including this one), which would likely hide any related (non-)errors.

node/router/src/handshake.rs
ljedrz: You mean the `upgrade_to_connected`? It's impossible for the peer to not belong in the peer pool at this point, as it's right underneath `if let Some(peer) = self.peer_pool.write().get_mut(&addr)`.

node/router/src/lib.rs
ljedrz: It's safe, because at the very beginning of `insert_candidate_peers` we perform a `!peer_pool.contains_key(&addr)` check for every entry. In other words, only completely new peers may be introduced to the peer pool via `insert_candidate_peers`.

node/router/src/lib.rs
ljedrz: It is in fact superfluous; I don't know why it was there originally, perhaps a remnant of back when it was delegated to some spawned task.

node/router/src/inbound.rs
ljedrz: Ah, fair, I didn't realize it constituted a protocol violation.

node/router/src/lib.rs
ljedrz: Great point; I decided to go with making the Aleo address optional in the `Resolver`, and only provide it for validators.

ljedrz:
All review comments have been addressed, and `staging` was merged into the PR.

node/bft/src/gateway.rs
vicsn: > It would be a great idea to soon do a fresh pass with tokio console to ensure that we're not spawning tasks too much etc.

Can you make an issue for this? And do note if this is something we should add as a CI check, or just a certain set of commands we should run manually before releases.

vicsn (dismissed):
LGTM, let's await Kai's review before merging

node/bft/src/gateway.rs
ljedrz: https://github.com/ProvableHQ/snarkOS/issues/3909

node/router/src/handshake.rs
kaimast: I mean for the case when `self.peer_pool.write().get_mut(&addr)` returns `None` it will still log   `Completed the handshake with '{peer_addr}'`. Does that make sense?

kaimast: Marking this as resolved, though, as this is something holding up merging the PR.

node/router/src/lib.rs
kaimast: Aren't you dropping the lock between this line and the initial check? Maybe you could upgrade the read to write lock instead, or hold a write lock throughout the entire duration of the function.

node/router/src/handshake.rs
ljedrz: ah, the bit underneath this snippet; it's impossible for a peer to not be in the peer pool at this point, as the handshake has succeeded, and the related `Peer` is in `Connecting` state, meaning any concurrent handshake attempts can't change that state until it's changed by this function; the log and the metrics update are put outside of the lifetime of the guard to reduce its grip on the peer pool

node/router/src/lib.rs
ljedrz: yes, and it's deliberate - the initial filtering might be longer than we'd like (especially if additional filtering criteria are added in the future), so the scope of the write guard is limited to the minimum; it used to be even more fine-grained, until [this comment](https://github.com/ProvableHQ/snarkOS/pull/3905#discussion_r2395302577)

kaimast: I still don't really understand how race conditions are avoided. 

My concern was something like this:
1. Task 1 calls `insert_candidate_peers` after receiving a `PeerResponse`
2. Task 1 grabs the read lock and determines a peer p is not in the pool.
3. Task 2 does the same as in (1)
4. Task 2 does the same as in (2)
5. Task 2 grabs the write lock and inserts p into the pool.
6. Task 3 runs the heartbeat logic and upgrades the peer to connecting.
7. Task 2 wakes up, grabs the write lock, and overwrites the entry for p.

ljedrz: this is not a concern with candidates - as long as it's a candidate peer, we can overwrite it as much as we like;

edit: I'll need to think about the connecting case, though

kaimast:
It still appears to me that there are potential race conditions. Read-write dependencies exist in a few places (e.g., when truncating the peers to add based on the existing number of candidate peers), but not for all of them a lock is held the entirety of the critical section.

If you think these race conditions are unlikely or the outcome is benign, e.g., a node might end up with some candidate peers too few/many, then I am okay with merging it.

For the line where it inserts new candidate peers, I would still prefer if it (double) checks it will not overwrite an existing connected peer. Using `match peer_pool.entry(addr)` should not be more expensive than `peer_pool.insert(addr, ..)`.

ljedrz:
I've given it some thought, and decided to hold the write guard for the entire duration of `insert_candidate_peers`; the `RwLock` we're using implements eventual fairness, and we can easily guard ourselves against unsolicited peer lists if we ever suspect it to be an issue.

vicsn (approved):

PR #3910, opened 2025-10-02
Propagate expensive transmissions over to the next batch.

### LINKED ISSUE #3761: [Feature] Iteratively pack proposals to increase TPS
The current logic in the ready queue stops filling batches when the spend limit is surpassed or when the batch size is attained. In the first case, batches may have remaining capacity in terms of size (a nice knapsack problem) and so we should
1. skip the transaction that brings the batch above the spend limit
2. try the next transaction in the ready queue if it's cheaper
3. repeat the process until we've packed the batch with transactions that don't surpass the spend limit
4. prioritise the skipped transactions in the next round

usagi32:
## Motivation

should close #3761 

## Test Plan

Not sure if this requires anything extra other than the CI but maybe specific test cases can be looked at, if any.

### DIFF
--- a/node/bft/src/primary.rs
+++ b/node/bft/src/primary.rs
@@ -510,6 +510,10 @@ impl<N: Network> Primary<N> {
 
         // Initialize the map of transmissions.
         let mut transmissions: IndexMap<_, _> = Default::default();
+        // Transmissions that caused proposal cost to overflow the spend limit budget in the current batch.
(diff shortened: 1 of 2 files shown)


vicsn:
Thank you for the contribution. The issue and solution needs more thought and analysis regarding the trade-offs. Marking the issue as "wontfix" because it's not offering a lot of value.

PR #3914, opened 2025-10-06
Enable dev nodes and tests to set bootstrap peers

vicsn:
## Motivation

So we can better test peering in our stress tests.

### DIFF
--- a/node/router/src/lib.rs
+++ b/node/router/src/lib.rs
@@ -698,8 +698,14 @@ impl<N: Network> CommunicationService for Router<N> {
 #[allow(clippy::if_same_then_else)]
 pub fn bootstrap_peers<N: Network>(is_dev: bool) -> Vec<SocketAddr> {
     if cfg!(feature = "test") || is_dev {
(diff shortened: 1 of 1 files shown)


ljedrz (approved):
While this is fine, perhaps we could introduce some new regular param that would overwrite any hardcoded bootstrap addresses? It might be useful for some non-dev users too.

vicsn:
> While this is fine, perhaps we could introduce some new regular param that would overwrite any hardcoded bootstrap addresses? It might be useful for some non-dev users too.

Can you explain/reword/give an example? I don't understand. Do you mean to add a --bootstrap <ip:port> flag to the CLI instead of using ENV variables?

ljedrz:
> Do you mean to add a --bootstrap ip:port flag to the CLI instead of using ENV variables?

Yes, as the bootstrap peers are currently all hardcoded - instead, perhaps they could just get their dedicated param like `--bootstrap` that would overwrite their list regardless of the network and `--dev` status.

PR #3915, opened 2025-10-07
[Feature] Improve signal and shutdown handling

kaimast:
These changes are part of #3874, but were moved to their own PR to reduce the size of the latter.

### Motivation
Before, #4021, snarkOS nodes would call `std::process::exit` before shutting down. This was problematic _"because this function never returns [and] terminates the process, no destructors on the current stack or any other thread’s stack will be run"_ ([source](https://doc.rust-lang.org/std/process/fn.exit.html)). 
The code worked around this by sleeping for multiple seconds during shutdown. Such a sleep may be too long in a test setting and too short in production. Additionally, the lack of a clean shutdown makes it more complicated to perform certain tasks at shutdown, such as caching the block tree on disk or flushing the log file.

#4021, as far as I understand, was intended to be a quick fix until this slightly cleaner version is merged. This PR improves on the former by encapsulating all signal handling logic in a dedicated module and avoids spawning the main node logic in a detached task, which makes catching panics easier.
For `snarkos-display` shutting down using `Escape` will now work the same as with Ctrl+C.

### Proposed Changes
In the current design, the code passes around an `AtomicBool` as a shutdown flag. This design generally works, but does not allow waiting for the flag to be set.  This PR introduces a `SignalHandler` struct and a `Stoppable` trait.

The `Stoppable` trait serves the same function as the `AtomicBool` did before, but it is a little clearer what its purpose is from its struct and function names alone.

The `SignalHandler` struct implements the `Stoppable` trait and also launches a background task that wait for Ctrl+C. 

### Testing
To test, I verified that shut down works correctly in multiple scenarios. "correctly" means that the process exists cleanly, and that the on-disk state is not corrupted. The latter was verified by starting the node again and ensuring it can load the stored state fine.

I went through the following scenarios
* started a production node and connected to canary
* ran the production node with and without terminal UI (display)
* ran production node sync with and without CDN

Additionally, I ran a development network and verified that nodes all nodes shut down cleanly without getting stuck or returning nonzero exit codes.

### DIFF
--- a/cli/src/commands/start.rs
+++ b/cli/src/commands/start.rs
@@ -288,26 +288,15 @@ impl Start {
         Self::runtime().block_on(async move {
             // Error messages.
             let node_parse_error = || "Failed to start node";
-            let signal_handler = SignalHandler::new();
 
             // Clone the configurations.
             let mut self_ = self.clone();
 
             // Parse the node arguments, start it, and block until shutdown.
             match self_.network {
-                MainnetV0::ID => self_
-                    .parse_node::<MainnetV0>(log_receiver, signal_handler.clone())
-                    .await
-                    .with_context(node_parse_error)?,
-
-                TestnetV0::ID => self_
-                    .parse_node::<TestnetV0>(log_receiver, signal_handler.clone())
-                    .await
-                    .with_context(node_parse_error)?,
-                CanaryV0::ID => self_
-                    .parse_node::<CanaryV0>(log_receiver, signal_handler.clone())
-                    .await
-                    .with_context(node_parse_error)?,
+                MainnetV0::ID => self_.parse_node::<MainnetV0>(log_receiver).await.with_context(node_parse_error)?,
+                TestnetV0::ID => self_.parse_node::<TestnetV0>(log_receiver).await.with_context(node_parse_error)?,
+                CanaryV0::ID => self_.parse_node::<CanaryV0>(log_receiver).await.with_context(node_parse_error)?,
                 _ => panic!("Invalid network ID specified"),
             };
 
@@ -577,7 +566,7 @@ impl Start {
 
     /// Start the node and blocks until it terminates.
     #[rustfmt::skip]
-    async fn parse_node<N: Network>(&mut self, log_receiver: mpsc::Receiver<Vec<u8>>, signal_handler: Arc<SignalHandler>) -> Result<()> {
+    async fn parse_node<N: Network>(&mut self, log_receiver: mpsc::Receiver<Vec<u8>>) -> Result<()> {
         if !self.nobanner {
             // Print the welcome banner.
             println!("{}", crate::helpers::welcome_message());
@@ -721,6 +710,9 @@ impl Start {
         if !self.nodisplay && !self.nocdn {
             println!("🪧 The terminal UI will not start until the node has finished syncing from the CDN. If this step takes too long, consider restarting with `--nodisplay`.");
         }
+
+        // Register the signal handler.
+        let signal_handler = SignalHandler::new();
 
         // Initialize the node.
         let node = match node_type {
--- a/cli/src/commands/start.rs
+++ b/cli/src/commands/start.rs
@@ -24,6 +24,8 @@ use snarkos_node::{
     rest::DEFAULT_REST_PORT,
     router::DEFAULT_NODE_PORT,
 };
+use snarkos_utilities::SignalHandler;
+
 use snarkvm::{
     console::{
         account::{Address, PrivateKey},
@@ -57,7 +59,7 @@ use std::{
 };
 use tokio::{
     runtime::{self, Runtime},
-    sync::oneshot,
+    sync::mpsc,
     task,
 };
 use tracing::warn;
@@ -267,7 +269,7 @@ pub struct Start {
 }
 
 impl Start {
-    /// Starts the snarkOS node.
+    /// Starts the snarkOS node and blocks until it terminates.
     pub fn parse(self) -> Result<String> {
         // Prepare the shutdown flag.
         let shutdown: Arc<AtomicBool> = Default::default();
@@ -285,56 +287,29 @@ impl Start {
         // Initialize the runtime.
         Self::runtime().block_on(async move {
             // Error messages.
-            let node_parse_error = || "Failed to parse node arguments";
-            let display_start_error = || "Failed to initialize the display";
-
-            let (shutdown_tx, shutdown_rx) = oneshot::channel();
+            let node_parse_error = || "Failed to start node";
+            let signal_handler = SignalHandler::new();
 
             // Clone the configurations.
-            let mut cli = self.clone();
-            // Parse the network.
-            match cli.network {
-                MainnetV0::ID => {
-                    // Parse the node from the configurations.
-                    let node = cli
-                        .parse_node::<MainnetV0>(shutdown.clone(), shutdown_tx)
-                        .await
-                        .with_context(node_parse_error)?;
-                    // If the display is enabled, render the display.
-                    if !cli.nodisplay {
-                        // Initialize the display.
-                        Display::start(node, log_receiver).with_context(display_start_error)?;
-                    }
-                }
-                TestnetV0::ID => {
-                    // Parse the node from the configurations.
-                    let node = cli
-                        .parse_node::<TestnetV0>(shutdown.clone(), shutdown_tx)
-                        .await
-                        .with_context(node_parse_error)?;
-                    // If the display is enabled, render the display.
-                    if !cli.nodisplay {
-                        // Initialize the display.
-                        Display::start(node, log_receiver).with_context(display_start_error)?;
-                    }
-                }
-                CanaryV0::ID => {
-                    // Parse the node from the configurations.
-                    let node = cli
-                        .parse_node::<CanaryV0>(shutdown.clone(), shutdown_tx)
-                        .await
-                        .with_context(node_parse_error)?;
-                    // If the display is enabled, render the display.
-                    if !cli.nodisplay {
-                        // Initialize the display.
-                        Display::start(node, log_receiver).with_context(display_start_error)?;
-                    }
-                }
+            let mut self_ = self.clone();
+
+            // Parse the node arguments, start it, and block until shutdown.
+            match self_.network {
+                MainnetV0::ID => self_
+                    .parse_node::<MainnetV0>(log_receiver, signal_handler.clone())
+                    .await
+                    .with_context(node_parse_error)?,
+
+                TestnetV0::ID => self_
+                    .parse_node::<TestnetV0>(log_receiver, signal_handler.clone())
+                    .await
+                    .with_context(node_parse_error)?,
+                CanaryV0::ID => self_
+                    .parse_node::<CanaryV0>(log_receiver, signal_handler.clone())
+                    .await
+                    .with_context(node_parse_error)?,
                 _ => panic!("Invalid network ID specified"),
             };
-
-            // Wait for the shutdown signal.
-            let _ = shutdown_rx.await;
 
             Ok(String::new())
         })
@@ -600,9 +575,9 @@ impl Start {
         }
     }
 
-    /// Returns the node type corresponding to the given configurations.
+    /// Start the node and blocks until it terminates.
     #[rustfmt::skip]
-    async fn parse_node<N: Network>(&mut self, shutdown: Arc<AtomicBool>, shutdown_tx: oneshot::Sender<()>) -> Result<Node<N>> {
+    async fn parse_node<N: Network>(&mut self, log_receiver: mpsc::Receiver<Vec<u8>>, signal_handler: Arc<SignalHandler>) -> Result<()> {
         if !self.nobanner {
             // Print the welcome banner.
             println!("{}", crate::helpers::welcome_message());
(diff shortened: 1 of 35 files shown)


node/src/traits.rs
```diff
@@ -55,67 +49,21 @@ pub trait NodeInterface<N: Network>: Routing<N> {
...
-        });
-
-        node
+    /// Blocks until a shutdown signal was received or manual shutdown was triggered.
+    async fn wait_for_signals(&self, handler: &SignalHandler) {
+        handler.wait_for_signals().await;
+
+        warn!("==========================================================================================");
+        warn!("⚠️  Attention - Starting the graceful shutdown procedure (ETA: 30 seconds)...");
+        warn!("⚠️  Attention - To avoid DATA CORRUPTION, do NOT interrupt snarkOS (or press Ctrl+C again)");
```
ljedrz: we probably no longer need to worry about any corruption since we're not using `process::exit`

utilities/src/signals.rs
```diff
@@ -0,0 +1,140 @@
...
+/// This is useful when no signal or complex shutdown handling is necessary (e.g., in a test environment).
+pub struct SimpleStoppable {
+    state: AtomicBool,
+}
+
+impl SimpleStoppable {
+    pub fn new() -> Arc<Self> {
+        Arc::new(Self { state: AtomicBool::new(false) })
+    }
+}
```
ljedrz, line 50: nit: this is just `Default`, which can be derived on `SimpleStoppable`

utilities/src/signals.rs
```diff
@@ -0,0 +1,140 @@
...
+
+impl Stoppable for SimpleStoppable {
+    fn stop(&self) {
+        self.state.store(true, Ordering::SeqCst);
```
ljedrz, line 54: do we really need `SeqCst`? it doesn't matter in which order this is set, and it will never be set to anything other than `true` post-creation

utilities/src/signals.rs
```diff
@@ -0,0 +1,140 @@
...
+    /// Spawns a background tasks that listens for Ctrl+C and returns `Self`.
+    pub fn new() -> Arc<Self> {
+        let obj = Arc::new(Self { stopped: AtomicBool::new(false), notify: Default::default() });
+
+        {
+            let obj = obj.clone();
+            tokio::spawn(async move {
+                obj.handle_signals().await;
+            });
+        }
```
ljedrz, line 89: nit: I'd move the spawning to `handle_signals`, so as not to confuse object creation with doing additional work

utilities/src/signals.rs
```diff
@@ -0,0 +1,140 @@
...
+            Ok(()) => {}
+            Err(error) => {
+                error!("tokio::signal encountered an error: {error}");
+            }
+        }
+
+        self.stop();
+    }
+
+    /// Blocks until the signal handler was invoked or the stopped flag was set some other way.
```
ljedrz: nit: I'd avoid the "blocks" term, as it is technically non-blocking

utilities/src/signals.rs
```diff
@@ -0,0 +1,140 @@
...
+    /// Note: This can only be called once, and must not be called concurrently.
+    pub async fn wait_for_signals(&self) {
+        while !self.is_stopped() {
+            self.notify.notified().await
+        }
```
ljedrz: this seems a little bit convoluted; wouldn't `self.notify.notified().await` suffice?

node/src/traits.rs
```diff
@@ -55,67 +49,21 @@ pub trait NodeInterface<N: Network>: Routing<N> {
...
+        warn!("⚠️  Attention - Please wait until the shutdown gracefully completes (ETA: 30 seconds)");
+        warn!("==========================================================================================");
+
+        // If the node is already initialized, then shut it down.
+        self.shut_down().await;
+
+        // A best-effort attempt to let any ongoing activity conclude.
+        tokio::time::sleep(Duration::from_secs(3)).await;
```
ljedrz: this should also be redundant

utilities/src/signals.rs
```diff
@@ -0,0 +1,140 @@
...
+    fn is_stopped(&self) -> bool {
+        self.state.load(Ordering::SeqCst)
+    }
+}
+
+/// Helper for signal handling that implements the `Stoppable` trait.
+///
+/// This struct will set itself to "stopped" as soon as the process receives Ctrl+C.
+/// It can also be manually stopped (e.g., when the node encounters a fatal error).
+pub struct SignalHandler {
```
ljedrz, line 66: if we only ever expect this to exist as a single shared instance, I generally prefer a clonable `struct Wrapper(Arc<ActualObject>);` that `Deref`s to `ActualObject` (rather than it being `Arc`ed) for maximum clarity, and to make sure that no one removes the `Arc` if the owning object becomes `Arc`ed itself

node/src/bootstrap_client/mod.rs
```diff
@@ -230,4 +231,21 @@ impl<N: Network> BootstrapClient<N> {
...
+
+    /// Blocks until a shutdown signal was received or manual shutdown was triggered.
+    pub async fn wait_for_signals(&self, handler: &SignalHandler) {
+        handler.wait_for_signals().await;
+
+        warn!("==========================================================================================");
+        warn!("⚠️  Attention - Starting the graceful shutdown procedure (ETA: 30 seconds)...");
+        warn!("⚠️  Attention - To avoid DATA CORRUPTION, do NOT interrupt snarkOS (or press Ctrl+C again)");
+        warn!("⚠️  Attention - Please wait until the shutdown gracefully completes (ETA: 30 seconds)");
+        warn!("==========================================================================================");
```
ljedrz: the `BootstrapClient` has no ledger, so this is unnecessary - same with the `sleep` below

node/bft/tests/gateway_e2e.rs
```diff
@@ -42,7 +41,7 @@ async fn new_test_gateway(
     let (accounts, committee) = new_test_committee(num_nodes, rng);
     let accounts_ = accounts.clone();
     let mut rng_ = TestRng::fixed(rng.r#gen());
-    let ledger = task::spawn_blocking(move || sample_ledger(&accounts_, &committee, &mut rng_)).await.unwrap();
+    let ledger = sample_ledger(&accounts_, &committee, &mut rng_);
```
ljedrz: the blocking task for ledger creation is unrelated to shutdowns - it's to not block the async executor (applies to other spots as well)

node/bft/src/sync/mod.rs
```diff
@@ -1252,18 +1253,18 @@ mod tests {
...
-        let storage_mode = StorageMode::new_test(None);
         let gateway = Gateway::new(
             account.clone(),
             storage.clone(),
             syncing_ledger.clone(),
             None,
             &[],
             false,
-            storage_mode,
+            StorageMode::Test(None),
```
```suggestion
            StorageMode::new_test(None),
```
this should always be used when a full ledger is involved

utilities/src/signals.rs
```diff
@@ -0,0 +1,140 @@
...
+            Err(error) => {
+                error!("tokio::signal encountered an error: {error}");
+            }
+        }
+
+        self.stop();
+    }
+
+    /// Blocks until the signal handler was invoked or the stopped flag was set some other way.
+    /// Note: This can only be called once, and must not be called concurrently.
```
ljedrz, line 133: for a strict single waiter and a single call, I'd prefer to use a `oneshot` channel for a stronger semantic contract

ljedrz: in that scenario we wouldn't even need an atomic boolean - a shared `Arc<RwLock<Option<oneshot::Sender<()>>>>` could be checked for the presence of the sender to determine if the node is shutting down

utilities/src/signals.rs
kaimast: Good point. If we assume the function is called at most once, this is additional check is indeed not needed.

utilities/src/signals.rs
kaimast: That's a good point. Initially, I thought multiple tasks might need to call this, but this additional complexity is not needed.

utilities/src/signals.rs
kaimast: It's returning `Arc<Self>`, so this cannot be derived by Default

utilities/src/signals.rs
kaimast: I usually set the highest ordering constraint, unless there is a performance argument to lower it.

utilities/src/signals.rs
kaimast: Would you want to spawn it in a separate function? There's really no point in creating this object without the signal handler, so it made sense to me to combine it.

utilities/src/signals.rs
ljedrz: no, I don't think it would be significant

cli/src/commands/start.rs
```diff
@@ -285,57 +287,30 @@ impl Start {
...
-            // Parse the network.
-            match cli.network {
-                MainnetV0::ID => {
-                    // Parse the node from the configurations.
-                    let node = cli
-                        .parse_node::<MainnetV0>(shutdown.clone(), shutdown_tx)
-                        .await
-                        .with_context(node_parse_error)?;
-                    // If the display is enabled, render the display.
-                    if !cli.nodisplay {
```
ljedrz, line 304: is this being deprecated?

cli/src/commands/start.rs
```diff
@@ -742,24 +717,28 @@ impl Start {
...
-            NodeType::Client => Node::new_client(node_ip, rest_ip, self.rest_rps, account, &trusted_peers, genesis, cdn, storage_mode, self.trusted_peers_only, self.dev, shutdown, shutdown_tx).await,
+                 let node = match node_type {
+            NodeType::Validator => Node::new_validator(node_ip, self.bft, rest_ip, self.rest_rps, account, &trusted_peers, &trusted_validators, genesis, cdn, storage_mode, self.trusted_peers_only, dev_txs, self.dev, signal_handler.clone()).await,
+            NodeType::Prover => Node::new_prover(node_ip, account, &trusted_peers, genesis, storage_mode, self.trusted_peers_only, self.dev, signal_handler.clone()).await,
+            NodeType::Client => Node::new_client(node_ip, rest_ip, self.rest_rps, account, &trusted_peers, genesis, cdn, storage_mode, self.trusted_peers_only, self.dev, signal_handler.clone()).await,
             NodeType::BootstrapClient => Node::new_bootstrap_client(node_ip, account, *genesis.header(), self.dev).await,
+        }?;
+
+        if !self.nodisplay {
```
ljedrz, line 725: ah, nevermind, it's still there

node/cdn/src/blocks.rs
```diff
@@ -87,11 +89,11 @@ impl CdnBlockSync {
     pub fn new<N: Network, C: ConsensusStorage<N>>(
         base_url: http::Uri,
         ledger: Ledger<N, C>,
-        shutdown: Arc<AtomicBool>,
+        stoppable: Arc<dyn Stoppable>,
```
ljedrz: this is perfectly fine, but I'm curious: why not use the concrete `SignalHandler` here?

utilities/Cargo.toml
```diff
@@ -0,0 +1,32 @@
...
+edition = "2024"
+
+[dependencies.tokio]
+workspace = true
+features = [ "macros", "signal", "sync" ]
+
+[dependencies.tracing]
+workspace = true
+
+[dependencies.locktick]
```
ljedrz, line 27: this means this new crate needs to go into the `locktick` feature in the workspace TOML

utilities/src/signals.rs
ljedrz: ah fair, I was thinking in terms of an `Arc` wrapped in a `struct`

utilities/src/signals.rs
ljedrz: yeah, it's definitely simpler to use this way, I was just thinking in terms of API clarity; still, it's difficult to misuse, so I don't feel strongly about it (unless other reviewers object to it)

ljedrz (commented):
Just one small change left, and it's good to go :+1:.

node/cdn/src/blocks.rs
kaimast: The `dyn Stoppable` mostly exists for testing puproses, but it is not needed here. I changed it to use the concrete type instead.

utilities/Cargo.toml
kaimast: Done in [aa6ddfc](https://github.com/ProvableHQ/snarkOS/pull/3915/commits/aa6ddfc570cf52a4e5459603837ebddb300af89b). I didn't add it to the top-level, though, as it does not depend directly on it.
The commit makes a few more feature cleanups, such as removing the remaining `parallel` features, and using `locktick`/`serial` in more places.

cli/src/commands/start.rs
```diff
@@ -285,57 +287,30 @@ impl Start {
...
-                        // Initialize the display.
-                        Display::start(node, log_receiver).with_context(display_start_error)?;
-                    }
-                }
+            let mut self_ = self.clone();
+
+            // Parse the node arguments, start it, and block until shutdown.
+            match self_.network {
+                MainnetV0::ID => self_
+                    .parse_node::<MainnetV0>(log_receiver, signal_handler.clone())
```
```suggestion
                    .parse_node::<MainnetV0>(log_receiver, signal_handler)
```
applies to the other spots too

utilities/src/signals.rs
```diff
@@ -0,0 +1,161 @@
...
+        let signal_listener = async move {
+            tokio::signal::ctrl_c().await?;
+            debug!("Received signal");
+
+            std::io::Result::<()>::Ok(())
+        };
+
+        // Block until the signal.
+        match signal_listener.await {
+            Ok(()) => {}
```
ljedrz: maybe we could log here that the signal has been received, just to provide proof of a clean shutdown when debugging? this would apply to all OSes

utilities/src/signals.rs
```diff
@@ -0,0 +1,161 @@
...
+                _ = s_hup.recv() => debug!("Received SIGHUP"),
+            );
+
+            std::io::Result::<()>::Ok(())
+        };
+
+        #[cfg(not(target_family = "unix"))]
+        let signal_listener = async move {
+            tokio::signal::ctrl_c().await?;
+            debug!("Received signal");
```
ljedrz:
```suggestion
```

utilities/src/signals.rs
```diff
@@ -0,0 +1,161 @@
...
+///
+/// This struct will set itself to "stopped" as soon as the process receives Ctrl+C.
+/// It can also be manually stopped (e.g., when the node encounters a fatal error).
+pub struct SignalHandler {
+    /// This sender is used to notify a waiting task that the node has been stopped.
+    /// If this is `None`, the node is in the process of shutting down.
+    stopped_sender: RwLock<Option<oneshot::Sender<()>>>,
+
+    /// This receiver is used to wait for the node to be stopped.
+    stopped_receiver: RwLock<Option<oneshot::Receiver<()>>>,
```
ljedrz:
```suggestion
    stopped_receiver: Mutex<Option<oneshot::Receiver<()>>>,
```
very much a nit: using a `Mutex` would make it abundantly clear that its purpose - combined with the `Option` - is only to facilitate the removal of this item from the parent object (as opposed to the `stopped_sender`, where we do want to read it to check if the signal has been intercepted yet)

ljedrz (dismissed):
Left a few final nits, but basically LGTM :ok_hand:.

ljedrz (approved):

vicsn (commented):
Can you note in the PR README which manual tests were run? Are they comprehensive?

kaimast:
> Can you note in the PR README which manual tests were run? Are they comprehensive?

Done!

vicsn (approved):
LGTM!

PR #3916, opened 2025-10-07
[Feature] Add flamegraph generation

kaimast:
With this PR, nodes create a `snarkos-trace.folded` file on shutdown if the `flamegraph` feature is enabled.

One can then create a flame graph from the file. For example, to generate an SVG of the flame graph using the `inferno` crate, run the following command:
```inferno-flamegraph snarkos-trace.folded > flamegraph.svg```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -587,9 +587,9 @@ dependencies = [
 
 [[package]]
 name = "cc"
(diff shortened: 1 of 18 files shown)


vicsn:
Can you show an example visualization?

PR #3917, opened 2025-10-07
Ensure --rotate-external-peers rotates all peers equally

vicsn:
## Motivation

When E2E testing a network with bootstrap peers, they won't accept new connections. I believe this may resolve the issue.

## Test Plan

- [ ] All clients can connect in E2E test

### DIFF
--- a/node/router/src/heartbeat.rs
+++ b/node/router/src/heartbeat.rs
@@ -119,8 +119,14 @@ pub trait Heartbeat<N: Network>: Outbound<N> {
     ///     - Validators are considered higher priority than provers or clients.
     ///     - Connections that have not been seen in a while are considered lower priority.
     fn get_removable_peers(&self) -> Vec<ConnectedPeer<N>> {
+        // Are we rotating external peers? Just return all connected peers.
+        if self.router().rotate_external_peers() {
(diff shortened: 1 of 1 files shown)


ljedrz (dismissed):
This is fine, though the doc comment might need updating (e.g. no sorting is involved in the new scenario).

ljedrz (approved):
This may result in much rotation, but in principle it's fine.

node/router/src/heartbeat.rs
```diff
@@ -119,8 +119,14 @@ pub trait Heartbeat<N: Network>: Outbound<N> {
     ///     - Validators are considered higher priority than provers or clients.
     ///     - Connections that have not been seen in a while are considered lower priority.
     fn get_removable_peers(&self) -> Vec<ConnectedPeer<N>> {
+        // Are we rotating external peers? Just return all connected peers.
+        if self.router().rotate_external_peers() {
```
kaimast, line 123: Does this mean we disconnect from trusted peers now?

vicsn:
Closing, as this is not needed and is superseeded by https://github.com/ProvableHQ/snarkOS/pull/3920

PR #3920, opened 2025-10-08
[Feat] Introduce NodeType::BootstrapClient

### LINKED ISSUE #3868: [Feature] Create BootstrapClient node type
## 🚀 Feature

This should be enabled with a --bootstrap flag

It should run a gateway and router, so it can gossip both clients and validators.

It should not have to create or sync any blocks.

There is some thinking we may have to do on the validator gossip front. Namely, ideally we only gossip around connections of our (dynamic...) validator set. At the moment, validators only accept connections from [other validators which are in the current validator set](https://github.com/ProvableHQ/snarkOS/blob/a50fb6c7bb7278c9d64914d835e256e5cef7dc59/node/bft/src/gateway.rs#L1515). This would have to be adjusted so:
- validators skip this check when the origin is a trusted Bootstrap IP address
- bootstrap nodes do check they only connect to validators
- review whether this approach opens up an easier MITM attack compared to validators specifying each other's IP addresses directly. I can't think of any attack at the moment. The IP packet origin address set by routers seems to me still the point of failure.

ljedrz:
This PR introduces a new `NodeType::BootstrapClient`, which is a light node that can serve peer lists to all kinds of nodes. It also adjusts the `OnConnect` behavior of other node types to immediately request a list of peers when connecting to one.

I tested these changes by starting a `--dev N --bootstrap-client` node, and connecting to it in both non-validator and validator handshake mode from other `--dev` nodes that had it specified as one of their `TEST_BOOTSTRAP_PEERS`; I also hardcoded the `mainnet` network once to ensure that the latest validator committee is downloaded and deserialized correctly.

Open points / suggestions:
- the `bootstrap_peers` function could now be removed in favor of introducing candidate peers at node's startup (`CandidatePeer` could have an `Option<NodeType>` member, which could also be useful when refreshing peers etc)
- the `Event::ChallengeRequest` object is lacking a `NodeType` member, making it both incompatible with the `Message::ChallengeRequest` one and forcing us to determine the node type based on our list of bootstrap nodes

Stress-tests also succeeded after several iterations.

Closes https://github.com/ProvableHQ/snarkOS/issues/3868.


node/router/messages/src/helpers/node_type.rs
```diff
@@ -77,6 +81,7 @@ impl FromBytes for NodeType {
             0 => Ok(Self::Client),
             1 => Ok(Self::Prover),
             2 => Ok(Self::Validator),
+            3 => Ok(Self::BootstrapClient),
             x => Err(error(format!("Invalid node type: expected 0, 1, or 2, got {x}."))),
```
vicsn:
```suggestion
            x => Err(error(format!("Invalid node type: expected 0..3, got {x}."))),
```

node/router/src/lib.rs
```diff
@@ -669,7 +694,7 @@ impl<N: Network> Router<N> {
         // - Validators and clients may accept older versions, depending on their current block height.
         let lowest_accepted_message_version = match self.node_type {
             // Provers should always use the latest version.
-            NodeType::Prover => Message::<N>::latest_message_version(),
+            NodeType::Prover | NodeType::BootstrapClient => Message::<N>::latest_message_version(),
```
vicsn, line 698: I think it's better to be a bit more lenient and to let the bootstrap client accept messages from lower versions based on migration height (i.e. move to the match arm below.

node/src/bootstrap_client/handshake.rs
```diff
@@ -0,0 +1,321 @@
...
+        let peer_addr = connection.addr();
+        let peer_side = connection.side();
+        let stream = self.borrow_stream(&mut connection);
+
+        // We don't know the listening address yet, as we don't initiate connections.
+        let mut listener_addr = if peer_side == ConnectionSide::Initiator {
+            debug!("Received a connection request from '{peer_addr}'");
+            None
+        } else {
+            unreachable!("The boostrapper clients don't initiate connections");
```
vicsn, line 112: Also not if peers are passed via the --peers or --validators flag? Maybe to prevent this footgun we should disallow setting peers or validators in the start command if --bootstrap-client is set.

node/src/bootstrap_client/network.rs
```diff
@@ -0,0 +1,173 @@
...
+                debug!("Sending {} peer address(es) to '{listener_addr}'", peers.len());
+                let msg = MessageOrEvent::Message(Message::PeerResponse(messages::PeerResponse { peers }));
+                let _ = self.unicast(peer_addr, msg)?.await;
+
+                debug!("Disconnecting from '{listener_addr}' - peers provided");
+                self.tcp().disconnect(peer_addr).await;
+            }
+            MessageOrEvent::Event(Event::ValidatorsRequest(_)) => {
+                debug!("Received a ValidatorsRequest from '{listener_addr}'");
+                let current_committee = match self.get_or_update_committee().await {
```
vicsn, line 149: In dev mode, i.e. in ephemeral devnets, we won't be able to get the committee from the block explorer.

Perhaps we can disable this filtering if dev mode is enabled

node/src/bootstrap_client/network.rs
```diff
@@ -0,0 +1,173 @@
...
+        let Some(listener_addr) = self.resolve_to_listener(peer_addr) else {
+            // Already disconnecting, ignore.
+            return Ok(());
+        };
+
+        // Handle the right peer request.
+        match message {
+            MessageOrEvent::Message(Message::PeerRequest(_)) => {
+                debug!("Received a PeerRequest from '{listener_addr}'");
+                let peers = self.get_best_connected_peers(Some(u8::MAX as usize));
```
vicsn: Can we re-use MAX_PEERS_TO_SEND?

node/src/bootstrap_client/network.rs
```diff
@@ -0,0 +1,173 @@
...
+                let peers = peers.into_iter().map(|peer| (peer.listener_addr, None)).collect::<Vec<_>>();
+
+                debug!("Sending {} peer address(es) to '{listener_addr}'", peers.len());
+                let msg = MessageOrEvent::Message(Message::PeerResponse(messages::PeerResponse { peers }));
+                let _ = self.unicast(peer_addr, msg)?.await;
+
+                debug!("Disconnecting from '{listener_addr}' - peers provided");
+                self.tcp().disconnect(peer_addr).await;
```
vicsn: Savage

node/src/bootstrap_client/network.rs
```diff
@@ -0,0 +1,173 @@
...
+
+                debug!("Sending {} validator address(es) to '{listener_addr}'", validators.len());
+                let msg = MessageOrEvent::Event(Event::ValidatorsResponse(events::ValidatorsResponse { validators }));
+                let _ = self.unicast(peer_addr, msg)?.await;
+
+                debug!("Disconnecting from '{listener_addr}' - peers provided");
+                self.tcp().disconnect(peer_addr).await;
+            }
+            _ => {
+                trace!("Ignoring an unhandled message from {listener_addr}");
```
vicsn: Maybe still nice to print the message id

vicsn (changes requested):

node/router/src/lib.rs
ljedrz: we don't know the current block height, though

vicsn: Good point, can you document that in the comment?

ljedrz:
@vicsn I think I've addressed all your comments; one of them also helped me squash a bug (wrong identification of validator-mode handshakes, which is now fixed with `validator_mode`), which is nice.

node/src/bootstrap_client/network.rs
```diff
@@ -129,20 +133,25 @@ impl<N: Network> Reading for BootstrapClient<N> {
...
+                let peers = self.get_best_connected_peers(Some(MAX_VALIDATORS_TO_SEND));
+                let mut validators =
+                    IndexMap::with_capacity(current_committee.as_ref().map(|c| c.len()).unwrap_or_default());
+                for validator in peers.into_iter().filter(|peer| {
+                    if let Some(committee) = &current_committee {
+                        // If possible, check the current committee for the prospect Aleo address.
+                        committee.contains(&peer.aleo_addr)
+                    } else {
+                        // Otherwise, filter out known peers connected in validator mode (via Gateway).
+                        self.resolve_to_aleo_addr(peer.listener_addr).is_some()
```
vicsn: Does this mean we are only sending validators which we are **currently** connected to? Given the bootstrap client's rapid disconnecting, that would be very little. We should just return all of the validators we know.

Let me know if I misunderstand.

ljedrz: No, you're spot on: https://github.com/ProvableHQ/snarkOS/pull/3920/commits/c7e37d3c72735b3a01e70ee3947fd4cf807b85f9.

vicsn (dismissed):

node/router/src/heartbeat.rs
```diff
@@ -119,8 +120,6 @@ pub trait Heartbeat<N: Network>: Outbound<N> {
     ///     - Validators are considered higher priority than provers or clients.
```
vicsn: This comment can be removed

vicsn (approved):
LGTM!

PR #3923, opened 2025-10-08
[Perf] Use snarkVM's block caching

kaimast:
This is a companion PR for [snarkVM #2971](https://github.com/ProvableHQ/snarkVM/pull/2971).

It removes the cache in `snarkos-node-bft-ledger-service` and instead uses the native mechanism in snarkVM. This allows cached blocks to be available for components that use the `Ledger` object directly and not through `LedgerService`.

This PR also fixes an issue with the `devnet_ci.sh` script, where the wrong `network_id` might be used for cleanup.


kaimast:
This should be ready to go now too and the change set is very small:  it removes caching at the snarkOS level and adds some additional log messages that I found useful for debugging

vicsn (approved):
LGTM!

vicsn:
Fsiling test is unrelated

Issue #2951, opened 2023-12-27
[Bug] Client nodes often fail to synchronize through CDN


elderhammer:
## 🐛 Bug Report

Client nodes often fail to synchronize through CDN, and can only be resynchronized after cleaning.

## Steps to Reproduce

1. Start node and sync via cdn
2. There is a chance that the following error will occur: "Failed to process block 16703: Failed to finalize transactions - Mismatch in finalize operations for an accepted execute"
3. After the above error occurs, block synchronization through CDN will no longer be possible. There are two ways to deal with this situation: the first is to continue synchronization through the p2p network (the speed is very slow), and the second is to clean the local data and then synchronize again through CDN (the same error may occur again). Either method is a waste of time.

## logs

[...]

PR #3925, opened 2025-10-09
[Fix] Ensure the `developer scan` command works on devnets

kaimast:
**Note:** [snarkVM PR #2951](https://github.com/ProvableHQ/snarkVM/pull/2951) needs to be merged first.

Most recent snarkVM breaks the devnet CI, because `developer scan` does not work with `test_targets` enabled.

```
 Failed to fetch records
    ↳ Invalid block metadata — Genesis block check failed — Invalid coinbase target for genesis block: Wa 536870911 but expected 31.
```

This is due to recent changes ensuring that the genesis block is valid when deserializing it. This PR changes the scan command to use `from_bytes_unchecked` when loading the genesis bytes.
The change is safe because the bytes are hardcoded and, thus, from a trusted source.

### DIFF
--- a/cli/src/commands/developer/scan.rs
+++ b/cli/src/commands/developer/scan.rs
@@ -103,7 +103,8 @@ impl Scan {
         let (start_height, end_height) = self.parse_block_range::<N>(&endpoint)?;
 
         // Fetch the records from the network.
-        let records = Self::fetch_records::<N>(private_key, &view_key, &endpoint, start_height, end_height)?;
+        let records = Self::fetch_records::<N>(private_key, &view_key, &endpoint, start_height, end_height)
+            .with_context(|| "Failed to fetch records")?;
 
         // Output the decrypted records associated with the view key.
         if records.is_empty() {
@@ -184,7 +185,7 @@ impl Scan {
     ) -> Result<Vec<Record<N, Plaintext<N>>>> {
         // Check the bounds of the request.
         if start_height > end_height {
-            bail!("Invalid block range");
+            bail!("Invalid block range. Start height ({start_height}) is not smaller than end height ({end_height}).");
         }
 
         // Derive the x-coordinate of the address corresponding to the given view key.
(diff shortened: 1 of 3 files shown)


cli/src/commands/developer/scan.rs
```diff
@@ -202,9 +203,13 @@ impl Scan {
...
         let genesis_block: Block<N> =
-            ureq::get(&format!("{endpoint}{}/block/0", N::SHORT_NAME)).call()?.into_body().read_json()?;
+            (|| ureq::get(&format!("{endpoint}{}/block/0", N::SHORT_NAME)).call()?.into_body().read_json())()
+                .with_context(|| "Failed to fetch genesis block")?;
+
         // Determine if the endpoint is on a development network.
-        let is_development_network = genesis_block != Block::from_bytes_le(N::genesis_bytes())?;
+        // Use the unchecked version to allow loading the regular genesis block, even if the CLI
+        // has `test_targets` enabled.
+        let is_development_network = genesis_block != Block::from_bytes_le_unchecked(N::genesis_bytes())?;
```
vicsn, line 212: test_network/test_targets are a dev-only feature.

I think the simplest change to fix scanning is to just assume we are on a development network if the test_network feature is enabled. And that also implies that the queried network has it enabled, because we were able to deserialize it's genesis block.

Generally I think it's good if we very tightly couple the concept of "dev" and "test_network", maybe even being the same thing one day.

kaimast: Sure, I can change it to that if you prefer.

vicsn: Yes please, let's aim for minimal and quick PRs that focus on getting the issue at hand solved.

kaimast:
Closing in favor of #3928.

In the long-term, this version would still be better, as the genesis block might change due to other feature flags, but, for now, #3928 is sufficient.

PR #3927, opened 2025-10-10
[Fix] Correctly resolve the peer address used in on_connect

ljedrz:
Fixes the `on_connect` implementations borked in https://github.com/ProvableHQ/snarkOS/pull/3920. Without this, the pings aren't correctly sent to peers upon connecting to them if they were the ones to initiate the connection.

This should fix the broken benchmarks: https://github.com/ProvableHQ/snarkOS/actions/runs/18400602850/job/52428757777

### DIFF
--- a/node/bft/src/gateway.rs
+++ b/node/bft/src/gateway.rs
@@ -1193,10 +1193,13 @@ impl<N: Network> Disconnect for Gateway<N> {
 #[async_trait]
 impl<N: Network> OnConnect for Gateway<N> {
     async fn on_connect(&self, peer_addr: SocketAddr) {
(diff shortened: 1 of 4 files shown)


vicsn (approved):
LGTM, let's await Kai's review too. Should we have TRACE logging in case the resolves or gets don't succeed or would that be too much spam?

ljedrz:
The difference between the connected address and the listener address for connections we've received (as opposed to initiated) is a potential footgun in many places - just adding a `TRACE` here wouldn't help.

What we should have instead is:
1. new and improved network tests that wouldn't omit features like `OnConnect`, which may not have been utilized in the past
2. a strongly typed `Connection/PeerID` that would make such an error impossible; it's a heavy lift, though

PR #3929, opened 2025-10-12
Add option to devnet.sh to run from local path

vicsn:
## Motivation

Found myself using this for the 20th time, so might as well merge upstream.

## Test Plan

Tested the happy path manually once.


ljedrz (approved):
Nice, I've also routinely changed it.

PR #3930, opened 2025-10-12
Track timing of consensus stages

vicsn:
## Motivation

This PR introduces a method to track consensus events for visualization, building on top of @kpandl 's earlier work. My understanding of the image below is that under load, block processing is the bottleneck, not certificate signing.

This "ugly" approach mirrors the way we track `pub static R1CS_HASHES` in snarkVM for certain tests only.

Open TODOs:
- [ ] Add a few logs so we can observe the timing in production

Screenshot 2025-10-12 at 21 31 33

Screenshot 2025-10-13 at 20 23 52


node/bft/src/helpers/timing.rs
```diff
@@ -0,0 +1,317 @@
...
+    SubdagProcessing,
+    CheckNextBlock,
+    AdvanceToNextBlock,
+}
+
+/// Timing event for a specific consensus stage
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct TimingEvent {
+    pub round: u64,
+    pub timestamp: SystemTime,
```
ljedrz, line 53: `SystemTime` is not monotonic; it would be more reliable to instead use the `Instant`, especially for the shorter events.

node/bft/src/helpers/timing.rs
```diff
@@ -0,0 +1,317 @@
...
+    pub event_type: String,
```
ljedrz, line 54:
```suggestion
    pub event_type: &'static str,
```
judging by the current use

node/bft/src/helpers/timing.rs
```diff
@@ -0,0 +1,317 @@
...
+    collections::HashMap,
+    sync::{Arc, RwLock},
+    time::SystemTime,
+};
+
+use once_cell::sync::Lazy;
+use serde::{Deserialize, Serialize};
+
+/// Global storage for round-based event data
+static ROUND_EVENTS: Lazy<Arc<RwLock<HashMap<u64, RoundEvents>>>> = Lazy::new(|| Arc::new(RwLock::new(HashMap::new())));
```
ljedrz:
```suggestion
static ROUND_EVENTS: Lazy<Arc<Mutex<HashMap<u64, RoundEvents>>>> = Lazy::new(|| Arc::new(RwLock::new(HashMap::new())));
```
I'm assuming we're only interested in the final values.

node/bft/src/helpers/timing.rs
```diff
@@ -0,0 +1,317 @@
...
+
+/// Global storage for subdag-based timing data (using lowest/highest rounds from subdag)
+static SUBDAG_TIMINGS: Lazy<Arc<RwLock<HashMap<(u64, u64), SubdagTimings>>>> =
```
ljedrz:
```suggestion
static SUBDAG_TIMINGS: Lazy<Arc<Mutex<HashMap<(u64, u64), SubdagTimings>>>> =
```
ditto: as long as we're not going to be actively inspecting these values with multiple readers, a `Mutex` will suffice

node/bft/src/helpers/timing.rs
```diff
@@ -0,0 +1,317 @@
...
+                if let Some((start, _)) = self.check_next_block {
+                    self.check_next_block = Some((start, Some(now)));
+                }
+            }
+            SubdagStage::AdvanceToNextBlock => {
+                if let Some((start, _)) = self.advance_to_next_block {
+                    self.advance_to_next_block = Some((start, Some(now)));
+                }
+            }
+        }
```
ljedrz, line 170: should we notify the user if `None` was found in any of those (indicating that a `start_stage` may have been omitted by accident)?

node/bft/src/helpers/timing.rs
```diff
@@ -0,0 +1,317 @@
...
+pub fn export_to_json(file_path: &str) -> Result<(), Box<dyn std::error::Error>> {
+    let snapshot = get_timing_snapshot();
+    let json = serde_json::to_string_pretty(&snapshot)?;
+    std::fs::write(file_path, json)?;
+    Ok(())
+}
+
+/// Clean up old timing entries to prevent memory growth
+/// Keeps only the most recent `keep_count` entries for each type
+pub fn cleanup_old_entries(keep_count: usize) {
```
ljedrz: Considering this use case, it would be more performant to use a `VecDeque` and use the round as the index (offset by the 1st round in the collection, which would be stored separately), but this should be good enough for ad-hoc dev applications.

ljedrz:
> dumping all of the data into a global lazy static, and only enabling this in dev/test_network mode. Would a pub static also be fine? Or something else?

For dev/test purposes it's perfectly fine; it is only in production where it's an anti-pattern, and should instead be incorporated into the workflow. The alternative would be to store it in the related objects, which would also work, though could be a bit trickier to implement.

> exporting the data to a JSON file, also only in dev/test_network mode.

As long as we don't expect very long measurements/outputs, this is fine.

Oh, and naturally, this still needs to be feature-gated.

node/bft/src/helpers/timing.rs
vicsn: When trying to replace it, compiler complains that it can't derive Serialize/Deserialize on Instant.

Is this an issue in practice? The created graph seems quite faithful and we don't care about even a millisecond difference at these scales.

node/bft/src/helpers/timing.rs
vicsn: Leaving as is for now as a bandaid for the fact that we inefficiently read out on every block.

node/bft/src/helpers/timing.rs
vicsn: Removed the function f462afde7

node/bft/src/helpers/timing.rs
ljedrz: hmm, as long as we don't perform uncheck subtraction between such values and we don't worry about tiny events, it should be fine

vicsn (commented):
TODO: also measure a stage for `commit_leader_certificate` onwards

node/bft/src/helpers/timing.rs
vicsn: Can't auto-deserialize into a static reference:

```
error: lifetime may not live long enough
  --> node/bft/src/helpers/timing.rs:62:5
   |
59 | #[derive(Debug, Clone, Serialize, Deserialize)]
   |                                   ----------- lifetime `'de` defined here
...
62 |     pub proposal_seen: Vec<TimingEvent>,
   |     ^^^ requires that `'de` must outlive `'static`
```

node/bft/src/helpers/timing.rs
vicsn: Good idea: d9492d025

kaimast:
Would it not be better to add a separate feature for this? Or do you think the performance overhead for test networks that do not require timing information is minimal?

vicsn:
> Would it not be better to add a separate feature for this? Or do you think the performance overhead for test networks that do not require timing information is minimal?

Agreed, made a new `test_consensus_tracking` feature. Specifically for long-running networks, these hacky JSON files might get huge.

node/bft/Cargo.toml
```diff
@@ -44,6 +44,7 @@ test = [
   "snarkos-node-bft-ledger-service/test",
   "snarkos-node-bft-storage-service/test"
 ]
+test_consensus_tracking = [ ]
```
ljedrz, line 48: nit: this should ideally include the new dependencies

node/consensus/src/lib.rs
```diff
@@ -524,12 +524,73 @@ impl<N: Network> Consensus<N> {
         #[cfg(feature = "metrics")]
         let current_block_timestamp = self.ledger.latest_block().header().metadata().timestamp();
 
+        // Extract lowest and highest rounds from the subdag
+        let rounds: Vec<u64> = subdag.keys().copied().collect();
+        let _lowest_round = rounds.iter().min().copied().unwrap_or(0);
+        let _highest_round = rounds.iter().max().copied().unwrap_or(0);
```
```suggestion
        let _lowest_round = rounds.first_key_value().map(|(k, _)| *k).unwrap_or(0);
        let _highest_round = rounds.last_key_value().map(|(k, _)| *k).unwrap_or(0);
```

node/bft/src/helpers/timing.rs
```diff
@@ -0,0 +1,300 @@
...
+    pub check_next_block: Option<(SystemTime, Option<SystemTime>)>,
+    pub advance_to_next_block: Option<(SystemTime, Option<SystemTime>)>,
+}
+
+/// Combined timing data for export
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct TimingSnapshot {
+    pub timestamp: SystemTime,
+    pub round_events: HashMap<String, RoundEvents>,
+    pub subdag_timings: HashMap<String, SubdagTimings>,
```
ljedrz, line 82: not sure it's worth it to lose the ordering just to have faster queries only in tests

node/bft/src/helpers/timing.rs
```diff
@@ -0,0 +1,300 @@
...
+#[cfg(feature = "locktick")]
+use locktick::parking_lot::RwLock;
+#[cfg(not(feature = "locktick"))]
+use parking_lot::RwLock;
+
+use once_cell::sync::Lazy;
+use serde::{Deserialize, Serialize};
+
+/// Global storage for round-based event data
+static ROUND_EVENTS: Lazy<Arc<RwLock<HashMap<u64, RoundEvents>>>> = Lazy::new(|| Arc::new(RwLock::new(HashMap::new())));
```
ljedrz:
```suggestion
static ROUND_EVENTS: Lazy<Arc<RwLock<HashMap<u64, RoundEvents>>>> = Lazy::new(|| Default::default());
```
nit

node/bft/src/helpers/timing.rs
```diff
@@ -0,0 +1,300 @@
...
+
+/// Global storage for subdag-based timing data (using lowest/highest rounds from subdag)
+static SUBDAG_TIMINGS: Lazy<Arc<RwLock<HashMap<(u64, u64), SubdagTimings>>>> =
+    Lazy::new(|| Arc::new(RwLock::new(HashMap::new())));
```
ljedrz:
```suggestion
    Lazy::new(|| Default::default());
```
same (as long as the complier doesn't find a reason to complain)

ljedrz (commented):
Left a few comments.

node/bft/src/helpers/timing.rs
kaimast: You might also be able to use [`LazyLock`](https://doc.rust-lang.org/std/sync/struct.LazyLock.html) in `std` instead of relying on `OnceCell`.

node/consensus/src/lib.rs
```diff
@@ -524,12 +524,73 @@ impl<N: Network> Consensus<N> {
...
+        snarkos_node_bft::helpers::end_subdag_stage(
+            _lowest_round,
+            _highest_round,
+            snarkos_node_bft::helpers::SubdagStage::AdvanceToNextBlock,
+        );
+
+        // Export timing data to JSON after block generation
+        #[cfg(feature = "test_consensus_tracking")]
+        {
+            let dev_index = self.bft().primary().gateway().dev().unwrap_or_default();
```
kaimast: Should this be `0` when not in dev mode? We could alternatively append nothing or `_nodev`.

Cargo.toml
```diff
@@ -281,6 +281,7 @@ serial = [ "snarkos-cli/serial" ]
 test_targets = [ "snarkos-cli/test_targets" ]
 test_consensus_heights = [ "snarkos-cli/test_consensus_heights" ]
 test_network = [ "snarkos-cli/test_network" ]
+test_consensus_tracking = [ "snarkos-node/test_consensus_tracking" ]
```
kaimast, line 307: It would be good to start documenting these new features. Either here or in the README similar to what I did in [`this PR`](https://github.com/ProvableHQ/snarkVM/pull/2887/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R48).

node/consensus/src/lib.rs
```diff
@@ -524,12 +524,73 @@ impl<N: Network> Consensus<N> {
         #[cfg(feature = "metrics")]
         let current_block_timestamp = self.ledger.latest_block().header().metadata().timestamp();
 
+        // Extract lowest and highest rounds from the subdag
+        let rounds: Vec<u64> = subdag.keys().copied().collect();
+        let _lowest_round = rounds.iter().min().copied().unwrap_or(0);
+        let _highest_round = rounds.iter().max().copied().unwrap_or(0);
+
+        #[cfg(feature = "test_consensus_tracking")]
```
kaimast, line 561: It might make sense to add a macro for the `start_subdag_stage`/`end_subdag_stage` pair. That macro could also be no-op if the feature flag isn't set, so we do not have to have this many feature gates in the code.

kaimast: (I can push a commit for with this change, if it is too much of a hassle for you)

node/bft/Cargo.toml
vicsn: Hmm that's a very tight coupling, I'll leave it out for now

node/bft/src/helpers/timing.rs
vicsn: Irrelevant at this stage

node/bft/src/helpers/timing.rs
vicsn: Why is that preferred?

node/consensus/src/lib.rs
vicsn: Agreed it was really ugly, made it more readable: 5da19d5b2

I don't like the marginal improvement of a new macro, but maybe once your other tracing logic works and we can decorate functions with it, it could replace this current approach entirely.

Cargo.toml
vicsn: ae0388458

node/consensus/src/lib.rs
vicsn: Good catch: 5da19d5b2

node/bft/src/helpers/timing.rs
vicsn: 0f89bbcd8

node/bft/src/helpers/timing.rs
vicsn: 0f89bbcd8

node/consensus/src/lib.rs
vicsn: 0f89bbcd8

PR #3936, opened 2025-10-14
[Feature] Do not sync when peer consensus version differs

### LINKED ISSUE #3822: [Bug] Prevent network from advancing on fork
## 🚀 Feature

Currently, if a node does not upgrade in time for a new consensus version, they may continue to produce blocks, and other clients may continue to sync from these nodes, until the point where they end up switching syncing to the canonical chain. Ideally, both honest and dishonest honest don't advance anymore as soon as the mismatch is observed.

We should decide which of the following options are feasible, and in which order we'll implement them.

# TL;DR starting goal
Individual BlockSync Block responses should note the `ConsensusVersion` of `blockresponse.last().height`, so the requester can disconnect if it doesn't match their expectation of what the ConsensusVersion should be at that height.

# Theoretical overview

## Consensus and Certificate Sync
1. Validators sign off on each other's advertised consensus version and/or recent block (height hash) in their certificate.
2. Validators optimistically announce their current consensus version and/or recent block (height, hash) in their handshake or a P2P message. Note that "rolling message versions" like we implement for clients might be tricky due to intricacies of how certificates work.

## Block Sync
1. Honest approach: blocks or sync messages announce the consensus version, so the requester can check this
2. Dishonest approach: we actually fetch and compare blocks from multiple nodes, which is a longer-term effort

kaimast:
This PR fixes #3822. It adds a `latest_consensus_version` to a block response to prevent nodes that did not upgrade from block advancement.

If the peers use different versions of snarkOS, it will either fail to deserialize the block response, or abort when it detects that the version do not match.

**Note:** This requires upgrading the message version, as the format for `BlockResponse` changed.


kaimast:
Strangely, I had to increase the machine size for another CI job. Some recent change caused a lot of our builds to take longer or use up more memory.

node/bft/events/src/block_response.rs
```diff
@@ -40,16 +46,18 @@ impl<N: Network> EventTrait for BlockResponse<N> {
 impl<N: Network> ToBytes for BlockResponse<N> {
     fn write_le<W: Write>(&self, mut writer: W) -> IoResult<()> {
         self.request.write_le(&mut writer)?;
-        self.blocks.write_le(&mut writer)
+        self.blocks.write_le(&mut writer)?;
+        self.latest_consensus_version.write_le(&mut writer)
```
vicsn: As the message is not versioned, upon upgrade, other nodes immediately won't be able to parse this message and sync from the upgraded node anymore.

As validatord and their core clients always upgrade first, this effectively means the rest of the network will halt until they upgrade.

Can/should we make this field Optional, and only write it once we are at a new ConsensusVersion? Note that if we do this, we also need a backwards compatible way to determine the message version when reading. Similar to the recent PR by Lukasz adjusting PeerResponse serialization, perhaps this can be handled by assuming an empty request (first bytes of this payload) is an implicit version indicator?

node/router/messages/src/block_response.rs
```diff
@@ -28,6 +28,9 @@ pub struct BlockResponse<N: Network> {
     pub request: BlockRequest,
     /// The blocks.
     pub blocks: Data<DataBlocks<N>>,
+    /// The consensus version at the height of the *last* block in this response.
+    /// This enables detecting if the current node, or the peer, missed an upgrade.
+    pub latest_consensus_version: ConsensusVersion,
```
vicsn: Same comment as above

vicsn (changes requested):
Core logic LGTM! Just a note on backwards compatibility

node/bft/events/src/block_response.rs
ljedrz: For reference, this is how it was done with the `PeerResponse` extension: https://github.com/ProvableHQ/snarkOS/pull/3860.

node/bft/src/helpers/channels.rs
```diff
@@ -253,22 +254,30 @@ impl<N: Network> SyncSender<N> {
     }
 
     /// Sends the request to insert a new block response.
-    pub async fn insert_block_response(&self, peer_ip: SocketAddr, blocks: Vec<Block<N>>) -> Result<()> {
+    pub async fn insert_block_response(
+        &self,
+        peer_ip: SocketAddr,
+        blocks: Vec<Block<N>>,
+        latest_consensus_version: ConsensusVersion,
```
ljedrz: this triplet appears several times in this diff; perhaps it would make sense to introduce at least a type alias for it, so any potential future extensions are more compact?

vicsn: Also less flexibility, don't think it'll be worth it in this case

ljedrz (changes requested):
Same point as @vicsn made; we should be able to make it backwards compatible, and only make the new field a hard requirement after a while.

kaimast:
Thanks for the feedback! I changed the PR to now interpret the `start_height` of the block request as a version number. For old messages, it is never zero, as nodes do not request the genesis block from other nodes.
For newer messages, it is set to zero for now. In the future, we can use the first byte(s) to encode an actual version.

To make this change simpler, `snarkos_node_router_messages` now re-export the `BlockResponse` and `BlockRequest` types from `snarkos_node_bft_events`, instead of re-implementing them.

Finally, I needed to make changes to the prop tests for these two message types because they didn't actually generate valid block requests or responses. For example, it picked two random numbers as start and end height, so that the ranges could be invalid. The number of blocks in the response also was not guaranteed to be equal to the size of the request range.

node/bft/events/src/block_response.rs
```diff
@@ -15,16 +15,30 @@
...
+    prelude::{FromBytes, ToBytes},
+    utilities::io_error,
+};
+
+use std::borrow::Cow;
+
+#[derive(Clone, Debug, PartialEq, Eq)]
 pub struct BlockResponse<N: Network> {
+    /// The message version.
+    version: u8,
```
ljedrz: nit: the usual convention is to not persist the version - only to serialize it and check it upon deserialization, then discard

node/router/messages/src/block_request.rs
```diff
@@ -15,72 +15,17 @@
 
 use super::*;
 
-use snarkvm::prelude::{FromBytes, ToBytes};
+use snarkos_node_bft_events::EventTrait;
```
ljedrz, line 18: nit: I wonder if we shouldn't reexport values from `node/router` to `bft`, as the latter already uses the `router`; granted, in this case there is no circular dependency, but it might be a good idea to have some convention, at least until we can isolate some common network-related crate that both `router` and `bft` could depend on without it depending on either of them

node/router/messages/src/block_request.rs
```diff
@@ -15,72 +15,17 @@
...
-        any::<(u32, u32)>().prop_map(|(start_height, end_height)| BlockRequest { start_height, end_height }).boxed()
-    }
-
-    #[proptest]
-    fn block_request_roundtrip(#[strategy(any_block_request())] block_request: BlockRequest) {
-        let mut bytes = BytesMut::default().writer();
-        block_request.write_le(&mut bytes).unwrap();
-        let decoded = BlockRequest::read_le(&mut bytes.into_inner().reader()).unwrap();
-        assert_eq![decoded, block_request];
+        EventTrait::name(self)
```
ljedrz, line 29: speaking of, this trait is literally the same as `MessageTrait` - there's no good reason to have both of them; this is beyond the scope of this PR, of course

ljedrz (dismissed):
Left some nits, but nothing blocking; LGTM 👌.

node/bft/events/src/block_response.rs
```diff
@@ -15,22 +15,30 @@
...
     pub blocks: Data<DataBlocks<N>>,
     /// The consensus version at the height of the *last* block in this response.
     /// This enables detecting if the current node, or the peer, missed an upgrade.
-    pub latest_consensus_version: ConsensusVersion,
+    latest_consensus_version: ConsensusVersion,
 }
 
 impl<N: Network> EventTrait for BlockResponse<N> {
-    /// Returns the event name.
+    /// Returns the message name.
```
vicsn: We categorize gateway messages as "Events" and router messages as "Messages". Why change this comment?

node/router/src/inbound.rs
```diff
@@ -125,17 +124,21 @@ pub trait Inbound<N: Network>: Reading + Outbound<N> {
                     false => bail!("Peer '{peer_ip}' sent an invalid block request"),
                 }
             }
-            Message::BlockResponse(BlockResponse { request, blocks, latest_consensus_version }) => {
+            Message::BlockResponse(response) => {
```
vicsn: In the Router case, can we disconnect from the peer if they're sending us a block response with an invalid ConsensusVersion? I can imagine a node connected to some outdated peers may experience a significant continued slowdown in syncing otherwise

node/router/messages/src/block_request.rs
vicsn: Let's cross that bridge when we get there

vicsn (changes requested):
Thx! We're almsot there!

node/bft/events/src/block_response.rs
vicsn: The backwards-compatible deserialization looks good!

We should only set `latest_consensus_version` from a new ConsensusVersion onwards. But I realize for that to happen, we first need to define a new ConsensusVersion... Made a snarkVM PR here: https://github.com/ProvableHQ/snarkVM/pull/2963

This was not important for the PeerResponse messages, but for block sync we should aim for 0 downtime.

node/bft/events/src/block_response.rs
kaimast: It was a copy-paste error I think...

node/router/src/inbound.rs
kaimast: It will fail in `Inbound::block_response` and should then disconnect the nod for protocol violation.

Considering the consensus heights are hardcoded, I can also move the check to here, so it fails earlier.

vicsn: If you're sure it will disconnect the node, then that sounds already fine, feel free to resolve this comment.

node/bft/events/src/block_response.rs
vicsn: Cc @kaimast

node/bft/events/src/block_response.rs
kaimast: Removed.

node/bft/events/src/block_response.rs
kaimast: I disabled the check for version numbers < 12. I can't test it until the V12 changes are merged, though.

node/src/client/router.rs
```diff
@@ -179,24 +180,40 @@ impl<N: Network, C: ConsensusStorage<N>> Inbound<N> for Client<N, C> {
     fn block_request(&self, peer_ip: SocketAddr, message: BlockRequest) -> bool {
         let BlockRequest { start_height, end_height } = &message;
 
+        // Get the latest consensus version, i.e., the one for the last block's height.
+        let latest_consensus_version = match N::CONSENSUS_VERSION(end_height - 1) {
```
raychu86: Nit: Not sure if the `end_height` can ever be 0, but this would cause an underflow. A `saturating_sub` would be safer.

```suggestion
        let latest_consensus_version = match N::CONSENSUS_VERSION(end_height.saturating_sub(1)) {
```

node/src/validator/router.rs
```diff
@@ -177,21 +178,36 @@ impl<N: Network, C: ConsensusStorage<N>> Inbound<N> for Validator<N, C> {
     fn block_request(&self, peer_ip: SocketAddr, message: BlockRequest) -> bool {
         let BlockRequest { start_height, end_height } = &message;
 
+        // Get the latest consensus version, i.e., the one for the last block's height.
+        let latest_consensus_version = match N::CONSENSUS_VERSION(end_height - 1) {
```
raychu86: Ditto above
```suggestion
        let latest_consensus_version = match N::CONSENSUS_VERSION(end_height.saturating_sub(1)) {
```

node/sync/src/block_sync.rs
```diff
@@ -447,7 +452,30 @@ impl<N: Network> BlockSync<N> {
...
+    pub fn insert_block_responses(
+        &self,
+        peer_ip: SocketAddr,
+        blocks: Vec<Block<N>>,
+        latest_consensus_version: Option<ConsensusVersion>,
+    ) -> Result<()> {
+        // Perform consensus version check, if possible.
+        // This check is only enabled after nodes have reached V12.
+        if let Some(latest_consensus_version) = latest_consensus_version
+            && latest_consensus_version > ConsensusVersion::V11
```
raychu86: Can a malicious node send a `None` value here any bypass this check?

node/bft/events/src/block_response.rs
```diff
@@ -37,25 +49,62 @@ impl<N: Network> EventTrait for BlockResponse<N> {
     }
 }
 
+impl<N: Network> BlockResponse<N> {
+    /// The version number that we use to serialize block responses,
+    /// and the highest known message valid version number.
+    const CURRENT_VERSION_NUMBER: u8 = 2;
+    /// The minimum supported message version number.
+    const MIN_VERSION_NUMBER: u8 = 1;
```
raychu86: When do these two values get updated? Is it every time we have a new consensus version or message version?

If so, we should probably have a test of some sort to enforce we are updating all the versions in tandem for each release.

node/src/client/router.rs
kaimast: This PR also adds a check when deserializing `BlockRequest` that `start_height > 0` and `end_height > start_height`. This is to ensure that nodes can never request the genesis block or an empty range.

Should I still add the check?

node/sync/src/block_sync.rs
kaimast: The idea here is that nodes can still sync from nodes that have not upgraded yet, and this change does not require nodes to update synchronously. Eventually we can remove this check.

node/bft/events/src/block_response.rs
kaimast: My hope is that we can bump the minimum to `2`  and drop support for responses that don't include the consensus version in the near future. Aside from that, changes to it should be fairly infrequent.

node/src/client/router.rs
raychu86: Would be nice, but it's a nit

node/sync/src/block_sync.rs
raychu86: That makes sense; I was mainly asking if the check should enforce a `Some` value if the blocks being processed are `> ConsensusVersion::V11`, but that may not be needed. I think the check is fine as is.

node/bft/events/src/block_response.rs
raychu86: Is there potentially a cleaner way to do this with the block heights and use the existing `ConsensusVersion`
s rather than introducing a new `BlockResponse::Version` we need to potentially track?

Otherwise, can this be removed entirely once we believe sufficient nodes have upgraded, and then we enforce that the `latest_consensus_version` is always included?

node/bft/events/src/block_response.rs
vicsn: V12 is merged now

node/bft/events/src/block_response.rs
kaimast: Done in [20c3dd8](https://github.com/ProvableHQ/snarkOS/pull/3936/commits/20c3dd854d787b58b2fd1df97473e910453878aa).

node/src/client/router.rs
kaimast: Done in [7c4b649](https://github.com/ProvableHQ/snarkOS/pull/3936/commits/7c4b6498bacb728a5bee29d82d54ad4214967e8f).

node/bft/events/src/block_response.rs
vicsn: Do we also have to prevent *writing* the ConsensusVersion in the BlockResponse until we hit V12?

kaimast:
I tested this finally, and you will get an output like the following starting at consensus V12 for nodes that have not upgraded:
```
2025-10-24T00:19:22.875901Z  WARN snarkvm_utilities::errors: Failed to insert block response — The peer did not send a consensus version
2025-10-24T00:19:22.875950Z  WARN snarkos_node::client::router: Failed to process inbound message from '127.0.0.1:4130' - Peer '127.0.0.1:4130' sent an invalid block response
```

Before merging/approving: take a look at the most recent change in [ddb841b](https://github.com/ProvableHQ/snarkOS/pull/3936/commits/ddb841bb46cff65bdc1f2f4a379c17bd046a8a9f), which ensures messages are always rejected if the *local* consensus version would be greater than 12.
I believe this is the behavior we want but please double check.

vicsn:
@kaimast thank you, this is the behaviour we want:
- before V12, old and new nodes can sync from each other
- after V12 old and new nodes can't sync from each other anymore due to the serialization mismatch

Please confirm

kaimast:
> @kaimast thank you, this is the behaviour we want:
> 
>     * before V12, old and new nodes can sync from each other
> 
>     * after V12 old and new nodes can't sync from each other anymore due to the serialization mismatch
> 
> 
> Please confirm

Yes. That's the behavior.

vicsn (dismissed):

node/bft/events/src/block_response.rs
```diff
@@ -37,25 +49,53 @@ impl<N: Network> EventTrait for BlockResponse<N> {
...
+        };
+
+        // Currently, we simply write four zero bytes as the version number,
+        // because we know a valid request start height is always non-zero.
+        // In the future we can encode the real version here.
+        0u32.write_le(&mut writer)?;
         self.request.write_le(&mut writer)?;
-        self.blocks.write_le(&mut writer)
+        self.blocks.write_le(&mut writer)?;
+        latest_consensus_version.write_le(&mut writer)
```
vicsn: Only serialize if it's not set

vicsn (approved):

Issue #3909, opened 2025-10-02
Perform a new run with `tokio console`


ljedrz:
It's been a while since we've done this, so it would be a good idea to do a fresh run just to make sure that we're still not blocking the executor or spawning too many tasks.

`tokio-console` is still under heavy development, so it's something that would still be best done manually; once it matures, and more features currently requiring `tokio_unstable` become stable, we might be able to make it into a CI run.

vicsn:
Some graphs on reference hardware (m7i.16xlarge) when submitting 6000 executions (alongside some solutions and deployments too)

Running this workload repeatedly is a tad bit expensive, so I hope we can take the appropriate shortcuts.

<img width="792" height="925" alt="Image" src="https://github.com/user-attachments/assets/826ac500-8269-4b19-b941-9b2d8c989a83" />
<img width="842" height="305" alt="Image" src="https://github.com/user-attachments/assets/ad608c26-0156-4ae2-8481-269e960794a4" />
<img width="849" height="309" alt="Image" src="https://github.com/user-attachments/assets/41efa2dd-33d3-4a48-8281-06eb40fd754c" />
<img width="853" height="303" alt="Image" src="https://github.com/user-attachments/assets/062a2ab7-a048-4501-b96b-1c1b0e325142" />

vicsn:
@ljedrz shall we close this issue?

ljedrz:
Yes, I think we're good for the time being 👍.

PR #3939, opened 2025-10-15
[Feat] Introduce a tokio_console feature

ljedrz:
This PR adds a `tokio_console` feature, which facilitates the use of `tokio-console` with the node. As with other features requiring a bit of extra setup, the necessary conditions are checked in `build.rs`.

In order to run a node with `tokio-console` support, it just needs to be **built** with `RUSTFLAGS="--cfg tokio_unstable"`. No other setup is required - the node can be run normally afterwards, and it will keep displaying its usual logs (`tokio-console` is set up as an additional subscriber). Finally, all that's left is to either run just `tokio-console` locally, or `tokio-console http://<IP>:6669` if running remotely.

Note: it is expected that only a single such node is running on a given machine - `console-subscriber` uses localhost and the default port (`6669`) to supply the relevant tracing data to be used by `tokio-console`.

CC https://github.com/ProvableHQ/snarkOS/issues/3909.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -259,11 +259,38 @@ checksum = "c08606f8c3cbf4ce6ec8e28fb0014a2c086708fe954eaa885384a6165172e7e8"
 
 [[package]]
 name = "axum"
+version = "0.7.9"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


vicsn (approved):
Exciting

PR #3940, opened 2025-10-15
[Tweak] Place per-connection tasks in the heap

ljedrz:
While `tokio` automatically places tasks with a future larger than 16KiB (in `release` mode, 2KiB in `debug`) on the heap via boxing, our tasks are typically smaller than that, but they can easily reach ~10KiB. Despite having relatively large stacks (8MiB) for both our `rayon` and `tokio` threads, they could still technically blow up under sufficient stress, especially if we scale up the number of connections in the future (for example for the `BootstrapperClient`s).

In order to avoid this, we can manually box the tasks that increase in numbers when the node scales up its operations, and the most obvious candidates for this treatment are long-running tasks that are spawned per each connection; in our case these are 2 `Reading` tasks and 1 `Writing` task, which this PR boxes unconditionally.

This was discovered with the aid of `tokio-console`.

### DIFF
--- a/node/tcp/src/protocols/reading.rs
+++ b/node/tcp/src/protocols/reading.rs
@@ -138,7 +138,7 @@ impl<R: Reading> ReadingInternal for R {
 
         // the task for processing parsed messages
         let self_clone = self.clone();
-        let inbound_processing_task = tokio::spawn(async move {
+        let inbound_processing_task = tokio::spawn(Box::pin(async move {
             let node = self_clone.tcp();
             trace!(parent: node.span(), "spawned a task for processing messages from {addr}");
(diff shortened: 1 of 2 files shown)


vicsn (approved):
Do I understand correctly that this may have a minor theoretical performance hit - but safety and stability first so it's ok?

> especially if we scale up the number of connections in the future (for example for the BootstrapperClients).

FWIW I don't see us needing to optimize anything in service of the bootstrap clients :)

ljedrz:
> Do I understand correctly that this may have a minor theoretical performance hit - but safety and stability first so it's ok?

A very tiny hit caused by extra 3 allocations per connection - completely negligible considering network delays and how often (full) connections are created. And yes, it's an extra safeguard.

Issue #3926, opened 2025-10-10
[Feature] Ensure bootstrap clients return validator routers only to validators


vicsn:
## 🚀 Feature

Validator routers should be able to find each other via the bootstrap clients without manual specification. However, bootstrap clients should not advertise validator peers to the rest of the network. One approach could be as follows:
- bootstrap clients routers check and cache whether their peer is a validator or not
- when receiving a PeerRequest from a validator, bootstrap clients routers only return validator peers
- when receiving a PeerRequest from a non-validator, bootstrap client routers should only return non-validator peers.
- validator routers should store just a single router peer per validator address in the current committee.

Security discusion informing the above design: validators should now both have their gateway and router port open to the public. However, they should not be able to receive random external peers from other validators. One way to set this up, is to ensure validators still do **not** specify --allow-external-peers, but regardless only fetch and accept peers from the trusted bootstrap peers. Even then there is still an attack to protect against: a malicious validator could re-use their address to advertise thousands of validator routers to the bootstrap peers, and thereby fill up all of the honest validator peer pools.

Engineering point of concern: the committee fetching should be fast enough not to break any handshake time-outs.

PR #3942, opened 2025-10-16
[Feat] Tighten BootstrapClient validator sharing

### LINKED ISSUE #3945: [Feature] Enable testing of get_or_update_committee in dev mode
## 🚀 Feature

Currently we skip bootstrap node P2P committee checks in dev mode because we can't reach any explorer.

But in dev mode, we could let `get_or_update_committee` query a validator IP address, set in an environment variable, similar to how BOOTSTRAP_PEERS is set during testing.

### LINKED ISSUE #3949: [Proposal] Make it easier to detect how a validator is connected
The distinction between the `Router` and `Gateway` is particularly important for the `BootstrapClient`, as it only has a single `Resolver`, can accommodate connections from both these sources, and needs to provide different addresses to validators based on their connection mode.

Right now the way it is discovered is somewhat roundabout, and bug-prone, but I believe we can make it a lot better without heavy lifting.

ljedrz:
This PR addresses https://github.com/ProvableHQ/snarkOS/issues/3926, though I'm not 100% sure that's the end of it.

First, it verifies whether the connecting validator belongs to the committee at the handshake stage, potentially rejecting both `Router` (if the peer's node type is the validator) and `Gateway` connections based on the Aleo address of the peer.

Second, it enhances the peer list filtering:
- non-validator peers receive no validator addresses at all
- validator peers connected in `Router` mode may receive `Router` validator addresses (among other peer addresses)
- validator peers connected in `Gateway` mode receive only `Gateway` validator addresses

Closes https://github.com/ProvableHQ/snarkOS/issues/3945.
Closes https://github.com/ProvableHQ/snarkOS/issues/3949.

### DIFF
--- a/node/src/bootstrap_client/handshake.rs
+++ b/node/src/bootstrap_client/handshake.rs
@@ -273,12 +273,38 @@ impl<N: Network> BootstrapClient<N> {
                     send_msg!(msg, framed, peer_addr)?;
                     return Ok(false);
                 }
+
+                // Reject validators that aren't members of the committee.
+                if msg.node_type == NodeType::Validator {
+                    // This check is skipped in dev mode (the None scenario).
+                    if let Some(current_committee) =
+                        self.get_or_update_committee().await.map_err(|_| error("Couldn't load the committee"))?
+                    {
+                        if !current_committee.contains(&msg.address) {
+                            let msg = Message::Disconnect::<N>(messages::DisconnectReason::ProtocolViolation.into());
+                            send_msg!(msg, framed, peer_addr)?;
+                            return Ok(false);
+                        }
+                    }
+                }
             }
             MessageOrEvent::Event(Event::ChallengeRequest(msg)) => {
                 if msg.version < Event::<N>::VERSION {
                     let msg = Event::Disconnect::<N>(events::DisconnectReason::OutdatedClientVersion.into());
                     send_msg!(msg, framed, peer_addr)?;
                     return Ok(false);
+                }
+
+                // Reject validators that aren't members of the committee.
+                // This check is skipped in dev mode (the None scenario).
+                if let Some(current_committee) =
+                    self.get_or_update_committee().await.map_err(|_| error("Couldn't load the committee"))?
+                {
+                    if !current_committee.contains(&msg.address) {
+                        let msg = Event::Disconnect::<N>(events::DisconnectReason::ProtocolViolation.into());
+                        send_msg!(msg, framed, peer_addr)?;
+                        return Ok(false);
+                    }
                 }
             }
             _ => unreachable!(),
--- a/node/src/bootstrap_client/handshake.rs
+++ b/node/src/bootstrap_client/handshake.rs
@@ -18,6 +18,7 @@ use crate::{
     bft::events::{self, Event},
     bootstrap_client::{codec::BootstrapClientCodec, network::MessageOrEvent},
     router::{
+        ConnectionMode,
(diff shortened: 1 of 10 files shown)


vicsn (changes requested):
Looks great!

To improve testing of this feature, could you also implement: https://github.com/ProvableHQ/snarkOS/issues/3945 ?

ljedrz:
I added 3 new commits:
- https://github.com/ProvableHQ/snarkOS/pull/3942/commits/7d7e8a56c002a30abcfa6f3a63bb29373d8458b6 - this addresses https://github.com/ProvableHQ/snarkOS/issues/3949, as the way the connection mode was determined was complicated, buggy, and unnecessarily affected non-`BootstrapClient` nodes
- https://github.com/ProvableHQ/snarkOS/pull/3942/commits/f6818a8fed351aac2c4589b6775f8eb89c86b32e - a tiny boostrapper-related fix that I found by accident
- https://github.com/ProvableHQ/snarkOS/pull/3942/commits/54a735300dec81b0ec273f162789b8cbebe72b35 - addresses https://github.com/ProvableHQ/snarkOS/issues/3945

node/src/bootstrap_client/handshake.rs
```diff
@@ -273,13 +283,39 @@ impl<N: Network> BootstrapClient<N> {
...
             }
             MessageOrEvent::Event(Event::ChallengeRequest(msg)) => {
                 if msg.version < Event::<N>::VERSION {
                     let msg = Event::Disconnect::<N>(events::DisconnectReason::OutdatedClientVersion.into());
                     send_msg!(msg, framed, peer_addr)?;
                     return Ok(false);
                 }
+
+                // Reject validators that aren't members of the committee.
+                // This check is skipped in dev mode (the None scenario).
```
vicsn: This comment appears twice

node/src/bootstrap_client/mod.rs
```diff
@@ -165,9 +169,19 @@ impl<N: Network> BootstrapClient<N> {
     /// Returns the current validator committee or updates it from the explorer, if
     /// we are capable of obtaining it from the network.
     pub async fn get_or_update_committee(&self) -> anyhow::Result<Option<HashSet<Address<N>>>> {
-        // The current committee can't be looked up in dev mode.
-        if self.is_dev() {
-            return Ok(None);
+        // Development testing may include a list of committee Aleo addresses loaded from the environment.
+        if cfg!(feature = "test") || self.is_dev() {
+            match std::env::var("TEST_COMMITTEE_ADDRS") {
```
vicsn, line 174: @meddle0x53 can you adjust the stress-tests to set the validator addresses as an env variable - only needed for the bootstrap client.

vicsn (changes requested):

node/src/bootstrap_client/handshake.rs
ljedrz: One is for validators is `Router` mode, the other is for those in `Gateway` mode, and someone doing debugging might only be interested in one of the modes, and only read the branch of interest to them. Should I remove both instances, or add clarity by stating that only validators are subject to the 2nd branch?

vicsn: I should be more specific. "This check is skipped in dev mode (the None scenario)." doesn't seem accurate because we are not skipping the check in dev mode anymore, so my understanding is that this line can be removed.

vicsn (approved):

PR #3946, opened 2025-10-16
[Logs] Print error when not connected to a quorum of validators

### LINKED ISSUE #3885: [Feature] Log lack of connected validators as error
## 🚀 Feature

A testnet validator had an outage because it wasn't connected to enough validators.

It logged it wasn't connected to some validators. We should log an error if we're not connected to a quorum majority of the validators.

kaimast:
With this change, snarkos validators log an error if they are not connected to a quorum of staked validators. Fixes #3885.


vicsn (approved):
LGTM!

PR #3947, opened 2025-10-16
[Fix] Create a better log message when not connected to any peers.

kaimast:
This is a simpler attempt at fixing #3885. The PR does not fix the issue of nodes starting as `synced`, but I will leave that for another PR.

The core change is that the sync state is never changed during block request generation, but only when processing responses, new block locators, or disconnects. For disconnects, there is a new function `SyncStatus::clear_greatest_peer_height` that is used when a peer disconnects and no block locators are remaining.
When invoking that function, the node's sync state will switch to `unsynced` until it receives new block locators.

The PR also adds more documentation to this delicate part of the codebase. Hopefully, the comments make the changes easy to review.

### DIFF
--- a/node/sync/src/block_sync.rs
+++ b/node/sync/src/block_sync.rs
@@ -730,8 +730,10 @@ impl<N: Network> BlockSync<N> {
     /// This should be called by at most one task at a time.
     ///
     /// # Usage
-    ///  - For validators, the primary spawns one task that periodically calls `bft::Sync::try_block_sync`. There is no possibility of multiple calls to it at a time.
-    ///  - For clients, `Client::initialize_sync` also spawns exactly one task that periodically calls this function.
+    ///  - For validators, the primary spawns exactly one task that periodically calls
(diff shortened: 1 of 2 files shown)


node/sync/src/block_sync.rs
```diff
@@ -1043,8 +1054,11 @@ impl<N: Network> BlockSync<N> {
                     return None;
                 };
 
-                // Retrieve the highest block height.
-                let greatest_peer_height = sync_peers.values().map(|l| l.latest_locator_height()).max().unwrap_or(0);
+                // Retrieve the greatest block height of any connected peer.
+                let Some(greatest_peer_height) = sync_peers.values().map(|l| l.latest_locator_height()).max() else {
+                    warn!("Cannot re-request blocks because no peers are connected");
```
kaimast, line 1059: This is another potential issue I cam across that we do not check for currently.

vicsn (approved):
LGTM!

PR #3951, opened 2025-10-17
Reduce excessive logging in block sync

kaimast:
This PR reduces log verbosity, and adds more documentation to the fairly convoluted `BlockSync::handle_block_request_timeouts`.

Concrete issues with block sync logging were the following:
* There were cases with excessive `Failed to issue new request for height` messages, which usually just means that the ledger advanced while the node was attempting to re-issue requests. This happened when we receive block responses around the same time the associated request times out.
* Validators that were too far behind would frequently print `Block sync is too far behind other validators. Syncing without BFT` at the INFO level. Now, validators will log `Try advancing block responses without BFT` at the TRACE level instead.
* Nodes sometimes logged `Re-requesting blocks starting at height` even if they did *not* re-request any blocks. 

[a811bf71](https://github.com/ProvableHQ/snarkOS/commit/a811bf71fce363405c7752baca7224cb5d9479e2) includes these log improvements and adds more documentation to the re-requesting logic. [a3a08d4](https://github.com/ProvableHQ/snarkOS/commit/a3a08d4b1c0a23c920e1566b736d50e911e0f5d9) additionally ensures that nodes do not attempt regular block sync if re-requesting blocks fails. I am unsure if this ever happened in practice, but I added it just to be safe.

### DIFF
--- a/node/sync/src/block_sync.rs
+++ b/node/sync/src/block_sync.rs
@@ -966,12 +966,16 @@ impl<N: Network> BlockSync<N> {
     /// This removes the corresponding block responses and returns the set of peers/addresses that timed out.
     /// It will ask the peer pool handling service to ban any timed-out peers.
     ///
-    /// Finally, it will return a set of new of block requests that replaced the timed-out requests (if needed).
+    /// # Return Value
+    /// On success it will return an empty set, if there is nothing to re-request, or a set of new of block requests that replaced the timed-out requests.
     /// This set of new requests can also replace requests that timed out earlier, and which we were not able to re-request yet.
+    ///
+    /// This function will return an error if it cannot re-request blocks due to a lack of peers.
+    /// In this case, the current iteration of block synchronization should not continue and the node should re-try later instead.
     pub fn handle_block_request_timeouts<P: PeerPoolHandling<N>>(
         &self,
         peer_pool_handler: &P,
-    ) -> Option<BlockRequestBatch<N>> {
+    ) -> Result<Option<BlockRequestBatch<N>>> {
         // Acquire the write lock on the requests map.
         let mut requests = self.requests.write();
 
@@ -1053,7 +1057,7 @@ impl<N: Network> BlockSync<N> {
         } else {
             // Nothing to do.
             // Do not log here as this check happens frequently.
-            return None;
+            return Ok(None);
         };
 
         // Set the maximum number of blocks, so that they do not exceed the end height.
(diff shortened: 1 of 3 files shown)


node/sync/src/block_sync.rs
```diff
@@ -367,7 +367,7 @@ impl<N: Network> BlockSync<N> {
             }
         };
 
-        debug!("Sending {len} block requests to peers {peers:?}", len = requests.len(), peers = sync_peers.keys());
+        debug!("Sending {len} block requests to peer(s) at {peers:?}", len = requests.len(), peers = sync_peers.keys());
```
ljedrz, line 370: Listing all the peers might be a bit verbose at `DEBUG` level, considering that all the individual requests would be visible at `TRACE`; perhaps the number of those peers would suffice?

node/sync/src/block_sync.rs
```diff
@@ -966,11 +966,16 @@ impl<N: Network> BlockSync<N> {
     /// This removes the corresponding block responses and returns the set of peers/addresses that timed out.
     /// It will ask the peer pool handling service to ban any timed-out peers.
     ///
-    /// Finally, it will return a set of new of block requests that replaced the timed-out requests (if needed).
+    /// # Return Value
+    /// On success it will return an empty set, if there is nothing to re-request, or a set of new of block requests that replaced the timed-out requests.
```
ljedrz: an empty set or a `None`?

node/sync/src/block_sync.rs
```diff
@@ -1169,12 +1185,9 @@ impl<N: Network> BlockSync<N> {
...
             let mut start_height = sync_height + 1;
 
-            loop {
-                if requests.contains_key(&start_height) {
-                    start_height += 1;
-                } else {
-                    break;
-                }
+            // Do not issue requests that already exist or would be obsolete.
+            while requests.contains_key(&start_height) || self.ledger.latest_block_height() >= start_height {
```
ljedrz: `Ledger::latest_block_height` involves an `RwLock::read` operation, so it would be beneficial if we moved it above the loop to obtain it just once; sure, the ledger could move in the meantime, but the likelihood is low, because:
1. block time is a lot greater than what this loop could take
2. we're attempting to sync already, so the ledger is most likely not advancing

ljedrz (commented):
Left a new batch of comments.

node/sync/src/block_sync.rs
kaimast: In production, the redundancy factor is still 1, so it should really only list one address. However, we can remove it, if you still find it too verbose.

node/sync/src/block_sync.rs
kaimast: That was indeed incorrect. Fixed now.

node/sync/src/block_sync.rs
kaimast: I am not sure if it affects performance much, but I made the suggested change.

node/sync/src/block_sync.rs
ljedrz: You've surely tested it more than I, so I'll let you be the judge - it just seemed potentially verbose to me.

node/sync/src/block_sync.rs
vicsn: I think 2. may be false, I recall we initiate sync aggressively. But I may ask for that to be changed in a separate issue soon.

vicsn (approved):

PR #3955, opened 2025-10-21
[Tweak] Validators shouldn't accept duplicate validator connections in router mode

ljedrz:
This PR resolves the following point from https://github.com/ProvableHQ/snarkOS/issues/3926:
> validator routers should store just a single router peer per validator address in the current committee.

The related method is moved from `Gateway` to `PeerPoolHandling`, as it is now applicable to the `Router` as well.

### DIFF
--- a/node/bft/src/gateway.rs
+++ b/node/bft/src/gateway.rs
@@ -371,12 +371,6 @@ impl<N: Network> Gateway<N> {
         self.worker_senders.get().and_then(|senders| senders.get(&worker_id))
     }
 
-    /// Returns `true` if the node is connected to the given Aleo address.
-    pub fn is_connected_address(&self, address: Address<N>) -> bool {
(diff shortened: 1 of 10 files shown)


ljedrz:
I had to update some `Router`-related tests so that they actually sample different accounts - otherwise the new rule causes some validator - validator connections to fail (confirming the efficacy of the main change).

vicsn (approved):
so PeerpoolHandling lives in Router? Would it make sense to move it to it's own folder (similar to how you noted shared message objects could be moved to a shared folder).

Kind of a canary in the coalmine for merging them completely one day. ;)

Can be follow-up work. If you feel like it and it's simple. But low priority.

ljedrz:
@vicsn Yes, I would definitely like to move it into some common network-related module, it shouldn't be an issue 👍.

PR #3958, opened 2025-10-22
[Metrics] Add BLOCK_LAG histogram

raychu86:
## Motivation

Adds a new `BLOCK_LAG` histogram metric to track the validator's system time with the block timestamp of the latest processed block. This is tracked in milliseconds.

### DIFF
--- a/node/bft/src/helpers/timestamp.rs
+++ b/node/bft/src/helpers/timestamp.rs
@@ -23,6 +23,16 @@ pub fn now() -> i64 {
     OffsetDateTime::now_utc().unix_timestamp()
 }
 
+/// Returns the current UTC epoch time.
+pub fn now_utc() -> OffsetDateTime {
+    OffsetDateTime::now_utc()
+}
+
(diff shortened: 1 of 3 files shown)


vicsn (approved):
This seems to work well. When running our stress-tests with 40 validators and 120 clients we see a block lag of:
- 3-5 seconds under normal conditions
- up to 30 seconds when being spammed with 6000 transactions, in line with the 20-30 second blocktimes we observe then. Lowering this blocktime is ongoing work.
- 0 seconds for the new bootstrap client which doesn't do syncing.

[Grafana link](https://aleostresstest.grafana.net/d/victor-devnet/victor-devnet?orgId=1&from=2025-10-23T09:02:37.827Z&to=2025-10-23T09:26:13.293Z&timezone=browser&refresh=5s)

Screenshot 2025-10-23 at 11 26 16

vicsn:
For reference, the cargo audit failure is benign and will be tackled in a separate PR, search for tracing in recent issues and PRs if you want to learn more.

PR #3964, opened 2025-10-23
[Feat] Enable sequential storage processing enforcement in snarkVM

ljedrz:
The snarkOS counterpart to https://github.com/ProvableHQ/snarkVM/pull/2975.

Filing as a draft due to the external branch dependency.

### DIFF
--- a/cli/src/commands/start.rs
+++ b/cli/src/commands/start.rs
@@ -55,7 +55,10 @@ use std::{
     path::PathBuf,
     sync::{Arc, atomic::AtomicBool},
 };
-use tokio::runtime::{self, Runtime};
+use tokio::{
+    runtime::{self, Runtime},
+    task,
+};
 use tracing::warn;
 use ureq::http;
 
@@ -560,7 +563,11 @@ impl Start {
             }
 
             // Construct the genesis block.
-            load_or_compute_genesis(dev_keys[0], committee, public_balances, bonded_balances, &mut rng)
+            std::thread::spawn(move || {
+                load_or_compute_genesis(dev_keys[0], committee, public_balances, bonded_balances, &mut rng)
+            })
+            .join()
+            .unwrap()
         } else {
             Block::from_bytes_le(N::genesis_bytes())
         }
@@ -621,7 +628,8 @@ impl Start {
         let cdn = self.parse_cdn::<N>().with_context(|| "Failed to parse given CDN URL")?;
 
         // Parse the genesis block.
-        let genesis = self.parse_genesis::<N>()?;
+        let start = self.clone();
+        let genesis = task::spawn_blocking(move || start.parse_genesis::<N>()).await??;
         // Parse the private key of the node.
         let account = self.parse_private_key::<N>()?;
         // Parse the node type.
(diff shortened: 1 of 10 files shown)


cli/src/commands/start.rs
```diff
@@ -556,7 +556,11 @@ impl Start {
             }
 
             // Construct the genesis block.
-            load_or_compute_genesis(dev_keys[0], committee, public_balances, bonded_balances, &mut rng)
+            std::thread::spawn(move || {
```
kaimast, line 566: Could you document why this is needed? It is not super clear to me.

node/bft/src/sync/mod.rs
```diff
@@ -1254,6 +1263,12 @@ mod tests {
         assert!(syncing_ledger.contains_block_height(1));
         assert!(syncing_ledger.contains_block_height(2));
 
+        tokio::task::spawn_blocking(move || {
+            drop(core_ledger);
```
kaimast, line 1286: Same here. Why do these objects have to be dropped in a dedicated task?

node/bft/src/sync/mod.rs
```diff
@@ -1432,6 +1457,13 @@ mod tests {
         }
         // Check that the set of pending certificates is equal to the set of candidate pending certificates.
         assert_eq!(pending_certificates, candidate_pending_certificates);
+
+        println!("1");
```
kaimast: This might be a leftover debug statement

kaimast:
I am testing these changes in this PR: #3983. However, I noticed both PRs still panic when running tests for `snarkos-node-bft`.

cli/src/commands/start.rs
ljedrz: It's a direct consequence of the use of `blocking_recv` in the dedicated sequential processing thread in snarkVM - the related function creates an instance of `VM`, and then runs `VM::genesis_quorum`, which triggers a sequential operation (and those are blocking). The owning function, `parse_node`, is called from an `async` context, which is why the wrapper is needed. That being said, perhaps using a blocking task in `parse_node` would be preferable - I'll check it out.

node/bft/src/sync/mod.rs
ljedrz: Same reason - the shutdown involves a blocking operation due to the signal involved; however, since this is only an issue in tests, perhaps it can be ignored by `catch_unwind` - I'll look into it.

ljedrz:
> I noticed both PRs still panic when running tests for snarkos-node-bft.

It's most likely just some leftover test cases where the drop wasn't wrapped. That being said, all should be fixed now.

node/bft/src/sync/mod.rs
ljedrz: Actually, it was [even easier](https://github.com/ProvableHQ/snarkVM/pull/2975/commits/c29f4309af57c38ca2caccb592c00e2e689552b6); I'll update the PR shortly.

kaimast:
I rebased the branch onto latest staging. It should be ready to merge now.

kaimast (dismissed):

kaimast (approved):

vicsn (approved):
Super happy we're seeing more stability with this approach!

PR #3965, opened 2025-10-23
[CI] Set correct sccache folder

kaimast:
The `cargo-audit` workflow has been failing. I realized it is because we cache `.cargo` instead of `~./cargo`, which means the workflow overwrites our `audit.toml` file.

This PR bumps the cache prefix so that we generated new caches. I also included the sccache fixes from snarkVM, so it is used correctly in snarkOS as well.


vicsn (dismissed):

kaimast:
Closing in favor of #3980.

PR #3966, opened 2025-10-23
[Chore] Update snarkVM rev

kaimast:
This PR bumps the snarkVM and locktick versions.


vicsn:
```
thread 'tests::test_latest_consensus_version' panicked at node/router/messages/src/lib.rs:328:9:
assertion `left == right` failed
  left: V11
 right: V12
```

vicsn (approved):

PR #3967, opened 2025-10-24
Correct primary ping processing logic

vicsn:
## Motivation

While browsing the code I noticed this logic error. Please correct me if I'm mistaken.

Even though this is unlikely to directly have caused a network halt, it may still degrade performance when nodes face networking issues and they miss out on certain certificates.

## Test Plan

Regular CI and periodic stress-testing should suffice.

### DIFF
--- a/node/bft/src/primary.rs
+++ b/node/bft/src/primary.rs
@@ -1256,7 +1256,7 @@ impl<N: Network> Primary<N> {
         self.spawn(async move {
             while let Some((peer_ip, primary_certificate)) = rx_primary_ping.recv().await {
                 // If the primary is not synced, then do not process the primary ping.
-                if self_.sync.is_synced() {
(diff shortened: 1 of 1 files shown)


kaimast:
I believe the current behavior, which ignores the ping if the node is synced, is correct.

vicsn:
> I believe the current behavior, which ignores the ping if the node is synced, is correct.

The current behavior, before this PR, ignores the ping if the node is not synced...

kaimast:
> > I believe the current behavior, which ignores the ping if the node is synced, is correct.
> 
> The current behavior, before this PR, ignores the ping if the node is not synced...

Sorry, I meant *not synced*. When the node is not synced (i.e., syncing), certificates are instead fetched by `Sync`/`BlockSync`.

PR #3971, opened 2025-10-24
[Feat] Provide repo sha over handshake

### LINKED ISSUE #3969: [Feature] Transmit commit hash on gateway and router handshake
## 🚀 Feature

As part of https://github.com/ProvableHQ/snarkOS/issues/3566 it would be great to collect more information from peers to help debug in case of a network outage.

It would be great if nodes (assuming honesty) transmit their snarkOS commit hash during the handshake, so the recipient can log this. This information should currently already be collected as it is also output on calling --version. We don't have to share build features.

We'll have to put some thought into making this backwards compatible.

ljedrz:
Closes https://github.com/ProvableHQ/snarkOS/issues/3969. It is backwards-compatible - if the value is missing, we just assume that it's unknown.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4256,6 +4256,7 @@ dependencies = [
  "indexmap 2.12.0",
  "proptest",
  "serde",
+ "snarkos-node-network",
  "snarkos-node-sync-locators",
  "snarkvm",
  "test-strategy 0.4.3",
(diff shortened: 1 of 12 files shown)


vicsn (dismissed):
Love it

ljedrz:
Hmm, not sure what went wrong with the `proptest`s, they passed for me locally; perhaps they became flaky, or some last-minute refactoring broke them; I'll check it out soon.

ljedrz:
The most recent CI failure is unrelated.

vicsn (approved):

PR #3972, opened 2025-10-24
[Fix] Do not tack transmissions that are added to the mempool

### LINKED ISSUE #3907: [Bug] Fix stale transmissions metric
## 🐛 Bug Report

When running stresstests against the latest staging branch, we observe lots of stale transmissions [in the metrics](https://aleostresstest.grafana.net/d/prerelease-devnet/prereleasedevnet?orgId=1&from=2025-10-02T10:45:55.528Z&to=2025-10-02T12:12:57.506Z&timezone=browser&viewPanel=panel-66):
- snarkos_consensus_stale_unconfirmed_solutions
- snarkos_consensus_stale_unconfirmed_transactions

We can ignore solutions (which are spammed a lot), but in the test there are also up to 500 stale transactions. We should investigate why and resolve this. Note that in production, there is a low but non-zero continuous amount of stale transmissions, so hopefully fixing the tests will fix production too.

Either the metric is still broken, or we have to fix the disappearance of transmissions. Given that 98% of transmissions land (based on comparison of *confirmed* and *unconfirmed* metrics), the metric is probably at fault.

Related issue: https://github.com/ProvableHQ/snarkOS/issues/3514

Image

Image

kaimast:
I found one instance in the code where a transaction can be added to the `transmissions_tracker` (when metrics are enabled) but will not be added to the mempool.

This might fix #3907. The PR also adds a few more comments to the code that should help the next person looking into this.

### DIFF
--- a/node/consensus/src/lib.rs
+++ b/node/consensus/src/lib.rs
@@ -372,6 +372,10 @@ impl<N: Network> Consensus<N> {
             if self.ledger.contains_transmission(&TransmissionID::Transaction(transaction_id, checksum))? {
                 bail!("Transaction '{}' exists in the ledger {}", fmt_id(transaction_id), "(skipping)".dimmed());
             }
+            // Check that the transaction is not in the mempool.
+            if self.contains_transaction(&transaction_id) {
+                bail!("Transaction '{}' exists in the memory pool", fmt_id(transaction_id));
+            }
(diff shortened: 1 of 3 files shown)


vicsn (approved):
Nice find! I'll check out the metric the next time we do a prerelease run

PR #3973, opened 2025-10-24
Reduce spurious ERROR logs

### LINKED ISSUE #3716: [Bug] Revisit which (unique) consensus and peering logs should be which log levels
## 🐛 Bug Report

In order to speed up debugging consensus issues, we should give logs more appropriate log levels. 

E.g. this should be a WARN (and there should not be another log saying the same thing):
```
snarkos[18506]: 2025-07-08T17:59:11.185621Z DEBUG snarkos_node_router::heartbeat: No connected peers
```

With that, we could first search for ERROR logs, then WARN, the DEBUG, etc.

vicsn:
## Motivation

Closes https://github.com/ProvableHQ/snarkOS/issues/3716

Some of the consensus-related logs are tricky to extract and don't happen that frequently anyway, so only changes some P2P log levels.

### DIFF
--- a/node/tcp/src/protocols/handshake.rs
+++ b/node/tcp/src/protocols/handshake.rs
@@ -69,7 +69,7 @@ where
                             Ok(conn)
                         }
                         Ok(Err(e)) => {
-                            debug!(parent: node.tcp().span(), "handshake with {} failed: {}", addr, e);
+                            debug!(parent: node.tcp().span(), "handshake with {addr} failed: {e}");
                             Err(e)
                         }
                         Err(_) => {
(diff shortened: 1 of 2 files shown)


node/tcp/src/protocols/handshake.rs
```diff
@@ -69,11 +69,11 @@ where
                             Ok(conn)
                         }
                         Ok(Err(e)) => {
-                            error!(parent: node.tcp().span(), "handshake with {} failed: {}", addr, e);
+                            debug!(parent: node.tcp().span(), "handshake with {} failed: {}", addr, e);
```
kaimast: This might trigger clippy warnings soon. Might be good to change it to this before merging.

```suggestion
                            debug!(parent: node.tcp().span(), "handshake with {addr} failed: {e}");
```

kaimast (dismissed):
LGTM! I always found the TCP warnings a little redundant.

ljedrz (dismissed):

ljedrz (approved):

PR #3974, opened 2025-10-24
Specify ConsensusVersion relevant to the task at hand

vicsn:
## Motivation

A small codestyle nit, this makes it easier to grep through the codebase to see the influence of a ConsensusVersion.


kaimast (approved):
LGTM! I think I used V11, because V12 did not exist yet in the codebase at that point.

Issue #3636, opened 2025-05-17
[Bug] Spurious error messages during sync


kaimast:
While syncing, we sometimes see the following error message.

```
2025-05-15T12:36:46.608927Z ERROR Unable to advance to the next block - Failed to speculate on transactions - Failed to post-ratify - Next round 2532 must be greater than current round 2532
2025-05-15T12:36:46.608943Z ERROR BFT failed to advance the subdag for round 2532 - Failed to speculate on transactions - Failed to post-ratify - Next round 2532 must be greater than current round 2532
```

These do not cause the network to halt, and do not seem to indicate any (long-term) problems. However, we should still investigate why this happens and fix it, as these error messages are confusing to users and clutter the logs.

My guess is this is due to some race condition. For example, a block might be processed/created too soon, causing the node to abort and retry later.

vicsn:
These logs exist 3.7.1

We should look into the logs of stress tests of your new re-opened sync refactor, with a bit of luck its already fixed

vicsn:
Separate similar instance appearing recently: 
```
336370: 2025-10-22T06:24:55.582428Z ERROR snarkos_node_consensus: Unable to advance to the next block - Failed to construct block template
336596: 2025-10-22T06:24:55.606636Z ERROR snarkos_node_bft::bft: BFT failed to advance the subdag for round 284 - Failed to construct block template
```

kaimast:
I ran a devnet with ten validators for quite a long time (~5000 blocks), but did not see either of these errors. My current guess is that these are caused by certain transactions. Devnet blocks do not contain many transactions, so that might be why I do not observe the error there.

kaimast:
This should be fixed with the recent sync improvements and usage of `check_logs` in CI.

PR #3978, opened 2025-10-27
[Logs] Print detailed errors in `bft.rs`

kaimast:
This PR does not change any functionality, but simply improves how errors in `bft.rs` are printed.

The change will be helpful to debug #3636. Merging it into staging will allow us to collect better logs from validators.


vicsn:
Running `prerelease-v4.4.4`

PR #3983, opened 2025-10-31
[Fix] Restore Pending Blocks

kaimast:
https://github.com/ProvableHQ/snarkVM/pull/3019 needs to be merged first, but it is ready to be reviewed otherwise.

This PR restores the `PendingBlock` changes and makes a small adjustment ([05d93a9](https://github.com/ProvableHQ/snarkOS/pull/3983/commits/05d93a97a8d8caeba506ecd86827d64a7d103e04)) so it works with the new sequential ops in snarkVM.


kaimast:
Closing this in favor of #4011, as we are testing those two together.

PR #3988, opened 2025-11-03
Add trusted_peers_only flag, deprecate allow_external_peers and rotate_external_peers

### LINKED ISSUE #3984: [Feature] Introduce --strict-peering flag
## 🚀 Feature

There is a particular edge case where a node should **only** use the peers specified by its --validators and --peers flags, not by its peer cache or hardcoded bootstrap peers. If the network forks, nodes need to be able to manually decide which node is at tip and who they want to sync from.

### LINKED ISSUE #3985: [Feature] remove --allow-external-peers
## 🚀 Feature

With the recent peering adjustments, I think we have a clear enough vision to just set sensible defaults for different types of nodes. Validators don't welcome external non-validator peers, clients do.

vicsn:
## Motivation

Closes https://github.com/ProvableHQ/snarkOS/issues/3984
Closes https://github.com/ProvableHQ/snarkOS/issues/3985
Deprecates rotate_external_peers which is no longer needed now that we have custom bootstrap peers.

Tackled them together in order to reduce the aggregate diff.

## Test Plan

- [x] Letting a local devnet produce a few blocks, turning off a validator, revive it as a client with --trusted-peers-only to see if it connects and syncs from just a single validator.

## Documentation

Validator onboarding docs will be adjusted accordingly.

## Backwards compatibility

The deprecated flags are still present to ensure nodes can still start with them, so as not to break any existing workflows.

### DIFF
--- a/node/bft/src/gateway.rs
+++ b/node/bft/src/gateway.rs
@@ -167,6 +167,8 @@ pub struct InnerGateway<N: Network> {
     handles: Mutex<Vec<JoinHandle<()>>>,
     /// The storage mode.
     storage_mode: StorageMode,
+    /// If the flag is set, the node will only connect to trusted peers.
+    trusted_peers_only: bool,
     /// The development mode.
     dev: Option<u16>,
 }
@@ -195,12 +197,14 @@ impl<N: Network> PeerPoolHandling<N> for Gateway<N> {
 
 impl<N: Network> Gateway<N> {
     /// Initializes a new gateway.
+    #[allow(clippy::too_many_arguments)]
     pub fn new(
         account: Account<N>,
         storage: Storage<N>,
         ledger: Arc<dyn LedgerService<N>>,
         ip: Option<SocketAddr>,
         trusted_validators: &[SocketAddr],
+        trusted_peers_only: bool,
         storage_mode: StorageMode,
         dev: Option<u16>,
     ) -> Result<Self> {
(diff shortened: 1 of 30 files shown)


vicsn:
Found https://github.com/ProvableHQ/snarkOS/issues/3991

node/bft/src/gateway.rs
```diff
@@ -1535,6 +1549,11 @@ impl<N: Network> Gateway<N> {
             warn!("{CONTEXT} Dropping '{peer_addr}' on version {version} (outdated)");
             return Some(DisconnectReason::OutdatedClientVersion);
         }
+        // If the node is in trusted peers only mode, ensure the peer is trusted.
+        if self.trusted_peers_only && !self.is_trusted(peer_addr) {
```
ljedrz: `peer_addr` is not necessarily the address we're recognizing the peer with - that one is constructed based on the listener port from the `ChallengeRequest`, and would need to be built here as well; the same applies to the check for bootstrap peers below, where it had been missed

vicsn: 1d1a645e9

ljedrz: that's probably the cause behind https://github.com/ProvableHQ/snarkOS/pull/3988#issuecomment-3480412132

ljedrz (changes requested):
Looks good, but I think we need two more things:
1. we should probably disable the requests for peers when connecting to bootstrap clients in the `OnConnect::on_connect` impls in `only_trusted_peers` mode
2. we should adjust `PeerPoolHandling::check_connection_attempt` to reject attempts to connect to untrusted peers (before the handshake)

vicsn:
>     we should probably disable the requests for peers when connecting to bootstrap clients in the OnConnect::on_connect impls in only_trusted_peers mode

Feels like overkill, we already perform the check in the gateway/router handshakes, and currently we don't have any such type of validation in on_connect.

>     we should adjust PeerPoolHandling::check_connection_attempt to reject attempts to connect to untrusted peers (before the handshake)

Added it: 9b65d976d

node/bft/src/gateway.rs
vicsn: I'm still seeing the issue on that commit

vicsn: Aha: https://github.com/ProvableHQ/snarkOS/issues/3991

ljedrz (dismissed):
The new changes are OK 👍.

ljedrz (approved):

Issue #3899, opened 2025-09-29
[Bug] Periodic DNS lookup required for trusted clients


vicsn:
## 🐛 Bug Report

We currently only lookup peer DNS entries at startup. This is mostly not a problem, but when trusted core clients update it may prevent them from connecting. We should periodically refresh with a new DNS lookup.

PR #3989, opened 2025-11-03
Disallow unknown peers as validator

vicsn:
## Motivation

Extension to https://github.com/ProvableHQ/snarkOS/issues/3888

Validator gateways will already only connect to other validators in the current committee. We didn't extend this logic to the Router because the Router messages aren't signed. However we can at least stick to trusted connections only.

Notes:
- this will be a pain for validators who have some managed nodes which frequently rotate, but if anyone complains we can tackle https://github.com/ProvableHQ/snarkOS/issues/3899
- even though gateways also broadcast transmissions via the `WorkerPing`, this PR may reduce validator transmission propagation overall and increase transmission inclusion latency, reducing the chance that transmissions land in a proposal. However, the validator onboarding guide _already_ implies not to connect to other validator's clients.

## Test Plan

- [x] CI passes

### DIFF
--- a/node/router/src/handshake.rs
+++ b/node/router/src/handshake.rs
@@ -339,7 +339,14 @@ impl<N: Network> Router<N> {
         if self.is_local_ip(listener_addr) {
             bail!("Dropping connection request from '{listener_addr}' (attempted to self-connect)");
         }
-        // Unknown peers are untrusted, so check if `trusted_peers_only` is true.
+        // As a validator, only accept connections from trusted peers and bootstrap nodes.
+        if self.node_type() == NodeType::Validator
+            && !self.is_trusted(listener_addr)
+            && !crate::bootstrap_peers::<N>(self.is_dev()).contains(&listener_addr)
+        {
+            bail!("Dropping connection request from '{listener_addr}' (untrusted)");
+        }
+        // If the node is in trusted peers only mode, ensure the peer is explicitly trusted.
         if self.trusted_peers_only() && !self.is_trusted(listener_addr) {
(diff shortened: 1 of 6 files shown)


ljedrz (commented):
LGTM, though some tests may require an adjustment.

node/router/src/handshake.rs
```diff
@@ -339,7 +339,14 @@ impl<N: Network> Router<N> {
         if self.is_local_ip(listener_addr) {
             bail!("Dropping connection request from '{listener_addr}' (attempted to self-connect)");
         }
-        // Unknown peers are untrusted, so check if `trusted_peers_only` is true.
+        // As a validator, only accept connections from trusted peers and bootstrap nodes.
+        if self.node_type() == NodeType::Validator
+            && !self.is_trusted(listener_addr)
+            && !crate::bootstrap_peers::<N>(self.is_dev()).contains(&listener_addr)
```
kaimast, line 345: Are bootstrap peers not considered trusted? It would be cleaner if validators always returned `true` for `trusted_peers_only` and we could avoid this special case entirely.

vicsn: The goal of `trusted_peers_only` is to be able to sync from the peers which you explicitly specified. And although the new `BootstrapClient` does not facilitate syncing, I think that one may not be deployed to mainnet yet so we might as well exclude bootstrap peers.

I think there may be several places in the code where trusted refers to "peers passed in via CLI flags", but we could consider unifying one day and letting `is_trusted` call into `bootstrap_peers` under the hood.

ljedrz (approved):

PR #3993, opened 2025-11-04
Fix DEFAULT_ENDPOINT in snarkos developer

### LINKED ISSUE #3992: [Bug] `snarkos developer execute` not working
## 🐛 Bug Report

Any attempts to execute a function using the `snarkos developer execute` command returns an error
```
⚠️ VM failed to execute transaction locally
     ↳ Failed to parse JSON error response
     ↳ json: expected value at line 1 column 1
```

## Steps to Reproduce

The above was produced by running:
```
snarkos developer execute --dry-run --private-key APrivateKey1zkp... credits.aleo transfer_public aleo1... 1000u64
```
although the specific function to be run doesn't seem to matter.

## Notes

@vicsn mentions that it happens "when you append /v1/ to the endpoint passed to `--query`"

vicsn:
## Motivation

Closes https://github.com/ProvableHQ/snarkOS/issues/3992 CC @AleoAlexander

## Test Plan

Tested the command from the issue manually.

Our CI also contains a few calls using `snarkos developer` but didn't catch this because we override the --query.

### DIFF
--- a/cli/src/commands/developer/mod.rs
+++ b/cli/src/commands/developer/mod.rs
@@ -68,8 +68,7 @@ pub enum DeveloperCommand {
 }
 
 /// Use the Provable explorer's API by default.
-/// Note, the `v1` here is not the API version, but the explorer's version.
-const DEFAULT_ENDPOINT: &str = "https://api.explorer.provable.com/v1";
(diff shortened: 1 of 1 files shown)


kaimast (approved):
Looks good!

PR #3994, opened 2025-11-04
Implement reporting unit test timings in CI

meddle0x53:
## Motivation

Implementation of the SnarkVM logic for measuring how much time different test take in SnarkOS.
See https://github.com/ProvableHQ/snarkVM/pull/2995

## Test Plan

The CI passes and the timings of the different tests could be viewed.

## Documentation

The PR updates the CI configuration and modifies how tests are ran.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -141,18 +141,183 @@ commands:
       flags:
         type: string
         default: ""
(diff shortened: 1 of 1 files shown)


vicsn (approved):
LGTM!

PR #3995, opened 2025-11-04
[Do Not Merge] Increase tps test

vicsn:
## Motivation

Opening to run CI.


vicsn:
With the latest change, we finally have on average 70 transmissions per proposal - but it makes block generation take longer and doesn't materially change TPS

Screenshot 2025-11-05 at 14 59 31

PR #3996, opened 2025-11-04
[Fix] Ensure `dev execute` works with a v1 API

kaimast:
The PR detects the API version in a given endpoint and adjust its logic accordingly. Depends on https://github.com/ProvableHQ/snarkVM/pull/3036.

### Overview of changes
* Adds a new `parse_custom_endpoint` method with unit tests.
* Expects errors as JSON for V2 and as strings for V1.
* Improvements to the `devnet_ci.sh` script, such as better log messages and a new test for the V2 API.
* Ensures that `--wait` works correctly for V1, which returns `500` not `404` for a missing transaction.

### Notes
* Some of the version parsing logic is also done in the corresponding snarkVM PR. I wanted to add as little on the snarkVM side as possible, though, so the `cli` does its own parsing of the version number.
* I simplified `parse_block_range` as it had two different branches that were fetching the latest block height.

### DIFF
--- a/cli/src/commands/developer/mod.rs
+++ b/cli/src/commands/developer/mod.rs
@@ -199,16 +199,17 @@ impl Developer {
         ensure!(!route.starts_with('/'), "path cannot start with a slash");
 
         // If the route already ends with a version segment (v1 or v2), don't prepend a version.
-        let (route_has_version_suffix, api_version) = {
+        let api_version = {
             let r = base_url.path().trim_end_matches('/');
 
             if r.ends_with(API_VERSION_V1) {
-                (true, ApiVersion::V1)
+                ApiVersion::V1
             } else if r.ends_with(API_VERSION_V2) {
-                (true, ApiVersion::V2)
+                ApiVersion::V2
             } else {
-                // Default to v2.
-                (false, ApiVersion::V2)
+                // Default to v1.
+                // Note: If the snarkos-node-rest switches default to v2, this needs to be updated.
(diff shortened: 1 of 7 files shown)


vicsn:
Clarifying a point from https://github.com/ProvableHQ/snarkOS/pull/4002 - we should not silently postfix versions because this is not what we can expect from explorers

kaimast:
> Clarifying a point from #4002 - we should not silently postfix versions because this is not what we can expect from explorers

Addressed in [7eb5f99](https://github.com/ProvableHQ/snarkOS/pull/3996/commits/7eb5f99a088dcfa5096204a70ee732767520b6da).

cli/src/commands/developer/mod.rs
```diff
@@ -183,14 +191,26 @@ impl Developer {
...
+    ///
+    /// # Returns
+    /// The full endpoint Uri and the API version used by it.
+    fn build_endpoint<N: Network>(base_url: &http::Uri, route: &str) -> Result<(String, ApiVersion)> {
         // This function is only called internally but check for additional sanity.
         ensure!(!route.starts_with('/'), "path cannot start with a slash");
 
         // If the route already ends with a version segment (v1 or v2), don't prepend a version.
-        let route_has_version_suffix = {
+        let api_version = {
```
kaimast, line 202: There is still one case where we need this.

The old API returns 500 in some cases where v2 returns 404.

vicsn (approved):

PR #4002, opened 2025-11-06
Fix developer execute

### LINKED ISSUE #3758: [Bug] `snarkos developer execute` doesn't work under consensus V8

vicsn:
## Motivation

Fixes https://github.com/ProvableHQ/snarkOS/issues/3758#issuecomment-3487937480

CC @HarukaMa 

- f219e719c7f923d913a51773868ac77a39809403 a clear bug to be fixed - manually tested on the example in the issue above
- 2d90c763138ea3f386a6868be8ada2059d52d91c doc improvement
- d6bba7a02fb995c205792d02c51b79bb3fb48957 somehow I could not fetch the fee amount. We first need to:
  - Merge this first to help debugging: https://github.com/ProvableHQ/snarkVM/pull/2999
  - Test on all explorers in the ecosystem that they at least support the snarkOS REST endpoints
- ee343af5df4a264d7e88ad6ff225f230f4d5da4e in order to support multiple explorers, we should not be amending the path other than what snarkOS' REST endpoints may expect (so appending the network name is fine)


HarukaMa:
Can confirm the command works with this PR.

For the explorer support, it did try to access `GET /v2/mainnet/block/height/latest` when i use `-e https://aleoscan.io/` where my setup doesn't have the versioning for the snarkOS REST API proxy.

kaimast:
This PR has been replaced by #4037.

PR #4004, opened 2025-11-08
Add cargo to the ubuntu build script

mikenike360:
## Motivation

When setting up a new node I had to install cargo manually

## Test Plan

I rebuilt snarkos with the ubuntu build script

## Documentation

add cargo as a dependency to the ubuntu build

## Backwards compatibility

No

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -19,6 +19,7 @@ use std::{
     fs::{self, File},
     io::Read,
     path::Path,
+    path::PathBuf,
     process,
 };
 use toml::Value;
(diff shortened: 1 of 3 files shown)


build_ubuntu.sh
```diff
@@ -19,7 +19,9 @@ sudo apt-get install -y \
 	pkg-config \
 	tmux \
 	xz-utils \
-	ufw
+	ufw \
+	cargo \
+	rustup
```
vicsn, line 24: This is downloaded below

build_ubuntu.sh
```diff
@@ -29,7 +31,7 @@ source $HOME/.cargo/env
 
 # Install snarkOS
 # cargo clean
-cargo install --locked --path .
+cargo install --locked --path . --features cuda
```
vicsn, line 34: This should not be the default for all users. Feel free to update the README for provers

build.rs
```diff
@@ -272,6 +273,46 @@ fn check_tokio_console_flags() {
     }
 }
 
+/// When the "cuda" feature is enabled, add CUDA library search paths and link the CUDA runtime.
+/// This keeps changes minimal and avoids touching any other build behavior.
+fn setup_cuda_linking_if_needed() {
```
vicsn, line 278: The current snarkOS puzzle can't use any of the CUDA code in snarkVM, it is a leftover from a previous puzzle and may become relevant again in the future due to ARC-43

vicsn (commented):
Appreciate the interest, some documentation should be improved but most of the PR is not applicable at the moment

mikenike360:
Well I close this since the puzzle doesn't even use cuda. Thanks!

PR #4008, opened 2025-11-11
[Fix] Make repo sha in handshake backwards compatible

vicsn:
## Motivation

Fixes an issue introduced in https://github.com/ProvableHQ/snarkOS/pull/3971
While upgraded nodes accept handshakes from old nodes without SHA, old nodes won't accept handshakes from new nodes.

## Test Plan

- [x] Run proper `update_validators_and_clients_one_by_one` E2E test

### DIFF
--- a/node/bft/src/gateway.rs
+++ b/node/bft/src/gateway.rs
@@ -50,6 +50,7 @@ use snarkos_node_network::{
     PeerPoolHandling,
     Resolver,
     bootstrap_peers,
+    built_info,
     log_repo_sha_comparison,
 };
 use snarkos_node_sync::{MAX_BLOCKS_BEHIND, communication_service::CommunicationService};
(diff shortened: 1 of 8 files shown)


node/bft/src/gateway.rs
```diff
@@ -1412,8 +1413,14 @@ impl<N: Network> Gateway<N> {
 
         // Sample a random nonce.
         let our_nonce = rng.r#gen();
+        // Determine the snarkOS SHA to send to the peer.
+        let current_block_height = self.ledger.latest_block_height();
+        let consensus_version = N::CONSENSUS_VERSION(current_block_height).unwrap();
+        let snarkos_sha = (consensus_version >= ConsensusVersion::V12)
+            .then(|| built_info::GIT_COMMIT_HASH.unwrap_or_default().into());
```
ljedrz, line 1420: nit: wouldn't this result in `Some("")` in case `GIT_COMMIT_HASH` is missing (shouldn't happen)?

ljedrz (approved):
LGTM with one nit.

node/bft/src/gateway.rs
vicsn: Yes, just like before this PR "" will be transmitted and the recipient will parse it as `unknown`. Perhaps we can argue there is some information in transmitting "", we know the counterparty updated 🤷

PR #4011, opened 2025-11-14
[Fix] Explicitely pause sync while committing leader certificate

### LINKED ISSUE #3911: [Bug] Errors during sync race conditions
This seems to be logged fairly frequently, but seems to be benign.

```
 Block synchronization failed  — Block has invalid height: Was 135 but expected 136.
```

Here, the ledger already advanced when block sync is trying to update it. This might happen because the sync passes the certificates to BFT before trying to advance the ledger. The BFT might then advance the ledger before sync does.
This behavior might also cause #3636.

We should avoid the race condition altogether by ensuring sync only advances the ledger when "syncing" is true, and BFT never advances the ledger when "syncing" is false.

kaimast:
This PR aims to fix #3911. It does so by pausing block sync when the BFT commits a new leader certificate. This "pausing" is achieved through an existing `sync_lock`.

Other minor changes in this PR:
* Adjusted`build.rs` to work correctly with MutexGuards.
* Removed the `response_lock` in block sync, as the changes in #3810 already ensure that there is always exactly one task processing responses.
* Error logging improvements in `primary.rs`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -209,7 +209,7 @@ checksum = "3b43422f69d8ff38f95f1b2bb76517c91589a924d1559a0e935d7c8ce0274c11"
 dependencies = [
  "proc-macro2",
  "quote 1.0.42",
- "syn 2.0.110",
+ "syn 2.0.111",
 ]
 
 [[package]]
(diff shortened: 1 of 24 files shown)


vicsn (dismissed):
Neat. Happy to merge if you have a passing prereleases test branch/link

kaimast:
I rebased this on the pending blocks PR (#3983), so I can stress test both.

PR #4014, opened 2025-11-17
Improve logging: log stake of unconnected validators

vicsn:
## Motivation

This helps to see how urgent it is to get some validators to become connected.

## Test Plan

- [x] Checked log message correctness on local network

### DIFF
--- a/node/bft/src/gateway.rs
+++ b/node/bft/src/gateway.rs
@@ -903,7 +903,7 @@ impl<N: Network> Gateway<N> {
                 connected_validator_addresses.insert(a);
                 a.to_string()
             });
-            debug!("{}", format!("Connected to: {peer_ip} - {address}").dimmed());
+            debug!("{}", format!("  Connected to: {peer_ip} - {address}").dimmed());
         }
 
         // Log the validators that are not connected.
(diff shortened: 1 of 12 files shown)


node/src/client/mod.rs
```diff
@@ -23,7 +23,11 @@ use snarkos_node_cdn::CdnBlockSync;
 use snarkos_node_network::NodeType;
 use snarkos_node_rest::Rest;
 use snarkos_node_router::{
-    Heartbeat, Inbound, Outbound, Router, Routing,
+    Heartbeat,
```
kaimast, line 26: Isn't the previous formatting the correct one? Either way, the formatting changes should probably not be included in this PR.

node/bft/src/gateway.rs
```diff
@@ -891,34 +891,42 @@ impl<N: Network> Gateway<N> {
...
-        // Log the connected validators and count the total connected stake.
-        info!("{connections_msg}");
+        // Include and log the connected validators.
         for peer_ip in &connected_validators {
             let address = self.resolve_to_aleo_addr(*peer_ip).map_or("Unknown".to_string(), |a| {
                 connected_validator_addresses.insert(a);
                 a.to_string()
             });
-            debug!("{}", format!("  {peer_ip} - {address}").dimmed());
+            debug!("{}", format!("Connected to: {peer_ip} - {address}").dimmed());
```
kaimast: Did you intentionally remove the indent here?

node/bft/src/gateway.rs
```diff
@@ -891,34 +891,42 @@ impl<N: Network> Gateway<N> {
...
         // Log the validators that are not connected.
         let num_not_connected = validators_total.saturating_sub(connected_validators.len());
         if num_not_connected > 0 {
-            info!("Not connected to {num_not_connected} validators {total_validators}");
             // Collect the committee members.
             let committee_members: HashSet<_> =
                 self.ledger.current_committee().map(|c| c.members().keys().copied().collect()).unwrap_or_default();
 
-            // Log the validators that are not connected.
+            let mut not_connected_stake = 0;
```
kaimast: Not necessary, but you could use `iter().map(..).sum()` on the loop instead.

node/bft/src/gateway.rs
```diff
@@ -891,34 +891,42 @@ impl<N: Network> Gateway<N> {
...
+                let address_stake = committee.get_stake(*address);
+                let address_stake_as_percentage = address_stake as f64 / committee.total_stake() as f64 * 100.0;
+                not_connected_stake += address_stake;
+                debug!(
+                    "{}",
+                    format!("  Not connected to {address} ({address_stake_as_percentage:.2}% of total stake)").dimmed()
+                );
             }
+            let not_connected_stake_as_percentage = not_connected_stake as f64 / committee.total_stake() as f64 * 100.0;
+            warn!(
```
kaimast, line 931: Should this warning be printed regardless of how low the non-connected stake is?

vicsn: IMO yes. It provides similar usefulness to other warns. Validators should just be connected all the time.

node/src/client/mod.rs
vicsn: I just followed cargo fmt. Separate commit theoretically better yes.

node/bft/src/gateway.rs
vicsn: Not sure and not sure it matters

kaimast (approved):
LGTM

PR #4015, opened 2025-11-17
[Logs] Avoid `transaction/solution already exists` warnings

kaimast:
### Motivation

I commonly see warnings like the following while testing:
```
WARN snarkos_node_consensus: Failed to add unconfirmed deployment transaction 'at1vxuhwu8z5qkj8..' to the memory pool - Transaction 'at1vxuhwu8z5qkj8...1006109566105467..' already exists.
```

This seems totally benign, and logging a warning may confuse users.

## Proposed change

The change in this PR is very similar to #3981. The relevant functions now return `Ok(false)`. For a valid transaction/solution that already exist in the mempool, in which case nodes print a debug message instead of a warning.

## Notes

The change set is a little bigger as the channels require quite a bit of boilerplate. I hope we can eventually merge something like #3858 to simplify this.

### DIFF
--- a/node/consensus/src/lib.rs
+++ b/node/consensus/src/lib.rs
@@ -355,8 +355,8 @@ impl<N: Network> Consensus<N> {
                     fmt_id(solution_id)
                 ),
                 Err(err) => {
-                    // If the BFT is synced, then log the warning.
-                    if self.bft.is_synced() {
+                    // If the node is synced and the occurs after the first 10 blocks of an epoch, log it as a warning, otherwise ignore.
+                    if self.bft.is_synced() && self.ledger.latest_block_height() % N::NUM_BLOCKS_PER_EPOCH > 10 {
                         let err = err.context(format!(
                             "Unable to add unconfirmed solution '{}' to the memory pool",
                             fmt_id(solution_id)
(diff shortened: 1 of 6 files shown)


node/consensus/src/lib.rs
```diff
@@ -347,13 +348,21 @@ impl<N: Network> Consensus<N> {
...
-            if let Err(e) = self.primary_sender.send_unconfirmed_solution(solution_id, Data::Object(solution)).await {
-                // If the BFT is synced, then log the warning.
-                if self.bft.is_synced() {
-                    // If error occurs after the first 10 blocks of the epoch, log it as a warning, otherwise ignore.
-                    if self.ledger.latest_block_height() % N::NUM_BLOCKS_PER_EPOCH > 10 {
-                        warn!("Failed to add unconfirmed solution '{}' to the memory pool - {e}", fmt_id(solution_id))
-                    };
+            match self.primary_sender.send_unconfirmed_solution(solution_id, Data::Object(solution)).await {
+                Ok(true) => {}
+                Ok(false) => debug!(
```
vicsn, line 353: Can we keep the 10 block ignore window where it makes sense? Either in the Ok(false) path and/or the Err(err) path

vicsn (changes requested):
1 nit

vicsn (dismissed):

vicsn:
@kaimast we may need to extend the connection deadline in the test, I've seen this one fail more often

vicsn (approved):

PR #4017, opened 2025-11-18
[Perf] Cache the block tree on shutdown

ljedrz:
This is the snarkOS counterpart to https://github.com/ProvableHQ/snarkVM/pull/3030, targeting https://github.com/ProvableHQ/snarkVM/issues/3026.

I've successfully run some local tests using it, and saving the block tree seems to work most of the time; I'll soon iron out some edge case that I've observed from time to time (with the worst case scenario being the status quo, i.e. a full rebuild of the block tree).

Filing as a draft, as it uses an external branch until the snarkVM changes are merged, plus that one edge case.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,9 +45,8 @@ version = "1.0.1"
 default-features = false
 
 [workspace.dependencies.snarkvm]
-#path = "../snarkVM"
-git = "https://github.com/ProvableHQ/snarkVM.git"
-rev = "fbd99608f"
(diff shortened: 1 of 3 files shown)


vicsn:
Amazing! @ljedrz let @meddle0x53 know when this is ready for him to give the mainnet ledger load test another try.

ljedrz:
Superseded by https://github.com/ProvableHQ/snarkOS/pull/4021.

PR #4019, opened 2025-11-20
[Build] Fix the scope of the "serial" feature

ljedrz:
The `serial` feature, useful e.g. for profiling, needs to involve more crates.


vicsn (approved):

kaimast:
Good catch! I thought `cli` enables the serial feature on all these crates, but it does not seem to.

PR #4021, opened 2025-11-21
[Fix] Drop the Node when shutting down

ljedrz:
This is a temporary solution that allows the `Node` to drop correctly when it's being shut down. It's a successor to https://github.com/ProvableHQ/snarkOS/pull/4017, which is no longer needed.

This will likely be superseded by https://github.com/ProvableHQ/snarkOS/pull/3915 in the future (as that one would also be able to support tests), but for regular node use, this should suffice.

### DIFF
--- a/cli/src/commands/start.rs
+++ b/cli/src/commands/start.rs
@@ -57,6 +57,7 @@ use std::{
 };
 use tokio::{
     runtime::{self, Runtime},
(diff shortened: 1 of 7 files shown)


kaimast (approved):
Thanks. This is a great first step to clean up shutdown.

Issue #4026, opened 2025-11-27
[Feature] Upgrade nodes in CI tests


vicsn:
## 🚀 Feature

This issue is essentially moving a test from `snarkos-stress-testing` into our CI.

In order to test backwards compatibility, we could upgrade our `run_devnet` tests in `.circleci/config.yml` with a new simple script which does the following:
1. download the latest snarkOS release: https://github.com/ProvableHQ/snarkOS/releases
2. start a devnet with that latest release
3. wait until the highest consensus version is reached
4. restart the nodes one by one to the version of the PR, with 60 seconds delay in between
5. wait until the highest consensus version is reached again

Note: for this to work, we'll have to adjust:
```
            cargo install --locked --path . --features test_network
```

To set an appropriate `CONSENSUS_VERSION_HEIGHTS` environment variable where the last consensus version is (2*4*60/5)= 100 blocks higher. (computed as: `REDUNDANCY_FACTOR * NUM_VALIDATORS * WAIT_BETWEEN_UPGRADES / EXPECTED_BLOCKTIME)`. We can set the last consensus version manually in `.circleci/config.yml`.

PR #4029, opened 2025-11-28
Upgrade nodes in CI tests

meddle0x53:
## Motivation

To test upgrading snarkos versions in a network we run a special denvet test as described in #4026

## Test Plan

Run the CI and the new test passes.

## Documentation

The test works following the steps:

1. Download latest snarkOS release source from GitHub.
2. Build it via `cargo install --locked --path . --features test_network` into a separate prefix (SNARKOS_RELEASE_DIR).
3. Probe latest_consensus_version for release and PR binaries.
4. Compute CONSENSUS_VERSION_HEIGHTS for release and PR.
5. Rebuild release snarkOS with its heights.
6. Rebuild PR snarkOS with its heights.
7. Start a devnet with the release binary.
8. Restart nodes one-by-one to the PR binary.

Added a new workflow for the new test, extracting it from the devnet tests, as it is time-taking with all the compilations and waiting on the consensus version.

### DIFF
--- a/.ci/upgrade_nodes_ci.sh
+++ b/.ci/upgrade_nodes_ci.sh
@@ -1,0 +1,660 @@
+#!/bin/bash
+
+####################################################
+# Backwards-compatibility upgrade test:
+#
+# 1. Download latest snarkOS release source from GitHub.
+# 2. Build it via `cargo install --locked --path . --features test_network`
+#    into a separate prefix (SNARKOS_RELEASE_DIR).
+# 3. Probe latest_consensus_version for release and PR binaries.
+# 4. Compute CONSENSUS_VERSION_HEIGHTS for release and PR.
+# 5. Rebuild release snarkOS with its heights.
+# 6. Rebuild PR snarkOS with its heights.
+# 7. Start a devnet with the release binary.
+# 8. Restart nodes one-by-one to the PR binary.
+####################################################
+
+set -eo pipefail  # error on any command failure
+
+# --- Parameters from CLI ---
+total_validators=$1
+total_clients=$2
+network_id=$3
+
+# Default values if not provided
+: "${total_validators:=4}"
+: "${total_clients:=2}"
+: "${network_id:=0}"
+
+# Node verbosity
+NODE_VERBOSITY=1
+
+# Expected highest consensus version (may be overridden after probing)
+EXPECTED_MAX_CONSENSUS_VERSION="${EXPECTED_MAX_CONSENSUS_VERSION:-}"
+
+# How long to wait between upgrades (seconds); used for block-height window
+WAIT_BETWEEN_UPGRADES="${WAIT_BETWEEN_UPGRADES:-60}"
+
+# Load shared helpers (is_integer, get_network_name, wait_for_nodes, stop_nodes, ...)
+# shellcheck disable=SC1091
+. ./.ci/utils.sh
+
+########################################
+# Logging helpers
+########################################
+
+# Log a message to the console.
+function log() {
+  echo "[$(date '+%Y-%m-%d %H:%M:%S')] $*"
+}
+
+# Checks that the given command is available in the PATH.
+function require_cmd() {
+  if ! command -v "$1" >/dev/null 2>&1; then
+    echo "ERROR: required command '$1' not found in PATH" >&2
+    exit 1
+  fi
+}
+
+# Reuse the same target dir for all builds (release + PR) to get incremental builds.
+export CARGO_TARGET_DIR="${CARGO_TARGET_DIR:-$PWD/.ci/target}"
+log "Using CARGO_TARGET_DIR=${CARGO_TARGET_DIR}"
+
+##################################################
+# Download + build latest stable snarkOS release
+##################################################
+
+# Release binary will be installed here via `cargo install --root`.
+SNARKOS_RELEASE_DIR="${SNARKOS_RELEASE_DIR:-$PWD/.ci/release-snarkos}"
+SNARKOS_RELEASE_BIN="${SNARKOS_RELEASE_BIN:-$SNARKOS_RELEASE_DIR/bin/snarkos}"
+SNARKOS_RELEASE_VERSION_FILE="${SNARKOS_RELEASE_VERSION_FILE:-$SNARKOS_RELEASE_DIR/VERSION}"
+
+download_and_build_latest_snarkos() {
+  require_cmd curl
+  require_cmd tar
+  require_cmd cargo
+
+  local force_build="${1:-0}"
+
+  mkdir -p "${SNARKOS_RELEASE_DIR}"
+
+  local repo="ProvableHQ/snarkOS"
+  local latest_url latest_tag tar_url
+  local tmpdir srcdir
+  local existing_tag=""
+
+  log "Resolving latest snarkOS release tag via redirect…"
+  latest_url="$(
+    curl -fsSL -o /dev/null -w '%{url_effective}' \
+      "https://github.com/${repo}/releases/latest"
+  )" || {
+    echo "ERROR: Failed to resolve latest release URL from GitHub." >&2
+    exit 1
+  }
+
+  latest_tag="${latest_url##*/}"
+
+  if [ -z "${latest_tag}" ] || [ "${latest_tag}" = "latest" ]; then
+    echo "ERROR: Failed to determine latest tag from URL: ${latest_url}" >&2
+    exit 1
+  fi
+
+  log "Latest stable tag resolved to: ${latest_tag}"
+
+  ########################################
+  # Cached binary exists & matches version
+  #         AND force_build = 0 → return early
+  ########################################
+  if [ "$force_build" != "1" ] &&
+     [ -x "${SNARKOS_RELEASE_BIN}" ] &&
+     [ -f "${SNARKOS_RELEASE_VERSION_FILE}" ]; then
+
+    existing_tag="$(cat "${SNARKOS_RELEASE_VERSION_FILE}" || true)"
+
+    if [ "${existing_tag}" = "${latest_tag}" ]; then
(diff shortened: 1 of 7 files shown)


.ci/upgrade_nodes_ci.sh
```diff
@@ -0,0 +1,417 @@
...
+# Default values if not provided
+: "${total_validators:=4}"
+: "${total_clients:=2}"
+: "${network_id:=0}"
+
+# Node verbosity
+NODE_VERBOSITY=1
+
+# Expected highest consensus version (optional, but recommended in CI)
+EXPECTED_MAX_CONSENSUS_VERSION="${EXPECTED_MAX_CONSENSUS_VERSION:-}"
```
vicsn, line 33: Idea: how about we return the latest consensus version from the /version endpoint?

This would require us to add a `latest_consensus_version` member to `VersionInfo`, set via `ConsensusVersion::latest()`.

I think this should be backwards compatible - because JSON parsers typically just ignore unexpected fields?

This would simplify the `wait_for_highest_consensus_version` logic in all of the scripts in this `snarkOS/.ci` folder.

meddle0x53: Yeah, sounds great, the scripts can use it, please if there is an issue that explains how to implement it (for a guy that is not so deep in the code), assign it to me and I can do it. 

I guess is just adding the latest consensus version to this struct that is then serialised to JSON. But where to get the latest consensus version from (in the code). I think it was defined somewhere in SnarkVM?

vicsn: I think this is small enough that there's no need for a new issue and you can include it in this PR.

The `enum ConsensusVersion` is defined in snarkVM, and it exposes a `pub fn latest()`

Let me know if you have any other questions.

.ci/upgrade_nodes_ci.sh
```diff
@@ -0,0 +1,595 @@
...
+  node_index=$((client_index + total_validators))
+  log_file="$log_dir/client-$client_index.log"
+  start_node "$SNARKOS_RELEASE_BIN" "$node_index" "client" "$log_file"
+  if (( client_index < total_clients-1 )); then
+    sleep 1
+  fi
+done
+
+wait_for_nodes "$total_validators" "$total_clients"
+wait_for_highest_consensus_version 900 10
```
vicsn, line 626: We are calling `wait_for_highest_consensus_version` both before and after the 1-by-1 upgrades. 

Is `EXPECTED_MAX_CONSENSUS_VERSION` set in both cases? Is it set to the same value? Should it be?

.ci/upgrade_nodes_ci.sh
```diff
@@ -0,0 +1,595 @@
...
+  if [ -z "$json" ]; then
+    log "WARN: Could not obtain version info from ${label} snarkos."
+    return 1
+  fi
+
+  echo "$json"
+  return 0
+}
+
+# Build heights string: for L, we want exactly L entries: 0,5,10,...,5*(L-1)
```
vicsn, line 342: If "current" has a new consensus version compared to "release", shouldn't the last height be, say, `5*(L-1) + 100` so that the nodes have time to update before hitting the new version?

.ci/upgrade_nodes_ci.sh
meddle0x53: Interesting question... So in the first case it is set to the value of the release version. In the second case - after the test, I think it is a forgotten bit form my local iterations. Will remove it.

We need to get it for the current branch snarkos to set its heights list, which we do, otherwise we don't wait on it, as we don't start anything after that.

.ci/upgrade_nodes_ci.sh
meddle0x53: Hmmm, that makes sense, right... Let me change and test that.

meddle0x53: Forgot to write here, but updated.

.ci/upgrade_nodes_ci.sh
meddle0x53: Removed

vicsn (dismissed):
LGTM, let's await Kai's review

.ci/upgrade_nodes_ci.sh
```diff
@@ -0,0 +1,639 @@
...
+done
+trusted_peers=${trusted_peers%,}
+
+start_node() {
+  local bin="$1"
+  local node_index="$2"
+  local role="$3"    # "validator" or "client"
+  local log_file="$4"
+
+  local flags=( "${common_flags[@]}" "--dev=$node_index" )
```
vicsn, line 221: We need to be able to support passing in custom flags when new required flags are added in new releases  CC https://github.com/ProvableHQ/snarkOS/pull/4033

Also, it might be easier to read if we just pass a list of flags to `start_node` like we do in other tests. Currently it is abstraction on top of abstraction (`common flags`, `role`, etc.). CC @kaimast what do you think?

kaimast:
Looks good to me! I pushed one more commit with some minor cleanups ([83355c6](https://github.com/ProvableHQ/snarkOS/pull/4029/commits/83355c6e11a2f431e856c371922909b170a95b37)).

I also rebased https://github.com/ProvableHQ/snarkOS/pull/3902 on it. That PR has some networking changes and the tests succeed.  We might still want to test it with a branch that has breaking changes to see that those are caught, or did you do that already?

Also, a side note, it would be great if we could keep the number of commits in PRs to a manageable amount. This one has 14 commits, for ~700 LOC. It's not a huge deal, but it can be a pain when reading the commit logs and or when using `git bisect`.

meddle0x53:
@kaimast @vicsn Squashed all into one commit now.

vicsn:
@kaimast good idea. You want to give it another go with a toy PR which changes the gateway handshake in a breaking way?

vicsn:
> We might still want to test it with a branch that has breaking changes to see that those are caught, or did you do that already?

@meddle0x53 can you try on a toy commit which makes a breaking change to both `struct ChallengeRequest` objects?

meddle0x53:
Yeah if I update `ChallengeRequest` the test fails and I get:

```
2026-01-07T13:03:59.066676Z DEBUG [MemoryPool] Shaking hands with 127.0.0.1:5003...
2026-01-07T13:03:59.066786Z DEBUG [MemoryPool] Shaking hands with 127.0.0.1:5002...
2026-01-07T13:03:59.066822Z DEBUG [MemoryPool] Shaking hands with 127.0.0.1:5001...
2026-01-07T13:03:59.068359Z  WARN [MemoryPool] Unable to connect to '127.0.0.1:5002' - I/O error: [MemoryPool] the peer disconnected before sending "Event::ChallengeResponse", likely due to peer saturation or shutdown
2026-01-07T13:03:59.068460Z  WARN [MemoryPool] Unable to connect to '127.0.0.1:5003' - I/O error: [MemoryPool] the peer disconnected before sending "Event::ChallengeResponse", likely due to peer saturation or shutdown
2026-01-07T13:03:59.068884Z  WARN [MemoryPool] Unable to connect to '127.0.0.1:5001' - I/O error: [MemoryPool] the peer disconnected before sending "Event::ChallengeResponse", likely due to peer saturation or shutdown
2026-01-07T13:04:00.579279Z DEBUG Skipping batch proposal for round 112 (node is syncing)
2026-01-07T13:04:01.410926Z  INFO Received 'GET /block/height/latest' from '127.0.0.1:54017'
2026-01-07T13:04:03.081466Z DEBUG Skipping batch proposal for round 112 (node is syncing)
2026-01-07T13:04:05.583033Z DEBUG Skipping batch proposal for round 112 (node is syncing)
2026-01-07T13:04:06.454622Z  INFO Received 'GET /block/height/latest' from '127.0.0.1:54020'
2026-01-07T13:04:08.085138Z DEBUG Skipping batch proposal for round 112 (node is syncing)
2026-01-07T13:04:10.587016Z DEBUG Skipping batch proposal for round 112 (node is syncing)
2026-01-07T13:04:11.504465Z  INFO Received 'GET /block/height/latest' from '127.0.0.1:54026'
2026-01-07T13:04:13.089705Z DEBUG Skipping batch proposal for round 112 (node is syncing)
2026-01-07T13:04:14.067889Z  INFO No connected validators
2026-01-07T13:04:14.068108Z DEBUG   Not connected to aleo1ashyu96tjwe63u0gtnnv8z5lhapdu4l5pjsl2kha7fv7hvz2eqxs5dz0rg (25.00% of total stake)
2026-01-07T13:04:14.068159Z DEBUG   Not connected to aleo12ux3gdauck0v60westgcpqj7v8rrcr3v346e4jtq04q7kkt22czsh808v2 (25.00% of total stake)
2026-01-07T13:04:14.068205Z DEBUG   Not connected to aleo1s3ws5tra87fjycnjrwsjcrnw2qxr8jfqqdugnf0xzqqw29q9m5pqem2u4t (25.00% of total stake)
2026-01-07T13:04:14.068246Z  WARN Not connected to 3 validators (of 3 bonded validators) (75.00% of total stake not connected)
```

Logs like this, @vicsn

.circleci/config.yml
```diff
@@ -682,6 +742,10 @@ workflows:
     jobs:
       - devnet-test
 
+  upgrade-workflow:
+    jobs:
+      - upgrade-test
```
vicsn: Given that it's a bit more expensive and unlikely to be an issue for most PRs, can we make sure to only run it on merges? See for example: https://github.com/ProvableHQ/snarkVM/pull/3078/commits/b5ddc0cb0f66cd3d34c042a98c734393fcb17b27

You can also move the `verify-windows` job in there to only run on merges

meddle0x53: Done and squashed!

vicsn: Can you add the `canary`, `testnet` and `mainnet` branches too? Thx!

meddle0x53: Done!

vicsn (approved):

PR #4031, opened 2025-12-01
[Fix] Allow handshake with peers that do not send commit hash

kaimast:
When testing local builds, I noticed I wasn't able to connect to some canary bootstrap peers because they are not sending the snarkOS commit hash.

This PR ensures handshake still succeed if no, or an invalid, commit hash is sent. We can reject such connections in the future, but, as of now, there are still too many nodes that do not send it.


ljedrz (approved):
LGTM

vicsn:
@kaimast great find. Should we backport this to canary/testnet/mainnet ? Validators don't necessarily have to update, but new entrants running their own nodes do. I can make the tags if you make the PRs.

PR #4032, opened 2025-12-02
[Feat] Add 2 new REST routes

ljedrz:
The snarkOS counterpart to https://github.com/ProvableHQ/snarkVM/pull/3047.

Filing as a draft due to the corresponding snarkVM branch not having been merged yet.

### DIFF
--- a/node/rest/src/routes.rs
+++ b/node/rest/src/routes.rs
@@ -936,4 +936,21 @@ impl<N: Network, C: ConsensusStorage<N>, R: Routing<N>> Rest<N, C, R> {
             None => Err(RestError::service_unavailable(anyhow!("Route isn't available for this node type"))),
         }
     }
+
+    /// GET /{network}/record_count
+    pub(crate) async fn get_number_of_records(State(rest): State<Self>) -> ErasedJson {
+        let (input, output) = rest.ledger.get_record_count();
+        ErasedJson::pretty(format!("input: {input}, output: {output}"))
+    }
+
+    /// GET /{network}/{block}/record_count
+    pub(crate) async fn get_number_of_block_records(
+        State(rest): State<Self>,
+        Path(height): Path<u32>,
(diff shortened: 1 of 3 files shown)


node/rest/src/routes.rs
```diff
@@ -936,4 +936,21 @@ impl<N: Network, C: ConsensusStorage<N>, R: Routing<N>> Rest<N, C, R> {
...
         }
     }
+
+    /// GET /{network}/record_count
+    pub(crate) async fn get_number_of_records(State(rest): State<Self>) -> ErasedJson {
+        let record_counts = rest.ledger.get_record_count();
+        ErasedJson::pretty(record_counts)
+    }
+
+    /// GET /{network}/{block}/record_count
```
niklaslong, line 946:
```suggestion
    /// GET /{network}/block/{block}/record_count
```

node/rest/src/routes.rs
```diff
@@ -936,4 +936,21 @@ impl<N: Network, C: ConsensusStorage<N>, R: Routing<N>> Rest<N, C, R> {
             None => Err(RestError::service_unavailable(anyhow!("Route isn't available for this node type"))),
         }
     }
+
+    /// GET /{network}/record_count
+    pub(crate) async fn get_number_of_records(State(rest): State<Self>) -> ErasedJson {
```
raychu86, line 941: This endpoint will be quite expensive, can we put it behind auth to prevent DoSing nodes? - https://github.com/ProvableHQ/snarkVM/pull/3047#discussion_r2590633826

ljedrz: as mentioned in the linked thread, this operation is in general less expensive than just obtaining a block; what should the criteria for auth-gating be? some fixed amount of time on reference hardware, or just algorithmic complexity (even if it's unlikely to be a concern until very large numbers are reached)?

PR #4033, opened 2025-12-03
[Feature] Filesystem reorganization

### LINKED ISSUE #4013: [Proposal] Make storage to filesystem less error-prone
## 💥 Proposal

Currently, the ledger folder contains the following (when using the snarkos start `--storage` flag`:
1. public ledger data
2. `.current-proposal-cache*`
3. `cached_*_peers`
4. `jwt_secret_*.txt`

Issues:
- if a node operator recovers from a local or remote snapshot, they are likely to overwrite the entire ledger folder
- if you don't use `--storage`, `StorageMode::Production` is used, causing more confusingly different behaviour: the `.current-proposal-cache*` will then be created in the parent directory of the ledger.

Proposal:
- We create a new `--storage-local-state` flag (name t.b.d.) which stores 2./3./4. , making it explicit these cannot be recovered without backup, and should not be overwritten by an external snapshot
- We ensure both `StorageMode::Production` and `StorageMode::Custom` write the appropriate files to this new folder.
- Consider renaming `current-proposal-cache` to e.g. `.current-active-proposal`

Design goals:
- Need to manually test all different storage modes
- Backwards compatible?
  - Only make the new flag **required** for validators from the next release
  - If you don't pass it, log a warning but use the old way of doing things

kaimast:
Fixes #4013.

This PR adds a new `node-data` directory that contains all node-specific data (essentially everything but the ledger). This is located at `~/.aleo/storage/node-data-{network_id}` for production or `./node-data-{network_id}-{dev}` for development nodes. Alternatively, users can also pass a value to `--node-data` and store the data at a custom path.

At startup, a node will check for any files that indicate that the old storage format is used. In that case, a message is printed in how to migrate, and the node will not start. Additionally, the FAQ in the README contains information on how to migrate.

Alternatively, nodes can also be started with `--auto-migrate-node-data` to do this automatically. This is not enabled by default out of extra precaution.


vicsn:
> We need to either provide a migration tool (e.g., snarkos migrate) or a document on how to move to the new layout.

I think documentation in the snarkOS README, release and release blog should be sufficient. `.current-proposal-cache*`, `cached_*_peers` and `jwt_secret_*.txt` are all written at every startup/shutdown.

> I picked the name node-config as a placeholder, and we can consider changing it.

I think this may not reflect how important it is for validators to keep `.current-proposal-cache*`. Please bring some alternatives to the next protocol weekly sync.

> Should nodes check for the new directory, so that even if --node-config-dir is not passed, the new layout is used?

Given that backups are essentially required are for validators, what about we make it a **mandatory** flag for `--validator` nodes so they're forced to think about it. I will be very clear about it in the next release.

cli/src/helpers/args.rs
```diff
@@ -74,6 +78,49 @@ pub(crate) fn parse_private_key<N: Network>(
...
 
+/// Returns the path to the node configuration directory.
+pub(crate) fn parse_node_config_dir(
+    node_config_dir: &Option<Option<PathBuf>>,
+    network: u16,
+    dev: Option<u16>,
+) -> Result<NodeConfigDir> {
+    let new_layout = match node_config_dir {
+        Some(Some(custom)) => return Ok(NodeConfigDir::new(custom.clone())),
+        Some(None) => true,
```
vicsn: What is this, when someone passes `--node-config-dir` without any arguments? I think allowing for this will be confusing.

utilities/src/lib.rs
```diff
@@ -17,3 +17,60 @@
...
+        Self { path, legacy: true, dev: None }
+    }
+
+    pub fn new_legacy_dev(path: PathBuf, dev: u16) -> Self {
+        Self { path, legacy: true, dev: Some(dev) }
+    }
+
+    pub fn new_development(dev: u16) -> Self {
+        Self { path: PathBuf::from(format!(".node-config-dev-{dev}")), legacy: false, dev: Some(dev) }
+    }
```
vicsn: That's a lot of options, would this be more succinct and easier to read if we pass in `StorageMode` directly?

kaimast: My hope was that we can remove the legacy layout the release after the next, and not maintain both nodes forever.

cli/src/helpers/args.rs
kaimast: It allows you to enable the new layout without specifying a custom directory. I can maybe add a new flag instead? That one would also be removed in the release after the next or so, when stop supporting the layout.

utilities/src/lib.rs
kaimast: I also considered having a unified `StorageMode` instead and have it refer to the parent directory of the ledger directory instead, e.g. `~/.aleo/storage/` or `~/.aleo/storage-0` instead of `~/.aleo/storage/ledger-0`. That would be a breaking change, though, but make things easier in the long run.

To clarify, the breaking change would be that when you pass a custom storage path, snarkOS would then append `ledger` to it for the rocksDB storage, which does not happen in the current implementation.

cli/src/helpers/args.rs
vicsn: I would not support the legacy format. Instead:
- when passing the --node-config-dir <path> it will be set to the path. It is mandatory for validators to set and therefore think about migration. For clients the folder contents are not that critical and nobody will care in the near-term.
- if you don't use the flag, use a hardcoded default path.

vicsn: @raychu86 does the above sound ok? I hope the PR README gives enough context. Naming is t.b.d.

vicsn: Result of discussion: 
- force use of new layout, no need for a new flag.
- keep ledger data in existing place, config data to a new subdirectory. Fail to start if we can't write to the desired file/folder.

utilities/src/lib.rs
kaimast: All legacy support was removed from the PR, as we discussed.

README.md
```diff
@@ -272,6 +272,20 @@ APrivateKey1xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
 
 Note, using the `--raw` flag with the command will sign plaintext messages as bytes rather than [Aleo values](https://developer.aleo.org/guides/aleo/language#data-types-and-values) such as `1u8` or `100field`.
 
+### 5. What the different between `node-data` and `ledger`?
```
vicsn: What's

cli/src/commands/clean.rs
```diff
@@ -35,29 +38,44 @@ pub struct Clean {
     /// dev).
     #[clap(long = "path")]
     pub path: Option<PathBuf>,
+
+    /// Sets a custom path for the node configuration.
+    #[clap(long)]
+    pub node_config_dir: Option<PathBuf>,
```
vicsn: Would `node_data_path` be more consistent with the existing `path` flag?

cli/src/commands/start.rs
```diff
@@ -237,6 +242,14 @@ pub struct Start {
     #[clap(long)]
     pub storage: Option<PathBuf>,
 
+    /// Set a custom path for the node-specific data.
+    /// This flag overrides the default path, even when `--dev` is set.
+    ///
+    /// That folder may contain sensitive data, such as the JWT secret, and should not be shared with untrusted parties.
+    /// For validators, it also contains the latest proposal cache, which is required to participate in consensus.
+    #[clap(long, verbatim_doc_comment)]
+    pub node_data: Option<PathBuf>,
```
vicsn: Revisiting a naming discussion, what about we name the flags `legder_storage` and `node_data_storage`. We keep `storage` as alias but hide it from documentation where possible and we can deprecate it in the future.

cli/src/commands/start.rs
```diff
@@ -695,13 +719,31 @@ impl Start {
...
 
-        // Helper function to store the JWT secret.
-        let store_jwt_secret = |network: u16, storage_mode: &StorageMode, address: &Address<N>, token: String| -> Result<()> {
-            let mut jwt_secret_path = aleo_std::aleo_ledger_dir(network, storage_mode);
-            std::fs::create_dir_all(&jwt_secret_path)?;
-            jwt_secret_path.push(format!("jwt_secret_{address}.txt"));
-            Ok(std::fs::write(jwt_secret_path, token)?)
-        };
+        if self.node_data.is_some() && !matches!(storage_mode, StorageMode::Custom(_)) {
+            warn!("Custom path set for `--storage`, but not for `--node-data`. The latter will use the default path.");
```
vicsn: Shall we just enforce via Clap that both flags are either set or not set together? Node operators will miss this log and not be aware of the implications.

cli/src/commands/start.rs
```diff
@@ -695,13 +719,31 @@ impl Start {
...
-            Ok(std::fs::write(jwt_secret_path, token)?)
-        };
+        if self.node_data.is_some() && !matches!(storage_mode, StorageMode::Custom(_)) {
+            warn!("Custom path set for `--storage`, but not for `--node-data`. The latter will use the default path.");   
+        } else if matches!(storage_mode, StorageMode::Custom(_)) && self.node_data.is_none() {
+            warn!("Custom path set for `--storage`, but `--node-data` is not set. The latter will use the default path.");
+        }
+
+        // Parse the node data directory.
+        let node_data_dir = parse_node_data_dir(&self.node_data, N::ID, self.dev).with_context(|| "Failed to setup node configuration directory")?;
```
vicsn: The ledger variable is called storage_mode. Can we make the naming and for the two different folders more consistent?

Should/could we use another instance of StorageMode in NodeData?

node/bft/src/bft.rs
```diff
@@ -99,7 +101,7 @@ impl<N: Network> BFT<N> {
         ip: Option<SocketAddr>,
         trusted_validators: &[SocketAddr],
         trusted_peers_only: bool,
-        storage_mode: StorageMode,
+        node_config_dir: NodeDataDir,
```
vicsn: Grep for 'config' throught the PR diff to fix

cli/src/commands/start.rs
```diff
@@ -795,6 +839,88 @@ impl Start {
...
+        address: &Address<N>,
+        node_data_dir: &NodeDataDir,
+        dev: Option<u16>,
+        auto_migrate: bool,
+    ) -> Result<()> {
+        let ledger_parent_dir = ledger_dir.parent().unwrap();
+
+        // Determine the old paths used for node configuration files.
+        let old_router_cache_path = ledger_dir.join(node_data::LEGACY_ROUTER_PEER_CACHE_FILE);
+        let old_gatway_cache_path = ledger_dir.join(node_data::LEGACY_GATWAY_PEER_CACHE_FILE);
```
vicsn: GATWAY

utilities/src/node_data.rs
```diff
@@ -0,0 +1,102 @@
...
+}
+
+impl NodeDataDir {
+    pub fn new(path: PathBuf) -> Self {
+        Self { path, dev: None }
+    }
+
+    pub fn new_test(dev: Option<u16>) -> Self {
+        if let Some(dev) = dev {
+            Self { path: PathBuf::from(format!(".node-config-test-{dev}")), dev: Some(dev) }
```
vicsn: See previous note about grepping for config

utilities/src/node_data.rs
```diff
@@ -0,0 +1,102 @@
...
+        } else {
+            Self { path: PathBuf::from(".node-config-test"), dev: None }
+        }
+    }
+
+    pub fn new_development(dev: u16) -> Self {
```
vicsn: Do we have to add the {network} too? What about for the production node data dir?

What do we do for the ledger folder and should we be consistent with it?

vicsn (changes requested):
Almost there!

cli/src/commands/start.rs
vicsn: Correction: only make it required for validators. So I guess we can stick with the current if statements.

vicsn:
Note this may need a minor merge to include https://github.com/ProvableHQ/snarkOS/pull/4054 once it's ready

cli/src/commands/start.rs
kaimast: Fixed

kaimast:
> Note this may need a minor merge to include #4054 once it's ready

[c8f5891](https://github.com/ProvableHQ/snarkOS/pull/4033/commits/c8f589170a0db83984dc4499b1efbb6e8afcd3ff) merged staging include #4054's changes.

cli/src/commands/start.rs
kaimast: Yeah, I was not sure if it makes sense to only specify a custom path for one in some cases. The warning is just out of additional caution.

cli/src/commands/start.rs
kaimast: They are slightly different. `storage_mode` is eventually passed to `aleo_std::aleo_ledger_dir` to retrieve the ledger path. `NodeDataDir` already contains the node data path.

utilities/src/node_data.rs
kaimast: Added. This function was only used in one of the tests, but it is working correctly now and is also used in `start.rs`.

cli/src/commands/start.rs
vicsn: Why did you resolve this? I suggested we make the flags depending on each other for validators and do not see argumentation

cli/src/commands/start.rs
vicsn: Can you not resolve comments when there's potential disagreement?

Would it be possible to re-use StorageMode and to add another node_data method in the aleo-std repo instead? It would make reading and maintaining the code a lot easier.

kaimast: That would make sense. So far, I have avoided changing `StorageMode` as that is also used in snarkVM.

Initially, my idea was that we would have `StorageMode` point to the parent directory of the two folders, e.g., `~/.aleo/storage`. This way, it would be straightforward to determine where they are located and there would only be a single command line argument to set the storage directory.
Such a change would be better long-term, IMHO, but it does involve breaking small parts of the API and CLI.

cli/src/commands/start.rs
```diff
@@ -488,6 +505,13 @@ impl Start {
         Ok(())
     }
 
+    /// Returns the path to where the JWT secret for the node is stored.
+    fn jwt_secret_path<N: Network>(node_data_dir: &NodeDataDir, address: &Address<N>) -> Result<PathBuf> {
+        let mut path = node_data_dir.path().to_path_buf();
+        path.push(format!("jwt_secret_{address}.txt"));
+        Ok(path)
+    }
```
```suggestion
    fn jwt_secret_path<N: Network>(node_data_dir: &NodeDataDir, address: &Address<N>) -> PathBuf {
        let mut path = node_data_dir.path().to_path_buf();
        path.push(format!("jwt_secret_{address}.txt"));
        path
    }
```

cli/src/commands/start.rs
```suggestion
            warn!("Custom path set for `--node-data`, but not for `--storage`. The latter will use the default path.");   
```

cli/src/commands/start.rs
```diff
@@ -781,9 +825,9 @@ impl Start {
 
         // Initialize the node.
         let node = match node_type {
-            NodeType::Validator => Node::new_validator(node_ip, self.bft, rest_ip, self.rest_rps, account, &trusted_peers, &trusted_validators, genesis, cdn, storage_mode, self.trusted_peers_only, dev_txs, self.dev, signal_handler.clone()).await,
-            NodeType::Prover => Node::new_prover(node_ip, account, &trusted_peers, genesis, storage_mode, self.trusted_peers_only, self.dev, signal_handler.clone()).await,
-            NodeType::Client => Node::new_client(node_ip, rest_ip, self.rest_rps, account, &trusted_peers, genesis, cdn, storage_mode, self.trusted_peers_only, self.dev, signal_handler.clone()).await,
+            NodeType::Validator => Node::new_validator(node_ip, self.bft, rest_ip, self.rest_rps, account, &trusted_peers, &trusted_validators, genesis, cdn,  storage_mode, node_data_dir, self.trusted_peers_only, dev_txs, self.dev, signal_handler.clone()).await,
```
```suggestion
            NodeType::Validator => Node::new_validator(node_ip, self.bft, rest_ip, self.rest_rps, account, &trusted_peers, &trusted_validators, genesis, cdn, storage_mode, node_data_dir, self.trusted_peers_only, dev_txs, self.dev, signal_handler.clone()).await,
```
`rustfmt` didn't get this one?

cli/src/commands/start.rs
```diff
@@ -795,6 +839,88 @@ impl Start {
...
+            if old_jwt_secret_path.exists() {
+                let new_jwt_secret_path = node_data_dir.jwt_secret_path(&address);
+                info!("Migrating node data file \"{old_jwt_secret_path:?}\" to \"{new_jwt_secret_path:?}\"");
+                fs::rename(old_jwt_secret_path, new_jwt_secret_path)
+                    .with_context(|| "Failed to migrate node data file")?;
+            }
+        } else {
+            if old_router_cache_path.exists() {
+                let new_router_cache_path = node_data_dir.router_peer_cache_path();
+                bail!(
```
ljedrz, line 902: nit: I'd prefer to flag in case of any legacy folders, log it, and only bail once all of them have been reported, but this works too

cli/src/helpers/args.rs
```diff
@@ -74,6 +77,27 @@ pub(crate) fn parse_private_key<N: Network>(
     Ok(private_key)
 }
 
+/// Returns the path to the node configuration directory.
+pub(crate) fn parse_node_data_dir(
+    node_data_dir: &Option<PathBuf>,
+    network: u16,
+    dev: Option<u16>,
+) -> Result<NodeDataDir> {
+    if let Some(custom) = node_data_dir {
```
ljedrz:
```suggestion
    custom_node_data_dir: &Option<PathBuf>,
    network: u16,
    dev: Option<u16>,
) -> Result<NodeDataDir> {
    if let Some(custom) = custom_node_data_dir {
```
nit: greater clarity

node/bft/src/gateway.rs
```diff
@@ -848,7 +842,7 @@ impl<N: Network> Gateway<N> {
     pub async fn shut_down(&self) {
         info!("Shutting down the gateway...");
         // Save the best peers for future use.
-        if let Err(e) = self.save_best_peers(&self.storage_mode, VALIDATOR_CACHE_FILENAME, None, true) {
+        if let Err(e) = self.save_best_peers(&self.node_data_dir.gateway_peer_cache_path(), None, false) {
```
ljedrz:
```suggestion
        if let Err(e) = self.save_best_peers(&self.node_data_dir.gateway_peer_cache_path(), None, true) {
```
the cached peers (both `Router` and `Gateway`) are expected to have port numbers

node/bft/src/helpers/proposal_cache.rs
```diff
@@ -79,15 +65,15 @@ impl<N: Network> ProposalCache<N> {
             && self.signed_proposals.is_valid(expected_signer)
     }
 
-    /// Returns `true` if a proposal cache exists for the given network and `dev`.
-    pub fn exists(storage_mode: &StorageMode) -> bool {
-        proposal_cache_path(N::ID, storage_mode).exists()
+    /// Returns `true` if a proposal cache  given network and `dev`.
```
ljedrz:
```suggestion
    /// Returns `true` if a proposal cache exists for the given network and `dev`.
```

node/bft/src/helpers/proposal_cache.rs
```diff
@@ -79,15 +65,15 @@ impl<N: Network> ProposalCache<N> {
...
+    pub fn exists(node_data_dir: &NodeDataDir) -> bool {
+        proposal_cache_path(node_data_dir).exists()
     }
 
     /// Load the proposal cache from the file system and ensure that the proposal cache is valid.
-    pub fn load(expected_signer: Address<N>, storage_mode: &StorageMode) -> Result<Self> {
-        // Construct the proposal cache file system path.
-        let path = proposal_cache_path(N::ID, storage_mode);
+    pub fn load(expected_signer: Address<N>, node_data_dir: &NodeDataDir) -> Result<Self> {
+        // Construct the proposal cache file system pa
```
ljedrz:
```suggestion
        // Construct the proposal cache file system path.
```

utilities/src/node_data.rs
```diff
@@ -0,0 +1,105 @@
...
+pub const ROUTER_PEER_CACHE_FILE: &str = "router-peer-cache";
+/// The old filename of the router peer cache.
+pub const LEGACY_ROUTER_PEER_CACHE_FILE: &str = "cached_router_peers";
+
+/// The filename of the proposal cache.
+pub const CURRENT_PROPOSAL_CACHE_FILE: &str = "current-proposal-cache";
+/// The filename of the validator whitelist.
+pub const VALIDATOR_WHITELIST_FILE: &str = "validator-whitelist";
+
+/// The old filename of the JWT secret for a given address.
```
ljedrz:
```suggestion
/// The filename of the JWT secret for a given address.
```
seems like it's not legacy?

ljedrz (changes requested):
I like the general direction, there are just a few errors left to iron out, especially [this one](https://github.com/ProvableHQ/snarkOS/pull/4033#discussion_r2667510972).

cli/src/commands/start.rs
kaimast: Yeah, I am not sure why it didn't get formatted or caught in the git-hook.

cli/src/commands/start.rs
kaimast: I changed the flags accordingly

cli/src/commands/start.rs
kaimast: I might have misread your last comment... I changed it `bail!` for validators.

cli/src/commands/start.rs
```diff
@@ -719,10 +719,21 @@ impl Start {
             },
         };
 
+        // Users may have unintentionally set a custom path for the ledger, but not for the node data.
+        // For validators, we make this an errors, so important files like the proposal cache are stored at the location
+        // exepcted by the node operator.
         if self.node_data.is_some() && !matches!(storage_mode, StorageMode::Custom(_)) {
-            warn!("Custom path set for `--storage`, but not for `--node-data`. The latter will use the default path.");   
+            if node_type == NodeType::Validator {
+                bail!("Custom path set for `--storage`, but not for `--node-data`.")
```
ljedrz: the path for storage is _not_ custom here - note the `!` in front of `matches!`

ljedrz: ah, it was changed in a newer commit

cli/src/commands/start.rs
```diff
@@ -684,21 +708,50 @@ impl Start {
...
 
-        // Helper function to store the JWT secret.
-        let store_jwt_secret = |network: u16, storage_mode: &StorageMode, address: &Address<N>, token: String| -> Result<()> {
-            let mut jwt_secret_path = aleo_std::aleo_ledger_dir(network, storage_mode);
-            std::fs::create_dir_all(&jwt_secret_path)?;
-            jwt_secret_path.push(format!("jwt_secret_{address}.txt"));
-            Ok(std::fs::write(jwt_secret_path, token)?)
-        };
+        // Users may have unintentionally set a custom path for the ledger, but not for the node data.
+        // For validators, we make this an errors, so important files like the proposal cache are stored at the location
```
ljedrz, line 720:
```suggestion
        // For validators, we treat this as an error, so important files like the proposal cache are stored at the location
```

ljedrz (approved):
Left one last comment, but LGTM otherwise :+1:.

vicsn (approved):
LGTM!

PR #4034, opened 2025-12-03
[Logs] Do not show debug logs on lowest verbosity level

kaimast:
I noticed that even on the lowest verbosity setting (i.e., `snarkos start --verbosity=0`), nodes will print some debug logs. This can be too verbose when running a devnet with many validators, and it should be possible to disable all debug messages.

This PR changes the logging behavior so that debug logs for `gateway`, `bft`, and `sync` are only show with verbosity set to 1 or greater.

### DIFF
--- a/cli/src/helpers/logger.rs
+++ b/cli/src/helpers/logger.rs
@@ -45,21 +45,29 @@ fn parse_log_verbosity(verbosity: u8) -> Result<EnvFilter> {
     // Now, set rules for specific crates.
     let filter = if verbosity >= 2 {
         filter.add_directive("snarkos_node_sync=trace".parse().unwrap())
-    } else {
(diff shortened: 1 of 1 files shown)


raychu86:
Log changes seem reasonable to me from a development standpoint, but node operators have come to expect certain log outputs based on their current operations and verbosity settings. 

@vicsn any suggestions on who needs what? Or how to effectively communicate these types of changes?

kaimast:
Do node operators set a verbosity this low usually? I assumed most of them have it set to 3 or 4.

vicsn (approved):
LGTM, especially because it only influences what one sees when `verbosity=0` which is an opinionated choice for a node operator. The default is 1.

The change will be reflexted in release notes.

PR #4035, opened 2025-12-03
[CI] Ensure devnet does not generate errors

### LINKED ISSUE #4030: [Proposal] Test for the absence of errors in devnet logs
`.ci/devnet_ci.sh` should check that node logs contain no errors, except for, maybe, a few select ones. This will allow us catching potential problematic code easier in the future and avoid introducing ERROR messages in cases where a node is not actually encountering a serious problem.

Right now, the only error printed is `Not connected to a quorum of validators`. This one we could allow, or change it, so the error is not printed until the node has been running for a few minutes.

The implementation could be as easy as `grep "ERROR" *.log | wc -l`.

kaimast:
Fixes #4030.

This PR makes two changes:
* [c9267f9](https://github.com/ProvableHQ/snarkOS/pull/4035/commits/c9267f96b84522fe08926b49f2510dcfcbd7d2ac) avoids errors about lacking sufficient validator connections within the first minute of a node's uptime. As of now, validators will virtually always print this error at startup, potentially confusing users.
* [a3606d1](https://github.com/ProvableHQ/snarkOS/pull/4035/commits/a3606d122d77b11d508682caa814c086953a3a32) adds a simple check (using `grep`) that no errors were generated when running the devnet tests.

### DIFF
--- a/.ci/utils.sh
+++ b/.ci/utils.sh
@@ -75,7 +75,7 @@ function check_heights() {
   fi
 }
 
-# Function checking that nodes created logs on disk.
+# Function checking that nodes created logs on disk and they contain no errors.
 function check_logs() {
   echo "Checking logs exist for all nodes..."
   local log_dir=$1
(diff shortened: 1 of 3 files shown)


vicsn (approved):
Nice!

PR #4037, opened 2025-12-04
[Fix] Use correct edition when recursively fetching programs

### LINKED ISSUE #3758: [Bug] `snarkos developer execute` doesn't work under consensus V8

kaimast:
This PR replaces #4002 and fixes #3758.

### Changes
* [83757e8](https://github.com/ProvableHQ/snarkOS/pull/4037/commits/83757e887b8bf2f9cef5fb91fea3b030e0082e27) ensure the correct edition for each import is fetched. The current behavior only fetches the latest edition for the top-level program and then incorrectly reuses that value.
*  [c74b12d](https://github.com/ProvableHQ/snarkOS/pull/4037/commits/c74b12d9e240d242a08c56c5e6c4b10314b49d4e) improves a doc comment.
* [7ba6481](https://github.com/ProvableHQ/snarkOS/pull/4037/commits/7ba6481aee92e2764c1fba54748c0df92e2aa576) changes the default endpoint/base_url to `https://api.explorer.provable.com/v2/`. The `v2` component is needed (otherwise the explorer returns a 404 error) and refers to the explorer version, not the snarkOS API version.
* [2db8047](https://github.com/ProvableHQ/snarkOS/pull/4037/commits/2db8047e32ca9b93728a1039d55f0ad12100d0b0) adds a new flag `--skip-funds-check`, so you can test these changes without requiring actual funds in your account. This flag is only intended for testing purposes and marked as hidden.

### Testing
Ensure the following works to verify the correctness of these changes:
```
~> snarkos developer execute --private-key APrivateKey1zkp8CZNn3yeCseEtxuVPbDCwSyhGW6yZKUYKfgXmcpoGPWH \
    token_registry.aleo transfer_public 1field \
    aleo1rhgdu77hgyqd3xjj8ucu3jj9r2krwz6mnzyd80gncr5fxcwlh5rsvzp9px 1000u128 \
    --broadcast https://api.explorer.provable.com/v1/mainnet/transaction/broadcast --skip-funds-check

📦 Creating execution transaction for 'token_registry.aleo/transfer_public'...
✅ Created execution transaction for 'token_registry.aleo/transfer_public'
⌛ Execution at1n73ffv5eewcnkfkfd7acyyaslnwvrjnewygu9wxudufev4veeqyq2juge7 ('token_registry.aleo/transfer_public') has been broadcast to https://api.explorer.provable.com/v1/mainnet/transaction/broadcast.
at1n73ffv5eewcnkfkfd7acyyaslnwvrjnewygu9wxudufev4veeqyq2juge7```

### DIFF
--- a/cli/src/commands/developer/execute.rs
+++ b/cli/src/commands/developer/execute.rs
@@ -105,6 +105,9 @@ pub struct Execute {
     /// Timeout in seconds when waiting for transaction confirmation. Default is 60 seconds.
     #[clap(long, default_value_t = 60, requires = "wait")]
     timeout: u64,
+    /// Send the transaction without checking if sufficient funds are available (intended for testing purposes only).
+    #[clap(long, hide = true)]
+    skip_funds_check: bool,
 }
 
 impl Drop for Execute {
@@ -184,7 +187,7 @@ impl Execute {
         };
 
         // Check if the public balance is sufficient.
-        if self.record.is_none() && !is_static_query {
+        if self.record.is_none() && !is_static_query && !self.skip_funds_check {
             // Fetch the public balance.
             let address = Address::try_from(&private_key)?;
             let public_balance = Developer::get_public_balance::<N>(&endpoint, &address)
--- a/cli/src/commands/developer/execute.rs
+++ b/cli/src/commands/developer/execute.rs
@@ -155,11 +155,9 @@ impl Execute {
                 let height = query.current_block_height().with_context(|| "Failed to retrieve current block height")?;
                 let version = N::CONSENSUS_VERSION(height)?;
                 debug!("At block height {height} and consensus {version:?}");
-                let edition = Developer::get_latest_edition(&endpoint, &program_id)
-                    .with_context(|| format!("Failed to get latest edition for program {program_id}"))?;
 
                 // Load the program and it's imports into the process.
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: cli/src/commands/developer/execute.rs [hunk 1]
```diff
@@ -122,7 +127,7 @@ impl Execute {
         // Specify the query
         let query = Query::<N, BlockMemory<N>>::from(endpoint.clone());
 
-        // TODO (kaimast): can this ever be true?
+        // Check if the query is a static query.
```
kaimast, line 130: I left the node because we instantiate the query as a `RestQuery` in the previous line.

Do we support passing a static query as an endpoint here? I never tested this and am unsure if it works.

vicsn: It works, but the variable naming needs to be cleaned up one day.

vicsn (approved):
LGTM!

PR #4039, opened 2025-12-05
[Fix] Pause BFT block advancement during sync

### LINKED ISSUE #3911: [Bug] Errors during sync race conditions

kaimast:
This PR aims to fix #3911 and #3636. The PR also contains changes from three smaller PRs (#4065, #4138, and #4139) that can be reviewed and merged separately.

## Motivation
Recently, we added `CheckBlockError` as a machine-readable response to failures in block checks. The proposed changes use this new error type to differentiate between invalid blocks, and cases where the ledger has already advanced, and the proposed block is contained in the ledger already.
The snarkVM changes also ensure that calls to the ledger are atomic, but provides *no transactional* guarantees across multiple invocations of `Ledger` functions.

`BFT` has a `lock` field that aims to protect, block advancement. This lock already existed before this change, but was not accessible from outside the struct. A problem with this setup is that `check_block_content` and `advance_to_next_block` are dependent operations. Generally, one assumes that if the former succeeds, the latter will also. However, without additional locking, `BFT` can call `advance_to_block` *between* `Sync`'s calls of `check_block_content` and `advance_to_next_block`. In this case, `Sync` second call will fail.

## Consistent usage of `PendingBlocks`
Before this PR, pending blocks were only used during DAG sync. This PR changes it to always use pending blocks during sync, preventing a validator from syncing a block that contains a leader certificate that was skipped by the majority of the network. It also means that the two sync modes BFT are now almost identical, except that "fast" sync does not update the DAG until it has caught up with the network. This makes the other changes (next section) a lot easier to implement.

[This README](https://github.com/ProvableHQ/snarkOS/pull/4039/changes#diff-88224c24f5c0399890a8698944fe09ebd9245fabce5e2e9d04a44695d98aa211) contains a longer write-up on how sync and block creation works in BFT. Please verify that it makes sense, and let me know if you notice inconsistencies between the write-up and the code.

## Improved Coordination between `Sync` and `Primary`
The implementation implements an approached based on optimistic concurrency control to address this. Sync builds a chain of `PendingBlock`s without grabbing the `BFT` lock (otherwise, Sync would hold the lock for long periods of time).
Once a chain of pending blocks has reached the availability threshold, it will grab the lock, check their contents for correctness, and advance the ledger. In this step, `check_block_content` will verify the height of the block again and, if the height is below the current ledger height, `Sync` will abort and retry.

The main change for `BFT` is that `update_dag` does not grab the `lock` anymore, but callers are required to grab the lock before calling `update_dag`. There is a new method `BFT::pause_for_block_sync` that `Sync` calls before trying to commit blocks.
I removed the `ALLOW_LEDGER_ACCESS` parameter from `update_dag` to remove complexity from this function. This flag was only set to `false` for some tests, and does not seem to be needed any more even in this case, as the `LedgerService` abstraction allows to simply ignore writes. I would like to eventually have separate `update_dag` functions for sync and BFT, but this PR already reduces the complexity of the function noticeably.

## Testing
[x] Pass CI
[x] Successfully run prerelease tests
[x] Successfully run prerelease tests again 
[x] Run benchmarks and ensured there are no performance regressions 

## Notes
* Sync on this branch is about three times as fast as on `staging`. I assume this is mostly due to the improvements in snarkVM and unrelated to the changes in this branch.


vicsn:
Pausing this for now, the latest stress-test run `prerelease-v4.4.20` had a halt. First tackling https://github.com/ProvableHQ/snarkOS/issues/4047

kaimast:
https://github.com/ProvableHQ/snarkVM/pull/3070 needs to be merged first.

vicsn:
Blocked by https://github.com/ProvableHQ/snarkOS/pull/4069 CC @ljedrz

kaimast:
This needs some more testing and requires #4065 to be merged first, but then it is ready to go 🚀

kaimast:
Rebased on #4138 and will run another round of stress tests.

node/bft/src/helpers/storage.rs
```diff
@@ -134,9 +134,11 @@ impl<N: Network> Storage<N> {
             batch_ids: Default::default(),
             transmissions,
         }));
+
         // Perform GC on the current round.
         // Since there are no certificates yet, this only sets `gc_round`.
-        storage.garbage_collect_certificates(current_round);
+        storage.garbage_collect_certificates(current_round).unwrap();
```
ljedrz: should this be unwrapped?

node/src/client/router.rs
```diff
@@ -89,14 +89,20 @@ impl<N: Network, C: ConsensusStorage<N>> Disconnect for Client<N, C> {
...
+            if was_fully_connected {
+                self.sync.remove_peer(&peer_ip);
+            }
 
             // Clear cached entries applicable to the peer.
             self.router.cache().clear_peer_entries(peer_ip);
             #[cfg(feature = "metrics")]
             self.router.update_metrics();
+        } else {
+            warn!("Got disconnect for a peer '{peer_addr}' that is not in the peer pool");
```
ljedrz, line 104: this is most likely impossible to trigger, as `handle_disconnect` is only called by `Tcp::disconnect`, and its disconnects are guarded with an atomic. This applies to all other `handle_disconnect` impls.

snarkos/main.rs
```diff
@@ -84,18 +84,20 @@ fn main() {
...
                     let location = &guard.location;
                     let kind = guard.kind;
                     let num_uses = guard.num_uses;
                     let active_users = guard.num_active_uses();
                     let avg_duration = guard.avg_duration();
                     let avg_wait_time = guard.avg_wait_time();
+                    let num_waiting = guard.num_waiting();
                     tracing::info!(
-                        "[locktick] {location} ({:?}): {num_uses}; {active_users} active; avg d: {:?}; avg w: {:?}",
+                        "[locktick] {location} ({:?}): {num_uses}; {active_users} active; {num_waiting} waiting; gavg d: {:?}; avg w: {:?}",
```
ljedrz:
```suggestion
                        "[locktick] {location} ({:?}): {num_uses}; {active_users} active; {num_waiting} waiting; avg d: {:?}; avg w: {:?}",
```

node/bft/src/gateway.rs
```diff
@@ -650,6 +651,11 @@ impl<N: Network> Gateway<N> {
                     // Send the blocks to the sync module.
                     match sync_sender.insert_block_response(peer_ip, blocks.0, latest_consensus_version).await {
                         Ok(_) => Ok(true),
+                        Err(err) if err.is_benign() => {
+                            // Do not disconnect for benign errors, such as the ledger already advanced past the height of the block.
+                            trace!("Unable to process block response from '{peer_ip}' - {err}");
```
ljedrz: don't we want the `CONTEXT` here as well?

kaimast:
The most recent commit fixed a minor disconnect issue and the stress tests were successful:

Screenshot 2026-03-09 at 8 55 08 PM

[link to Grafana](https://aleostresstest.grafana.net/d/prerelease-devnet/prerelease-devnet?from=2026-03-09T19:30:00.000Z&to=2026-03-09T21:00:32.000Z&refresh=5s&timezone=browser&orgId=1)

node/src/client/router.rs
kaimast: Isn't it still good to have the check for sanity?

ljedrz: sure, that's ok - I was initially worried it might suggest that it's a possibility, but it's a `WARN`, so it implies that it's out of the ordinary

node/bft/src/sync/mod.rs
```diff
@@ -421,31 +357,38 @@ impl<N: Network> Sync<N> {
 
 // Methods to manage storage.
 impl<N: Network> Sync<N> {
-    /// Syncs the storage with the ledger at bootup.
-    ///
-    /// This is called when starting the validator and after finishing a sync without BFT.
+    /// This functions inserts the certificates from the most recent blocks into the BFT DAG.
+    /// It is called when starting the validator and after finishing a sync without BFT.
     async fn sync_storage_with_ledger_at_bootup(&self) -> Result<()> {
```
ljedrz: I no longer see any `await` points, so this should no longer be `async`, no?

node/bft/src/sync/mod.rs
```diff
@@ -454,14 +397,17 @@ impl<N: Network> Sync<N> {
...
+            .garbage_collect_certificates(latest_block.round())
+            .with_context(|| "Failed to garbage collect certificates")?;
+
         // Iterate over the blocks.
-        for block in &blocks {
-            // If the block authority is a sub-DAG, then sync the batch certificates with the block.
-            // Note that the block authority is always a sub-DAG in production;
-            // beacon signatures are only used for testing,
-            // and as placeholder (irrelevant) block authority in the genesis block.
+        for block in blocks.clone() {
```
ljedrz: it's late in the day so I might just be missing it, but where are we consuming the blocks? do we need this clone?

node/sync/src/block_sync.rs
```diff
@@ -239,7 +273,9 @@ impl<N: Network> BlockSync<N> {
     /// (due to the latter having a tolerance of one block).
     #[inline]
     pub fn can_block_sync(&self) -> bool {
-        self.sync_state.read().can_block_sync() || self.has_pending_responses()
+        !self.failed_requests.lock().is_empty()
+            || self.sync_state.read().can_block_sync()
```
ljedrz: is this the preferable order considering short-circuiting? I wouldn't expect failed requests to appear that often

node/sync/src/block_sync.rs
```diff
@@ -239,7 +273,9 @@ impl<N: Network> BlockSync<N> {
     /// (due to the latter having a tolerance of one block).
     #[inline]
     pub fn can_block_sync(&self) -> bool {
```
ljedrz: nit: considering the presence of failed requests, I'd name it sth like `should_block_sync` instead (as it's not really about the capability)

node/sync/src/block_sync.rs
```diff
@@ -460,6 +521,46 @@ impl<N: Network> BlockSync<N> {
...
+            trace!("Nothing to sync. Will not issue new block requests");
+            return;
+        }
+
+        let batches = self.prepare_block_requests();
+
+        if batches.is_empty() {
+            let total_requests = self.num_total_block_requests();
+            let num_outstanding = self.num_outstanding_block_requests();
+            if total_requests > 0 {
```
ljedrz: nit: `clippy` might someday complain that it's not `!= 0`

node/sync/src/block_sync.rs
```diff
@@ -1103,58 +1342,9 @@ impl<N: Network> BlockSync<N> {
...
-        // The given height may also be greater `start_height` due to concurerent block advancement.
-        if let Some((height, _)) = requests.as_slice().first() {
-            debug!("Re-requesting blocks starting at height {height}");
-            Ok(Some((requests, sync_peers)))
-        } else {
-            // Do not log here as this constitutes a benign race condition.
-            Ok(None)
+        // Mark the non-obsolete requests that timed out as failed.
+        for (height, e) in timed_out_requests.into_iter() {
+            self.failed_requests.lock().insert(height, e);
```
ljedrz: `self.failed_requests.lock().extend(timed_out_requests.into_iter())` would probably be better, no?

node/bft/src/sync/mod.rs
```diff
@@ -482,40 +428,42 @@ impl<N: Network> Sync<N> {
...
-        // If a BFT sender was provided, send the certificates to the BFT.
-        if let Some(cb) = self.sync_callback.get() {
-            // Await the callback to continue.
-            cb.sync_dag_at_bootup(certificates).await;
+                // Insert (but do not commit) pending block certificates.
+                for block in pending_blocks.iter() {
+                    if let Authority::Quorum(subdag) = block.authority() {
+                        for round in subdag.values() {
+                            for certificate in round {
+                                cb.add_certificate(certificate.clone());
```
ljedrz: we could avoid the clone here if we did `into_iter` on the `pending_blocks`

node/bft/src/sync/mod.rs
```diff
@@ -889,66 +876,109 @@ impl<N: Network> Sync<N> {
...
+            return Ok(true);
+        };
 
-            // Create a more detailed log message if we are committing more than one block at a time.
-            if num_blocks > 1 {
-                trace!(
-                    "Attempting to commit {chain_length} pending block(s) starting at height {start_height}.",
-                    chain_length = pending_blocks.len(),
-                );
+        let ledger = self.ledger.clone();
```
```suggestion
        let ledger = &self.ledger;
```
nit: it's a bit cheaper, and I don't see it consumed

ljedrz (commented):
I'm not intimately familiar with the syncing logic, but I've traversed it all and left some comments.

node/sync/src/block_sync.rs
kaimast: I removed this helper function as it is only used in one place, and renamed the one in `SyncState` to `can_issue_new_block_requests`.

node/sync/src/block_sync.rs
kaimast: I swapped the order

node/bft/src/sync/mod.rs
kaimast: You are right. Interesting that clippy did not catch this.

node/bft/src/sync/mod.rs
kaimast: The pending blocks aren't copied beforehand, and we cannot remove them from `self.pending_blocks` until their availability threshold is reached.

node/sync/src/block_sync.rs
kaimast: The loop now also prints a warning if the failed request already existed. I don't think there is a way to do this with `extend`.

kaimast:
I ran another set of stress tests on top of the two commits from today. Still passes.

[Grafana Link](https://aleostresstest.grafana.net/d/prerelease-devnet/prerelease-devnet?from=2026-03-11T23:07:04.000Z&to=2026-03-12T00:07:04.000Z&refresh=5s&timezone=browser&orgId=1)

node/bft/src/helpers/storage.rs
raychu86: +1, `garbage_collect_certificates` now returns a Result, so either we ignore the results here or have `Storage::new` return a Result as well.

node/bft/src/helpers/storage.rs
```diff
@@ -811,6 +829,8 @@ impl<N: Network> Storage<N> {
         block: &Block<N>,
         certificate: BatchCertificate<N>,
         unconfirmed_transactions: &HashMap<N::TransactionID, Transaction<N>>,
+        pending_aborted_transactions: &HashSet<N::TransactionID>,
+        pending_aborted_solutions: &HashSet<SolutionID<N>>,
```
raychu86: Why are these necessary? 

Pending blocks I assume are the list of blocks that have not been committed and processed yet. The aborted transmission should be processed in the first instance the `transmission_id` is seen, they should not be in the future blocks that have not been processed.

node/bft/src/bft.rs
```diff
@@ -102,7 +103,7 @@ impl<N: Network> BFT<N> {
             leader_certificate: Default::default(),
             leader_certificate_timer: Default::default(),
             consensus_sender: Default::default(),
-            lock: Default::default(),
+            block_advancement_lock: Default::default(),
```
raychu86: I don't see this `block_advancement_lock` being used despite the doc above 
> Note: Always aquire a lock to `block_advancement_lock` before write-locking the DAG.`

 Only instance I found is in `shutdown`.

node/bft/src/bft.rs
```diff
@@ -510,14 +513,8 @@ impl<N: Network> BFT<N> {
...
 impl<N: Network> BFT<N> {
-    /// Stores the certificate in the DAG, and attempts to commit one or more anchors.
-    async fn update_dag<const ALLOW_LEDGER_ACCESS: bool, const IS_SYNCING: bool>(
-        &self,
-        certificate: BatchCertificate<N>,
-    ) -> Result<()> {
-        // Acquire the BFT lock.
-        let _lock = self.lock.lock().await;
-
+    /// Callers of this function are required to hold the write lock to the `LedgerService`, by calling [`Self::pause_for_block_sync`].
```
raychu86: This `Self::pause_for_block_sync` doesn't exist?

node/bft/src/sync/mod.rs
```diff
@@ -421,31 +363,48 @@ impl<N: Network> Sync<N> {
...
-    /// Syncs the storage with the ledger at bootup.
-    ///
-    /// This is called when starting the validator and after finishing a sync without BFT.
-    async fn sync_storage_with_ledger_at_bootup(&self) -> Result<()> {
-        // Retrieve the latest block in the ledger.
-        let latest_block = self.ledger.latest_block();
+    /// This functions inserts the certificates from the most recent blocks into the BFT DAG.
+    /// It is called when starting the validator and after finishing a sync without BFT.
+    fn sync_storage_with_ledger_at_bootup(&self) -> Result<()> {
+        let mut pending_blocks = self.pending_blocks.lock();
```
raychu86, line 364: Previously, this lock fetched was not active while the ledger was reading from DB, but now this function keeps the lock for longer.

node/sync/src/block_sync.rs
ljedrz: Where is that warning? It's just a `BTree::insert` operation. I'm mostly interested in not acquiring the lock separately for every item in the loop.

node/bft/src/bft.rs
kaimast: You are right. This is an outdated comment.

All the locking logic is now in `LedgerService::begin_ledger_update`, which returns a `LedgerUpdate` object that is somewhat like a database transaction.
It ensures only one ledger update is active at a time and `check_next_block` and `check_block_content` are only accessible through `LedgerUpdate`. This ensures that the ledger is not updated between running the checks and advancing it (essentially a read-write dependency, if you think of it as a database transaction).

I updated the doc comments and the READMEs.

node/bft/src/bft.rs
kaimast: Same as the previous comment. `BFT::pause_for_sync` has been replaced by `LedgerService::begin_ledger_update`.

node/bft/src/sync/mod.rs
kaimast: I don't it is an issue. There is only one task that performs block advancement, and that is also the only tasks that modifies `pending_blocks`.

node/bft/src/helpers/storage.rs
kaimast: I removed this and tested again, it seems to not be needed indeed.

snarkos/main.rs
kaimast: Fixed.,

node/bft/README.md
```diff
@@ -31,6 +31,80 @@ f + 1 vertices that vote for the anchor, or 2f + 1 vertices that do not, or a ti
...
+
+##### Within GC Range (Normal Sync)
+
+When the node is within the garbage collection range of the network tip:
+
+1. **Block Reception**: `bft::Sync` requests and receives blocks from peer nodes via `BlockSync`.
+2. **Block Verification**: Blocks are verified using `check_block_subdag()` and added to a queue of `pending_blocks`.
+3. **Certificate Insertion**: Each certificate from the block's subDAG is added to storage via `sync_certificate_with_block()` and sent to the BFT. This populates the DAG with the certificates needed for consensus.
+4. **BFT-Driven Ledger Advancement**: The BFT module handles block creation through its normal consensus path -- when enough certificates are added to the DAG, the BFT commits leader certificates and creates blocks just as it does during normal operation.
+5. **Pending Block Cleanup**: When the ledger advances because of leader commits (see 4), pending blocks are removed from the queue.
```
raychu86, line 66: Is there any safety check to ensure that the BFT generated block is equivalent from the incoming block from the peer?

kaimast: It could only happen in case of a bug, as the certificates are the same and the block building process is deterministic.
If you want I can add a check, that prints a warning if the hashes mismatch.

node/sync/src/block_sync.rs
kaimast: I changed it to only grab the lock once in [af65c9b](https://github.com/ProvableHQ/snarkOS/pull/4039/commits/af65c9b8882613d9a0fc116d30ae63ae28dbce29)

kaimast:
I addressed all comments and resolved conflicts with mainnet.

kaimast:
I added the changes from #4179 to this PR, which should make merging easier.

node/bft/README.md
raychu86: Would be nice, since we can detect and avoid forking cases

raychu86 (approved):
Logic looks sound to me. Please run proper stress tests before merging

node/bft/README.md
vicsn: If this is not done yet, I'd suggest this can be added in a separate PR

vicsn:
Stresstest `prerelease-v4.5.29` succeeded

PR #4042, opened 2025-12-09
Add Nix flake for reproducible builds

IGI-111:
Add a flake.nix to enable Nix users to build and develop snarkOS in a reproducible environment. This provides:

- Development shell with all necessary dependencies
- Reproducible builds across different systems
- Easy integration with NixOS, nix-darwin and other Nix-based systems
- Consistent toolchain versions for all contributors

This change does not affect non-Nix users and maintains backward compatibility with existing build processes.

Nix flakes provide a standardized way to define inputs and outputs for Nix projects, making it easier for users of NixOS, nix-darwin, and other Nix-based systems to work with snarkOS.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,78 @@
+{
+  "nodes": {
+    "nixpkgs": {
+      "locked": {
+        "lastModified": 1764950072,
+        "narHash": "sha256-BmPWzogsG2GsXZtlT+MTcAWeDK5hkbGRZTeZNW42fwA=",
+        "owner": "NixOS",
+        "repo": "nixpkgs",
+        "rev": "f61125a668a320878494449750330ca58b78c557",
+        "type": "github"
+      },
+      "original": {
+        "owner": "NixOS",
+        "ref": "nixos-unstable",
+        "repo": "nixpkgs",
+        "type": "github"
+      }
+    },
+    "nixpkgs_2": {
+      "locked": {
+        "lastModified": 1744536153,
+        "narHash": "sha256-awS2zRgF4uTwrOKwwiJcByDzDOdo3Q1rPZbiHQg/N38=",
+        "owner": "NixOS",
+        "repo": "nixpkgs",
+        "rev": "18dd725c29603f582cf1900e0d25f9f1063dbf11",
+        "type": "github"
+      },
+      "original": {
+        "owner": "NixOS",
+        "ref": "nixpkgs-unstable",
+        "repo": "nixpkgs",
+        "type": "github"
+      }
+    },
+    "root": {
+      "inputs": {
+        "nixpkgs": "nixpkgs",
+        "rust-overlay": "rust-overlay",
+        "systems": "systems"
+      }
+    },
+    "rust-overlay": {
+      "inputs": {
+        "nixpkgs": "nixpkgs_2"
+      },
+      "locked": {
+        "lastModified": 1765248027,
+        "narHash": "sha256-ngar+yP06x3+2k2Iey29uU0DWx5ur06h3iPBQXlU+yI=",
+        "owner": "oxalica",
+        "repo": "rust-overlay",
+        "rev": "7b50ad68415ae5be7ee4cc68fa570c420741b644",
+        "type": "github"
+      },
+      "original": {
+        "owner": "oxalica",
+        "repo": "rust-overlay",
+        "type": "github"
+      }
+    },
+    "systems": {
+      "locked": {
+        "lastModified": 1681028828,
+        "narHash": "sha256-Vy1rq5AaRuLzOxct8nz4T6wlgyUR7zLU309k9mBC768=",
+        "owner": "nix-systems",
+        "repo": "default",
+        "rev": "da67096a3b9bf56a91d16901293e51ba5b49a27e",
+        "type": "github"
+      },
+      "original": {
+        "owner": "nix-systems",
+        "repo": "default",
+        "type": "github"
+      }
+    }
+  },
+  "root": "root",
+  "version": 7
+}
--- a/flake.nix
+++ b/flake.nix
@@ -1,0 +1,137 @@
+{
+  description = "snarkOS - A decentralized operating system";
+
+  inputs = {
+    nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
+    rust-overlay.url = "github:oxalica/rust-overlay";
+    systems.url = "github:nix-systems/default";
+  };
+
+  outputs =
+    {
+      self,
+      nixpkgs,
+      rust-overlay,
+      systems,
+    }:
+    let
+      eachSystem = f: nixpkgs.lib.genAttrs (import systems) (system: f system);
+
+      # Common configuration factory that takes pkgs as input
+      mkCommonConfig = pkgs: rec {
+        # Rust toolchain matching rust-toolchain.toml
+        rustToolchain = pkgs.rust-bin.fromRustupToolchainFile ./rust-toolchain.toml;
+
+        # Native build inputs for the project
+        nativeBuildInputs = with pkgs; [
+          rustToolchain
+          curl
+          openssl
+          zlib
+          pkg-config
+          clang
+          gcc
+          lld
+          libclang
+          git # Required for build.rs (built crate with git2 feature)
+        ];
+
+        # Runtime dependencies
+        buildInputs = with pkgs; [
+          openssl
+        ];
+
+        # Environment variables
+        LIBCLANG_PATH = "${pkgs.libclang.lib}/lib";
+        GIT = "${pkgs.git}/bin/git";
+        LD_LIBRARY_PATH = pkgs.lib.makeLibraryPath [
+          pkgs.zlib
+          pkgs.openssl
+          pkgs.libclang.lib
+        ];
+      };
+    in
+    {
+      devShells = eachSystem (
+        system:
+        let
+          overlays = [ (import rust-overlay) ];
+          pkgs = import nixpkgs {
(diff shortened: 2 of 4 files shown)


vicsn (commented):
To prevent regressions, should we add a CI job or manually run a Nix command before new releases?

Did any partner or ecosystem developer ask about this?

IGI-111:
There's a few people in the org using Nix so I thought it might be a good idea to share the flake I use to run snarkos anyways, but it's not prompted by a specific ecosystem demand.

There's precedent though, nixpkgs has [a snarkos package](https://github.com/NixOS/nixpkgs/blob/7241bcbb4f099a66aafca120d37c65e8dda32717/pkgs/by-name/sn/snarkos/package.nix#L56) but it's very out of date. At least a source based flake is fresh.

One could add a CI job I suppose, but this is essentially similar to the Ubuntu script which didn't seem to require one and would require similarly infrequent changes.

flake.nix
```diff
@@ -0,0 +1,82 @@
...
+            echo "To build the project, run: cargo build --release"
+            echo "To run tests, run: cargo test"
+          '';
+        };
+
+        # Package for building snarkOS binary
+        # Note: For workspace projects, it's often easier to use `cargo build` in the devShell
+        packages.default = pkgs.rustPlatform.buildRustPackage {
+          pname = "snarkos";
+          version = "4.4.0";
```
vicsn: Is there a light-weight approach to catch updating this in `build.rs` or CI, or can it be left out?

flake.lock
```diff
@@ -0,0 +1,96 @@
+{
```
vicsn, line 1: Should we be worried about cluttering the repo too much, and move the flake files into a build folder or add the `flake.lock` in `.gitignore`. Noting this instance is fine, but maybe people come up with new build methods in the future.

Arguably, `build_ubuntu.sh` should also live in `scripts` but we may also break people's builds if we move it around.

vicsn (commented):
Last comments from me, can you ask another Nix expert to review?

flake.nix
IGI-111: We can pick it up directly from the Cargo.toml

flake.lock
IGI-111: All for categorizing instead of clutter, but in this case you want the lockfile committed and static to get the reproducibility benefit.

flake.nix
vicsn: Very fancy!

flake.nix
```diff
@@ -0,0 +1,96 @@
...
+          openssl
+        ];
+
+        # Read version from Cargo.toml
+        cargoToml = builtins.fromTOML (builtins.readFile ./Cargo.toml);
+        version = cargoToml.package.version;
+      in
+      {
+        devShells.default = pkgs.mkShell {
+          inherit nativeBuildInputs buildInputs;
```
mitchmindtree: One handy pattern for `devShells` can be to declare them in terms of the package you're exposing. E.g. this is how my leo-dev devShell looks:

- https://github.com/mitchmindtree/leo.nix/blob/a2a47806cde794d261a34469bcf2822d47e739f1/pkgs/leo-dev.nix#L1-L16

i.e. we could use:

- `inputsFrom = [ snarkos ];` to provide all `buildInputs`, `nativeBuildInputs` from `snarkos`.
- `inherit` env vars directly from `snarkos`. 

That said, sharing the inputs directly is also fine! But can get a bit unwieldy when you want a more modular setup with different mods per package or shell.

flake.nix
```diff
@@ -0,0 +1,96 @@
...
+          GIT = "${pkgs.git}/bin/git";
+
+          shellHook = ''
+            echo "snarkOS development environment"
+            echo "Rust version: $(rustc --version)"
+            echo "Cargo version: $(cargo --version)"
+            echo ""
+            echo "To build the project, run: cargo build --release"
+            echo "To run tests, run: cargo test"
+          '';
```
mitchmindtree: This is sweet - I don't make much use of shellHook but this a nice use-case!

I wonder we should just keep it informative and minimal?

```suggestion
          shellHook = ''
            echo "snarkOS development environment"
            echo "Rust version: $(rustc --version)"
            echo "Cargo version: $(cargo --version)"
          '';
```

flake.nix
```diff
@@ -0,0 +1,96 @@
+{
+  description = "snarkOS - A decentralized operating system";
+
+  inputs = {
+    nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
+    rust-overlay.url = "github:oxalica/rust-overlay";
+    flake-utils.url = "github:numtide/flake-utils";
```
mitchmindtree: I find you can get away without flake-utils if you declare one function for populating an attrset per system, e.g.

https://github.com/mitchmindtree/leo.nix/blob/a2a47806cde794d261a34469bcf2822d47e739f1/flake.nix#L25-L29

Then you can just have regular flake outputs rather than the custom flake-utils wrapper:

https://github.com/mitchmindtree/leo.nix/blob/a2a47806cde794d261a34469bcf2822d47e739f1/flake.nix#L66-L80

This makes it a little easier to include both system-agnostic declarations (like `overlays`) *and* `packages` and `devShells` per system. With flake-utils, you end up having to merge the result of `eachDefaultSystem` with another attrset containing overlays which feels a bit gross, e.g.

```nix
flake-utils.lib.eachDefaultSystem { ... } // { overlays = { ... }; }
```

I've mostly avoided flake-utils for this reason, but I'm happy to go with it if you're familiar with it!

flake.nix
```diff
@@ -0,0 +1,96 @@
...
+            echo ""
+            echo "To build the project, run: cargo build --release"
+            echo "To run tests, run: cargo test"
+          '';
+        };
+
+        # Package for building snarkOS binary
+        # Note: For workspace projects, it's often easier to use `cargo build` in the devShell
+        packages.default = pkgs.rustPlatform.buildRustPackage {
+          pname = "snarkos";
```
mitchmindtree: could maybe also use the `cargoToml` here, e.g. `pname = cargoToml.package.name;`

flake.nix
```diff
@@ -0,0 +1,96 @@
...
+          inherit version;
+
+          src = ./.;
```
mitchmindtree: When your nix rust package lives inside your repo like this, I find it can be handy to filter the source files down to what the package really needs - otherwise Nix will consider it a cache invalidation everytime even unrelated files change (e.g. changing `README.md` or `build_ubuntu.sh` would require rebuilding the nix pkg).

You can do something like this:

```suggestion
          src = lib.sourceFilesBySuffices ./. [ ".lock" ".rs" ".toml" ]; # plus any other required file exts
```

mitchmindtree (commented):
Nice one @IGI-111! Overall looks great, I've left a few optional thoughts and tips inline and below.

### `buildFeatures`

One thing that might be useful could be to allow customising the set of features with which snarkos is built.

For leo dev, we'd need to at least expose a build of `snarkos` that includes the `test_network` feature as I think we need this for running the leo test suite. E.g. in my leo-dev shell I have a [`snarkos-testnet` package as a buildInput](https://github.com/mitchmindtree/leo.nix/blob/a2a47806cde794d261a34469bcf2822d47e739f1/pkgs/leo-dev.nix#L11).

To address this in [mitchmindtree/leo.nix](https://github.com/mitchmindtree/leo.nix), I exposed [the `buildFeatures` as an input](https://github.com/mitchmindtree/leo.nix/blob/a2a47806cde794d261a34469bcf2822d47e739f1/pkgs/snarkos.nix#L2-L3), so you can easily declare [custom snarkos packages for different sets](https://github.com/mitchmindtree/leo.nix/blob/a2a47806cde794d261a34469bcf2822d47e739f1/flake.nix#L55-L62). Alternatively, we might be able to achieve the same using `override` or `overrideAttrs` with `snarkos` to override the `buildFeatures` that way.



### Formatter

Maybe we could add a `formatter` output using the new standard `nixfmt-tree`? This would allow us to `nix fmt` locally, and do `nix fmt -- --ci` to validate in CI.

### Overlay

Maybe we could provide an overlay too? This isn't strictly necessary, but I find them useful for merging sets of packages into a nixpkgs set. I tend to [declare my packages in an overlay](https://github.com/mitchmindtree/leo.nix/blob/a2a47806cde794d261a34469bcf2822d47e739f1/flake.nix#L32-L64), and then [expose the `packages` outputs from the merged set](https://github.com/mitchmindtree/leo.nix/blob/a2a47806cde794d261a34469bcf2822d47e739f1/flake.nix#L66-L72) for convenience. I wouldn't block this PR on this though - I think this comes down to personal preference, and I'd be happy to contribute an overlay in a future PR.

### CI

I agree with @vicsn that ideally we'd test this in CI too. Normally I'll have something like this job in the github workflows for my nix repos, maybe we can do something similar with circleci?
```yml
jobs:
  nix:
    permissions:
      id-token: "write"
      contents: "read"
    runs-on: ubuntu-latest
    strategy:
      fail-fast: false
      matrix:
        include:
          - command: build --print-build-logs --no-update-lock-file .#leo
          - command: develop --no-update-lock-file .#leo-dev
          - command: flake check --all-systems --no-update-lock-file
          - command: fmt -- --ci
    steps:
      - uses: actions/checkout@v5
      - uses: DeterminateSystems/nix-installer-action@main
      - uses: DeterminateSystems/magic-nix-cache-action@v13
      - run: nix ${{ matrix.command }}
```
If we don't want to block snarkos dev on nix failures, maybe we can allow failure somehow? But at least we'll still be able to keep an eye on it.

IGI-111:
Closing this in favor of a separate leo.nix repo for such configs

PR #4045, opened 2025-12-17
Add metric for total connected stake

vicsn:
## Motivation

Adding a metric for total stake of validators a node is connected to, so we can alert on it if it's goes too low (below 66% the network will halt).

A stylistic comment: it's a bit wonky to compute unconnected stake and then convert back to connected stake, but it's not too complex and it's likely the most efficient because there will be fewer non-connected than connected validators.

CC @mike2194


raychu86 (approved):

raychu86:
CI looks to be failing, but looks unrelated to this PR

PR #4050, opened 2025-12-19
Improve tx propagation

### LINKED ISSUE #3962: [Feature] Improve transaction propagation
## 🚀 Feature

A core feature of Narwhal is that transmissions may be included in multiple proposals. 

image

The current setup of snarkOS has an overreliance on safety, because clients and validators propagate valid seen transmissions to all of their peers, incurring compute, bandwidth, and generally a lot of latency. The graph below shows that certificate generation slows down significantly under load. Previous measurements have shown that this in turn is caused by nodes waiting on transmission fetching.

Image

We shouldn't get rid of the propagation entirely either. When running a load test with 6000 transmissions on reference hardware and [no propagation at all](https://github.com/ProvableHQ/snarkOS/pull/3493), sometimes all transmissions would land within a few rounds, but more often only around 5750 would land, which is an indication some older certificates get left behind under stress and we need at least some propagation.

To improve network throughput, I propose the following:
- We add a `propagate: bool` field to `struct UnconfirmedTransaction`
- When clients receive a transaction via `/broadcast/transaction`, broadcast to all peers with `propagate: true`
- When clients receive a transaction via the P2P network, they broadcast to all peers with `propagate: false`, but only if `propagate: true` on the incoming message
- When validators receive a transaction via `/broadcast/transaction`, broadcast to all validators with `propagate: false`.
- When validators receive a transaction via the P2P network, they broadcast to all peers with `propagate: false`, but only if `propagate: true` on the incoming message. Receiving validators should immediately add the transmission into `cache_transmissions` so they don't have to fetch it.
- In order to ensure all transmissions land, validators periodically - say every `PRIMARY_PING_IN_MS` - include transmissions which they have not seen in any proposal/certificate/ledger yet from `cache_transmissions`. We may want to tackle this last point only after https://github.com/ProvableHQ/snarkOS/issues/3961 is done. This can also be done based on the validator index.

The above is also applicable to solutions.

Note that the above approach to optimistic broadcast only works as when validator's router's are well-connected. With large networks and peer limits of 21, that may not be the case, so we may want to also move transmission broadcasts to the Gateway.

vicsn:
## Motivation

Closes https://github.com/ProvableHQ/snarkOS/issues/3962

## Test Plan

- [ ] Run stress-testing after review is completed.

## Backwards compatibility

- [ ] Review whether nodes which didn't upgrade yet handle the new unknown message gracefully (without closing the connection). This should be done automatically by https://github.com/ProvableHQ/snarkOS/pull/4029

### DIFF
--- a/node/bft/src/gateway.rs
+++ b/node/bft/src/gateway.rs
@@ -65,9 +65,9 @@ use snarkvm::{
     console::prelude::*,
     ledger::{
         committee::Committee,
-        narwhal::{BatchHeader, Data},
+        narwhal::{BatchHeader, Data, Transmission, TransmissionID},
     },
-    prelude::{Address, Field},
+    prelude::{Address, Field, Transaction},
 };
 
 use colored::Colorize;
@@ -738,6 +738,31 @@ impl<N: Network> Gateway<N> {
                 }
                 Ok(true)
             }
+            Event::UnconfirmedTransaction(event) => {
+                // Perform the deferred non-blocking deserialization of the transaction.
+                let transaction = match event.transaction.deserialize().await {
+                    Ok(transaction) => transaction,
+                    Err(error) => bail!("[UnconfirmedTransaction] {error}"),
+                };
+                // Calculate the transmission checksum.
+                let checksum = Data::<Transaction<N>>::Buffer(transaction.to_bytes_le()?.into()).to_checksum::<N>()?;
+                // Construct the transmission ID.
+                let transmission_id = TransmissionID::Transaction(transaction.id(), checksum);
+                // Construct the transmission.
+                let transmission = Transmission::Transaction(Data::Object(transaction));
+
+                // Determine the worker ID.
+                let Ok(worker_id) = assign_to_worker(transmission_id, self.num_workers()) else {
+                    warn!("{CONTEXT} Unable to assign transmission ID '{}' to a worker", transmission_id);
+                    return Ok(true);
+                };
+                // Send the unconfirmed transmission to the worker.
+                if let Some(sender) = self.get_worker_sender(worker_id) {
+                    // Send the unconfirmed transmission to the worker.
+                    let _ = sender.tx_unconfirmed_transmission.send((peer_ip, transmission_id, transmission)).await;
+                }
+                Ok(true)
+            }
             Event::ValidatorsRequest(_) => {
                 let mut connected_peers = self.get_best_connected_peers(Some(MAX_VALIDATORS_TO_SEND));
                 connected_peers.shuffle(&mut rand::thread_rng());
@@ -1104,6 +1129,14 @@ impl<N: Network> Gateway<N> {
     // Remove addresses whose ban time has expired.
     fn handle_banned_ips(&self) {
         self.tcp.banned_peers().remove_old_bans(IP_BAN_TIME_IN_SECS);
+    }
+
+    /// Broadcast an unconfirmed transaction so other validators can cache it.
+    pub fn broadcast_unconfirmed_transaction(&self, transaction: Transaction<N>) {
+        let event = Event::UnconfirmedTransaction(crate::events::UnconfirmedTransaction {
+            transaction: Data::Object(transaction),
+        });
+        Transport::broadcast(self, event);
     }
 }
 
--- a/node/bft/src/gateway.rs
+++ b/node/bft/src/gateway.rs
@@ -739,13 +739,13 @@ impl<N: Network> Gateway<N> {
                 Ok(true)
             }
             Event::UnconfirmedTransaction(event) => {
+                // Calculate the transmission checksum.
+                let checksum = event.transaction.to_checksum::<N>()?;
                 // Perform the deferred non-blocking deserialization of the transaction.
                 let transaction = match event.transaction.deserialize().await {
                     Ok(transaction) => transaction,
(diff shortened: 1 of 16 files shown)


node/bft/storage-service/src/persistent.rs
```diff
@@ -102,7 +99,7 @@ impl<N: Network> StorageService<N> for BFTPersistentStorage<N> {
     /// If the transmission ID does not exist in storage, `None` is returned.
     fn get_transmission(&self, transmission_id: TransmissionID<N>) -> Option<Transmission<N>> {
         // Try to get the transmission from the cache first.
-        if let Some((transmission, _)) = self.cache_transmissions.lock().get_mut(&transmission_id) {
+        if let Some(transmission) = self.cache_transmissions.lock().get(&transmission_id) {
```
vicsn, line 102: In hindsight not sure if this will really have that much impact, the idea is to make the cache more FIFO, because typically older transactions won't be needed in the future anymore.

node/bft/storage-service/src/persistent.rs
```diff
@@ -52,10 +52,8 @@ pub struct BFTPersistentStorage<N: Network> {
     transmissions: DataMap<TransmissionID<N>, (Transmission<N>, IndexSet<Field<N>>)>,
     /// The map of `aborted transmission ID` to `certificate IDs` entries.
     aborted_transmission_ids: DataMap<TransmissionID<N>, IndexSet<Field<N>>>,
-    /// The LRU cache for `transmission ID` to `(transmission, certificate IDs)` entries that are part of the persistent storage.
-    cache_transmissions: Mutex<LruCache<TransmissionID<N>, (Transmission<N>, IndexSet<Field<N>>)>>,
-    /// The LRU cache for `aborted transmission ID` to `certificate IDs` entries that are part of the persistent storage.
-    cache_aborted_transmission_ids: Mutex<LruCache<TransmissionID<N>, IndexSet<Field<N>>>>,
+    /// The LRU cache for `transmission ID` to `transmission` entries that are part of the persistent storage.
+    cache_transmissions: Mutex<LruCache<TransmissionID<N>, Transmission<N>>>,
```
vicsn, line 56: Removed `cache_aborted_transmission_ids` because we never read from it.

vicsn:
Opened an issue about CI failing: https://github.com/ProvableHQ/snarkOS/issues/4052

node/bft/src/gateway.rs
```diff
@@ -738,6 +738,35 @@ impl<N: Network> Gateway<N> {
...
                 Ok(true)
             }
+            Event::UnconfirmedTransaction(event) => {
+                // Perform the deferred non-blocking deserialization of the transaction.
+                let transaction = match event.transaction.deserialize().await {
+                    Ok(transaction) => transaction,
+                    Err(error) => bail!("[UnconfirmedTransaction] {error}"),
+                };
+                // Calculate the transmission checksum.
+                let checksum = Data::<Transaction<N>>::Buffer(transaction.to_bytes_le()?.into()).to_checksum::<N>()?;
```
ljedrz: note: this operation involves plenty of redundant work (since we're serializing a transaction that was _just_ deserialized, even if it was in a blob form); it should be possible to optimize it so that the checksum can be calculated based on the `event`

node/src/validator/router.rs
```diff
@@ -278,18 +278,27 @@ impl<N: Network, C: ConsensusStorage<N>> Inbound<N> for Validator<N, C> {
...
+        };
+
+        // Cache the unconfirmed transaction.
+        if let Err(error) = self.consensus.cache_unconfirmed_transaction(transaction.clone()).await {
+            trace!("[UnconfirmedTransaction] {error}");
+            return true; // Maintain the connection.
+        };
+
+        // Broadcast the unconfirmed transaction so other validators can cache it.
+        self.consensus.bft().primary().gateway().broadcast_unconfirmed_transaction(transaction);
```
ljedrz: we should instead use the `serialized` transaction here, as it is almost free to clone and write to the socket (as opposed to the deserialized transaction); it could also be used to cheaply construct the transmission id for caching purposes (unless it's already available at the callsite, in which case it would be practical to pass it on to the `cache_unconfirmed_transaction` call above)

node/bft/src/gateway.rs
vicsn: d29ba9118

node/src/validator/router.rs
vicsn: d29ba9118

reducing the redundant id computation we do all over the place is out of scope but looking forward to the day we can do it

node/network/src/lib.rs
```diff
@@ -44,7 +44,12 @@ pub fn bootstrap_peers<N: Network>(is_dev: bool) -> Vec<SocketAddr> {
         // Development testing contains optional bootstrap peers loaded from the environment.
         match std::env::var("TEST_BOOTSTRAP_PEERS") {
             Ok(peers) => peers.split(',').map(|peer| SocketAddr::from_str(peer).unwrap()).collect(),
+            Err(VarError::NotPresent) => {
```
kaimast, line 47: I slightly changed this, so we still log other errors.

kaimast:
Looks good to me. I merged `staging` and tested it locally.

kaimast (approved):

node/network/src/lib.rs
vicsn: Cool!

PR #4054, opened 2025-12-26
[Feat] Introduce a dynamic validator whitelist

### LINKED ISSUE #4053: [Feature] Ensure gateway peers are frequently correctly updated
## 🚀 Feature

In order for validators to automatically connect to gossipped peers in a secure manner, they should run a firewall with a strict whitelist for other validators. To make sure the whitelist is up to date, nodes should write the gateway peers file on every heartbeat, not just during shutdown.

And just to double check, when a malicious bootstrap peer or validator supplies us with many false peers, do we ensure we at least keep the peers we're currently connected to?

ljedrz:
This PR introduces a new ledger-adjacent file containing the list of all the currently connected validators, and updated with every validator heartbeat. The format is a space-delimited list of IP-port pairs for trivial script processing for potential use in commands like `ipset`.

Closes https://github.com/ProvableHQ/snarkOS/issues/4053.

### DIFF
--- a/node/bft/src/gateway.rs
+++ b/node/bft/src/gateway.rs
@@ -70,7 +70,6 @@ use snarkvm::{
     prelude::{Address, Field},
 };
 
-use aleo_std::aleo_ledger_dir;
 use colored::Colorize;
 use futures::SinkExt;
 use indexmap::IndexMap;
@@ -84,7 +83,6 @@ use rand::{
 };
 use std::{
     collections::{HashMap, HashSet},
-    fs,
     future::Future,
     io,
     net::{Ipv4Addr, SocketAddr, SocketAddrV4},
@@ -850,7 +848,7 @@ impl<N: Network> Gateway<N> {
     pub async fn shut_down(&self) {
         info!("Shutting down the gateway...");
         // Save the best peers for future use.
-        if let Err(e) = self.save_best_peers(&self.storage_mode, VALIDATOR_CACHE_FILENAME, None) {
+        if let Err(e) = self.save_best_peers(&self.storage_mode, VALIDATOR_CACHE_FILENAME, None, true) {
             warn!("Failed to persist best validators to disk: {e}");
         }
         // Abort the tasks.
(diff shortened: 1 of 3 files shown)


node/bft/src/gateway.rs
```diff
@@ -1119,6 +1126,30 @@ impl<N: Network> Gateway<N> {
     fn handle_banned_ips(&self) {
         self.tcp.banned_peers().remove_old_bans(IP_BAN_TIME_IN_SECS);
     }
+
+    // Update the dynamic validator whitelist.
+    fn update_validator_whitelist(&self) {
+        // Collect the addresses of all the connected validators.
+        let connected_validator_addrs = self.connected_peers();
```
vicsn: We should include validators suggested to us from the bootstrap node, otherwise those can never connect to us

node/bft/src/gateway.rs
```diff
@@ -1119,6 +1126,30 @@ impl<N: Network> Gateway<N> {
...
+            Ok(file) => file,
+            Err(e) => {
+                warn!("Couldn't create the validator whitelist file at {VALIDATOR_WHITELIST_FILENAME}: {e}");
+                return;
+            }
+        };
+
+        // Save all the addresses in a space-delimited format for trivial script extraction.
+        for addr in connected_validator_addrs {
+            if let Err(e) = writeln!(whitelist, "{} {}", addr.ip(), addr.port()) {
```
vicsn: Wouldn't just whitelisting the ip make more sense? We don't know the port through which they will connect

node/bft/src/gateway.rs
ljedrz: You mean the candidates? So we should optimistically whitelist them even if we're not yet sure if they are legitimate?

node/bft/src/gateway.rs
ljedrz: Some validators may have custom firewall setups; also, you may need administrator privileges to alter them.

ljedrz:
Not sure what went wrong with the devnet CI workflow, but it seems that it's not new.

node/bft/src/gateway.rs
vicsn: Indeed, as long as they don't crowd out the validator's we're sure about. So I think the list may just be the exact same as for the cached gateway peers?

node/bft/src/gateway.rs
vicsn: The main usecase we discussed in the issue to have a new separate file was that it would be easy for validator's to use as a dynamic whitelist for the default use-case, which is to whitelist the validator peer's ip.

I don't understand the relevance of the comment about admin privileges, but yes I'm assuming the validator which uses this list for their own firewall has permission to do so.

ljedrz: > I don't understand the relevance of the comment about admin privileges

Ah, I just misread it as the node performing the whitelisting operation itself instead ("just whitelisting the IP" as opposed to "whitelisting just the IP"). As for just using the IP, yeah, that'll work for the purpose of allowing the inbound connection, and will cover the other applications too.

node/bft/src/gateway.rs
ljedrz: Yep, the verified validators will always come first.

ljedrz:
Updated to use a minimally modified `PeerPoolHandling::save_best_peers`.

vicsn (approved):
LGTM!

PR #4056, opened 2026-01-05
[Draft] Upgrade to Rust 1.92

kaimast:
An upgrade to Rust 1.92 may slightly improve compile times. This PR tests if such an upgrade would break any CI.


.cargo/config.toml
```diff
@@ -1,9 +1,2 @@
-[target.x86_64-unknown-linux-gnu]
-rustflags = [
```
kaimast, line 2: This is enabled by default now, and we do not need to set it manually anymore.

PR #4057, opened 2026-01-05
Check missing previous certificates when syncing with batch header

vicsn:
## Motivation

^

## Test Plan

- [x] stress-tests need to pass after code review: prerelease-v4.5.3


raychu86 (commented):
Logic for this looks correct. Can this be unified by using the `IS_SYNCING` const generic?

If not, then it would be nice to unify by making the new flag a const generic as well.

vicsn:
> If not, then it would be nice to unify by making the new flag a const generic as well.

Done

raychu86 (approved):

PR #4059, opened 2026-01-07
[Feat] Add a network delay script

### LINKED ISSUE #4047: [Feature] Simulate network latency locally
## 🚀 Feature

It would be great if we can better test robustness of consensus logic locally. Not only is this cheaper, but it also makes it possible to do testing and iteration a lot faster.

To make this possible we'd need to introduce some kind of configurable delays for gateway/router messages. 

What are your thoughts on the following open questions:
- Where are the delays inserted?
- How do we specify the delays, can we change them at runtime?
- Which combination of delays do we care about? i.e. slow down some set of messages and validators.
- What factors could still cause this not to be representable? I.e. does this measurement cause a slow-down?

ljedrz:
This PR introduces a script which can be used to introduce kernel-level network delays, in order to test local nodes under more real-life network conditions (i.e. with some base latency), how the nodes run under high network stress, or when one or more of the nodes are less reliable than the rest.

The usage is described in the script itself.

Notes:
- the script is Linux-only
- it requires `sudo` to run (since we're adjusting kernel settings)
- these changes are ephemeral, and don't persist across OS restarts
- it doesn't matter if the nodes are already running or not
- the active config can be changed at any time
- every execution starts with a clear slate, i.e. to add an extra port, you need to include it with the ones from before

Closes https://github.com/ProvableHQ/snarkOS/issues/4047.

### DIFF
--- a/scripts/delay-network.sh
+++ b/scripts/delay-network.sh
@@ -1,0 +1,199 @@
+#!/bin/bash
+
+# Configuration
+INTERFACE="lo"
+PARENT="1:3"
+HANDLE="30:"
+
+function usage() {
+    cat <<EOF
+Usage: sudo ./delay-network.sh [COMMAND] [PORTS] [OPTIONS]
+
+Description:
+  A wrapper for Linux Traffic Control (tc) to simulate network conditions 
+  on specific ports of the loopback interface.
+
+Commands:
+  clear               Remove all delays and reset to normal.
+  status              Show current active rules and monitored ports.
+
+Presets (Simplest Usage):
+  lan [PORTS]         Congested LAN:    5ms  ± 2ms  (Normal dist)
+  wan [PORTS]         Trans-Oceanic:    200ms ± 50ms (Normal dist)
+  spikes [PORTS]      Lag Spikes:       20ms ± 800ms (Pareto dist)
+  bad-wifi [PORTS]    Packet Loss:      50ms ± 30ms + 2% Loss
+
+Custom Configuration:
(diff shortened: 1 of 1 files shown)


scripts/delay-network.sh
```diff
@@ -0,0 +1,199 @@
...
+
+Examples:
+  1. Simple Uniform Delay (100ms to 200ms):
+     sudo ./delay-network.sh custom 8080 150ms 50ms uniform
+
+  2. "Real" Internet (Correlation + Normal Dist):
+     sudo ./delay-network.sh custom 8000-8005 100ms "20ms 25%" normal
+
+  3. Extreme Stress (Pareto Spikes + Loss):
+     sudo ./delay-network.sh custom 8080 20ms "1000ms 25%" pareto 5%
```
kaimast, line 49: Would it be better if the suggested port ranges here would be something like `4000-6000`, so that snarkOS connections are affected by the filter?

ljedrz: These aren't suggestions, just examples; I don't expect this script to be popular with anyone other than the devs, and they'll know which ports interest them. I'd also rather not suggest anything specific in case other repos in our ecosystem want to use it for their needs - it's not snarkOS-specific after all.

kaimast: Okay. We can probably just keep it as is then. I thought we would like to have a script that provides a specific scenario that we test against for snarkOS, but that can also be added later.

kaimast (approved):

PR #4063, opened 2026-01-12
Fix the upgrade test with --auto-migrate-node-data

meddle0x53:
## Motivation

Fix the upgrade test to use --auto-migrate-node-data when running the current version as this is required for the upgrade

## Test Plan

Run the CI

### DIFF
--- a/.ci/upgrade_nodes_ci.sh
+++ b/.ci/upgrade_nodes_ci.sh
@@ -242,6 +242,10 @@ function start_node() {
     if [ "$node_index" -eq 0 ]; then
       flags+=( --metrics --no-dev-txs )
     fi
+
+    if [ "$bin" = "$SNARKOS_CURRENT_BIN" ]; then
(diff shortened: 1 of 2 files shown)


.ci/upgrade_nodes_ci.sh
```diff
@@ -242,6 +242,10 @@ function start_node() {
     if [ "$node_index" -eq 0 ]; then
       flags+=( --metrics --no-dev-txs )
     fi
+
+    if [ "$bin" = "$SNARKOS_CURRENT_BIN" ]; then
+      flags+=( --auto-migrate-node-data )
```
vicsn, line 248: Can you add a comment writing this won't be needed once old nodes are no longer on v4.4.0?

vicsn (dismissed):
LGTM! You can revert the changes to the CI file so we don't pollute origin/staging

vicsn (approved):

PR #4065, opened 2026-01-12
[Refactor] Remove BFT channels

kaimast:
The following changes are included in #4039, but I moved them to a separate PR so they are easier to review.

snarkOS relies on channels for communication between modules in the `consensus` and `bft` crates. This indirection exists mostly to avoid circular dependencies between modules, but has the following drawbacks.

* The code is generally harder to read as, instead of a simple function call, it relies on two channels: one to send the call and one to return the result to.
* A single operation now involves multiple `tokio` tasks, as the receiving side of each channel has a distinct task waiting on it.
* Error handling is more complicated, because panics do not propagate through channels easily.
* Channels can return errors, e.g., while waiting on a `oneshot::Receiver`, that we then need to handle in the code.
* Waiting on a channel is an async task and, thus, requires `await`.

The last point, especially, makes it harder to improve locking/coordination between different tasks as regular locks cannot be held across awaits and blocking tasks cannot call async functions. The former can be solved through async locks, but those are generally less efficient regular locks. The latter is harder to solve, as we use blocking tasks in multiple places while advancing the ledger.

This PR removes `BFTSender` and `BFTReceiver`, but leaves other channels in place. The BFT channels are replaced by two callback traits, one for the `Primary` and one for `Sync`. Instead of waiting on `BFTReceiver` the `BFT` struct now implements those to traits and registers itself as a callback for each.


node/bft/src/sync/mod.rs
```diff
@@ -59,10 +58,17 @@ use std::{
...
-    sync::{OnceCell, oneshot},
-    task::JoinHandle,
-};
+use tokio::{sync::oneshot, task::JoinHandle};
+
+/// This callback trait allows listening to synchronization updates, such as discorvering new `BatchCertificate`s.
+/// This is currently used by BFT.
+#[async_trait::async_trait]
+pub trait SyncCallback<N: Network>: Send + std::marker::Sync {
+    async fn sync_dag_at_bootup(&self, certificates: Vec<BatchCertificate<N>>) -> Result<()>;
```
```suggestion
    async fn sync_dag_at_bootup(&self, certificates: Vec<BatchCertificate<N>>);
```
I didn't see any fallible implementations

node/bft/src/sync/mod.rs
```diff
@@ -88,8 +94,8 @@ pub struct Sync<N: Network> {
     block_sync: Arc<BlockSync<N>>,
     /// The pending certificates queue.
     pending: Arc<Pending<Field<N>, BatchCertificate<N>>>,
-    /// The BFT sender.
-    bft_sender: Arc<OnceCell<BFTSender<N>>>,
+    /// The sync callback (used by [`BFT`]).
+    sync_callback: Arc<CallbackHandle<Arc<dyn SyncCallback<N>>>>,
```
ljedrz, line 98: I get why this was called a callback, but I can't help feeling that it's a misleading name - I was expecting this to be some function pointer.

node/bft/src/bft.rs
```diff
@@ -299,6 +283,41 @@ impl<N: Network> BFT<N> {
         is_ready
     }
 
+    /// Notification about a new certificated generated by `Primary` or received by the `Primary` from a peer.
+    async fn add_new_certificate(&self, certificate: BatchCertificate<N>) -> Result<()> {
+        // Update the DAG with the certificate.
+        self.update_dag::<false, true>(certificate).await
```
ljedrz: Shouldn't this be `true, false`? This is how primary certificate processing has worked so far.

ljedrz (commented):
The PR is directionally OK, but the devnet script appears to be failing (perhaps due to [this](https://github.com/ProvableHQ/snarkOS/pull/4065#discussion_r2804520291)?).

node/bft/src/sync/mod.rs
kaimast: Honestly, I just could not come up with a better name. Dynamic traits are essentially just fancy function pointers, though.

node/bft/src/bft.rs
kaimast: Good catch!

node/bft/src/sync/mod.rs
kaimast: Let me know if I should rename it to something else. "EventHandler" or "EventListener" is the only other option that I can think of.

ljedrz (approved):
LGTM 👍

node/bft/src/sync/mod.rs
```diff
@@ -59,10 +58,17 @@ use std::{
...
-use tokio::{
-    sync::{OnceCell, oneshot},
-    task::JoinHandle,
-};
+use tokio::{sync::oneshot, task::JoinHandle};
+
+/// This callback trait allows listening to synchronization updates, such as discorvering new `BatchCertificate`s.
+/// This is currently used by BFT.
+#[async_trait::async_trait]
+pub trait SyncCallback<N: Network>: Send + std::marker::Sync {
```
vicsn, line 66: Is it even worth having this trait if it only has a single impl? Do we foresee more uses?

vicsn (commented):
Shall we merge the pipeline of sync-changes into a feature branch, so that after they have been approved we can 1. run the prerelease-tests on the combined result and 2. merge them into staging in one go

node/bft/src/sync/mod.rs
kaimast: One reason for this abstraction was to avoid circular dependencies. For this specific channel it seems to be all contained within one crate, so this indirection might not be needed.

I still find it cleaner to have clearly defined "messages" between different components, similar to how the current channel design works.

node/bft/src/sync/mod.rs
kaimast: Assuming this is resolved, because you approved the PR.

kaimast:
@vicsn this should be good to merge. I ran multiple stress-tests on top of this branch, including a full prerelease test, when testing #4039.

PR #4066, opened 2026-01-12
[Feature] REST endpoint to query connected validators

kaimast:
## Motivation
Currently, one can query the connected router peers (through `{network}/peers/..`) but not validators connected through the gateway.

This PR adds a new set of endpoints `{network}/connection/bft/..` and `{network}/connections/p2p/..`. The former allows querying connected validators, while the latter is just an alias for the `{network}/peers/..` endpoint.

## Test Plan
The new endpoint is used in the sync benchmark to test that validators are fully connected to each other. Successfully running that benchmark validates that the endpoint works as expected.

## Backwards compatibility
Initially, I hoped we could just include the validators in the existing `peers` endpoint, but I added the new endpoints to avoid breaking existing functionality.

## Benchmark fixes
Recent changes on staging broke the benchmark workflow. This PR also contains a commit that fixes the workflow.

### DIFF
--- a/.ci/bench_bft_sync.sh
+++ b/.ci/bench_bft_sync.sh
@@ -66,7 +66,7 @@ PIDS[0]=$!
 # Stores the list of all validators.
 validators="127.0.0.1:5000"
 
-# Spawn the clients that will sync the ledger
+# Spawn the validator that will sync the ledger
 for node_index in $(seq 1 "$num_nodes"); do
   name="validator-$node_index"
 
@@ -125,6 +125,6 @@ while (( SECONDS < max_wait )); do
 done
 
 log "❌ Benchmark failed! Validators did not sync within 30 minutes."
-print_client_logs "$log_dir" "$num_validators" "$num_nodes"
+print_validator_logs "$log_dir" "$num_validators" "$num_nodes"
 
 exit 1
--- a/.ci/bench_bft_sync.sh
+++ b/.ci/bench_bft_sync.sh
@@ -91,12 +91,12 @@ wait_for_nodes $((num_nodes+1)) 0 "$network_name"
 
 SECONDS=0
 
-# TODO add API call for number of connected validators.
-#for ((node_index = 0; node_index < num_nodes+1; node_index++)); do
-#  if ! (wait_for_peers "$node_index" $num_nodes); then
-#    exit 1
-#  fi
-#done
+# Wait for all validators to be connected to each other via the gateway.
+for ((node_index = 0; node_index < num_nodes+1; node_index++)); do
+  if ! (wait_for_bft_connections "$node_index" $num_nodes "$network_name"); then
+    exit 1
+  fi
+done
 
 connect_time=$SECONDS
 echo "ℹ️ Nodes are fully connected (took $connect_time secs). Starting block sync measurement."
--- a/.ci/bench_bft_sync.sh
+++ b/.ci/bench_bft_sync.sh
@@ -25,17 +25,17 @@ poll_interval=1 # Check block heights every second
 #shellcheck source=SCRIPTDIR/utils.sh
 . .ci/utils.sh
 
-branch_name=$(git rev-parse --abbrev-ref HEAD)
-echo "On branch: ${branch_name}"
-
-network_name=$(get_network_name $network_id)
-echo "Using network: $network_name (ID: $network_id)"
-
-snapshot_info=$(<info.txt)
-echo "Snapshot_info: ${snapshot_info}"
-
 # Create log directory
 init_log_dir
+
+branch_name=$(git rev-parse --abbrev-ref HEAD)
+log "On branch: ${branch_name}"
+
+network_name=$(get_network_name $network_id)
+log "Using network: $network_name (ID: $network_id)"
+
+snapshot_info=$(<info.txt)
+log "Snapshot_info: ${snapshot_info}"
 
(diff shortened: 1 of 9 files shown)


.ci/utils.sh
```diff
@@ -272,6 +272,37 @@ function wait_for_peers() {
...
+  
+  while (( total_wait < max_wait )); do
+    result=$(curl -s "http://localhost:$port/v2/$network_name/connections/bft/count")
+
+    if ! (is_integer "$result"); then
+      log "Failed to get number of BFT connections for node #${node_index} (port=$port). Will retry..."
+    elif (( result < min_peers )); then
+      log "Node #${node_index} (port=$port) has $result BFT connections, expected at least $min_peers. Will wait and retry..."
+    else
+      return 0
```
esren0x, line 294: I might not have full context, but looks like this return would potentially show a "sucess" even tho it failed. By logging and then returning 0 it would appear to have worked and not get tot he message at the end for 

`  log "❌ BFT connections did not reach $min_peers within 5 minutes."`

.ci/bench_bft_sync.sh
```diff
@@ -124,7 +124,7 @@ while (( SECONDS < max_wait )); do
   sleep $poll_interval
 done
 
-echo "❌ Benchmark failed! Validators did not sync within 30 minutes."
+log "❌ Benchmark failed! Validators did not sync within 30 minutes."
 print_client_logs "$log_dir" "$num_validators" "$num_nodes"
```
esren0x:
```suggestion
print_validator_logs "$log_dir" $((num_nodes+1)) 0
```

esren0x (commented):
Noticed README also has the list of routes, but not these new ones. 
The new ones from a structure POV look good, just noticed a few extra bits when reviewing. And not sure if you guys usually add integration tests for these?

.ci/utils.sh
kaimast: I think this is correct. It first checks that it is an integer, then if the value is large enough. If both of these checks succeed, the return should show a success.

Otherwise, it will sleep and retry, or timeout.

.ci/bench_bft_sync.sh
kaimast: Thanks! Fixed it.

kaimast:
> Noticed README also has the list of routes, but not these new ones. The new ones from a structure POV look good, just noticed a few extra bits when reviewing. And not sure if you guys usually add integration tests for these?

I am actually not sure what list you are referring to. Do you mean the README for the crate, the repo, or something else?

.ci/bench_bft_sync.sh
```diff
@@ -66,7 +66,7 @@ PIDS[0]=$!
 # Stores the list of all validators.
 validators="127.0.0.1:5000"
 
-# Spawn the clients that will sync the ledger
+# Spawn the validator that will sync the ledger
```
ljedrz:
```suggestion
# Spawn the validators that will sync the ledger
```

.ci/devnet_ci.sh
```diff
@@ -17,7 +17,7 @@ min_height=$4
 max_warnings=$5
 
 # The verobsity of snarkos nodes.
-NODE_VERBOSITY=3
+NODE_VERBOSITY=0
```
ljedrz: why the lower verbosity? won't this obscure the nature of any potential issues?

.ci/devnet_ci.sh
```diff
@@ -101,13 +101,19 @@ done
 # Ensure all nodes are up and running.
 wait_for_nodes "$total_validators" "$total_clients" "$network_name"
 
-# Wait for all clients to be connected to another client or a validator.
-# TODO(kaimast): also check that validators are connected to each other.
+# Wait for validators to be fully connected.
```
ljedrz, line 104: if this is waiting for a full mesh, then the `log` output below should be adjusted; otherwise the comment is a bit misleading

node/rest/src/lib.rs
```diff
@@ -200,10 +200,13 @@ impl<N: Network, C: ConsensusStorage<N>, R: Routing<N>> Rest<N, C, R> {
             .route("/find/transactionID/{transition_id}", get(Self::find_transaction_id_from_transition_id))
             .route("/find/transitionID/{input_or_output_id}", get(Self::find_transition_id))
 
-            // GET ../peers/..
+            // GET ../peers/.. (with /connections/p2p/.. aliases)
```
ljedrz: nit: I'd say that the `/peers` are aliases, since they are less specific

ljedrz (dismissed):
LGTM in general, left some suggestions

raychu86 (dismissed):
REST logic looks fine. Did not review scripts

.ci/devnet_ci.sh
kaimast: Fixed

node/rest/src/lib.rs
kaimast: Good point

.ci/devnet_ci.sh
kaimast: Lowered it while I was debugging the script, but will set it back to 3.

kaimast:
All comment should have been addressed.

ljedrz (approved):
LGTM; the devnet CI job error seems to be transient/unrelated,

vicsn (approved):

Issue #4062, opened 2026-01-09
[Feature] Use network delay script in CI


vicsn:
## 🚀 Feature

Once https://github.com/ProvableHQ/snarkOS/pull/4059 is approved, it would be great if we can incorporate it into CI to catch consensus issues. What has worked reliably for us in the past, was to:
- repeatedly remove the ledger of a quorum **minority** (f out of (3f+1)) of the validators and restart them.
- once stop a quorum **majority** (f+1 out of (3f+1)) for 30 seconds and restart

We should be able to re-use large parts of `.ci/upgrade_nodes_ci.sh` for this new script, very GPT-able.

We should try to use a github runner (see https://github.com/ProvableHQ/snarkVM/pull/3074 for an example) and comment out all other jobs so we don't burn through money as we test this in the CI environment. 

Ideally we also pin cores to specific validator nodes. The basic free runner has only 4 cores, so that might not be sufficient for a proper test. We already have access to the `ubuntu-latest-l` machine. We can also request permission to use even larger github runners.`.

PR #4069, opened 2026-01-13
[CI] Chaotic devnet test

ljedrz:
This PR introduces a `chaotic-network-runner.sh` script which can be used to simulate chaotic network conditions while running other scripts. As a an example, it is currently applied to `devnet_ci.sh` as a new CI job.

All the other CI jobs are temporarily removed until this is ready.

Filing this as a draft until it's decided which CI should be used for it, and which jobs to enhance with it (and whether those would be new jobs, or updates to the existing ones).


scripts/chaotic-network-runner.sh
```diff
@@ -0,0 +1,90 @@
...
+if [ $# -eq 0 ]; then
+    echo "Error: No test command provided."
+    exit 1
+fi
+
+# Chaos Timing (in seconds)
+MIN_WAIT=10
+MAX_WAIT=30
+MIN_DURATION=15
+MAX_DURATION=30
```
vicsn, line 30: I think we can simplify with fixed waits and durations. Despite all the chaos, we should aim for reproducible results as much as possible.

scripts/chaotic-network-runner.sh
```diff
@@ -0,0 +1,90 @@
...
+    echo "[Chaotic Network Runner] 🎲 Background chaos loop started on ports $TARGET_PORTS."
+
+    while true; do
+        # 1. Stay Clean
+        WAIT_TIME=$(shuf -i $MIN_WAIT-$MAX_WAIT -n 1)
+        echo "[Chaos Runner] 🟢 Network healthy for ${WAIT_TIME}s..."
+        sleep $WAIT_TIME
+
+        # 2. Select Random Preset
+        RANDOM_PRESET=${PRESETS[$RANDOM % ${#PRESETS[@]}]}
```
vicsn, line 57: Instead of a random preset, can we deterministically iterate through them?

scripts/chaotic-network-runner.sh
```diff
@@ -0,0 +1,90 @@
...
+fi
+
+# Chaos Timing (in seconds)
+MIN_WAIT=10
+MAX_WAIT=30
+MIN_DURATION=15
+MAX_DURATION=30
+
+# Presets from delay-network.sh
+PRESETS=("lan" "wan" "spikes") # "bad-wifi" might be too volatile for a generic test
```
vicsn, line 33: Just checking, do all three of them invoke meaningfully abnormal behaviour? In the original discussion we only talked about invoking multi-second message delays to trigger race conditions and bugs.

.github/workflows/chaotic-devnet.yml
```diff
@@ -0,0 +1,66 @@
...
+        run: |
+          # Make it executable (permissions may get lost)
+          chmod +x bin/snarkos
+
+          # Add the 'bin' folder to the system PATH
+          echo "$(pwd)/bin" >> $GITHUB_PATH
+
+      - name: Run Tests with Chaos
+        # Arguments: <PORT_RANGE> <TEST_COMMAND>
+        run: ./scripts/chaotic-network-runner.sh 5000-5003 ./.ci/devnet_ci.sh 4 2 0 45
```
vicsn: The original issue asks for a specific set of behaviours which is not the same as what `devnet_ci.sh` does: https://github.com/ProvableHQ/snarkOS/issues/4062

vicsn (changes requested):

scripts/chaotic-network-runner.sh
ljedrz: it would be simpler, yes, but I would argue that in such a test, the only reproducibility that we want is "it doesn't break" - otherwise we might accidentally trigger the problematic conditions exactly during the times when the node can easily survive it, while neglecting a situation where it would break in production

scripts/chaotic-network-runner.sh
ljedrz: I would argue the same as with the random intervals here - a deterministic cycle might fail to uncover some edge cases

scripts/chaotic-network-runner.sh
ljedrz: no, but that's part of the point - the "plain" `devnet_ci` job works under perfect network conditions, while in theory, even a small LAN delay could be impactful; I've already run all of these scenarios for extended periods of time, but it's not the point of such a CI job to trigger issues - it's to discover them if introduced by some new changes

ljedrz: in other words, if we see an `ERROR` in a `spikes` phase, it's cause for investigation; if it happens under `LAN`, it's cause for concern

.github/workflows/chaotic-devnet.yml
ljedrz: do we already have a test which restarts validators and potentially removes some of their ledgers? if not, we should have one that's separate from this script - it's designed to apply random network stress to existing scenarios, which allows us to use it in tandem with any base scenario

vicsn: Not yet, can you write such a script? `.ci/upgrade_nodes_ci.sh` comes closest in functionality.

There's an open question of what machine size / number of validators / runtime reliably triggers issues. If we can reliably trigger issues on a heavy machine within 30 minutes, that would be amazing, and a 40x improvement over our status quo (where we launch 40 machines).

 I would first test the basic functionality on the free default github actions machine, and then migrate it back to CircleCI where we make part of the `release` job (and we can manually trigger it).

scripts/chaotic-network-runner.sh
kaimast: To add to this, these tests will always be random due to scheduling etc.

I don't think the goal should be for us to reproduce specific failures, but for it to serve as an early detection mechanism that something is wrong.

ljedrz:
Updated the description and rebased now that the PR this had been built on (https://github.com/ProvableHQ/snarkOS/pull/4059) has been merged. There were no changes to the contents of the script or the job.

ljedrz:
I added a [script](https://github.com/ProvableHQ/snarkOS/pull/4069/commits/8ebadaecb998f7f858d57f7d8d71072e89b7311a) according to the following:
> repeatedly remove the ledger of a quorum minority (f out of (3f+1)) of the validators and restart them.

ljedrz:
I added a [script](https://github.com/ProvableHQ/snarkOS/pull/4069/commits/53af408e43f91dcb5373ce3e3664d4d2743c26a8) according to the following:

> once stop a quorum majority (f+1 out of (3f+1)) for 30 seconds and restart

scripts/chaotic-network-runner.sh
```diff
@@ -0,0 +1,90 @@
+#!/bin/bash
+
```
meddle0x53, line 2: It is a good idea to add strict mode at the top to catch unset vars/pipeline failures/whitespace issues. If there is a bug it will make it easier to fail and debug. I learned this from bad experience with my last scripts.

```suggestion
set -euo pipefail
IFS=$'\n\t'
```

scripts/chaotic-network-runner.sh
```diff
@@ -0,0 +1,90 @@
...
+# ==========================================
+# CONFIGURATION
+# ==========================================
+
+# 1. Require Port Range (Argument 1)
+TARGET_PORTS=$1
```
meddle0x53, line 8: It is good to quote all the vars (as I did in the PR for the delay-network.sh), because unquoted variables change meaning depending on their contents, and Bash won’t warn us. We can pass them in a bad way in the CI and then debug why some white spaces are there.

scripts/chaotic-network-runner.sh
```diff
@@ -0,0 +1,90 @@
...
+
```
meddle0x53, line 9: We can add 

```
die() {
  echo "Error: $*" >&2
  exit 1
}
```

function for the errors and also log function for clear logs - like I did for the delay-network.sh script. Helps debugging, not important, of course.

Also 

```
need_cmd sudo
need_cmd shuf

NETWORK_SCRIPT="./scripts/delay-network.sh"
[[ -f "$NETWORK_SCRIPT" ]] || die "Network script not found: $NETWORK_SCRIPT"
[[ -x "$NETWORK_SCRIPT" ]] || die "Network script is not executable: $NETWORK_SCRIPT"

```

would be a good check.

scripts/chaotic-network-runner.sh
```diff
@@ -0,0 +1,90 @@
...
+CHAOS_PID=$!
+
+# Run the actual test command (passed as arguments)
+echo "[Chaos Runner] 🚀 Starting main test command: $@"
+"$@"
+EXIT_CODE=$?
+
+# Cleanup
+echo "[Chaos Runner] 🏁 Main command finished with exit code $EXIT_CODE."
+kill $CHAOS_PID 2>/dev/null
```
meddle0x53, line 87: Maybe it can be part of a trap. It is not important as it is a CI run and the CI job will just exit and all will be cleared, I guess, but for local runs later is a good way to cleanup:

```
cleanup() {
    if [ -n "${CHAOS_PID:-}" ]; then
        kill "$CHAOS_PID" 2>/dev/null || true
        wait "$CHAOS_PID" 2>/dev/null || true
    fi
    reset_network
}

trap cleanup EXIT
```

.ci/utils.sh
```diff
@@ -156,6 +156,49 @@ function stop_nodes() {
...
+  echo "🚨 Stopping ${#indices[@]} selected node(s)..."
+
+  for i in "${indices[@]}"; do
+    # Get the PID from the global PIDS array using the index
+    local pid="${PIDS[$i]}"
+
+    # Check if PID exists (is not empty) and is currently running
+    if [[ -n "$pid" ]] && kill -0 "$pid" 2>/dev/null; then
+      echo "Killing PIDS[$i] -> $pid"
+      kill -9 "$pid" 2>/dev/null || true
```
meddle0x53, line 188: Do you want to kill the nodes with `-9`, better SIGTERM? Maybe it can lead to some problems with the ledger and random failures, maybe not, just asking?

Otherwise these two functions seem great to me.

.ci/reset_quorum_minority.sh
```diff
@@ -0,0 +1,84 @@
+#!/bin/bash
+
```
meddle0x53, line 2: The script itself seems to do its idea right - in a great way.

If you want you can do some of my suggestions I added to `chaotic-network-runner.sh`.
The strict mode, variables use helpers for logging, actually shared helper functions for logging and exiting are coming from utils.sh in [this PR](https://github.com/ProvableHQ/snarkOS/pull/4079). Maybe use them + strict mode + quoting the `$` vars.

.ci/reset_quorum_majority.sh
```diff
@@ -0,0 +1,85 @@
+#!/bin/bash
+
```
meddle0x53, line 2: Same comment as on the minority script.

.ci/reset_quorum_majority.sh
```diff
@@ -0,0 +1,85 @@
...
+#shellcheck source=SCRIPTDIR/utils.sh
+. ./.ci/utils.sh
+
+# Network parameters
+total_validators=7
+majority=$(( (total_validators - 1) / 3 + 1 ))
```
meddle0x53, line 8: Shouldn't this be over 50% to be majority?

meddle0x53 (commented):
So some suggestions/comments from me.

kaimast:
I am closing this in favor of #4095.

PR #4072, opened 2026-01-14
deduplicate peers when storing to a file

### LINKED ISSUE #4071: [Bug] validator-whitelist contains duplicates
## 🐛 Bug Report

List of validator IP addresses contains duplicates. 

## Steps to Reproduce

1. Run validator
2. Check validator-whitelist:
```
./snarkos/node-data$ cat validator-whitelist | wc -l
67
./snarkos/node-data$ cat validator-whitelist | uniq |wc -l
65
```

## Expected Behavior

List contains only unique IP addresses

## Your Environment

SnarkOS: `snarkos unknown_branch b1080b24ff45e7f5d9b5a72e6c951e2d0baa904b features=[default,snarkos_node_metrics,telemetry]`

14MR:
## Motivation

Resolves #4071

## Test Plan

I tested the change locally with unit test (test is not included in the diff)

## Backwards compatibility

The code is backwards compatible

### DIFF
--- a/.ci/upgrade_nodes_ci.sh
+++ b/.ci/upgrade_nodes_ci.sh
@@ -242,6 +242,10 @@ function start_node() {
     if [ "$node_index" -eq 0 ]; then
       flags+=( --metrics --no-dev-txs )
     fi
+
+    if [ "$bin" = "$SNARKOS_CURRENT_BIN" ]; then
+      flags+=( --auto-migrate-node-data )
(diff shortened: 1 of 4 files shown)


ljedrz (approved):
LGTM, though note that this only makes a difference for addresses stored without ports, as the source list of peers is already a `HashMap<SocketAddr, Peer<N>>`.

PR #4073, opened 2026-01-15
[Fix] committee query endpoint

vicsn:
## Motivation

Improve debugging experience.


ljedrz (approved):
More importantly, this fixes the query address; LGTM.

PR #4074, opened 2026-01-16
[Fix] Introduce `InsertBlockResponseError`

kaimast:
This PR introduces a new `InsertBlockResponseError` that allows differentiating between the more-severe consensus version mismatch and other block insertion errors.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -288,7 +288,7 @@ dependencies = [
  "rustversion",
  "serde",
  "sync_wrapper",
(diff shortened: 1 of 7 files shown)


vicsn (dismissed):

kaimast:
I removed the `snarkvm` upgrade from this PR as it causes tests `snarkvm-node-router-messages` to fail. We can open a separate PR for it.

vicsn (approved):

PR #4082, opened 2026-01-23
Suggest improvements on the delay-network.sh

meddle0x53:
## Motivation

Some suggestions to improve the delay-network.sh for stability, debugging and local usage.

### DIFF
--- a/scripts/delay-network.sh
+++ b/scripts/delay-network.sh
@@ -15,14 +15,16 @@ function die() {
   exit 1
 }
 
-# Check if the a command this script depends on exist (I do that for all bash scripts I write that depends on commands)
+# Check if the a command this script depends on exist
 function need_cmd() {
     command -v "$1" >/dev/null 2>&1 || die "Missing required command: $1"
 }
 
-# Print the command before executing - helps with debugging if something goes wrong:
+# Print the command before executing
 function run() {
-    echo "  + $*"
+    # We use a subshell ( ... ) to temporarily set IFS to space for the echo command only.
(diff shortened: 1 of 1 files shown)


scripts/delay-network.sh
```diff
@@ -15,14 +15,16 @@ function die() {
   exit 1
 }
 
-# Check if the a command this script depends on exist (I do that for all bash scripts I write that depends on commands)
+# Check if the a command this script depends on exist
```
meddle0x53, line 18: Yeah, I added the comment to explain to you why I do it with the idea that you'll remove it.

scripts/delay-network.sh
```diff
@@ -15,14 +15,16 @@ function die() {
...
     command -v "$1" >/dev/null 2>&1 || die "Missing required command: $1"
 }
 
-# Print the command before executing - helps with debugging if something goes wrong:
+# Print the command before executing
 function run() {
-    echo "  + $*"
+    # We use a subshell ( ... ) to temporarily set IFS to space for the echo command only.
+    # This ensures "  + tc qdisc..." prints on one line despite the global IFS setting.
+    ( IFS=' '; echo "  + $*" )
```
meddle0x53, line 27: Great!

meddle0x53 (commented):
I can not approve it as it is mine, but fixes the problems I introduced, which is the important thing.

ljedrz (approved):
LGTM; I'm not a `bash` expert, but as far as I can tell this makes the script stricter, validates some of the main inputs, and makes it easier to debug. I've tested it locally with a few of the commands and it works like before.

Issue #4000, opened 2025-11-06
[Bug] Streamline peer banning and block sync


vicsn:
## 🐛 Bug Report

- We should ban in dev mode
- We should revisit banning on BlockResponse timeouts, because we ban peers when we restart our node
- We should add some tests to make sure it is handled correctly and we don't regress
- We should fix the following edge-case:
  - A validator sends out block requests to a peer
  - The peer's Router disconnects due to peer refresh
  - Despite this log `Block sync is removing all block requests to peer` - no block requests are removed
  - 10 minutes later, the block requests time out and we see `IP-banning 44.251.223.68 reason: timed out on block requests`

Tested on commit hash: 7a3f235e3

kaimast:
It might be good to postpone this until https://github.com/ProvableHQ/snarkOS/pull/3670 is done. There are some interesting subtleties to consider, such as differentiating between malicious nodes (sending invalid blocks) and other faulty/slow nodes (simply not responding in time).

PR #4085, opened 2026-01-28
[Fix] Ban peers with invalid/outdated consensus versions

kaimast:
This PR ensures that nodes do not repeatedly attempt to sync from a peer that does not have the correct consensus version.

### DIFF
--- a/node/sync/src/block_sync.rs
+++ b/node/sync/src/block_sync.rs
@@ -473,36 +473,47 @@ impl<N: Network> BlockSync<N> {
         blocks: Vec<Block<N>>,
         latest_consensus_version: Option<ConsensusVersion>,
     ) -> Result<(), InsertBlockResponseError> {
-        let Some(last_height) = blocks.as_slice().last().map(|b| b.height()) else {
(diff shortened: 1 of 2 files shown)


node/sync/src/block_sync.rs
```diff
@@ -473,36 +473,47 @@ impl<N: Network> BlockSync<N> {
         blocks: Vec<Block<N>>,
         latest_consensus_version: Option<ConsensusVersion>,
     ) -> Result<(), InsertBlockResponseError> {
-        let Some(last_height) = blocks.as_slice().last().map(|b| b.height()) else {
-            return Err(InsertBlockResponseError::EmptyBlockResponse);
-        };
+        // Attempt to insert the block responses, and break if we encounter an error.
```
kaimast, line 476: Before this change, we did not remove the block requests to the peer on invalid consensus versions.
Not sure if that made a difference, because the requests should be removed on disconnect anyway, but I made the change to be safe.

vicsn:
Previous issue on the topic: https://github.com/ProvableHQ/snarkOS/issues/4000

kaimast:
Closing in favor of #4087.

PR #4086, opened 2026-01-28
Increase max transaction size 

raychu86:
## Motivation

This PR updates the snarkVM rev and adds support for the new features in the snarkVM [sister PR](https://github.com/ProvableHQ/snarkVM/pull/3087):

- Increase max TX size
- Increase max writes
- Increase max program size

Do not merge this PR until the snarkVM PR is merged.

### DIFF
--- a/node/bft/src/primary.rs
+++ b/node/bft/src/primary.rs
@@ -605,23 +605,9 @@ impl<N: Network> Primary<N> {
                             }
                         })?;
 
-                        // TODO (raychu86): Remove this logic after the next migration height is reached.
-                        // ConsensusVersion Migration logic -
-                        // Do not include deployments in a batch proposal.
+                        // Fetch the current block height and consensus version.
                         let current_block_height = self.ledger.latest_block_height();
-                        let consensus_version_v7_height = N::CONSENSUS_HEIGHT(ConsensusVersion::V7)?;
-                        let consensus_version_v8_height = N::CONSENSUS_HEIGHT(ConsensusVersion::V8)?;
                         let consensus_version = N::CONSENSUS_VERSION(current_block_height)?;
-                        if current_block_height > consensus_version_v7_height
-                            && current_block_height <= consensus_version_v8_height
-                            && transaction.is_deploy()
-                        {
-                            trace!(
-                                "Proposing - Skipping transaction '{}' - Deployment transactions are not allowed until Consensus V8 (block {consensus_version_v8_height})",
-                                fmt_id(transaction_id)
-                            );
-                            continue;
-                        }
 
                         // Compute the transaction spent cost (in microcredits).
                         // Note: We purposefully discard this transaction if we are unable to compute the spent cost.
@@ -916,20 +902,9 @@ impl<N: Network> Primary<N> {
                         }
                     })?;
 
-                    // TODO (raychu86): Remove this logic after the next migration height is reached.
-                    // ConsensusVersion Migration logic -
-                    // Do not include deployments in a batch proposal.
-                    let consensus_version_v7_height = N::CONSENSUS_HEIGHT(ConsensusVersion::V7)?;
-                    let consensus_version_v8_height = N::CONSENSUS_HEIGHT(ConsensusVersion::V8)?;
+                    // Fetch the current consensus version.
                     let consensus_version = N::CONSENSUS_VERSION(block_height)?;
-                    if block_height > consensus_version_v7_height
-                        && block_height <= consensus_version_v8_height
-                        && transaction.is_deploy()
-                    {
-                        bail!(
-                            "Invalid batch proposal - Batch proposals are not allowed to include deployments until Consensus V8 (block {consensus_version_v8_height})",
-                        )
-                    }
+                    // TODO (raychu86): Remove this logic after the next migration height is reached.
                     // Enforce maximum transaction size.
                     if consensus_version < ConsensusVersion::V14 {
                         // Retrieve the maximum transaction size for the consensus version.
--- a/node/bft/src/primary.rs
+++ b/node/bft/src/primary.rs
@@ -622,25 +622,6 @@ impl<N: Network> Primary<N> {
                             );
                             continue;
                         }
-                        // Enforce maximum transaction size.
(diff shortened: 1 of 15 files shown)


node/bft/src/primary.rs
```diff
@@ -928,6 +947,23 @@ impl<N: Network> Primary<N> {
...
+                            // Ensure the transaction does not exceed the maximum size.
+                            if transaction.to_bytes_le()?.len() > max_tx_size as usize {
+                                trace!(
+                                    "Proposing - Skipping transaction '{}' - Exceeds maximum transaction size of {max_tx_size} bytes",
+                                    fmt_id(transaction_id),
+                                );
+                                continue;
+                            }
+                        }
+                    }
```
vicsn, line 921: After additional thought, I'm not certain this approach is safe.

First, when we propose we use `let current_block_height = self.ledger.latest_block_height();` whereas when we sign we use `let block_height = self.ledger.latest_block_height() + 1;`. This seems off.

Second, even if we use the same heights, I think there is a theoretical race condition where all validators propose at height X, the network advances N blocks, and now the signers are at height X+1, meaning no certificates are created. Perhaps there is some obscure detail which prevents this from happening, but I don't think anyone knows this by heart at the moment and a formal investigation would take days.

If you agree, can you prune the outdated (broken) spend limit logic from here and document these findings in this part of the code?

I think we're better off relying on consensus gated checks in `check_transaction` only. As soon as nodes upgrade, they can technically deserialize large transactions, but the honest quorum will only include them in their proposals at the new consensus height.

node/bft/src/primary.rs
```diff
@@ -622,6 +622,25 @@ impl<N: Network> Primary<N> {
...
+                                // Ensure the transaction does not exceed the maximum size.
+                                if transaction.to_bytes_le()?.len() > max_tx_size as usize {
+                                    trace!(
+                                        "Proposing - Skipping transaction '{}' - Exceeds maximum transaction size of {max_tx_size} bytes",
+                                        fmt_id(transaction_id),
+                                    );
+                                    continue;
+                                }
+                            }
+                        }
```
vicsn: ~Same as the other comment I think we can remove~

UPDATE: my latest thought still is that we can remove this one but keep the other one.

node/bft/src/primary.rs
vicsn: Ugh sorry the above recommendation is incomplete.


Only relying on `check_transaction` does not guard against a malicious proposer who includes a large transaction while only half of the honest quorum is upgraded.

raychu86: It is known that on the snarkOS side, these checks are not full guarantees, but more probabilistic with the intention of having all validators advancing to the newest block before starting to allow transactions on the new rule. 

The issue is because the validators operate on rounds rather than concrete block heights, causing the issue here.

raychu86: One better approach is for validators to not include those transactions until they are at least 50 blocks past the migration height, which would guarantee that all validators should be past the migration before starting to process them.

raychu86: Actually i think it is safe without these checks at all,  our `check_transaction` logic will abort malicious transactions during block construction. 

Case 1: Validator 1 has not committed block height V14, while others have.
1. See's other proposals with large txes.
2. Will not sign proposal.
3. Will sync up with other certificates and advance to V14 before signing.

Case 2: Malicious validator injects a large TX (or a TX with new rules) when nodes are still on V13
1. Validators will attempt to commit that TX to a block
2. That TX will be aborted 
3. Nodes continue.

I believe the only reason we have the V7 - V8 deployment prevention was for UX and reducing the number of programs that needed VK upgrades, rather than safety.

vicsn: I think we can't fully rely on `check_transaction` because the transaction size directly impacts proposal signing at the snarkOS layer. The edge case I'll share below is an unlikely worst case scenario - but given how much time it takes us to reach a solid conclusion I'm mostly hoping we can set the right precedent for documentation and code structure so we save time in the future and lose less money on bug bounty findings.

The attack in more detail: 
1. `2f+1` nodes upgraded, including a malicious validator, everyone is still on V13. Note that this doesn't break any quorum assumptions on **honesty**, but because the honest quorum is not **available**, it will lead to a **safety** issue regardless in the following steps.
2. the malicious validator creates a proposal with 1 valid and 1 invalid (large) TX.
3. The `2f+1` quorum accepts the valid TX and aborts the invalid (large) TX, the `f` nodes which didn't upgrade yet can't deserialize the invalid transmission and will stall until the GC limit is reached and they switch to block syncing.
4. The malicious validator now has 100 rounds to wreak havoc.

I think the **minimum** defense is the check which this PR introduces in `fn process_batch_propose_from_peer`. IMO this can use the actual consensus migration height which is communicated with all stakeholders as cutoff, adding an extra 50 blocks leeway is just adding complexity.

And I would remove the check in `fn propose_batch` because honest validators call `check_transaction` already on the transmissions they include.

Also note that if in the future we were to hypothetically introduce a **lowering** of the allowed maximum deserialization size, then we run into the race condition I mentioned at the top of this comment chain, and we may want to add a 50 block leeway. If you agree, maybe you can explain in a comment as a reminder for us?

raychu86: 3. Isn't a problem, because the assumption is that 2f+1 honest validators are required to upgrade to the latest version before the new ConsensusVersion even kicks in. We've operated under the assumption of human validation of upgrades rather than a programatic one for these cases.

I'll leave the check in `process_batch_propose_from_peer` for now, but I still believe it is not necessary. The one in `propose_batch` will be removed. 

We already have a comment in `snarkVM` reminding us about implications of decreasing the max transaction size, but I will leave another in snarkOS as well.

```
    // Note: This value must **not** decrease without considering the impact on transaction validity.
```

vicsn: > Isn't a problem, because the assumption is that 2f+1 honest validators are required to upgrade to the latest version before the new ConsensusVersion even kicks in.

In my latest example, the new consensus version didn't kick in yet, it may be 2 weeks or months out still.

vicsn (approved):

PR #4087, opened 2026-01-29
Test partial network upgrade in CI

kaimast:
Replaces #4085.

This PR makes two changes:
1. Nodes will ban peers if they run an outdated consensus version. Peers are still not banned if they time out, as we need to do more testing/design work for that.
2. Added a script that tests a partial upgrade of the network, where some nodes did not update in time. This is not enabled in CI yet, but can be run by invoking `.ci/partial_upgrade_test.sh`.


.ci/partial_upgrade_test.sh
```diff
@@ -0,0 +1,362 @@
...
+SNARKOS_RELEASE_DIR="${SNARKOS_RELEASE_DIR:-$PWD/.ci/release-snarkos}"
+SNARKOS_RELEASE_BIN="${SNARKOS_RELEASE_BIN:-$SNARKOS_RELEASE_DIR/bin/snarkos}"
+SNARKOS_RELEASE_VERSION_FILE="${SNARKOS_RELEASE_VERSION_FILE:-$SNARKOS_RELEASE_DIR/VERSION}"
+
+download_and_build_latest_snarkos() {
+  require_cmd curl
+  require_cmd tar
+  require_cmd cargo
+
+  local force_build="${1:-0}"
```
vicsn: I think `upgrade_nodes_ci.sh` was unfortunately built with a lot of AI which added complexity and makes the script harder to maintain and read. Can you remove variables which aren't actually used, e.g. `force_build` is immutable.

Normally I'd avoid the scope creep but I'm having a hard time reviewing the script logic.

.ci/partial_upgrade_test.sh
```diff
@@ -0,0 +1,362 @@
...
+
+########################################
+# MAIN
+########################################
+
+log "Starting partial_upgrade_test.sh"
+log "total_validators=${total_validators}, network_id=${network_id}"
+log "WAIT_BETWEEN_UPGRADES=${WAIT_BETWEEN_UPGRADES}"
+
+# 1. Build release snarkos once (without heights override yet).
```
vicsn: > (without heights override yet).

We concluded the heights override was a run-time not a build-time setting right? Can you reword and fix it also in `upgrade_nodes_ci.sh`

.ci/partial_upgrade_test.sh
```diff
@@ -0,0 +1,362 @@
...
+latest_height=$(get_block_height "$last_node" "$network_name")
+
+log "Clearing ledger for node $last_node and restarting with updated binary"
+stop_node "$last_node"
+log_file="$log_dir/validator-$last_node.log"
+"$SNARKOS_RELEASE_BIN" clean "--dev=$last_node" "--network=$network_id"
+start_node "$SNARKOS_CURRENT_BIN" "$last_node" "validator" "$log_file"
+
+log "Waiting for node $last_node to reach new consensus version.."
+if ! wait_for_new_consensus_version "$last_node" "$network_name" "$latest_release_version"; then
```
vicsn: I only see `latest_release_version` being set as a `local` within a function. Does the variable leak outside of it? If that is shell being silly shell, I'd consider rewriting so it's readable by folks who are not shell experts.

.ci/partial_upgrade_test.sh
```diff
@@ -0,0 +1,362 @@
...
+####################################################
+# Partial network ugprade test:
+#
+# 1. Download latest snarkOS release source from GitHub.
+# 2. Build it via `cargo install --locked --path . --features test_network`
+#    into a separate prefix (SNARKOS_RELEASE_DIR).
+# 3. Probe latest_consensus_version for latest release, and assume the PR has a consensus version that is one higher.
+# 5. Start the devnet with latest release binaries.
+# 6. Wait for consensus version to stabilize
+# 7. Upgrade a majority of the network (2f+1) nodes to the new version
```
vicsn: Where is this happening? I only see "Upgrade one node to the new version"

node/sync/src/block_sync.rs
```diff
@@ -473,36 +473,47 @@ impl<N: Network> BlockSync<N> {
...
+            for block in blocks {
+                if let Err(error) = self.insert_block_response(peer_ip, block) {
+                    break 'outer Err(error.into());
+                }
             }
+
+            Ok(())
+        };
+
+        // On failure, remove all block requests to the peer.
```
vicsn, line 511: Nice catch

vicsn (changes requested):

.ci/partial_upgrade_test.sh
```diff
@@ -0,0 +1,362 @@
+#!/bin/bash
```
vicsn: Can you add a run of this script to the `upgrade-workflow` in `.circleci/config.yml`?

node/router/messages/src/lib.rs
```diff
@@ -298,8 +298,13 @@ mod tests {
     /// Ensure that *message versions* are unique and incrementing by 1.
     fn consensus_constants_increasing_heights<N: Network>() {
         let mut previous_message_version = Message::<N>::VERSIONS.first().unwrap().1;
-        for (_, message_version) in Message::<N>::VERSIONS.iter().skip(1) {
-            assert_eq!(*message_version, previous_message_version + 1);
+        for (consensus_version, message_version) in Message::<N>::VERSIONS.iter().skip(1) {
+            if consensus_version == &ConsensusVersion::V13 {
+                // Message version did not increment for ConsensusVersion::V13.
```
ljedrz, line 303: such exceptions could get annoying in the future; I propose that we instead just
```
assert!(*message_version >= previous_message_version);
```
as this is only a sanity check

.ci/db_backup_ci.sh
```diff
@@ -52,32 +60,32 @@ function create_checkpoints() {
...
-wait_for_nodes "$total_validators" 0 
+wait_for_nodes "$total_validators" 0 "$network_name"
 
 # Check heights periodically with a timeout
 total_wait=0
 checkpoint_created=false
 while (( total_wait < 600 )); do  # 10 minutes max
   # Apply short-circuiting
-  if [[ $checkpoint_created = true ]] || check_heights  "$total_validators" 0 "$checkpoint_height" "$network_name"; then
+  if [[ $checkpoint_created = true ]] || check_heights  0 "$total_validators" "$checkpoint_height" "$network_name"; then
```
ljedrz, line 74: oof good one, dunno how it got swapped

node/bft/src/gateway.rs
```diff
@@ -655,6 +655,7 @@ impl<N: Network> Gateway<N> {
                         | Err(err @ InsertBlockResponseError::NoConsensusVersion)
                         | Err(err @ InsertBlockResponseError::ConsensusVersionMismatch { .. }) => {
                             error!("Peer '{peer_ip}' sent an invalid block response - {err}");
+                            self.ip_ban_peer(peer_ip, Some(&err.to_string()));
```
ljedrz, line 658: I recall that we recently had some issues related to IP-level bans; these changes (all the bans) should be moved to a separate PR which includes a test for this feature.

ljedrz (changes requested):
+1 regarding the complex upgrade script, left some suggestions, and I propose that we move the bans to a separate PR which also adds a test for it.

node/router/messages/src/lib.rs
vicsn: Yeah it'll be annoying temporarily (until we can prune this logic as the network advanced), but the test actually served as a reminder for us to upgrade, and it was my mistake for not running CI before releasing.

node/bft/src/gateway.rs
vicsn: Indeed this PR therefore explicitly doesn't do banning because of timeouts. As a result I think it's fine to keep here. Any particular test you'd recommend? CC https://github.com/ProvableHQ/snarkOS/issues/4000

ljedrz: just basics, really, like having 2 nodes which:
1. connect successfully
2. one IP-bans the other, could be completely arbitrarily
3. verify that neither node is able to initiate a connection with the other
4. optionally wait until the ban expires (which should be a lot quicker in `test`/`dev` mode) and connect again

.ci/partial_upgrade_test.sh
kaimast: I actually changed this. I only upgrade one other node now, so it is more likely that the syncing node connects to one of the outdated ones

.ci/partial_upgrade_test.sh
kaimast: Done. I cannot run it on my machine right now to test, though. Will let you know once I verify it works in CircleCI.

.ci/partial_upgrade_test.sh
kaimast: I actually removed this step now. It should be sufficient to just wait for the block height.

node/bft/src/gateway.rs
vicsn: @kaimast to what extent can we leverage `partial_upgrade_test.sh` for this?

What are even the test expectations / invariants for `partial_upgrade_test.sh`? My assumption is:
- If the latest consensus version increases in the "current" release, `$penultimate_index` and `$last_node` will ban the rest of the nodes but `$last_node` will still sync from `$penultimate_index`.
- If the latest consensus version does not increase in the "current" release, then no banning happens and `$last_node` will sync from any node. Which is fine, but we should document that this is a possible outcome of the test.

kaimast: How about we do the following:

0. Perform a sanity check that the consensus version actually changes between the previous release and the current commit. If not, return early with "success" (NEW)
1. Start all nodes with the previous release's binary
2. Wait for block height to be greater than that of the new consensus version.
3. Stop the penultimate node but do restart it yet (NEW)
4. Start the last node with the new binary and ensure it does *not* sync (because all other nodes are outdated) (NEW)
5. Start the penultimate node with the new binary and ensure the last node syncs.

kaimast:
The most recent commits fix the `test_partial_upgrade` script and introduces a revised `test_full_upgrade` script (formerly `upgrade_nodes_ci`).  I removed most duplicated code between the two and simplified their logic. The naming also makes the `.ci` folder less cluttered: all benchmarks are prefixed with `bench_` and most tests are not prefixed with `test_`. 

It seems like the consensus version and height information is not available in the latest release. The scripts work around this by spawning a devnet, waiting for the consensus version to stabilize, and the records its number and block height.
We can eventually remove this with a follow-up PR.

Finally, I pushed two other small commits to fix CI failures unrelated to the PR.

vicsn:
> It seems like the consensus version and height information is not available in the latest release. The scripts work around this by spawning a devnet, waiting for the consensus version to stabilize, and the records its number and block height.

The /version endpoint returns the consensus heights, does that help?

vicsn (approved):

PR #4095, opened 2026-02-04
[CI] Add chaotic devnet workflow

### LINKED ISSUE #4062: [Feature] Use network delay script in CI

kaimast:
Resolves #4062.

This PR adds two new tests `.ci/test_reset_minority.sh` and `.ci/test_reset_majority.sh`.  Both of these scripts set up a devnet and then periodically reset (stop and remove the ledger) a subset of the nodes. The test succeeds if the nodes keep advancing to a given high.
`reset_minority` resets exactly `f = (N-1)/3` nodes, where `N` is the total number of validators, and `reset_majority` resets `N-f` nodes. 

Both of these scripts are run in the new `chaotic-devnet-test` job, that also adds network delays and message drops through `script/delay-network.sh`. The job runs on a virtual machine, as CircleCI does not allow changing the behavior of network devices in its containers.

For both tests, I selected seven validators (`N=7` and `f=2`), which is a good middle ground between "big enough" to reproduce issues that may appear in production and "not too big" to be run on a single CI machine.
The test run up to 250 block and reset validators every 50 blocks. I also used the most aggressive network delay emulation, where I set the network to contain spikes through the entire run of the test. This was the only setting where I could reliably reproduce halts, using [this branch](https://github.com/ProvableHQ/snarkOS/tree/fix/sync-race-conditions-broken) which contains a known halt. [This](https://app.circleci.com/pipelines/github/ProvableHQ/snarkOS/3411/workflows/7a8b4c28-857e-45d8-a449-296e5aa52b84/jobs/69122) is an instance where the tests for the branch failed.

The PR also contains the `chaotic-devnet-runner` script, but it is not used right now, as I just set the most aggressive network delays at the beginning.
Further, the PR also contains a script that restarts all validators, which restarts all validators without clearing the ledger. It is useful to reproduce certain issues that happen when restoring the ledger state at startup. For now, this test is also disabled in CI as each of the reset tests takes close to 20 minutes. It might still be a useful test in the future.


.github/workflows/chaotic-devnet.yml
```diff
@@ -0,0 +1,88 @@
+name: Chaotic Devnet Test
```
vicsn: I would prefer if we do this one in CircleCI, it is more consistent with the other devnet scripts. In hindsight it was a mistake to even start to build this one in github actions, but we had a panic about CircleCI costs that day.

.github/workflows/chaotic-devnet.yml
```diff
@@ -0,0 +1,88 @@
...
+          # Make it executable (permissions may get lost)
+          chmod +x bin/snarkos
+
+          # Add the 'bin' folder to the system PATH
+          echo "$(pwd)/bin" >> $GITHUB_PATH
+
+      - name: Run Devnet Test with Chaos
+        # Run with f=1 and no clients, so the nodes do not overwhelm the runner.
+        # Allow for up to 50 warnings per node, as network issues will create those. 
+        run: ./scripts/chaotic-network-runner.sh $PORT_RANGE ./.ci/test_devnet.sh 4 0 0 45 50
```
vicsn: I don't see the value in running the chaos script on `test_devnet.sh` and never communicated that we should do this. I think it makes sense for the minority/majority reset scripts, so we can mimic results from `snarkos-stress-testing` at lower cost

vicsn (changes requested):

.github/workflows/chaotic-devnet.yml
kaimast: This is, more or less, the way it was set up in the original branch. But I will change it.

.github/workflows/chaotic-devnet.yml
kaimast: No worries. It's not hard to change.

.github/workflows/chaotic-devnet.yml
kaimast: Changed it.

vicsn:
>  The job is run with f=2 and N=7, which is a good middle ground between "big enough" to reproduce issues that may appear in production and "not too big" to be run on a single CI machine.

Great! How reliably do these specifc CI jobs trigger the halts observed in this thread? https://github.com/ProvableHQ/snarkOS/issues/4047#issuecomment-3683541464

.circleci/config.yml
```diff
@@ -592,9 +555,58 @@ jobs:
...
+     - run:
+          name: "Run Reset Minority Devnet Test"
+          timeout: 20m  # Allow 20 minutes total
+          command: |
+            ./.ci/test_reset_minority.sh 7 0 50 250 3
+     - run:
+          name: "Run Reset Majority Devnet Test"
+          timeout: 20m  # Allow 20 minutes total
+          command: |
+            ./.ci/test_reset_majority.sh 7 0 20 250 3
```
vicsn, line 622: Did this one actually surface any issues? Total runtime for the job is currently 45 minutes... Maybe we can lower this one to just 5 minutes as well with fewer blocks. Or if it's instrumental, we could split into two concurrent jobs.

vicsn (changes requested):

.circleci/config.yml
kaimast: Yes that one is the one that gets produces test failures more reliably with known issues. We could disable the minority reset one maybe?

vicsn: That's surprising! Let's make them concurrent

vicsn (approved):
This is great!

PR #4096, opened 2026-02-04
[Fix] Update snarkVM to support atomic `prepare_next_block`

kaimast:
Prerequisite for #4039.

Depends on #4095 and https://github.com/ProvableHQ/snarkVM/pull/3070.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -788,7 +788,7 @@ workflows:
           filters:
             branches:
               only:
-                - ci/updated_network_delay_test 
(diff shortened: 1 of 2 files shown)


vicsn (dismissed):

kaimast:
The `clippy` task on this still fails due to the changes in the `history` features. I will wait for those to be resolved.

vicsn (approved):

kaimast:
Seems like we already updated the snarkVM rev, so I will just close this PR.

PR #4102, opened 2026-02-11
Don't disable CDN if nodetype was not given

vicsn:
## Motivation

Our documentation says we default to `--client`, but if we do, we should enable CDN sync.

CC @14MR

### DIFF
--- a/cli/src/commands/start.rs
+++ b/cli/src/commands/start.rs
@@ -345,15 +345,11 @@ impl Start {
 
     /// Returns the CDN to prefetch initial blocks from, from the given configurations.
     fn parse_cdn<N: Network>(&self) -> Result<Option<http::Uri>> {
-        // Determine if the node type is not declared.
-        let is_no_node_type = !(self.validator || self.prover || self.client);
-
         // Disable CDN if:
         //  1. The node is in development mode.
         //  2. The user has explicitly disabled CDN.
         //  3. The node is a prover (no need to sync).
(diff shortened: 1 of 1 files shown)


kaimast (dismissed):
Looks good. I rebased so that `cargo audit` stops failing.

cli/src/commands/start.rs
```diff
@@ -345,15 +345,11 @@ impl Start {
...
-        // Determine if the node type is not declared.
-        let is_no_node_type = !(self.validator || self.prover || self.client);
-
         // Disable CDN if:
         //  1. The node is in development mode.
         //  2. The user has explicitly disabled CDN.
         //  3. The node is a prover (no need to sync).
-        //  4. The node type is not declared (defaults to client) (no need to sync).
-        if self.dev.is_some() || self.nocdn || self.prover || is_no_node_type {
+        if self.dev.is_some() || self.nocdn || self.prover {
```
kaimast: Should we DEBUG or INFO log here why CDN was disabled? Or are we logging this already somewhere?

vicsn: Good idea: b9874b233

ljedrz (changes requested):
The CDN-related logic looks solid, but what are those blobs of serialized blocks?

vicsn:
> but what are those blobs of serialized blocks?

Awkward, that's what they are

ljedrz (approved):

PR #4101, opened 2026-02-10
[Feat] Revamp historical mapping query

ljedrz:
The snarkOS counterpart to https://github.com/ProvableHQ/snarkVM/pull/3143. The query format changes from
```
GET /{network}/block/{blockHeight}/history/{mapping}
```

to 
```
GET /{network}/program/{id}/mapping/{name}/{key}/history/{height}
```

And returns the historical `Value` applicable at the given height.

Filing as a draft due to the use of a temporary snarkVM branch.

PR #4104, opened 2026-02-12
[Feat] Enhanced historical mapping retrieval

ljedrz:
A reopened and rebased https://github.com/ProvableHQ/snarkOS/pull/4101.

### DIFF
--- a/node/rest/src/routes.rs
+++ b/node/rest/src/routes.rs
@@ -35,6 +35,9 @@ use rayon::prelude::*;
 
 use version::VersionInfo;
 
+#[cfg(feature = "history")]
+type HistoricalMappingKey<N> = (ProgramID<N>, Identifier<N>, Plaintext<N>, u32);
+
 /// Deserialize a CSV string into a vector of strings.
 fn de_csv<'de, D>(de: D) -> std::result::Result<Vec<String>, D::Error>
 where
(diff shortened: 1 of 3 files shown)


node/rest/src/routes.rs
```diff
@@ -905,19 +908,19 @@ impl<N: Network, C: ConsensusStorage<N>, R: Routing<N>> Rest<N, C, R> {
         Ok(ErasedJson::pretty(()))
     }
 
-    /// GET /{network}/block/{blockHeight}/history/{mapping}
+    /// GET /{network}/program/{id}/mapping/{name}/{key}/history/{height}
     #[cfg(feature = "history")]
     pub(crate) async fn get_history(
         State(rest): State<Self>,
-        Path((height, mapping)): Path<(u32, snarkvm::synthesizer::MappingName)>,
+        Path((program_id, mapping_name, mapping_key, height)): Path<HistoricalMappingKey<N>>,
```
vicsn, line 915: @miazn although the way we update and retrieve the history state is not backwards compatible - how much of an issue is it for the endpoints to not be backwards compatible? We could also let these return an empty 200 and make new endpoints

kaimast:
Rebased this PR, so I can rebase #4096 on top of it.

node/rest/src/routes.rs
miazn: Still confirming with stakeholders, but i dont see why its an issue, this solution is much cleaner

vicsn (approved):
@miazn I'll merge this now so as not to block Kai's development - my bad for being too eager with merging the snarkVM side already. I assume ww'll hear back from your contact within 2 weeks - the next release surely won't be sooner than that

PR #4115, opened 2026-02-16
Initialize the block height metric on startup

vicsn:
## Motivation

When nodes restart, their height is briefly reported as 0, making Grafana dashboards harder to read.

CC @mike2194


kaimast (approved):
Looks good. The ledger service is initialized after the ledger is loaded from disk, so the height should be >0.

PR #4121, opened 2026-02-18
[Logs] Show a warning if we detect a forked peer

kaimast:


node/sync/src/block_sync.rs
```diff
@@ -708,38 +708,52 @@ impl<N: Network> BlockSync<N> {
...
-                Some((PeerPair(peer_ip, *other_ip), ancestor))
-            })
-            .collect();
+                    Some((PeerPair(peer_ip, *other_ip), ancestor))
+                })
+                .collect();
 
-        // Update the map of common ancestors.
+            // If most other peers differ from this one, the peer most likely forked.
+            if fork_count > all_locators.len() {
```
vicsn: If I read this correctly, this would not trigger if the peer forked just the most recent block.

Also, typically if there's a fork, we may expect more than one node to be on it.

I would remove this block for now, we were able to trace who forked from the logs pretty well.

kaimast: The original problem was that a single forked peer creates a lot of debug logs, sometimes more than once a second. In the logs the offending peer can appear, both, as "other_ip" and "peer_ip" (depending on which locators are updated). 

This would create a warning if there is single (or a minority of) forked peers. However, the line above should be `fork_count * 2 > all_locators.len()`, so it triggers when a peer is forked from most of the other peers.

vicsn: I get that this is a warnings meant to only print when we observe a lot of forks. I don't have time at the moment to understand whether it is targeted/reliable enough, maybe you can leave it as a separate issue.

kaimast: Okay, will do.

kaimast: Without this message, all the PR does is promote the fork log to a warning and improve documentation. Might still be worth merging, but it is up to you.

vicsn (approved):

PR #4122, opened 2026-02-18
[Feat] Automated ledger checkpoints

### LINKED ISSUE #4103: [Feature] Automate local ledger snapshotting
## 🚀 Feature

Currently we expose `db_backup` so a user can manually create a rocksdb snapshot. As this uses hard links to existing files, very little extra space is required.

However, few node operators are using this. No one wants to do more work than just running `snarkos start`.

To make this feature more accessible, we propose adding the following flags to `snarkos start` (names are t.b.d.): 
- `--backups-path` specifies where rocksdb backups are stored
- `--backups-frequency` specifies how often a backup is made, e.g. daily
- `--backups-max-age` specifies how old backups are allowed to be, older ones will be removed

@ljedrz any comments on the design? What are the best configuration flags of other famous backup tools? Should we encode an exponential backoff? (e.g. we always store a backup no older than `1 day, 1 week ánd 1 month`

@raychu86 keen to hear your thoughts - this would obviate the need for any complex non-black-box db rollback functionality.

ljedrz:
This PR introduces an optional feature (enabled via `--auto_db_checkpoints=<path>`), which automatically creates ledger backups (via RocksDB checkpoints, i.e. hardlink-based) at a frequency of every 1k blocks, and with a maximum of 5 checkpoints present at any time (a "rolling" set). The applicable block tree is also stored alongside them.

The auto-checkpoint eligibility interval is every 500ms, basically guaranteeing the "height precision" of the checkpoints and block tree dumps.

I successfully tested it with a local devnet and a lower checkpoint threshold.

Closes https://github.com/ProvableHQ/snarkOS/issues/4103.

### DIFF
--- a/node/src/node.rs
+++ b/node/src/node.rs
@@ -339,7 +339,6 @@ impl<N: Network> Node<N> {
         }
 
         // Spawn a loop that will periodically create the checkpoints.
-        let node = self.clone();
         let handle = tokio::spawn(async move {
             info!("Starting the automatic ledger checkpoint routine...");
 
@@ -350,13 +349,10 @@ impl<N: Network> Node<N> {
             block_tree_path.push("block_tree");
 
             loop {
-                // A small delay that's smaller than block time.
+                // A small delay that's smaller than block time. There are technically situations when
+                // blocks can be inserted one after the other more quickly (syncing, multiple blocks in
+                // a Subdag), those are edge cases unlikely to be encountered under normal conditions.
                 tokio::time::sleep(Duration::from_millis(500)).await;
-
-                // Don't create checkpoints while syncing.
-                if !node.is_block_synced() {
-                    continue;
-                }
 
                 // Skip if we've already created a checkpoint during this run, and the
                 // number of blocks baked since then is lower than the configured threshold.
@@ -388,7 +384,7 @@ impl<N: Network> Node<N> {
                     // Copy the block tree file to the new checkpoint.
                     checkpoint_path.push("block_tree");
                     if let Err(e) = fs::copy(source_block_tree_path, checkpoint_path) {
-                        warn!("Couldn't move the block tree file to a ledger checkpoint: {e}");
+                        warn!("Couldn't copy the block tree file to a ledger checkpoint: {e}");
                     }
                 });
 
--- a/node/src/node.rs
+++ b/node/src/node.rs
@@ -37,14 +37,21 @@ use snarkvm::prelude::{
     store::helpers::{memory::ConsensusMemory, rocksdb::ConsensusDB},
 };
 
-use aleo_std::StorageMode;
-use anyhow::Result;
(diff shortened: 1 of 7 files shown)


node/src/node.rs
```diff
@@ -290,4 +321,130 @@ impl<N: Network> Node<N> {
...
+            block_tree_path.push("block_tree");
+
+            loop {
+                // A small delay that's smaller than block time.
+                tokio::time::sleep(Duration::from_millis(500)).await;
+
+                // Don't create checkpoints while syncing.
+                if !node.is_block_synced() {
+                    continue;
+                }
```
vicsn: I think we should also create checkpoints while syncing. A validator usually at tip may briefly go into syncing mode and we want them to still be able to create accurate snapshots and rollback.

node/src/node.rs
```diff
@@ -290,4 +321,130 @@ impl<N: Network> Node<N> {
...
+            info!("Starting the automatic ledger checkpoint routine...");
+
+            // Prepare some object that will be useful throughout the routine.
+            let mut last_checkpoint_height = None;
+            let mut existing_checkpoints = Vec::with_capacity(MAX_AUTO_CHECKPOINTS + 1);
+            let mut block_tree_path = aleo_ledger_dir(N::ID, ledger.vm().block_store().storage_mode());
+            block_tree_path.push("block_tree");
+
+            loop {
+                // A small delay that's smaller than block time.
```
vicsn: Technically blocks can be inserted really quickly after another, e.g. when syncing or when a single `Subdag` produces multiple blocks. While the current design is probably fine most of the time, you may want to document this in the comment for clarity.

node/src/node.rs
```diff
@@ -290,4 +321,130 @@ impl<N: Network> Node<N> {
...
+                // Create a checkpoint.
+                let mut checkpoint_path = auto_checkpoint_path.clone();
+                checkpoint_path.push(format!("checkpoint_{current_height}"));
+                if let Err(e) = ledger.backup_database(&checkpoint_path) {
+                    warn!("Couldn't automatically store a checkpoint at {}: {e}", checkpoint_path.display());
+                    continue;
+                }
+                last_checkpoint_height = Some(current_height);
+
+                // Immediately procure and copy the applicable block tree in the background.
```
vicsn, line 375: For consistency, can you also make the block tree caching happen when someone manually calls the backup endpoint?

vicsn (changes requested):

ljedrz:
@vicsn All done :ok_hand:.

vicsn (approved):

raychu86:
Do you know how much additional storage this incurs?

ljedrz:
It should be very little for such a conservative block range and snapshot count, **as long as it's local storage** (in order for hardlinks to work).

PR #4126, opened 2026-02-20
[Fix] Do not start nodes if proposal cache and ledger rounds are too far apart

### LINKED ISSUE #4106: [Bug] Nodes load proposal cache before syncing
If a node restarts after loosing its ledger or reverting to an older snapshot, it will not be able to find the committee of its proposal cache. 

Nodes should delay loading the cache until they have synced within GC rounds of the cached proposal.

kaimast:
Partially fixes #4106.

An ideal solution to the proposal cache being too far in the "future" would be to defer it from being loaded until later. This approach is much simpler: it simply will tell the user to fetch a more recent ledger snapshot. 

Generally, it is unlikely for this to happen in production, nodes would need to revert to a much older snapshot or delete their ledger entirely. However, it is good to add this check, so that nodes do not start with outdated proposal caches and get stuck with confusing error messages.

### DIFF
--- a/node/bft/src/primary.rs
+++ b/node/bft/src/primary.rs
@@ -179,6 +179,20 @@ impl<N: Network> Primary<N> {
                     let (latest_certificate_round, proposed_batch, signed_proposals, pending_certificates) =
                         proposal_cache.into();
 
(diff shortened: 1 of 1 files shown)


vicsn (approved):
LGTM!

PR #4132, opened 2026-02-26
[Fix] Stop CDN processing cleanly when receiving SIGINT

### LINKED ISSUE #4089: [Bug] When snarkos is started without "--nocdn" the block tree is not cached
## 🐛 Bug Report

If snarkos is started with CDN (without the `--nocdn`) option the block tree is not cached and on every start it will do:

```
DEBUG snarkvm_ledger_store::helpers::rocksdb::block: Creating the block tree from scratch
```

This makes the startup way slower every time. If we start it with `--nocdn`, on the next restart it will print:

```
DEBUG snarkvm_ledger_store::helpers::rocksdb::block: Loading the cached block tree from <PATH-TO-LEDGER>/block_tree
```

and will start way faster (startup time with the block tree cache is 2s, without - 158s).

## Steps to Reproduce

Start snarkos without the `--nocdn` option and then restart it - it will always print that it is creating the block tree from scratch.

Another way is to trigger the load-ledger test and check that the startup time is over 150s.

## Expected Behavior

Start snarkos without the `--nocdn` option and then restart it - it will should print that it is Loading the cached block tree and start very fast.

Another way is to trigger the load-ledger test and check that the startup time is around 2-3s.

## Your Environment

- e8547620c51405530e4b6ba23678110a0fcaf2b1
- Cargo 0.88.0
- Ubuntu 22

ljedrz:
This ensures that the node shuts down cleanly, which makes the block tree be cached etc. As a result, the shutdown is only marginally slower.

Fixes https://github.com/ProvableHQ/snarkOS/issues/4089.

### DIFF
--- a/node/cdn/src/blocks.rs
+++ b/node/cdn/src/blocks.rs
@@ -243,7 +243,7 @@ pub async fn load_blocks<N: Network>(
         if stoppable.is_stopped() {
             info!("Stopping block sync at {} - shutting down", current_height);
             // We can shut down cleanly from here, as the node hasn't been started yet.
-            std::process::exit(0);
+            return Ok(current_height);
         }
 
         let mut candidate_blocks = pending_blocks.lock().await;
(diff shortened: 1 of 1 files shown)


vicsn (approved):
Cool

kaimast (approved):
Great!

PR #4133, opened 2026-02-27
[Perf] Use unchecked deserialization in CDN sync

ljedrz:
This was a lot simpler than I expected - we can easily tap into the setup we're already using for retrieving trusted data from local storage. This results in a pretty solid boost in performance, most likely (hard to measure accurately due to the variance in download speed block size etc.) around 15% or more, depending on committee size (since most of the omitted checks are related to signature verification).


vicsn (approved):
#madgains

PR #4134, opened 2026-02-27
[Perf] Don't recreate the rayon threadpool in the CDN loop

ljedrz:
Tricky to find, trivial to correct. This must have been introduced by complete accident, as there is no reason to recreate the `rayon` threadpool.

Locally, this has resulted in nothing short of mad gains in CDN sync speed.


vicsn (approved):
#madgains

PR #4136, opened 2026-03-02
[Feat] Log the combined stake at our SHA (in validator mode)

### LINKED ISSUE #4135: [Feature] Expose a log for % of connected stake which is on the same commit hash
## 🚀 Feature

We now log whether certain peers have the same or different commit hashes.

It would be very useful if we could log and expose a metric for which % of stake is on the same commit hash as us, similar to how we're logging `not_connected_stake_as_percentage`.

ljedrz:
Closes https://github.com/ProvableHQ/snarkOS/issues/4135.

### DIFF
--- a/node/bft/src/gateway.rs
+++ b/node/bft/src/gateway.rs
@@ -24,6 +24,7 @@ use crate::{
     helpers::{Cache, PrimarySender, Storage, SyncSender, WorkerSender, assign_to_worker},
     spawn_blocking,
 };
+use smol_str::SmolStr;
 use snarkos_account::Account;
 use snarkos_node_bft_events::{
     BlockRequest,
@@ -50,6 +51,7 @@ use snarkos_node_network::{
     bootstrap_peers,
     get_repo_commit_hash,
     log_repo_sha_comparison,
+    shorten_snarkos_sha,
 };
 use snarkos_node_sync::{InsertBlockResponseError, MAX_BLOCKS_BEHIND, communication_service::CommunicationService};
 use snarkos_node_tcp::{
@@ -475,6 +477,7 @@ impl<N: Network> Gateway<N> {
                 address,
                 NodeType::Validator,
                 0,
+                get_repo_commit_hash(),
                 ConnectionMode::Gateway,
             );
         }
@@ -886,7 +889,7 @@ impl<N: Network> Gateway<N> {
     /// Logs the connected validators.
     fn log_connected_validators(&self) {
         // Retrieve the connected validators and current committee.
-        let connected_validators = self.connected_peers();
+        let connected_validators = self.get_connected_peers();
         let committee = match self.ledger.current_committee() {
             Ok(c) => c,
             Err(err) => {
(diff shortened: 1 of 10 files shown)


node/bft/src/gateway.rs
```diff
@@ -908,15 +911,30 @@ impl<N: Network> Gateway<N> {
 
         // Collect the connected validator addresses and stake.
         let mut connected_validator_addresses = HashSet::with_capacity(connected_validators.len());
+        let mut connected_validator_shas: HashMap<SmolStr, u64> = HashMap::with_capacity(connected_validators.len());
+        // Insert our sha.
+        let our_sha = shorten_snarkos_sha(&get_repo_commit_hash());
```
kaimast, line 916: Is this conversion only done, so the printed string is shorter? It might be a little cleaner to do it when formatting the log message.

node/bft/src/gateway.rs
```diff
@@ -908,15 +911,30 @@ impl<N: Network> Gateway<N> {
...
-        for peer_ip in &connected_validators {
-            let address = self.resolve_to_aleo_addr(*peer_ip).map_or("Unknown".to_string(), |a| {
-                connected_validator_addresses.insert(a);
-                a.to_string()
-            });
-            debug!("{}", format!("  Connected to: {peer_ip} - {address}").dimmed());
+        for peer in &connected_validators {
+            let peer_ip = peer.listener_addr;
+            // Register the Aleo address, which is guaranteed to be present for validators.
+            let address = self.resolve_to_aleo_addr(peer_ip).unwrap();
```
kaimast: You can do `peer.aleo_addr` here to avoid the `unwrap` entirely, right?

node/bft/src/gateway.rs
ljedrz: mostly, yes, but this also reduces the size of the keys in the hashmap, and simplifies them (as they'd otherwise be `Option<[u8; 40]>`)

node/bft/src/gateway.rs
```diff
@@ -908,15 +911,30 @@ impl<N: Network> Gateway<N> {
...
+            let address_stake = committee.get_stake(address);
+            let short_peer_sha = shorten_snarkos_sha(&peer.snarkos_sha);
+            *connected_validator_shas.entry(short_peer_sha.clone()).or_default() += address_stake;
+            // Log the connected validator.
+            debug!("{}", format!("  Connected to: {peer_ip} - {address} @ {short_peer_sha}").dimmed());
+        }
+
+        if let Some(combined_stake) = connected_validator_shas.get(&our_sha) {
+            let percentage = *combined_stake as f64 / committee.total_stake() as f64 * 100.0;
+            debug!("{}", format!("  Combined stake @ {our_sha}: {percentage:.2}%").dimmed());
```
vicsn, line 937: Can you also track a metric, we could call it `STAKE_WITH_MATCHING_SHA`

ljedrz: done in https://github.com/ProvableHQ/snarkOS/pull/4136/commits/7be6802706f2ce3e88e6d7a5202eaa41dc6eb59d, with the name `snarkos_bft_connected_stake_with_matching_sha_as_percentage` (following the naming of the existing `snarkos_bft_connected_stake_as_percentage`)

vicsn (approved):

PR #4138, opened 2026-03-03
[Fix] Ensure aborted transactions are handled properly on sync

kaimast:
This PR introduces two small, but important, improvements to block sync.

https://github.com/ProvableHQ/snarkOS/commit/c1b4d7aae1b5c604f94f2c3cb1ebcb22e57be1ea updates the snarkVM rev to leverage the new `try_get` functionality for transmissions (https://github.com/ProvableHQ/snarkVM/pull/3117). This change also makes errors propagate to the calling function, instead of simply logging a warning.

https://github.com/ProvableHQ/snarkOS/commit/362e9e96bfca8aacf65c1d6c174a4ee9d9f46825 ensures that the sync logic checks whether a transaction was aborted before querying the ledger. In the current implementation the ledger returns an error for aborted transmissions, as there exists a reference to it in the ledger but the transmission itself is not stored in the ledger.
This should probably be fixed on the snarkVM-side, but checking the aborted set first is also cheap and avoids expensive rocksdb lookups in some cases.

I made this a separate PR as the changes are self-standing and provide a useful improvement.

### DIFF
--- a/node/bft/src/helpers/storage.rs
+++ b/node/bft/src/helpers/storage.rs
@@ -855,9 +855,9 @@ impl<N: Network> Storage<N> {
                         // Otherwise, try to load the solution from the ledger.
                         None => match self.ledger.get_solution(solution_id) {
                             // Insert the solution.
-                            Ok(solution) => missing_transmissions.insert(*transmission_id, solution.into()),
+                            Ok(Some(solution)) => missing_transmissions.insert(*transmission_id, solution.into()),
                             // Check if the solution is in the aborted solutions.
-                            Err(_) => {
+                            Ok(None) | Err(_) => {
                                 // Insert the aborted solution if it exists in the block or ledger.
                                 match aborted_solutions.contains(solution_id)
                                     || self.ledger.contains_transmission(transmission_id).unwrap_or(false)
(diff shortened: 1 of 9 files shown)


node/bft/src/helpers/storage.rs
```diff
@@ -848,54 +848,43 @@ impl<N: Network> Storage<N> {
...
+                    } else if aborted_solutions.contains(solution_id) {
+                        aborted_transmissions.insert(*transmission_id);
+                    } else if let Some(solution) = self
+                        .ledger
+                        .get_solution(solution_id)
+                        .with_context(|| "Failed to fetch missing solution from ledger")?
+                    {
+                        // The solution was already contained in the ledger.
+                        missing_transmissions.insert(*transmission_id, solution.into());
+                    } else {
```
vicsn, line 862: I'm missing the ledger.contains_transmission call for aborted transmissions

node/bft/src/helpers/storage.rs
```diff
@@ -848,54 +848,43 @@ impl<N: Network> Storage<N> {
...
-                                continue;
-                            }
-                        },
-                    };
+                    // The transaction may exists either in the block itself, or stored in the ledger.
+                    // Note that the ledger does *not* store aborted transmissions, so we need to check if the transaction
+                    // was aborted before querying the ledger.
+                    if let Some(transaction) = unconfirmed_transactions.get(transaction_id) {
+                        missing_transmissions.insert(*transmission_id, transaction.clone().into());
+                    } else if aborted_transactions.contains(transaction_id) {
```
vicsn, line 891: Same here about ledger.contains

vicsn (changes requested):

kaimast:
I updated the PR and addresses @vicsn's comments. It now includes the workaround to log a warning when a transaction is not found, instead of aborting.

Cargo.toml
```diff
@@ -48,7 +48,7 @@ default-features = false
 [workspace.dependencies.snarkvm]
 #path = "../snarkVM"
 git = "https://github.com/ProvableHQ/snarkVM.git"
-rev = "8df377d"
+rev = "bb5c8ae6"
```
ljedrz: This revision needs to be updated, as we already have snarkOS changes depending on very recent snarkVM revisions. Actually, it seems that the current value is fine, this line just needs to be reverted.

ljedrz (commented):
LGTM, but needs rebasing.

Cargo.toml
kaimast: I did a rebase and removed this change.

ljedrz (approved):

kaimast:
I made some fixes to this logic, but I don't want to rebase #4039 again, so I will close this PR.

PR #4146, opened 2026-03-09
Harden TCP connection config

vicsn:
## Motivation

We recently observed an updated validator taking dozens of minutes before it was connected to > 90% of other nodes. We also observed that connections were refused due to too many pending connections, potentially the *same* peers filling up the precious pending TCP cache.

This PR introduces:
- 9cdfcd5025a52ad4c2dae49779204d411045d383 We catch and cleanup if issues were to happen in socket creation or binding.
- edcf75657ff4e2ee6781ebd0d2396cde71576d3f A 10x increase in the TCP connection limit. The worst case impact on resource usage should be negligible, and we ultimately only connect to validators in the committee anyway.

## Test Plan

- [x] CI should be sufficient.

## Documentation

This needs a new snarkOS mainnet release

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -788,7 +788,7 @@ workflows:
           filters:
             branches:
               only:
-                - ci/log-ci-runner-canary
+                - patch/tcp-socket-creation-2
                 - canary
                 - testnet
                 - mainnet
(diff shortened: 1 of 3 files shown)


ljedrz (dismissed):

ljedrz:
The `node-workflow` failure appears unrelated.

kaimast (dismissed):

raychu86 (approved):

PR #4149, opened 2026-03-10
[Feature] Batch creation and round advancement improvements

kaimast:
This PR changes the batch creation task, to trigger on certain events, instead of only periodically. The goal is to avoid cases where a node is not ready to propose and then waits another `MIN_BATCH_DELAY` (1s) instead of only for the relevant conditions. It also ensures that the batch delay timer is reset whenever the node advances.

The most important changes in this PR are in `Primary::start_handlers`. The [`README`](https://github.com/ProvableHQ/snarkOS/pull/4149/changes#diff-88224c24f5c0399890a8698944fe09ebd9245fabce5e2e9d04a44695d98aa211) details when the batch creation task is woken up.

### DIFF
--- a/node/bft/README.md
+++ b/node/bft/README.md
@@ -31,6 +31,17 @@ f + 1 vertices that vote for the anchor, or 2f + 1 vertices that do not, or a ti
 ```
 Note that in this quote `2f + 1` should really be `n - f`.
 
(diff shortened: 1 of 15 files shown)


kaimast:
This PR was replaced by #4190.

PR #4152, opened 2026-03-10
[Fix] Robust broadcast counters

ljedrz:
No need to manually keep track of broadcast limits ourselves; using a `Semaphore` we don't have to worry about the atomic ordering involved, or about cleanups.


ljedrz:
The [cargo-workflow](https://github.com/ProvableHQ/snarkOS/pull/4152/checks?check_run_id=66450271394) failure is unrelated.

vicsn (approved):

PR #4153, opened 2026-03-10
Postrelease merge mainnet

vicsn:
## Motivation

Regular merge back into staging. Please review the files changed. They contain some hotfixes which went into canary/testnet. We should make sure that the correct logic ends up in staging.


node/router/messages/src/lib.rs
```diff
@@ -116,6 +116,9 @@ impl<N: Network> Message<N> {
         (ConsensusVersion::V10, 21),
         (ConsensusVersion::V11, 22),
         (ConsensusVersion::V12, 23),
+        // For ConsensusVersion::V13, we forgot to run CI and increment the
+        // message version before the canary release, so we keep it the same for
+        // Canary.  We can bump it again on Canary for V14.
```
ljedrz: wouldn't this imply that the version should be the same in `V13` as in `V12`?

ljedrz (dismissed):
LGTM with one question regarding a comment.

.ci/utils.sh
```diff
@@ -436,6 +436,12 @@ function wait_for_bft_connections() {
   return 1
 }
 
+# Succeeds if the node with the given index has the specified number of BFT connections (or greater)
```
kaimast: This should be removed. The proper implementation is above.

kaimast (dismissed):
One minor change is needed. Everything else looks good.

.ci/utils.sh
kaimast: To fix the lint warning

```suggestion
    sleep "$poll_interval"
```

node/bft/src/gateway.rs
```diff
@@ -219,7 +219,7 @@ impl<N: Network> Gateway<N> {
             (Some(ip), _) => ip,
         };
         // Initialize the TCP stack.
```
kaimast, line 221: It would be good to document why we multiply by 10 here. Something like the following would, maybe.

```suggestion
        // Initialize the TCP stack.
        //
        // This allows for more TCP connections than the maximum committee size to prevent "connection refused"
        // errors when two nodes simultaneous attempt to connect to each other.
        // Note, that that later, during handshake, the Gateway applies its own limit to the number of
        // active connections and removes duplicates.
```

vicsn: 5e1301644

.ci/utils.sh
vicsn: b81d40f15

.ci/utils.sh
vicsn: b81d40f15

node/router/messages/src/lib.rs
vicsn: Glad you caught it 5ae20801f

ljedrz (approved):
LGTM 👌

PR #4160, opened 2026-03-12
Update release version

vicsn:
## Motivation

to v4.5.3

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4404,7 +4404,7 @@ dependencies = [
 
 [[package]]
 name = "snarkos"
-version = "4.5.1"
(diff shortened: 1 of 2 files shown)


ljedrz (approved):
Just noting that it's `.3`, don't recall there having been a `.2`.

vicsn:
> Just noting that it's .3, don't recall there having been a .2.

Yeah I pushed an immutable tag for .2 without updating the Cargo.toml. 🤦

ljedrz:
Just one of those days - I for one forgot the lockfile updates with my backports; however, all should be good now :muscle:.

PR #4161, opened 2026-03-12
[Logs] Remove empty log entries

ljedrz:
While practical locally, the empty log lines are unhelpful when browsing them in bulk:
provable grafana net_explore_schemaVersion=1 panes=%7B%22fy0%22%3A%7B%22datasource%22%3A%22bf6hh0chl7xtse%22%2C%22queries%22%3A%5B%7B%22refId%22%3A%22A%22%2C%22datasource%22%3A%7B%22type%22%3A%22googlecloud-logging-datasource%22%2C%22uid%22%3A%22bf6hh
Since the devs can easily apply extra ad-hoc visibility to any logs at any point in time, I propose that we remove the additional newlines by default.

### DIFF
--- a/node/bft/ledger-service/src/ledger.rs
+++ b/node/bft/ledger-service/src/ledger.rs
@@ -395,7 +395,7 @@ impl<N: Network, C: ConsensusStorage<N>> LedgerService<N> for CoreLedgerService<
             metrics::update_block_metrics(block);
         }
 
-        tracing::info!("\n\nAdvanced to block {} at round {} - {}\n", block.height(), block.round(), block.hash());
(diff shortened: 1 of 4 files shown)


ljedrz:
I'm also wondering if we shouldn't downgrade the log related to the inability of obtaining the commit SHA (currently `WARN`), as it's popping up very often, while technically not being an issue.

The same perhaps could be said about the `Unable to add unconfirmed solution '<ID>' to the memory pool — Invalid Solution '<ID>' - Prover '<address>' has reached their solution limit for the current epoch`.

vicsn (approved):

PR #4162, opened 2026-03-12
[Fix] Correct the Router heartbeat interval

ljedrz:
During log analysis I've realized that the `Router` heartbeat fires twice in a row every time, which is caused by a sneaky miscalculation in its sleep time. This may cause not only confusion while debugging, but  also issues with `Router` connections. I've confirmed locally that this change fixes it.

As a drive-by, remove one empty line log missed in https://github.com/ProvableHQ/snarkOS/pull/4161.

### DIFF
--- a/node/router/src/routing.rs
+++ b/node/router/src/routing.rs
@@ -50,21 +50,17 @@ pub trait Routing<N: Network>:
     fn initialize_heartbeat(&self) {
         let self_clone = self.clone();
         self.router().spawn(async move {
-            let mut last_update = Instant::now();
-
             loop {
+                let start = Instant::now();
                 // Process a heartbeat in the router.
                 self_clone.heartbeat().await;
+                let end = Instant::now();
 
-                // The heartbeat itself may take a while (as it is waiting for connection attempts to complete).
-                // Take this time into account when rate-limiting the heartbeat.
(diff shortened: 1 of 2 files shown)


node/router/src/routing.rs
```diff
@@ -50,21 +50,17 @@ pub trait Routing<N: Network>:
     fn initialize_heartbeat(&self) {
         let self_clone = self.clone();
         self.router().spawn(async move {
-            let mut last_update = Instant::now();
-
             loop {
+                let start = Instant::now();
                 // Process a heartbeat in the router.
                 self_clone.heartbeat().await;
+                let end = Instant::now();
```
kaimast:
```suggestion
                let took = start.elapsed();
```

kaimast (dismissed):
Looks good to me. I think the previous version was intended to account for sleep taking longer than the given Duration, but it seems too confusing and not worth it.

With the new version it still can fire right after the previous one, if the call to `Self::heartbeat` takes very long. If that is a problem, we should add a minimum sleep time as well.
However, the code already has some logic to prevent too frequent connection attempts to specific peers, so it should not cause any issues AFAIK.

ljedrz:
Review comment addressed.

vicsn (approved):

PR #4168, opened 2026-03-16
feat(node/bft): add more metrics

vicsn:
## Motivation

More metrics to determine slower block production on mainnet

## Test Plan

- [ ] Run extended CI tests.


vicsn:
Duplicate with https://github.com/ProvableHQ/snarkOS/pull/4169

PR #4170, opened 2026-03-16
Revert "[Feat] Add more block creation metrics to mainnet"

vicsn:
Reverts ProvableHQ/snarkOS#4169

On second thought, these should not go directly into mainnet, we can make a feature branch.


vicsn:
Too late, the new patch uses the metrics: https://github.com/ProvableHQ/snarkOS/pull/4171

PR #4175, opened 2026-03-17
[Feature] Add Slipstream Plugin Support

dgrkotsonis:
# Summary                                                                                                                                                                 
   
   Wires the new [Slipstream plugin system](https://github.com/ProvableHQ/snarkVM/pull/3190) into snarkOS so that node operators can load one or more streaming plugins at startup and manage them at runtime via authenticated REST endpoints. Validator and Client nodes both support plugins; Prover nodes do not (they do not finalize blocks).                                                                                                                                                                                                 
   
  ---                                                                                                                                                                     
  # Changes                                                                                                                                                                                                                                                                                                                   

 ### Cargo / feature wiring
  - All Slipstream functionality is gated behind --features slipstream-plugins. Nodes compiled without the feature are unaffected. The flag propagates through the workspace: snarkos → snarkos-cli → snarkos-node → snarkos-node-rest → snarkvm.           
  - Cargo.toml — Added snarkvm-slipstream-plugin-manager as a new workspace dependency                                                                                                                                                                 
  - node/Cargo.toml — Added snarkvm-slipstream-plugin-manager as an optional dep, propagate new `slipstream-plugins` feature flag to enable it alongside the corresponding snarkvm features.                                                                                                                                         
  - node/rest/Cargo.toml — Same optional dep + feature wiring.                                                                                                            
  - cli/Cargo.toml — Added slipstream-plugins features forwarding to snarkos-node.
                                                                                                                                                                          
  ### CLI                                                                                                                                                                     
  - cli/src/commands/start.rs — Added --slipstream-config PATH flag (repeatable; feature-gated). Empty slice is passed when the feature is disabled so the call site is always uniform.                                                                                                                                                         
                                                                                                                                                                        
  ### Node layer                                                                                                                                                              
  - node/src/node.rs — Threaded slipstream_configs: &[PathBuf] through new_validator and new_client.                                                                    
  - node/src/validator/mod.rs / node/src/client/mod.rs — After Ledger::load, if plugin configs are provided, a SlipstreamPluginManager is initialized and injected into the FinalizeStore                                                                                                       
                                                                                                                                                                        
 ### REST API                                                                                                                                                                
  - node/rest/src/lib.rs — Registered three slipstream routes. Added DELETE to CORS allowed methods.     
  - node/rest/src/routes.rs — Implemented three handlers:                                                                                                                  
                                                        
  | Method | Path                                     | Description                              |
|--------|------------------------------------------|------------------------------------------|
| GET    | /{network}/slipstream/plugins            | List loaded plugins                      |
| POST   | /{network}/slipstream/plugins            | Load a plugin from a config file         |
| DELETE | /{network}/slipstream/plugins/{name}     | Unload a plugin by name                  |       

 PUT (reload) is not yet implemented and is left as a TODO.                                                            
   
  ### Error mapping: PluginNotLoaded → 404, PluginAlreadyLoaded → 422, all others → 500.                                                                                      
                                                                                                                                                                        
  ### Tests / docs                                                                                                                                                            
  - node/tests/common/node.rs — Updated client() and validator() test helpers to pass &[] for the new parameter.                                                        
  - docs/slipstream_plugins.md — New operator guide covering: what Slipstream is, plugin ABI and build instructions, JSON5 config format, startup flags, full REST API reference, and curl examples. (NOTE: Couldn't think of a better place to put it so happy to move elsewhere)
                                                                                                                                                                          
  ---                                                                                                                                                                   
  # Usage                                                                                                                                                                   
                                                                                                                                                                        
  #### Build with plugin support
  cargo build --release --features slipstream-plugins                                                                                                                               
   
  #### Start a client node with two plugins                                                                                                                                  
  snarkos start --client  \                                                                                                                                              
    --slipstream-config ~/.aleo/plugins/postgres/plugin.json5 \                                                                                                           
    --slipstream-config ~/.aleo/plugins/metrics/plugin.json5                                                                                                              
                                                                                                                                                                          
  #### List loaded plugins                                                                                                                                                   
  curl http://localhost:3030/mainnet/slipstream/plugins                                                                                                 
                                                                                                                                                                          
  #### Load a new plugin at runtime                                                                                                                                              
  curl -X POST -H "Content-Type: application/json" \                                                                                                                               
    -d '{"config_file":"/path/to/plugin.json5"}' \                                                                                                                        
    http://localhost:3030/mainnet/slipstream/plugins
    
    
 #### Delete/Unload a new plugin at runtime                                                                                                                                              
  curl -X DELETE  \                                                                                                                                                                                                                                        
    http://localhost:3030/mainnet/slipstream/plugins/{plugin_name}                                                                                                                    
                                                                                                                                                                          
  ---                                                                                                                                                                     
  # Notes                                                                                                                                                                   
                                                                                                                                                                        
  - All plugin-related code is gated on #[cfg(feature = "slipstream-plugins")], so builds without these features are unaffected.
  - Plugin manager lock errors (poisoned RwLock) surface as 500 errors; plugin call errors during finalize are logged as warnings and never propagated to the node.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,7 +47,7 @@ default-features = false
 
 [workspace.dependencies.snarkvm]
 git = "https://github.com/ProvableHQ/snarkVM.git"
-rev = "10845556178ac"
+rev = "60322dd"
 #version = "=4.6.0"
 default-features = false
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -49,13 +49,13 @@ default-features = false
 #path = "../snarkVM"
 git = "https://github.com/ProvableHQ/snarkVM.git"
 # Latest stream_plugin_testing commit
-rev = "0d50724d180f02cc9973cbb07c1cb8ec0c10212a"
+rev = "68e72e6c68a87425e8a406e078dae1cea46f2bc1"
 default-features = false
 
 [workspace.dependencies.snarkvm-slipstream-plugin-manager]
 git = "https://github.com/ProvableHQ/snarkVM.git"
 # Latest stream_plugin_testing commit
-rev = "0d50724d180f02cc9973cbb07c1cb8ec0c10212a"
+rev = "68e72e6c68a87425e8a406e078dae1cea46f2bc1"
 default-features = false
 
 [workspace.dependencies.anyhow]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,14 +47,14 @@ default-features = false
 
 [workspace.dependencies.snarkvm]
 git = "https://github.com/ProvableHQ/snarkVM.git"
-# Latest stream_plugin_testing commit
-rev = "951cb1bc5ea305747cad56ef7ef2b5f1d0dfe01e"
+# Latest staking commit
+rev = "159f04ce49a29674e0fe63530e0212a71d58417f"
 default-features = false
 
 [workspace.dependencies.snarkvm-slipstream-plugin-manager]
 git = "https://github.com/ProvableHQ/snarkVM.git"
-# Latest stream_plugin_testing commit
-rev = "951cb1bc5ea305747cad56ef7ef2b5f1d0dfe01e"
+# Latest staking commit
+rev = "159f04ce49a29674e0fe63530e0212a71d58417f"
 default-features = false
 
 [workspace.dependencies.anyhow]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,9 +48,12 @@ default-features = false
 [workspace.dependencies.snarkvm]
 #path = "../snarkVM"
 git = "https://github.com/ProvableHQ/snarkVM.git"
-rev = "8c5ef1f5849cf12e2e00b2b4ca5cea152db48402"
-#version = "=4.6.0"
+branch = "stream_plugin_testing"
 default-features = false
+
+[workspace.dependencies.snarkvm-slipstream-plugin-manager]
+git = "https://github.com/ProvableHQ/snarkVM.git"
+branch = "stream_plugin_testing"
 
 [workspace.dependencies.anyhow]
 version = "1.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -49,13 +49,13 @@ default-features = false
 #path = "../snarkVM"
 git = "https://github.com/ProvableHQ/snarkVM.git"
 # Latest stream_plugin_testing commit
-rev = "1eeb80dac6e9333ccffbeb5be57c6ed9438ed4bb"
+rev = "3e09bd1c6f8bcd39310836dbbe0a853c0431fcf7"
 default-features = false
 
 [workspace.dependencies.snarkvm-slipstream-plugin-manager]
 git = "https://github.com/ProvableHQ/snarkVM.git"
 # Latest stream_plugin_testing commit
-rev = "1eeb80dac6e9333ccffbeb5be57c6ed9438ed4bb"
+rev = "3e09bd1c6f8bcd39310836dbbe0a853c0431fcf7"
 default-features = false
 
 [workspace.dependencies.anyhow]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,15 +47,10 @@ default-features = false
 
 [workspace.dependencies.snarkvm]
 git = "https://github.com/ProvableHQ/snarkVM.git"
-rev = "5f6b502578cc"
+rev = "10845556178ac"
 #version = "=4.6.0"
 default-features = false
 
-[workspace.dependencies.snarkvm-slipstream-plugin-manager]
-git = "https://github.com/ProvableHQ/snarkVM.git"
-rev = "5f6b502578cc"
-#version = "=4.6.0"
-default-features = false
 
 [workspace.dependencies.anyhow]
 version = "1.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,13 +48,13 @@ default-features = false
 [workspace.dependencies.snarkvm]
 git = "https://github.com/ProvableHQ/snarkVM.git"
 # Latest stream_plugin_testing commit
-rev = "41023939401d9a2f5667cd1e0c211c582b269c95"
+rev = "37b18a3bd4bdb0a02145df4d39a550dce1cf53bf"
 default-features = false
 
 [workspace.dependencies.snarkvm-slipstream-plugin-manager]
 git = "https://github.com/ProvableHQ/snarkVM.git"
 # Latest stream_plugin_testing commit
-rev = "41023939401d9a2f5667cd1e0c211c582b269c95"
+rev = "37b18a3bd4bdb0a02145df4d39a550dce1cf53bf"
 default-features = false
 
 [workspace.dependencies.anyhow]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -49,13 +49,13 @@ default-features = false
 #path = "../snarkVM"
 git = "https://github.com/ProvableHQ/snarkVM.git"
 # Latest stream_plugin_testing commit
-rev = "3e09bd1c6f8bcd39310836dbbe0a853c0431fcf7"
+rev = "2811e1a524ff23a5e937ac4c1e8453520569b789"
 default-features = false
 
 [workspace.dependencies.snarkvm-slipstream-plugin-manager]
 git = "https://github.com/ProvableHQ/snarkVM.git"
 # Latest stream_plugin_testing commit
-rev = "3e09bd1c6f8bcd39310836dbbe0a853c0431fcf7"
+rev = "2811e1a524ff23a5e937ac4c1e8453520569b789"
 default-features = false
 
 [workspace.dependencies.anyhow]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -49,13 +49,13 @@ default-features = false
 #path = "../snarkVM"
 git = "https://github.com/ProvableHQ/snarkVM.git"
 # Latest stream_plugin_testing commit
-rev = "147b33914743874c26f9285a2ef5baf3cf478f56"
+rev = "1eeb80dac6e9333ccffbeb5be57c6ed9438ed4bb"
 default-features = false
 
 [workspace.dependencies.snarkvm-slipstream-plugin-manager]
 git = "https://github.com/ProvableHQ/snarkVM.git"
 # Latest stream_plugin_testing commit
-rev = "147b33914743874c26f9285a2ef5baf3cf478f56"
+rev = "1eeb80dac6e9333ccffbeb5be57c6ed9438ed4bb"
 default-features = false
 
 [workspace.dependencies.anyhow]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,12 +48,14 @@ default-features = false
 [workspace.dependencies.snarkvm]
 #path = "../snarkVM"
 git = "https://github.com/ProvableHQ/snarkVM.git"
-branch = "stream_plugin_testing"
+# Latest stream_plugin_testing commit
+rev = "8edbf9fc4a7bd0092ffa63b7a816ad9a76494440"
 default-features = false
 
 [workspace.dependencies.snarkvm-slipstream-plugin-manager]
 git = "https://github.com/ProvableHQ/snarkVM.git"
-branch = "stream_plugin_testing"
+# Latest stream_plugin_testing commit
+rev = "8edbf9fc4a7bd0092ffa63b7a816ad9a76494440"
 
 [workspace.dependencies.anyhow]
 version = "1.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -56,6 +56,7 @@ default-features = false
 git = "https://github.com/ProvableHQ/snarkVM.git"
 # Latest stream_plugin_testing commit
 rev = "0d50724d180f02cc9973cbb07c1cb8ec0c10212a"
+default-features = false
 
 [workspace.dependencies.anyhow]
 version = "1.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,13 +48,13 @@ default-features = false
 [workspace.dependencies.snarkvm]
 git = "https://github.com/ProvableHQ/snarkVM.git"
 # Latest stream_plugin_testing commit
-rev = "dbcc803158311d7c4ae88a387b426b3e96b1c63d"
+rev = "7fb6b76886c6a6daf359de380b4a5be2b923e11e"
 default-features = false
 
 [workspace.dependencies.snarkvm-slipstream-plugin-manager]
 git = "https://github.com/ProvableHQ/snarkVM.git"
 # Latest stream_plugin_testing commit
-rev = "dbcc803158311d7c4ae88a387b426b3e96b1c63d"
+rev = "7fb6b76886c6a6daf359de380b4a5be2b923e11e"
 default-features = false
 
 [workspace.dependencies.anyhow]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,13 +48,13 @@ default-features = false
 [workspace.dependencies.snarkvm]
 git = "https://github.com/ProvableHQ/snarkVM.git"
 # Latest stream_plugin_testing commit
-rev = "7fb6b76886c6a6daf359de380b4a5be2b923e11e"
+rev = "790bc32d54619db93ac35f3f2d0fae333e1082d6"
 default-features = false
 
 [workspace.dependencies.snarkvm-slipstream-plugin-manager]
 git = "https://github.com/ProvableHQ/snarkVM.git"
 # Latest stream_plugin_testing commit
-rev = "7fb6b76886c6a6daf359de380b4a5be2b923e11e"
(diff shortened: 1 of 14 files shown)


node/rest/src/lib.rs
```diff
@@ -137,7 +137,7 @@ impl<N: Network, C: ConsensusStorage<N>, R: Routing<N>> Rest<N, C, R> {
     fn build_routes(&self, rest_rps: u32) -> axum::Router {
         let cors = CorsLayer::new()
             .allow_origin(Any)
-            .allow_methods([Method::GET, Method::POST, Method::OPTIONS])
+            .allow_methods([Method::GET, Method::POST, Method::PUT, Method::DELETE, Method::OPTIONS])
```
dgrkotsonis: Would we prefer to keep these methods as-is? I can rework the routes if so

docs/slipstream_plugins.md
```diff
@@ -0,0 +1,198 @@
+# Slipstream Plugins
```
dgrkotsonis, line 1: Is there a better place to put this file rather than creating this new docs dir? (Unlike snarkVM there isn't a specific plugins crate to put this in)

node/rest/src/lib.rs
```diff
@@ -267,6 +267,17 @@ impl<N: Network, C: ConsensusStorage<N>, R: Routing<N>> Rest<N, C, R> {
...
+        // If the `slipstream-plugins` feature is enabled,
+        // enable the Slipstream plugin management endpoints (no auth required).
+        #[cfg(feature = "slipstream-plugins")]
+        let routes = routes
+            .route("/slipstream/plugins", get(Self::slipstream_list_plugins).post(Self::slipstream_load_plugin))
+            .route(
+                "/slipstream/plugins/{name}",
+                // TODO: PUT (reload) is not yet implemented.
+                axum::routing::delete(Self::slipstream_unload_plugin),
+            );
```
dgrkotsonis: Would it make sense to put these behind the auth middleware? Could foresee this being a potential issue if in theory anyone is able to hit a node and unload plugins

vicsn: Yes good idea

Cargo.toml
```diff
@@ -46,10 +46,15 @@ version = "1.0.1"
 default-features = false
 
 [workspace.dependencies.snarkvm]
-#path = "../snarkVM"
 git = "https://github.com/ProvableHQ/snarkVM.git"
-rev = "8c5ef1f5849cf12e2e00b2b4ca5cea152db48402"
-#version = "=4.6.0"
+# Latest staking commit
```
vicsn: Nits: this is a weird comment and can you leave the other comments intact to make local development easier? I always forget the right syntax

Cargo.toml
```diff
@@ -46,10 +46,15 @@ version = "1.0.1"
...
+rev = "159f04ce49a29674e0fe63530e0212a71d58417f"
+default-features = false
+
+[workspace.dependencies.snarkvm-slipstream-plugin-manager]
```
vicsn: Is there any way that we don't have to import snarkVM twice?

Cargo.toml
```diff
@@ -46,10 +46,15 @@ version = "1.0.1"
 default-features = false
 
 [workspace.dependencies.snarkvm]
-#path = "../snarkVM"
 git = "https://github.com/ProvableHQ/snarkVM.git"
-rev = "8c5ef1f5849cf12e2e00b2b4ca5cea152db48402"
-#version = "=4.6.0"
+# Latest staking commit
+rev = "159f04ce49a29674e0fe63530e0212a71d58417f"
```
dgrkotsonis: TODO: Update before merge

node/rest/src/routes.rs
```diff
@@ -1111,4 +1111,88 @@ impl<N: Network, C: ConsensusStorage<N>, R: Routing<N>> Rest<N, C, R> {
...
+        let manager = mgr_guard
+            .as_ref()
+            .ok_or_else(|| RestError::service_unavailable(anyhow!("No Slipstream plugin manager is installed")))?;
+        let plugins = manager
+            .list_plugins()
+            .map_err(|e: SlipstreamPluginManagerError| RestError::internal_server_error(anyhow!(e)))?;
+        Ok((StatusCode::OK, ErasedJson::pretty(plugins)))
+    }
+
+    /// POST /{network}/slipstream/plugins
```
vicsn, line 1216: Is it common practice to use API instead of flags to load?

node/rest/src/routes.rs
```diff
@@ -1111,4 +1111,88 @@ impl<N: Network, C: ConsensusStorage<N>, R: Routing<N>> Rest<N, C, R> {
...
+    #[cfg(feature = "slipstream-plugins")]
```
vicsn, line 1217: Is using API endpoints instead of flags common practice for (slipstream) plugins?

vicsn (changes requested):

node/rest/src/lib.rs
```diff
@@ -172,13 +172,25 @@ impl<N: Network, C: ConsensusStorage<N>, R: Routing<N>> Rest<N, C, R> {
...
+        #[cfg(feature = "slipstream-plugins")]
+        let auth_routes = auth_routes
+            .route("/slipstream/plugins", get(Self::slipstream_list_plugins).post(Self::slipstream_load_plugin))
+            .route(
+                "/slipstream/plugins/{name}",
+                // TODO: PUT (reload) is not yet implemented.
+                axum::routing::delete(Self::slipstream_unload_plugin),
+            );
+
+        let routes = axum::Router::new()
```
ljedrz, line 192: it would be nice to note here that the routes below are no longer protected, just to improve visibility

node/rest/src/routes.rs
```diff
@@ -1111,4 +1111,88 @@ impl<N: Network, C: ConsensusStorage<N>, R: Routing<N>> Rest<N, C, R> {
...
+                    SlipstreamPluginManagerError::PluginNotLoaded(_) => anyhow!("404: {e}"),
+                    other => anyhow!("{other}"),
+                },
+            )
+        })
+        .await
+        .map_err(|e| RestError::internal_server_error(anyhow!("Task join error: {e}")))?
+        .map_err(|e| {
+            let msg = e.to_string();
+            if msg.starts_with("404: ") {
```
ljedrz: nit: the numeric error code should be exposed in the error?

Cargo.toml
```diff
@@ -46,10 +46,15 @@ version = "1.0.1"
...
-rev = "8c5ef1f5849cf12e2e00b2b4ca5cea152db48402"
-#version = "=4.6.0"
+# Latest staging commit
+rev = "159f04ce49a29674e0fe63530e0212a71d58417f"
+default-features = false
+
+[workspace.dependencies.snarkvm-slipstream-plugin-manager]
+git = "https://github.com/ProvableHQ/snarkVM.git"
+# Latest staging commit
+rev = "159f04ce49a29674e0fe63530e0212a71d58417f"
```
ljedrz:
```suggestion
rev = "159f04ce49a29674e0fe63530e0212a71d58417f"
default-features = false
features = [ "slipstream-plugins" ]
```
wouldn't this work instead? fewer SHAs to keep track of

node/src/validator/mod.rs
```diff
@@ -469,6 +481,12 @@ impl<N: Network, C: ConsensusStorage<N>> NodeInterface<N> for Validator<N, C> {
         // Shut down the node.
         trace!("Shutting down the node...");
 
+        // Shut down the Slipstream plugin manager.
+        #[cfg(feature = "slipstream-plugins")]
+        if let Some(manager) = self.ledger.vm().finalize_store().slipstream_plugin_manager().write().as_mut() {
+            manager.unload();
```
ljedrz, line 502: is this operation fast enough to be done in an `async` context without blocking? I'm asking since similar operations were blocking in RPC

node/src/validator/mod.rs
```diff
@@ -103,6 +104,17 @@ impl<N: Network, C: ConsensusStorage<N>> Validator<N, C> {
         }
         .with_context(|| "Failed to initialize the ledger")?;
 
+        // Initialize the Slipstream plugin manager (if any config files were provided).
+        #[cfg(feature = "slipstream-plugins")]
+        if !slipstream_configs.is_empty() {
+            let manager =
+                snarkvm_slipstream_plugin_manager::SlipstreamPluginManager::from_config_files(slipstream_configs)
+                    .context("Failed to initialize Slipstream plugin manager")?;
+            ledger.vm().finalize_store().set_slipstream_plugin_manager(manager);
```
ljedrz, line 115: same question as [here](https://github.com/ProvableHQ/snarkOS/pull/4175/changes#r3209398476)

node/src/prover/mod.rs
```diff
@@ -200,7 +200,7 @@ impl<N: Network, C: ConsensusStorage<N>> Prover<N, C> {
             // If the node is not connected to any peers, then skip this iteration.
             if self.router.number_of_connected_peers() == 0 {
                 debug!("Skipping an iteration of the puzzle (no connected peers)");
-                tokio::time::sleep(Duration::from_secs(N::ANCHOR_TIME as u64)).await;
+                tokio::time::sleep(Duration::from_secs(N::REWARD_ANCHOR_TIME as u64)).await;
```
ljedrz: I think this should depend on the current consensus version, as this is 35s starting from `ConsensusVersion::V15` (as opposed to `REWARD_ANCHOR_TIME`, which is 25s)

node/src/client/mod.rs
```diff
@@ -544,6 +556,12 @@ impl<N: Network, C: ConsensusStorage<N>> NodeInterface<N> for Client<N, C> {
         // Shut down the node.
         trace!("Shutting down the node...");
 
+        // Shut down the Slipstream plugin service.
+        #[cfg(feature = "slipstream-plugins")]
+        if let Some(manager) = self.ledger.vm().finalize_store().slipstream_plugin_manager().write().as_mut() {
```
ljedrz, line 561: same question as [here](https://github.com/ProvableHQ/snarkOS/pull/4175/changes#r3209398476)

node/src/client/mod.rs
```diff
@@ -155,6 +156,17 @@ impl<N: Network, C: ConsensusStorage<N>> Client<N, C> {
         }
         .with_context(|| "Failed to initialize the ledger")?;
 
+        // Initialize the Slipstream plugin manager (if any config files were provided).
+        #[cfg(feature = "slipstream-plugins")]
+        if !slipstream_configs.is_empty() {
+            let manager =
+                snarkvm_slipstream_plugin_manager::SlipstreamPluginManager::from_config_files(slipstream_configs)
+                    .context("Failed to initialize Slipstream plugin manager")?;
+            ledger.vm().finalize_store().set_slipstream_plugin_manager(manager);
```
ljedrz, line 165: same question as [here](https://github.com/ProvableHQ/snarkOS/pull/4175/changes#r3209398476)

ljedrz (commented):
Left some comments, needs a rebase too.

node/rest/src/routes.rs
dgrkotsonis: We use flags for loading on startup, this is just for dynamic loading at runtime without interrupting the node

Cargo.toml
dgrkotsonis: This was intentional because the snarkVM crate doesn't declare the slipstream manager as a dependency because it uses libloading so I wanted to keep that separate from the core VM but I can change that in snarkVM if we think it's fine?

node/src/validator/mod.rs
dgrkotsonis: Good question, it depends on the plugin itself so hard to say. For the plugins that I've been working on which we plan to eventually publicize as "official" plugins, unloading is quick and simple, but in theory someone could create their own plugin that takes a while to unload, but it would only impact their own client. Is that something we're willing to accept/does potential blocking behavior matter much if it's on shutdown?

node/rest/src/routes.rs
dgrkotsonis: That's not an actual error code, that's just how I was identifying not found errors (see a few lines above)

node/src/client/mod.rs
dgrkotsonis: Similar answer to [here](https://github.com/ProvableHQ/snarkOS/pull/4175/changes#r3221147969) but it depends. Have had no issues thus far in testing but I could foresee someone using numerous inefficient plugins encountering some blocking. If this is just on initialization how low/high is our tolerance for potential delays?

node/src/client/mod.rs
dgrkotsonis: Answered [here](https://github.com/ProvableHQ/snarkOS/pull/4175/changes#r3209398476)

node/src/validator/mod.rs
dgrkotsonis: Answered [here](https://github.com/ProvableHQ/snarkOS/pull/4175#discussion_r3221480825)

Cargo.toml
dgrkotsonis: Answered [here](https://github.com/ProvableHQ/snarkOS/pull/4175#discussion_r3221038291) but I believe we would need to slightly rework snarkVM if we wanted to do this consolidation

node/src/validator/mod.rs
```diff
@@ -92,6 +92,7 @@ impl<N: Network, C: ConsensusStorage<N>> Validator<N, C> {
         trusted_peers_only: bool,
         dev_txs: bool,
         dev: Option<u16>,
+        slipstream_configs: &[std::path::PathBuf],
```
dgrkotsonis: Clippy is failing because this isn't used unless the slipstream-configs feature flag is enabled. I assume we don't want to add an underscore for cleanliness, should I just go ahead and conditionally pass/compile the param or could that get hairy in the future? @vicsn

node/src/client/mod.rs
ljedrz: as a general rule, the millisecond scale is already blocking territory

Cargo.toml
vicsn: Yeah I think it's fine to change in snarkVM, makes our development easier if we only have a single SHA to adjust in snarkOS.

node/src/validator/mod.rs
vicsn: underscore is fine for now, we're gonna refactor all these million parameters at some point with some config object

node/src/client/mod.rs
dgrkotsonis: @vicsn is this something worth reworking/revisiting? The plugins feature is opt-in so I assume any brief blocking one might encounter on startup/shutdown is acceptable for someone wanting to use the features. Plus, I'm not sure it would make sense to have these methods async: on startup we want to ensure plugins are connected to external sources and fully initialized before processing blocks (so we don't miss any data), and on shutdown we need to ensure that plugins are fully shut down before the shared library is unloaded. If this sounds reasonable, I can add spawn_blocking() around the callsites to these synchronous methods to make it more explicit that they're technically blocking

Cargo.toml
dgrkotsonis: Okay sounds good, I'll have a quick follow-up snarkVM PR shortly

dgrkotsonis: Addressed [here](https://github.com/ProvableHQ/snarkVM/pull/3256)

node/src/client/mod.rs
vicsn: The access to these endpoints is limited, so I'm also fine if we just create a new snarkOS issue to revisit all REST endpoints simultaneously on this topic in the future

vicsn (dismissed):

node/src/validator/mod.rs
ljedrz: For shutdown we should be fine, as we do it cleanly; as for runtime, usually millisecond scale is too much, but anything faster should be fine.

vicsn: Made an issue for the future: https://github.com/ProvableHQ/snarkOS/issues/4264

I think we can resolve for now

cli/src/commands/start.rs
```diff
@@ -858,11 +864,18 @@ impl Start {
...
+        // Collect slipstream plugin config paths (empty slice when feature is disabled).
+        #[cfg(feature = "slipstream-plugins")]
+        let slipstream_configs: &[PathBuf] = &self.slipstream_configs;
+        #[cfg(not(feature = "slipstream-plugins"))]
+        let slipstream_configs: &[PathBuf] = &[];
+
         // Initialize the node.
         let node = match node_type {
-            NodeType::Validator => Node::new_validator(node_ip, self.bft, rest_ip, self.rest_rps, account, &trusted_peers, &trusted_validators, genesis, cdn, storage_mode, node_data_dir, self.trusted_peers_only, self.auto_db_checkpoints.clone(), dev_txs, self.dev, dev_num_validators_for_committee_hotswap, signal_handler.clone()).await,
+            // NodeType::Validator => Node::new_validator(node_ip, self.bft, rest_ip, self.rest_rps, account, &trusted_peers, &trusted_validators, genesis, cdn, storage_mode, node_data_dir, self.trusted_peers_only, self.auto_db_checkpoints.clone(), dev_txs, self.dev, slipstream_configs, signal_handler.clone()).await,
```
vicsn, line 875: leftover from the rebase, you can fix in a new PR

vicsn (approved):

PR #4176, opened 2026-03-18
Prevent old block responses from causing a gateway disconnect

vicsn:
## Motivation

A mainnet validator lost quorum multiple times today. The following log was observed:
```
[MemoryPool] Disconnecting from '34.186.58.65:5000' - The sync request was removed because we already advanced —
```

## Test Plan

- CI, which is running chaotic devnets with ledger resets of quorum minority, should be sufficient
- Reviewers please evaluate the surrounding logic, `fn inbound` and `fn insert_block_response`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -788,7 +788,7 @@ workflows:
           filters:
             branches:
               only:
-                - feat/vk-migration
+                - prevent_block_response_race_disconnect
                 - canary
                 - testnet
                 - mainnet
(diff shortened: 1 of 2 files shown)


ljedrz (approved):
LGTM 👍

kaimast (approved):
LGTM

We should eventually add another `InsertBlockResponseError` type for this, so real error like invalid blocks are still handled properly.

vicsn:
> We should eventually add another InsertBlockResponseError type for this, so real error like invalid blocks are still handled properly.

Can you make an issue for this?

PR #4177, opened 2026-03-18
Prevent old block responses from causing a gateway disconnect

vicsn:
## Motivation

Duplicate of https://github.com/ProvableHQ/snarkOS/pull/4176
See the reasoning and test plan in that PR README.


ljedrz (approved):

mike2194:
watching this

kaimast (approved):

raychu86:
It would be better to just return `Ok(())` in `insert_block_response`.

So updating this:
```
        if self.ledger.contains_block_height(height) {
            bail!("The sync request was removed because we already advanced");
        }
```

Instead of returning Ok(true) where there is an error

PR #4179, opened 2026-03-18
[Fix] Do not attempt to sync when only one block behind

kaimast:
## Motivation

It looks like nodes always sync when `can_block_sync` is true, which can also be the case when a node is `Synced` and only one block behind.

Most likely it is more noticeable because somewhat recently we changed block sync to run more frequently (#3810).

## Proposed changes

* [e8c42bc](https://github.com/ProvableHQ/snarkOS/pull/4179/commits/e8c42bc0da64d4b9fb7798e4845e1a4b15d930b5) replaces the call to `can_block_sync` with a call to `is_block_synced`
* [5b84d80](https://github.com/ProvableHQ/snarkOS/pull/4179/commits/5b84d803b13c97f8b1c9969fb094daa9bf717570) extends `InsertBlockReponseError` to handle benign cases, such as the request becoming obsolete, better.

### DIFF
--- a/node/src/client/router.rs
+++ b/node/src/client/router.rs
@@ -30,7 +30,6 @@ use snarkos_node_router::{
         UnconfirmedTransaction,
     },
 };
-use snarkos_node_sync::InsertBlockResponseError;
 use snarkos_node_tcp::{Connection, ConnectionSide, Tcp};
 use snarkvm::{
     console::network::{ConsensusVersion, Network},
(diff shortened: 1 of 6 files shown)


node/src/client/router.rs
```diff
@@ -220,21 +219,32 @@ impl<N: Network, C: ConsensusStorage<N>> Inbound<N> for Client<N, C> {
...
+            Err(err) => {
+                let err: anyhow::Error = err.into();
+                let err = err.context(format!("Failed to insert block response from '{peer_ip}'"));
+                warn!("{}", flatten_error(err));
 
-            false
-        } else {
-            true
+                // TODO(kaimast): This needs more testing to ensure disconnect is the correct action.
+                true
```
ljedrz, line 246: I don't understand this comment; `true` doesn't indicate a disconnect, and would not impact the related connection

ljedrz (approved):
Left one comment, LGTM otherwise.

node/src/client/router.rs
kaimast: This is meant to say that we should maybe disconnect here, but it seems dangerous to make the change without additional testing.

I want to revisit this for 4.7.

vicsn (approved):
LGTM!

PR #4120, opened 2026-02-17
[CI] Do not delete node data in reset tests

kaimast:

PR #4183, opened 2026-03-25
[Fix] Only count validators for `SnarkOS BFT Connected Total` metric

kaimast:
Fixes #4120, by adding a new `PeerPoolHandling::number_of_connected_validators` method.

The PR is slightly bigger as it adds tests to `snarkos-node-network`. I don't plan to end-to-end test this in isolation, and hope the unit tests will suffice.

@ljedrz let me know if there are other peering changes in the works that should be merged first.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -504,6 +504,14 @@ jobs:
           workspace_member: node/consensus
           cache_key: node-consensus
 
+  node-network:
+    executor: rust-docker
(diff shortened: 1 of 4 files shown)


ljedrz:
There are no no pending changes that would clash with these :ok_hand:.

vicsn (approved):
LGTM!

vicsn:
Making a new issue for the failing job

PR #4190, opened 2026-03-30
[Feature] Batch Triggers

kaimast:
This PR changes the batch creation task, to trigger on certain events, instead of only periodically. The goal is to avoid cases where a node is not ready to propose and then waits another MIN_BATCH_DELAY (1s) instead of only for the relevant conditions. It also ensures that the batch delay timer is reset whenever the node advances.

The most important changes in this PR are in `Primary::start_handlers` and the new `proposal_task.rs` module. The [README](https://github.com/ProvableHQ/snarkOS/pull/4149/changes#diff-88224c24f5c0399890a8698944fe09ebd9245fabce5e2e9d04a44695d98aa211) details when the batch creation task is woken up.

## Overview of Changes
The major change of this PR is to only have proposals occur in a single location -- the proposal task. It gets woken up on specific events, such as timer expiration or receiving sufficient certificates for a round.
Unlike before, where it woke up every second and retired independent of round advancement, it now correctly resets timers after advancing rounds. 

**Minor Changes**
* [6167485](https://github.com/ProvableHQ/snarkOS/pull/4190/commits/6167485d5abeb4c947c1f05551b7fc57a002eb30) changes some of the time constants to use `Duration`. This avoids the potential for incorrect conversion of these constants
* [cefe21b](https://github.com/ProvableHQ/snarkOS/pull/4190/commits/cefe21b96c6e4041fd1b711739d7359f7642fffc) moves the proposal logic to a dedicated `proposal_task` module. This enables unit testing its behavior and when it will be woken up.

## Testing
End-to-end stress tests passed:
Screenshot 2026-03-30 at 5 08 16 PM

### DIFF
--- a/node/bft/src/lib.rs
+++ b/node/bft/src/lib.rs
@@ -55,9 +55,16 @@ pub const MEMORY_POOL_PORT: u16 = 5000; // port
 
 /// The maximum time to wait before proposing a batch.
 pub const MAX_BATCH_DELAY: Duration = Duration::from_millis(2500);
+
 /// The minimum time that needs to elapse between two consecutive batch proposals.
 /// This creates a lower bound on the block interval, and ensures the network will not be overwhelmed with too many blocks/certificates.
 pub const MIN_BATCH_DELAY: Duration = Duration::from_secs(1);
+
+/// The time a primary waits between attempts to create a new batch (only relevant after `MIN_BATCH_DELAY` has passed).
+/// This only serves as a failsafe in case the task does not get woken up through other means.
+/// eowering it too much will most likely waste
+pub const CREATE_BATCH_INTERVAL: Duration = Duration::from_millis(250);
+
 /// The maximum time to wait before timing out on a fetch.
 pub const MAX_FETCH_TIMEOUT: Duration = Duration::from_secs(3 * MAX_BATCH_DELAY.as_secs());
 /// The maximum time allowed for the leader to send their certificate.
--- a/node/bft/src/lib.rs
+++ b/node/bft/src/lib.rs
@@ -62,7 +62,7 @@ pub const MIN_BATCH_DELAY: Duration = Duration::from_secs(1);
 
 /// The time a primary waits between attempts to create a new batch (only relevant after `MIN_BATCH_DELAY` has passed).
 /// This only serves as a failsafe in case the task does not get woken up through other means.
-/// eowering it too much will most likely waste
+/// Lowering it too much would be wasteful.
 pub const CREATE_BATCH_INTERVAL: Duration = Duration::from_millis(250);
 
 /// The maximum time to wait before timing out on a fetch.
--- a/node/bft/src/lib.rs
+++ b/node/bft/src/lib.rs
@@ -66,21 +66,27 @@ pub const MIN_BATCH_DELAY: Duration = Duration::from_secs(1);
 pub const CREATE_BATCH_INTERVAL: Duration = Duration::from_millis(250);
 
 /// The maximum time to wait before timing out on a fetch.
-pub const MAX_FETCH_TIMEOUT: Duration = Duration::from_secs(3 * MAX_BATCH_DELAY.as_secs());
+/// TODO(kaimast): directy multiply by constant once the `const_trait_impl` feature is stable.
+pub const MAX_FETCH_TIMEOUT: Duration = Duration::from_millis(3 * (MAX_BATCH_DELAY.as_millis() as u64));
+
 /// The maximum time allowed for the leader to send their certificate.
 /// After this time, the node will consider the leader as failed and try to advance the round without it.
-pub const MAX_LEADER_CERTIFICATE_DELAY: Duration = Duration::from_secs(2 * MAX_BATCH_DELAY.as_secs());
+pub const MAX_LEADER_CERTIFICATE_DELAY: Duration = Duration::from_millis(2 * (MAX_BATCH_DELAY.as_millis() as u64));
+
 /// The maximum difference allowed between our local time and a certificate's timestamp, for the node to sign the certificate.
 /// This prevents malicious actors from proposing certificates with timestamps that are too log or too far in the future)
+/// w
 pub const MAX_TIMESTAMP_DELTA: Duration = Duration::from_secs(10);
+
 /// The maximum number of workers that can be spawned.
 pub const MAX_WORKERS: u8 = 1; // worker(s)
 
 /// The interval at which each primary broadcasts a ping to every other node.
 /// Note: If this is updated, be sure to update `MAX_BLOCKS_BEHIND` to correspond properly.
-pub const PRIMARY_PING_INTERVAL: Duration = Duration::from_secs(2 * MAX_BATCH_DELAY.as_secs());
+pub const PRIMARY_PING_INTERVAL: Duration = Duration::from_millis(2 * (MAX_BATCH_DELAY.as_millis() as u64));
+
 /// The interval at which each worker broadcasts a ping to every other node.
-pub const WORKER_PING_INTERVAL: Duration = Duration::from_secs(4 * MAX_BATCH_DELAY.as_secs());
+pub const WORKER_PING_INTERVAL: Duration = Duration::from_millis(4 * (MAX_BATCH_DELAY.as_millis() as u64));
 
 /// A helper macro to spawn a blocking task.
 #[macro_export]
--- a/node/bft/src/lib.rs
+++ b/node/bft/src/lib.rs
@@ -29,6 +29,8 @@ pub use snarkos_node_bft_events as events;
 pub use snarkos_node_bft_ledger_service as ledger_service;
 pub use snarkos_node_bft_storage_service as storage_service;
 
+use std::time::Duration;
+
 pub mod helpers;
 
 mod bft;
@@ -51,24 +53,27 @@ pub const CONTEXT: &str = "[MemoryPool]";
 /// The port on which the memory pool listens for incoming connections.
 pub const MEMORY_POOL_PORT: u16 = 5000; // port
 
-/// The maximum number of milliseconds to wait before proposing a batch.
-pub const MAX_BATCH_DELAY_IN_MS: u64 = 2500; // ms
-/// The minimum number of seconds to wait before proposing a batch.
-pub const MIN_BATCH_DELAY_IN_SECS: u64 = 1; // seconds
-/// The maximum number of milliseconds to wait before timing out on a fetch.
-pub const MAX_FETCH_TIMEOUT_IN_MS: u64 = 3 * MAX_BATCH_DELAY_IN_MS; // ms
-/// The maximum number of seconds allowed for the leader to send their certificate.
-pub const MAX_LEADER_CERTIFICATE_DELAY_IN_SECS: i64 = 2 * MAX_BATCH_DELAY_IN_MS as i64 / 1000; // seconds
-/// The maximum number of seconds before the timestamp is considered expired.
-pub const MAX_TIMESTAMP_DELTA_IN_SECS: i64 = 10; // seconds
+/// The maximum time to wait before proposing a batch.
+pub const MAX_BATCH_DELAY: Duration = Duration::from_millis(2500);
+/// The minimum time that needs to elapse between two consecutive batch proposals.
+/// This creates a lower bound on the block interval, and ensures the network will not be overwhelmed with too many blocks/certificates.
+pub const MIN_BATCH_DELAY: Duration = Duration::from_secs(1);
+/// The maximum time to wait before timing out on a fetch.
+pub const MAX_FETCH_TIMEOUT: Duration = Duration::from_secs(3 * MAX_BATCH_DELAY.as_secs());
+/// The maximum time allowed for the leader to send their certificate.
+/// After this time, the node will consider the leader as failed and try to advance the round without it.
+pub const MAX_LEADER_CERTIFICATE_DELAY: Duration = Duration::from_secs(2 * MAX_BATCH_DELAY.as_secs());
+/// The maximum difference allowed between our local time and a certificate's timestamp, for the node to sign the certificate.
+/// This prevents malicious actors from proposing certificates with timestamps that are too log or too far in the future)
+pub const MAX_TIMESTAMP_DELTA: Duration = Duration::from_secs(10);
 /// The maximum number of workers that can be spawned.
 pub const MAX_WORKERS: u8 = 1; // worker(s)
 
 /// The interval at which each primary broadcasts a ping to every other node.
 /// Note: If this is updated, be sure to update `MAX_BLOCKS_BEHIND` to correspond properly.
-pub const PRIMARY_PING_IN_MS: u64 = 2 * MAX_BATCH_DELAY_IN_MS; // ms
+pub const PRIMARY_PING_INTERVAL: Duration = Duration::from_secs(2 * MAX_BATCH_DELAY.as_secs());
 /// The interval at which each worker broadcasts a ping to every other node.
-pub const WORKER_PING_IN_MS: u64 = 4 * MAX_BATCH_DELAY_IN_MS; // ms
+pub const WORKER_PING_INTERVAL: Duration = Duration::from_secs(4 * MAX_BATCH_DELAY.as_secs());
 
 /// A helper macro to spawn a blocking task.
 #[macro_export]
--- a/node/bft/src/primary.rs
+++ b/node/bft/src/primary.rs
@@ -70,7 +70,7 @@ use indexmap::{IndexMap, IndexSet};
 #[cfg(feature = "locktick")]
 use locktick::{
     parking_lot::{Mutex, RwLock},
-    tokio::Mutex as TMutex,
+    tokio::RwLock as TRwLock,
 };
 #[cfg(not(feature = "locktick"))]
 use parking_lot::{Mutex, RwLock};
@@ -86,11 +86,11 @@ use std::{
     time::Duration,
 };
 #[cfg(not(feature = "locktick"))]
-use tokio::sync::Mutex as TMutex;
-use tokio::task::JoinHandle;
+use tokio::sync::RwLock as TRwLock;
+use tokio::{sync::Notify, task::JoinHandle};
 
-/// A helper type for an optional proposed batch.
-pub type ProposedBatch<N> = RwLock<Option<Proposal<N>>>;
+/// A helper type to keep track of the latest proposed batch.
+pub(crate) type ProposedBatch<N> = RwLock<Option<Proposal<N>>>;
 
 /// This callback trait allows listening to changes in the Primary, such as round advancement.
 /// This is currently used by BFT.
@@ -119,22 +119,31 @@ pub struct Primary<N: Network> {
     workers: Arc<OnceLock<Vec<Worker<N>>>>,
     /// The primary callback (used by [`BFT`]).
     primary_callback: Arc<CallbackHandle<Arc<dyn PrimaryCallback<N>>>>,
+
     /// The batch proposal, if the primary is currently proposing a batch.
     proposed_batch: Arc<ProposedBatch<N>>,
-    /// The timestamp of the most recent proposed batch.
-    latest_proposed_batch_timestamp: Arc<RwLock<i64>>,
+
     /// The instant at which the current batch was proposed (used to measure certification latency).
     /// (used for higher precision in the metrics compared to the batch timestamp)
     #[cfg(feature = "metrics")]
     batch_propose_start: Arc<Mutex<Option<Instant>>>,
+
+    /// Holds the most recent round and timestamp that the primary proposed a batch for.
+    /// TODO(kaimast): avoiding using an async lock here, so this can be merged with the `proposed_batch`,
+    /// to have a unified `primary_state` field.
+    latest_proposed_batch: Arc<TRwLock<Option<(u64, i64)>>>,
+
     /// The recently-signed batch proposals.
     signed_proposals: Arc<RwLock<SignedProposals<N>>>,
+
     /// The handles for all background tasks spawned by this primary.
     handles: Arc<Mutex<Vec<JoinHandle<()>>>>,
-    /// The lock for propose_batch. It holds the most recent round that was proposed for.
-    propose_lock: Arc<TMutex<u64>>,
+
     /// The node configuration directory.
     node_data_dir: NodeDataDir,
+
+    /// Allows notifying the proposal taks when a new batch can be proposed.
+    is_ready_notify: Arc<Notify>,
 }
 
 impl<N: Network> Primary<N> {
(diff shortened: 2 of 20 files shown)


node/sync/src/block_sync.rs
```diff
@@ -1109,15 +1195,19 @@ impl<N: Network> BlockSync<N> {
     /// This is a no-op if `new_height` is equal or less to the current sync height.
     pub fn set_sync_height(&self, new_height: u32) {
         // Scope state lock to avoid locking state and metrics at the same time.
-        let fully_synced = {
+        let (synced, fully_synced) = {
             let mut state = self.sync_state.write();
             state.set_sync_height(new_height);
-            !state.can_issue_new_block_requests()
+            (state.is_block_synced(), !state.can_issue_new_block_requests())
```
vicsn, line 1209: I may have asked this before, but when would we ever be synced but still able to issue new block requests?

node/bft/src/lib.rs
```diff
@@ -51,24 +53,34 @@ pub const CONTEXT: &str = "[MemoryPool]";
...
+/// The maximum time to wait before proposing a batch.
+pub const MAX_BATCH_DELAY: Duration = Duration::from_millis(2500);
+
+/// The minimum time that needs to elapse between two consecutive batch proposals.
+/// This creates a lower bound on the block interval, and ensures the network will not be overwhelmed with too many blocks/certificates.
+pub const MIN_BATCH_DELAY: Duration = Duration::from_secs(1);
+
+/// The time a primary waits between attempts to create a new batch (only relevant after `MIN_BATCH_DELAY` has passed).
+/// This only serves as a failsafe in case the task does not get woken up through other means.
+/// eowering it too much will most likely waste
```
vicsn:
```suggestion
/// Lowering it too much would be wasteful
```

node/bft/src/sync/mod.rs
```diff
@@ -984,12 +1006,6 @@ impl<N: Network> Sync<N> {
 impl<N: Network> Sync<N> {
     /// Returns `true` if the node is synced and has connected peers.
     pub fn is_synced(&self) -> bool {
-        // Ensure the validator is connected to other validators,
-        // not just clients.
-        if self.gateway.number_of_connected_peers() == 0 {
```
vicsn, line 989: Why is this safe to remove / why was this here in the first place?

node/consensus/src/lib.rs
```diff
@@ -501,13 +507,17 @@ impl<N: Network> Consensus<N> {
 
         // Process the unconfirmed transactions in the memory pool.
         //
-        // TODO (kaimast): This shouldn't happen periodically but only when new batches/blocks are accepted
-        // by the BFT layer, after which the worker's ready queue may have capacity for more transactions/solutions.
+        // This loop wakes up either when a block is committed (signaled via notify) or after a timeout.
+        // When a block is committed, the BFT workers have freed capacity for more transmissions.
```
vicsn, line 511: An even tighter feedback loop would be when a proposal is created (and the workers are drained) right? If you agree can you leave a new TODO?

node/metrics/src/names.rs
```diff
@@ -99,6 +99,12 @@ pub mod consensus {
     pub const COMMITTED_CERTIFICATES: &str = "snarkos_consensus_committed_certificates_total";
     pub const BLOCK_LATENCY: &str = "snarkos_consensus_block_latency_secs";
     pub const BLOCK_LAG: &str = "snarkos_consensus_block_lag_ms";
+    /// Time spent in prepare_advance_to_next_quorum_block (block construction).
+    pub const PREPARE_ADVANCE_SECS: &str = "snarkos_consensus_prepare_advance_secs";
```
vicsn, line 103: Is this granular enough? Should we track milliseconds?

vicsn (changes requested):
Cool, do you see a lower average blocktime in your stress tests or in the CI devnets?

node/sync/src/block_sync.rs
kaimast: If you are one block behind you are still considered "synced" but could also request one more block.

node/bft/src/sync/mod.rs
kaimast: Now `BlockSync` switches to "Unsynced" if no peers are connected, so we do not need this additional check in the BFT layer anymore.

node/consensus/src/lib.rs
kaimast: Added a comment

node/metrics/src/names.rs
kaimast: The metrics is stored as a float, so it should be fine-grained enough.

It's a little awkward that block lag is in milliseconds, though. Ideally, they would all use the same unit.

node/bft/src/lib.rs
kaimast: Fixed

node/bft/src/lib.rs
```diff
@@ -51,24 +53,34 @@ pub const CONTEXT: &str = "[MemoryPool]";
...
+/// The time a primary waits between attempts to create a new batch (only relevant after `MIN_BATCH_DELAY` has passed).
+/// This only serves as a failsafe in case the task does not get woken up through other means.
+/// Lowering it too much would be wasteful.
+pub const CREATE_BATCH_INTERVAL: Duration = Duration::from_millis(250);
+
+/// The maximum time to wait before timing out on a fetch.
+pub const MAX_FETCH_TIMEOUT: Duration = Duration::from_secs(3 * MAX_BATCH_DELAY.as_secs());
+/// The maximum time allowed for the leader to send their certificate.
+/// After this time, the node will consider the leader as failed and try to advance the round without it.
+pub const MAX_LEADER_CERTIFICATE_DELAY: Duration = Duration::from_secs(2 * MAX_BATCH_DELAY.as_secs());
```
raychu86: I don't believe this value is the same as before. This is now 4 seconds (2 * 2 = 4). Whereas it was 5 seconds before (2 * 2500 / 1000 = 5000 / 1000 = 5) .

node/bft/src/lib.rs
```diff
@@ -51,24 +53,34 @@ pub const CONTEXT: &str = "[MemoryPool]";
...
+/// This creates a lower bound on the block interval, and ensures the network will not be overwhelmed with too many blocks/certificates.
+pub const MIN_BATCH_DELAY: Duration = Duration::from_secs(1);
+
+/// The time a primary waits between attempts to create a new batch (only relevant after `MIN_BATCH_DELAY` has passed).
+/// This only serves as a failsafe in case the task does not get woken up through other means.
+/// Lowering it too much would be wasteful.
+pub const CREATE_BATCH_INTERVAL: Duration = Duration::from_millis(250);
+
+/// The maximum time to wait before timing out on a fetch.
+pub const MAX_FETCH_TIMEOUT: Duration = Duration::from_secs(3 * MAX_BATCH_DELAY.as_secs());
```
raychu86: Same truncation issue here

node/bft/src/primary.rs
```diff
@@ -117,24 +126,38 @@ pub struct Primary<N: Network> {
     ledger: Arc<dyn LedgerService<N>>,
     /// The workers.
     workers: Arc<OnceLock<Vec<Worker<N>>>>,
+
```
raychu86, line 131: nit: why the new lines here?

node/bft/src/primary.rs
```diff
@@ -396,25 +398,47 @@ impl<N: Network> Primary<N> {
     }
 }
 
-impl<N: Network> Primary<N> {
+#[async_trait::async_trait]
```
raychu86: nit: Duplicate line

node/bft/src/primary.rs
```diff
@@ -472,44 +498,42 @@ impl<N: Network> Primary<N> {
...
-            return Ok(());
+            return Ok(false);
         }
 
         #[cfg(feature = "metrics")]
         metrics::gauge(metrics::bft::PROPOSAL_ROUND, round as f64);
 
         // Ensure that the primary does not create a new proposal too quickly.
-        if let Err(err) = self.check_proposal_timestamp(previous_round, self.gateway.account().address(), now()) {
-            debug!(
```
raychu86, line 485: Was the removal of this debug message intentional?

node/bft/src/primary/proposal_task.rs
```diff
@@ -0,0 +1,472 @@
...
+                if let Some(sync_fut) = primary.wait_for_synced_if_syncing() {
+                    futures.push(sync_fut);
+                }
+
+                // If the maximum batch delay has not been reached, wait for it.
+                // Otherwise, add a small timeout to avoid busy waiting.
+                if reached_min_batch_delay {
+                    // Add a timeout to prevent nodes from getting stuck
+                    // (removing this would require more testing)
+                    let _ = tokio::time::timeout(CREATE_BATCH_INTERVAL, futures::future::join_all(futures)).await;
```
raychu86: If `futures` is empty here, does this skip the `CREATE_BATCH_INTERVAL` and run this on a loop (without sleeps) if `primary.propose_batch()` returns `Ok(false)`? 

Is this a possible scenario?

raychu86:
Direction looks good, it'll be nice to have invocation-based activity rather than just timer-based ones. 

Do you know what kind of speed-ups we should expect in consensus? Or if there is meaningful impact at all to block time?

node/consensus/src/lib.rs
```diff
@@ -648,49 +688,51 @@ impl<N: Network> Consensus<N> {
...
-                Err(err) => warn!("{}", flatten_error(err.context("Failed to get latest committee for telemetry"))),
             }
         }
 
         Ok(true)
     }
 
     /// Reinserts the given transmissions into the memory pool.
-    async fn reinsert_transmissions(&self, transmissions: IndexMap<TransmissionID<N>, Transmission<N>>) {
+    async fn reinsert_transmissions(&self, transmissions: IndexMap<TransmissionID<N>, Transmission<N>>) -> Result<()> {
```
ljedrz: why the `Result` if this is still infallible?

node/bft/src/primary/proposal_task.rs
```diff
@@ -0,0 +1,472 @@
...
+}
+
+/// The inner state of a [`ProposalTask`], shared via `Arc`.
+struct ProposalTaskInner {
+    /// Whether the primary is ready to propose a new batch.
+    ///
+    /// Initialized to `true` so round 1 can be proposed immediately without an explicit signal.
+    /// Set to `true` by [`ProposalTask::signal`] when a new round starts,
+    /// and reset to `false` after a batch is successfully proposed.
+    is_proposal_ready: RwLock<bool>,
```
ljedrz: wouldn't an `AtomicBool` suffice?

ljedrz (commented):
LGTM in general, though it is a non-negligible increase in the amount of logic; it would be great if in the near future we could generalize the "trigger X if Y or Z happens or a timeout occurs" logic across our use cases.

node/sync/src/block_sync.rs
vicsn: Can you write or link up some docblocks or README paragraph giving examples with specific block heights, for either validators and clients, for how the two concepts evolve?

node/bft/src/primary.rs
kaimast: It makes the multi-line doc comments easier to read IMHO, but I can revert it.

node/bft/src/primary.rs
kaimast: It will now return an error which is then logged as a warning. The intuition behind this is that the changes to the proposal task should prevent calling `propose_batch` too soon. If the error still appears there is something wrong with the proposal loop, and it should not be downgraded to a deubg message.

node/bft/src/primary/proposal_task.rs
kaimast: For it to work correctly in conjunction with `is_ready_notify` you need to hold a look to `is_proposal_ready` while registering the task to wait for the notification.

node/bft/src/lib.rs
kaimast: FIxed in [942daae](https://github.com/ProvableHQ/snarkOS/pull/4190/commits/942daae957eb1019b346da4ac2313eca422586c3). I also added a note to change it to simply `= 2 * MAX_BATCH_DELAY` once [`const_trait_impl`](https://github.com/rust-lang/rust/issues/143874) is stable.

node/bft/src/primary/proposal_task.rs
kaimast: I added a case for this in [3a10cef](https://github.com/ProvableHQ/snarkOS/pull/4190/commits/3a10cef4bb7cbbada537a4e6d846a4a6018c58a5). Not sure how likely it is but in that case it will simply sleep now.

kaimast:
> Do you know what kind of speed-ups we should expect in consensus? Or if there is meaningful impact at all to block time?

I am still trying to figure out a straightforward way of measuring this. My expectation is that the average block time stays about the same but the tail end of block times will be lower.

vicsn:
Forwarding from @kaimast , this PR introduces a significant improvement in commit latency for light networks (5 validators, weak machines, see image below), and at least no regression for normal prerelease runs.

Screenshot 2026-04-13 at 10 47 50 PM

node/bft/src/primary/proposal_task.rs
```diff
@@ -204,9 +174,12 @@ impl<N: Network> ProposalTask<N> {
                     }
                     Ok(false) => (), // retry
                     Err(err) => {
+                        // Log warning and retry
                         warn!("{}", flatten_error(err.context("Cannot propose a batch")));
                     }
                 }
+
+                attempt += 1;
```
vicsn, line 160: Do we or should we ever decrement `attempt`?

node/bft/src/primary/proposal_task.rs
```diff
@@ -144,53 +118,49 @@ impl<N: Network> ProposalTask<N> {
...
+                                    },
+                                    _ = self.wait().boxed() => {},
+                                    _ = sleep(CREATE_BATCH_INTERVAL) => {
+                // Retry if the timeout triggered (but do not count it as a proper attempt).
+                                    debug!("Skipping batch proposal for round {round} {}", "(not ready yet)".dimmed());
+                                    continue;
+                                    }
+                                };
+
+                if attempt > 1 {
```
vicsn, line 135: Am I reading correctly that when `attempt > 1`, we first sleep for at least `MIN_BATCH_DELAY`, and then again for at least `CREATE_BATCH_INTERVAL`? That seems like too much. Or did you intend for this to be `sleep_until(round_start + CREATE_BATCH_INTERVAL)` or something?

Maybe you can document the role of `attempt` better

vicsn (changes requested):

node/sync/src/block_sync.rs
vicsn: Ping

node/bft/src/primary/proposal_task.rs
```diff
@@ -0,0 +1,439 @@
...
+                                        }
+                                        leader_timeout_reached = true;
+                                    },
+                                    _ = self.wait().boxed() => {},
+                                    _ = sleep(CREATE_BATCH_INTERVAL) => {
+                // Retry if the timeout triggered (but do not count it as a proper attempt).
+                                    debug!("Skipping batch proposal for round {round} {}", "(not ready yet)".dimmed());
+                                    continue;
+                                    }
+                                };
```
ljedrz: I think something wrong went with the indentation here; `rustfmt` likely missed it because it's in a macro

node/bft/src/primary/proposal_task.rs
ljedrz: this is fine, it gets reset with every iteration of the outer loop, once the round advances or a proposal is cut

node/bft/src/primary.rs
```diff
@@ -117,24 +126,38 @@ pub struct Primary<N: Network> {
...
+
     /// The instant at which the current batch was proposed (used to measure certification latency).
     /// (used for higher precision in the metrics compared to the batch timestamp)
     #[cfg(feature = "metrics")]
     batch_propose_start: Arc<Mutex<Option<Instant>>>,
+
+    /// Holds the most recent round and timestamp that the primary proposed a batch for.
+    /// TODO(kaimast): avoiding using an async lock here, so this can be merged with the `proposed_batch`,
+    /// to have a unified `primary_state` field.
+    latest_proposed_batch: Arc<TRwLock<Option<(u64, i64)>>>,
```
ljedrz: I'd change this name to `proposed_batch_timestamp` or something else which would make it more distinct from the neighboring `proposed_batch`

node/consensus/src/lib.rs
```diff
@@ -538,20 +550,31 @@ impl<N: Network> Consensus<N> {
...
+            }
+            Ok(false) | Err(_) => {
+                if let Err(err) = self.reinsert_transmissions(transmissions).await {
+                    error!(
+                        "{}",
+                        flatten_error(
+                            err.context("Failed to reinsert transmissions after unsuccessful block advancement")
+                        )
+                    );
+                }
```
ljedrz: this is currently impossible, as `reinsert_transmissions` is infallible

node/bft/src/primary/proposal_task.rs
```diff
@@ -0,0 +1,439 @@
...
+            let mut fut = std::pin::pin!(self.inner.is_ready_notify.notified());
+            {
+                if *self.inner.is_proposal_ready.read() {
+                    return;
+                }
+                fut.as_mut().enable();
+            }
+            fut.await;
+        }
+    }
```
ljedrz: this is pretty complex, I'd expect such code in `Future` trait impls; wouldn't the following work?
```
async fn wait(&self) {
    loop {
        let fut = self.inner.is_ready_notify.notified();

        if *self.inner.is_proposal_ready.read() {
            return;
        }

        fut.await;
    }
}
```
alternatively, this might be a good use case for [watch](https://docs.rs/tokio/latest/tokio/sync/watch/index.html)

ljedrz (commented):
Left a new round of comments.

node/bft/src/primary/proposal_task.rs
kaimast: I manually tried to correct this before pushing the commit, but rustfmt actually adds these indentations. Not sure if this a bug or what causes it.

node/bft/src/primary.rs
kaimast: I changed it to `latest_proposal_timestamp`. The `latest` makes it clear that this is also a valid value when `proposed_batch` is none. The naming is not perfect, because this also contains a round number, but it should be fine for now.

node/consensus/src/lib.rs
kaimast: I made `reinsert_transmissions` infallible to simplify/remove this part of the code.

node/bft/src/primary/proposal_task.rs
kaimast: `watch` is slightly less efficient because the waiting task will also wake up when `is_ready` changes to `false`.
However, I still changed it to use `watch`, because that should make the code easier to maintain. Thanks for the suggestion!

node/router/tests/cleanups.rs
```diff
@@ -97,6 +97,12 @@ async fn test_connection_cleanups() {
     // Register final heap use.
     let heap_after_loop = PEAK_ALLOC.current_usage();
 
-    // Final heap use should equal that after the first connection.
-    assert_eq!(heap_after_one_conn.unwrap(), heap_after_loop);
+    // Final heap use should be close to that after the first connection.
+    // We allow up to 1 KiB of growth to accommodate small per-task allocations that
+    // tokio's runtime retains internally across connections (e.g. sharded queue state).
+    let heap_growth = heap_after_loop.saturating_sub(heap_after_one_conn.unwrap());
```
kaimast, line 103: I had this test failed and got the following explanation by Claude:

> The ~700-byte heap growth across 9 extra connect/disconnect cycles comes from `Data::deserialize()` calling `tokio::task::spawn_blocking()` during signature
> verification in the handshake (`verify_challenge_response`). When a signature arrives from the network it is a `Data::Buffer(bytes)` variant, so snarkvm
> offloads the cryptographic deserialization to a blocking thread.
> 
> Each full connect/disconnect cycle triggers this twice (once on each side of the handshake), so 9 extra cycles = 18 `spawn_blocking` calls beyond the baseline.
> 
> In tokio 1.52.0 (upgraded from 1.51.1 by the `cargo update` on this branch), [PR #7757 ](https://github.com/tokio-rs/tokio/pull/7757
) replaced the blocking pool's injection queue with a sharded structure o improve scalability. The sharded queue retains a small amount of internal bookkeeping per submitted task even after the task completes — roughly 39 bytes per call × 18 calls ≈ 700 bytes. All of our application-level data structures (peer_pool, resolver, cache, conn.tasks, tcp.tasks) are fully cleaned up after each cycle; the residual allocation is entirely inside tokio's runtime internals.
> 
> The 1 KiB assertion bound accounts for this overhead while still being tight enough to catch any real per-connection leak (even accumulating a small struct  across 9 connections would far exceed 1 KiB).

@lukasz the fix makes sense to me, but maybe you can also take quick look

kaimast:
I think I addressed all the comments. However, the most recent stresstest run for this branch was not successful, and I need to address that first before we can merge the branch.

node/bft/src/primary/proposal_task.rs
kaimast: The second and subsequent calls to sleep should return immediately, because `round_start` is only reset when the round number changes.

node/router/tests/cleanups.rs
ljedrz: this is perfectly in line with my past experiences with `tokio`, where some allocated memory is retained due to plumbing: https://github.com/tokio-rs/tokio/issues/4031

ljedrz (approved):

ljedrz (approved):
LGTM with the stresstest issue resolved.

raychu86 (approved):

kaimast:
I fixed an issue with batches being rebroadcasted too often. It passed the most recent stress test run ([Grafana link](https://aleostresstest.grafana.net/d/prerelease-devnet/prerelease-devnet?from=2026-05-01T20:19:18.193Z&to=2026-05-02T00:19:18.193Z&refresh=5s&timezone=browser&orgId=1)). 

It should be good to merge now :)

node/bft/src/primary/proposal_task.rs
```diff
@@ -88,96 +88,154 @@ impl<N: Network> ProposalTask<N> {
...
+            }
+
+            // Enforce the minimum inter-proposal delay.
+            // This is a no-op once the deadline has already passed.
+            sleep_until(round_start + MIN_BATCH_DELAY).await;
+
+            // Wait for a readiness signal, the MAX_BATCH_DELAY deadline, or a short heartbeat
+            // that lets the round-change check at the top of the loop fire regularly.
+            tokio::select! {
+                _ = sleep_until(round_start + MAX_BATCH_DELAY) => {
```
ljedrz, line 148: can this branch ever not fire, considering that we've already ensured that this point in time has already passed in L143?

ljedrz: ah no, I misread - it's `MAX` and not `MIN` this time

ljedrz (approved):

node/sync/src/block_sync.rs
vicsn: Made an issue: https://github.com/ProvableHQ/snarkOS/issues/4249

vicsn (approved):
Great work, LGTM!

Given the scope of the PR, I think our current benchmarking suffices, but just want to highlight that given our recent increases in blocktime we should prioritize https://github.com/ProvableHQ/snarkOS/issues/4250

PR #4201, opened 2026-04-02
[Feature] initial commit of snarkos grafana dashboard

mike2194:
## Motivation

Provide starting point for node administration observability.

## Test Plan

N/A

## Documentation

add first draft of Snarkos Grafana Dashboard
add first draft of Snarkos Prometheus Alert Rules

## Backwards compatibility

N/A

### DIFF
--- a/scripts/observability/alerts.conf.yaml
+++ b/scripts/observability/alerts.conf.yaml
@@ -1,0 +1,28 @@
+---
+groups:
+  - name: snarkos_alerts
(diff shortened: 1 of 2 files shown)


vicsn (changes requested):
We also have https://github.com/ProvableHQ/snarkOS/blob/staging/node/metrics/snarkOS-grafana.json - I guess that's outdated and can be removed. CC @miazn

PR #4202, opened 2026-04-03
[Fix] Properly handle late-arriving signatures

### LINKED ISSUE #4092: [Bug] Incorrect logging of Unknown batch ID
## 🐛 Bug Report

This is low priority but `Unknown batch ID` is incorrectly frequently printed in production. This is due to a benign race condition: a validator evicts a proposal from the Primary (`proposed_batch.take()`) before it lands in the certificate storage checked by `self_.storage.contains_batch`

kaimast:
This PR fixes #4092 by introducing a `ProposedBatchState` enum that properly handles the different states of certification a new batch can be in. Requires #4190 to be merged first.

It also moves signatures handling to a dedicated `Primary::add_signature_to_batch` method and adds unit tests for it. The test coverage is almost a little excessive, but it is a critical part of the code and the tests only take about two seconds to run on my machine.

### DIFF
--- a/node/bft/src/primary.rs
+++ b/node/bft/src/primary.rs
@@ -65,7 +65,7 @@ use snarkvm::{
         narwhal::{BatchCertificate, BatchHeader, Data, Transmission, TransmissionID},
         puzzle::{Solution, SolutionID},
     },
-    prelude::{ConsensusVersion, committee::Committee},
+    prelude::{ConsensusVersion, Signature, committee::Committee},
     utilities::flatten_error,
 };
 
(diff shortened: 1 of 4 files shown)


node/bft/src/primary.rs
```diff
@@ -2064,7 +2169,11 @@ impl<N: Network> Primary<N> {
         self.handles.lock().drain(..).for_each(|handle| handle.abort());
         // Save the current proposal cache to disk.
         let proposal_cache = {
-            let proposal = self.proposed_batch.write().take();
+            // Only persist a Certifying batch; a Certified batch will appear in pending_certificates.
```
vicsn, line 2172: Can this theoretically be in a race condition with certification/storage insertion? If yes would having this as a duplicate in the cache as proposal and certificate be a problem? If so this would have been an issue independent already before this PR

node/bft/src/primary.rs
```diff
@@ -1100,60 +1249,11 @@ impl<N: Network> Primary<N> {
...
-                None => return Ok(None),
-            };
-            // Retrieve the batch proposal, clearing the proposed batch.
-            match proposed_batch.take() {
-                Some(proposal) => Ok(Some(proposal)),
-                None => Ok(None),
-            }
+            let (result, new_state) =
+                self_.add_signature_to_batch(std::mem::take(&mut *proposed_batch), peer_ip, batch_id, signature);
+            *proposed_batch = new_state;
```
vicsn, line 1255: Are there existing or hypothetical scenario's where it's a problem that `proposed_batch` is temporarily empty?

kaimast: It's fine because a write lock is held to `proposed_batch`while this happens. So there cannot be any concurrent accesses to it.

node/bft/src/primary.rs
kaimast: The `shut_down` logic first stops all handlers, so there can't be any concurrent updates to this anymore.

vicsn (dismissed):

vicsn (approved):

PR #4203, opened 2026-04-04
[Fix] Ensure `TCP_TASKS` is always decremented properly

### LINKED ISSUE #4200: [Bug] Metric tcp_tasks should be recorded as counter, it is an incrementing value
## 🐛 Bug Report

low-severity issue with how metric [`TCP_TASKS` is registered](https://github.com/ProvableHQ/snarkOS/blob/c2f52badcd1be67ac607ea4983fb8b50102dba4c/node/metrics/src/names.rs#L47) is a value that increments forever over the lifetime of the running process as there are no decrement calls.  However, the series is registered as a "gauge" while "counter" is a better fit.   This issue affects certain functional defaults within GCP observability tools, Prometheus functions, and how data is recorded in the tsdb.

## Steps to Reproduce
N/A

## Expected Behavior

Value for metric series should decrement as well as increment if declared as a gauge.

## Your Environment

- 4.6.0
- N/A
- N/A

kaimast:
I noticed a bug in the TCP logic that prevented the gauge for `TCP_TASKS` to be decremented properly when the channel `inbound_message` was closed. This PR fixes the bug and should also resolve #4200.

The changes add a simple `QueuedMessageGuard` that decrements the counter when its dropped, so that when the channel is closed, and the messages within the channel are dropped, the gauge is also decremented accordingly.

### DIFF
--- a/node/tcp/src/protocols/reading.rs
+++ b/node/tcp/src/protocols/reading.rs
@@ -139,7 +139,8 @@ impl<R: Reading> ReadingInternal for R {
             framed.read_buffer_mut().reserve(Self::INITIAL_BUFFER_SIZE);
         }
 
-        let (inbound_message_sender, mut inbound_message_receiver) = mpsc::channel(self.message_queue_depth());
+        let (inbound_message_sender, mut inbound_message_receiver) =
+            mpsc::channel::<(R::Message, QueuedMessageGuard)>(self.message_queue_depth());
 
         // use a channel to know when the processing task is ready
         let (tx_processing, rx_processing) = oneshot::channel::<()>();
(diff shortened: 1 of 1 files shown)


node/tcp/src/protocols/reading.rs
```diff
@@ -216,8 +216,6 @@ impl<R: Reading> ReadingInternal for R {
                             warn_about_dropped_messages(&node, addr, &mut dropped_count, &mut last_drop_log);
                             debug!(parent: node.span(), "the inbound queue for {addr} is no longer saturated");
                         }
-                        #[cfg(feature = "metrics")]
```
kaimast, line 219: This increment also happened on failure, which was another issue with the previous implementation.

vicsn (approved):
Cool approach

ljedrz (approved):
LGTM; note: I wouldn't call this TCP_TASKS (as there are more, mostly persistent), but perhaps sth like PENDING_INBOUND_MESSAGES).

kaimast:
> LGTM; note: I wouldn't call this TCP_TASKS (as there are more, mostly persistent), but perhaps sth like PENDING_INBOUND_MESSAGES).

Should we make a new issue for this?

ljedrz:
@kaimast this is ultimately a UX concern, so it could also be adjusted directly in Grafana; it's not strictly wrong either, just imprecise.

PR #4206, opened 2026-04-08
[Fix] Improve parsing errors for --peers & --validators

### LINKED ISSUE #3991: [Bug] Invalid peer address fails with opaque error message
## 🐛 Bug Report

The following is all on a local devnet when passing values to `--peers` or `--validators`

I still have to isolate when this happens, but it is either when we fail to prepend `http://`, or when we use `localhost` instead of `127.0.0.1`, or it is just happening randomly:

```2025-11-03T13:00:28.012609Z  WARN snarkos_node_network::peering: [Router] Unable to connect to '[::1]:4131' - I/O error: Address family not supported by protocol family (os error 47)```

To add to the fun, for `--validators` I can't prepend `http://`:
```
     ↳ The supplied trusted hostname or IP ('http://localhost:5000') is malformed: failed to lookup address information: nodename nor servnam
e provided, or not known                                                                                                                     
```

ljedrz:
Potential parsing errors for `--peers` and `--validators` are currently opaque. This PR makes the most likely ones more clear, slightly clarifies the docs for both parameters, and references the underlying [`getaddrinfo`](https://en.wikipedia.org/wiki/Getaddrinfo#getaddrinfo).

Fixes https://github.com/ProvableHQ/snarkOS/issues/3991.

### DIFF
--- a/cli/src/commands/start.rs
+++ b/cli/src/commands/start.rs
@@ -155,7 +155,7 @@ pub struct Start {
     #[clap(long, requires = "validator")]
     pub bft: Option<SocketAddr>,
 
-    /// Specify the IP address and port of the peer(s) to connect to (as a comma-separated list).
+    /// Specify the host:port address pairs of the peer(s) to connect to (as a comma-separated list).
     ///
     /// These peers will be set as "trusted", which means the node will not disconnect from them when performing peer rotation.
     ///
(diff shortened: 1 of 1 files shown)


vicsn (approved):
LGTM

PR #4207, opened 2026-04-08
[Feat] Account for stake when determining request redundancy

### LINKED ISSUE #3857: [Bug] Certificate fetching redundancy must be increased
## 🐛 Bug Report

I think this line might be technically wrong: https://github.com/ProvableHQ/snarkOS/blob/mainnet/node/bft/src/helpers/pending.rs#L55

If stake is spread among the validators unevenly, and a rich node is offline, then we might want to fetch from all of the remaining validators.

Short term fix: re-evaluate how long it takes for these requests to time out and re-issue. We should not dumbly increase the number of requests as this will negatively impact performance.

Long term fix: consider adjusting the Pending queue to account for the stake of the respective peers and adjust redundancy accordingly.

ljedrz:
This PR guards against a scenario where a BFT worker would only request certificates or transmissions from "weak" validators - in other words, it ensures that even if the maximum number of redundant requests has been sent, the collective stake of those who have received them is greater than 33%.

Closes https://github.com/ProvableHQ/snarkOS/issues/3857.

### DIFF
--- a/node/bft/src/helpers/pending.rs
+++ b/node/bft/src/helpers/pending.rs
@@ -13,14 +13,15 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use crate::MAX_FETCH_TIMEOUT_IN_MS;
+use crate::{Gateway, MAX_FETCH_TIMEOUT_IN_MS};
 use snarkos_node_bft_ledger_service::LedgerService;
+use snarkos_node_network::PeerPoolHandling;
 use snarkvm::{
     console::network::{Network, consensus_config_value},
     prelude::Result,
 };
 
-use anyhow::anyhow;
+use anyhow::{anyhow, bail};
 #[cfg(feature = "locktick")]
 use locktick::parking_lot::RwLock;
 #[cfg(not(feature = "locktick"))]
@@ -109,6 +110,13 @@ impl<T: Copy + Clone + PartialEq + Eq + Hash, V: Clone> Pending<T, V> {
     /// Returns the peer IPs for the specified `item`.
     pub fn get_peers(&self, item: impl Into<T>) -> Option<HashSet<SocketAddr>> {
         self.pending.read().get(&item.into()).map(|map| map.keys().cloned().collect())
+    }
+
+    /// Returns the peer IPs for the specified `item` who have received a request for it.
+    pub fn get_sent_peers(&self, item: impl Into<T>) -> Option<HashSet<SocketAddr>> {
+        self.pending.read().get(&item.into()).map(|map| {
+            map.iter().filter_map(|(addr, cbs)| cbs.iter().any(|(_, _, sent)| *sent).then_some(*addr)).collect()
+        })
     }
 
     /// Returns the number of pending callbacks for the specified `item`.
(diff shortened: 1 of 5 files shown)


gai1680857-prog (approved):

node/bft/src/helpers/pending.rs
```diff
@@ -111,6 +112,13 @@ impl<T: Copy + Clone + PartialEq + Eq + Hash, V: Clone> Pending<T, V> {
         self.pending.read().get(&item.into()).map(|map| map.keys().cloned().collect())
     }
 
+    /// Returns the peer IPs for the specified `item` who have received a request for it.
+    pub fn get_sent_peers(&self, item: impl Into<T>) -> Option<HashSet<SocketAddr>> {
+        self.pending.read().get(&item.into()).map(|map| {
+            map.iter().filter_map(|(addr, cbs)| cbs.iter().any(|(_, _, sent)| *sent).then_some(*addr)).collect()
```
vicsn, line 118: Nit: maybe we can avoid collecting, but that would also be more complex for unclear benefit at this stage

node/bft/src/worker.rs
```diff
@@ -497,9 +503,16 @@ impl<N: Network> Worker<N> {
...
+        #[cfg(test)]
+        let stake_redundancy_reached = true;
+        #[cfg(not(test))]
+        let stake_redundancy_reached = self.pending.request_stake_redundancy_reached(&self.gateway, transmission_id)?;
         // Determine if we should send a transmission request to the peer.
-        // We send at most `num_redundant_requests` requests and each peer can only receive one request at a time.
-        let should_send_request = num_sent_requests < num_redundant_requests && !contains_peer_with_sent_request;
+        // Each peer can only receive one request at a time.
+        // We send at most `num_redundant_requests` requests, unless the stake redundancy factor hasn't been reached.
+        let should_send_request = !contains_peer_with_sent_request
```
vicsn, line 514: Can we short circuit in a clean way so we don't have to do unnecessary computations?

node/bft/src/worker.rs
```diff
@@ -497,9 +503,16 @@ impl<N: Network> Worker<N> {
         let contains_peer_with_sent_request = self.pending.contains_peer_with_sent_request(transmission_id, peer_ip);
         // Determine the maximum number of redundant requests.
         let num_redundant_requests = max_redundant_requests(self.ledger.clone(), self.storage.current_round())?;
+        // Establish whether the peers who already got the request collectively hold sufficient stake.
+        #[cfg(test)]
```
vicsn, line 507: Given our existing nice unit tests, it would be nice if we can test the behaviour of either without complex compile time guards.

Maybe we can pass in a `redundancy_variant` as function argument or generic

vicsn (changes requested):

node/bft/src/worker.rs
```diff
@@ -51,6 +53,9 @@ pub struct Worker<N: Network> {
     /// The worker ID.
     id: u8,
     /// The gateway.
+    #[cfg(not(test))]
```
kaimast, line 56: Is this just to avoid dynamic dispatch in production? Would be good to add a doc comment here. 

Couldn't we also just change the struct to `Worker<N: Network, G: Transport<N>>`?

node/bft/src/helpers/pending.rs
```diff
@@ -244,6 +252,31 @@ impl<T: Copy + Clone + PartialEq + Eq + Hash, V: Clone> Pending<T, V> {
             !peer_map.is_empty()
         });
     }
+
+    /// Ensures that the combined stake of the peers that have received the request is
+    /// greater than the redundancy threshold (33%).
+    pub fn request_stake_redundancy_reached<N: Network>(&self, gateway: &Gateway<N>, desired_item: T) -> Result<bool> {
+        let committee = match gateway.ledger().current_committee() {
```
kaimast, line 259: How expensive is this function, especially fetching the committee? It looks like it will be called often.

node/bft/src/worker.rs
ljedrz: no, this is because in tests, there's a `MockGateway`, which only implements `Transport`; we can't have that field be generic (with the current generics), as I lose access to the needed methods

node/bft/src/helpers/pending.rs
ljedrz: collapsing to https://github.com/ProvableHQ/snarkOS/pull/4207#discussion_r3053623042

node/bft/src/helpers/pending.rs
ljedrz: not really (as it's behind a lock), but this isn't a concern with the short-circuiting enhancement

node/bft/src/worker.rs
ljedrz: https://github.com/ProvableHQ/snarkOS/pull/4207/changes/c6e7980a7de51ec5bbd80271dae82237e7d7c035

ljedrz:
I'm not sure how to work around the `#[cfg(test)]`; the new method requires the `Gateway` in order to have access to `PeerPoolHandling::resolve_to_aleo_addr` (`LedgerService::current_committee` could be obtained from the `ledger` member, which would have been redundant if a full `Gateway` was available under `Worker`).

As far as I can tell, the only reason for the `Gateway` being type-erased as the `Transport` is testing (which uses a `MockGateway` as the `Transport` for the `Worker`).

node/bft/src/worker.rs
ljedrz: note that this _is_ being tested wherever a "full" (containing a `Gateway`) `Worker` is used, like in the 2 tests that required adjustments as part of these changes

vicsn (approved):

PR #4220, opened 2026-04-17
[Chore] Update snarkVM rev

kaimast:
Companion PR for https://github.com/ProvableHQ/snarkVM/pull/3223/. Depends on https://github.com/ProvableHQ/snarkVM/pull/3241

### DIFF
--- a/cli/src/commands/developer/execute.rs
+++ b/cli/src/commands/developer/execute.rs
@@ -26,7 +26,7 @@ use snarkvm::{
         Address,
         Identifier,
         Locator,
-        Process,
+        ProcessExclusiveGuard,
         ProgramID,
         VM,
         Value,
@@ -162,7 +162,7 @@ impl Execute {
                 debug!("At block height {height} and consensus {version:?}");
 
                 // Load the program and it's imports into the process.
-                load_program(&query, &mut vm.process().write(), &program_id, &endpoint)?;
+                load_program(&query, &vm.process().lock(), &program_id, &endpoint)?;
             }
 
             // Prepare the fee.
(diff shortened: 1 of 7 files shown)


cli/src/commands/developer/execute.rs
```diff
@@ -269,6 +269,7 @@ fn load_program<N: Network>(
     if !process.contains_program(program.id()) {
         debug!("Adding program {program_id} with edition {edition}");
         process
+            .lock()
```
kaimast: @ljedrz with this change the lock is not held throughout `execution_cost`. This is because the lock guard does not seem to expose some of the methods needed above, e.g., `Process::contains_program`.

Do you think this change could cause issues? If so, I can make a snarkVM PR to expose those methods.

ljedrz: no, as long as the `ProcessExclusiveGuard` is acquired, the user should be free to also use any methods available to `Process`; therefore, you could even add a `impl Deref<Process> for ProcessExclusiveGuard`, which should cover this

vicsn: Just so we're on the same page, this discussion is not about `cli/src/commands/developer/execute.rs:272` perse, but about the general safety of the new Process locking right?

Indeed if a program upgrades in the middle of a call to `execution_cost`, its result will be stale.

We overcome this in snarkVM's `fn check_transaction` by checking at the end that no program upgrades happened since the start of the call.

However, I now realise that we may need similar protection for snarkOS's `fn transaction_spend_in_microcredits` might need similar protection. Removed it here: https://github.com/ProvableHQ/snarkOS/pull/4254

kaimast: I will make a small PR for `impl Deref<Process>`, so all read-only functions are exposed for `ProcessExclusiveGuard`. That seems the cleanest and most future-proof way to do it.

ljedrz (approved):

vicsn (approved):

Issue #4228, opened 2026-04-23
[Bug] Connections to lots of already connected validators


vicsn:
## 🐛 Bug Report

Most of our logs may show that we're only disconnected from e.g. "Not connected to 4 validators"

And yet, we're seeing:
- Lots of "Connecting to <IP>:5000..."
- "Reconnnecting to 11 out of ...total... trusted validators"

Open question: 
- `check_connection_attempt` might not be comprehensive enough
- do we see this behaviour on the other networks as well?

PR #4229, opened 2026-04-24
[Fix] Handle changing IPs of trusted peers

kaimast:
This PR aims to tackle (some of) #4228. But there are some open questions that need to be addressed.

It makes the following changes:
* `handle_trusted_validators` now properly waits for connection attempts to succeed/fail and prints any associated errors. We already do this for bootstrap peers, and most likely forgot to make the change for the other function.
* `CandidatePeers` now track the last known Aleo address. This way, we can detect when a peer changes IP addresses.
* The PR now also generated a warning if a node connects with an Aleo address that is associated with a trusted peer, but from a different IP

Discussion/Open Questions:
* Are there legitimate cases where multiple nodes appear in the network with the same Aleo address? If so, the solution in this PR might be too simplistic.
* Does this add a new potential attack vector? The handshake should prevent a third party from using someone's Aleo address, but we should still ensure that this new behavior cannot be exploited in some way.

### DIFF
--- a/node/bft/src/gateway.rs
+++ b/node/bft/src/gateway.rs
@@ -1051,7 +1051,27 @@ impl<N: Network> Gateway<N> {
 
     /// This function attempts to connect to any disconnected trusted validators.
     fn handle_trusted_validators(&self) {
-        let trusted_peers = self.trusted_peers();
+        // Collect trusted peer addresses that genuinely need a reconnection attempt.
+        // Skip peers whose last known Aleo address is already connected via a different IP,
+        // which prevents spurious reconnection attempts when a validator's IP changes.
+        let trusted_peers: Vec<_> = {
+            let pool = self.peer_pool().read();
+            pool.iter()
+                .filter_map(|(addr, peer)| {
+                    if !peer.is_trusted() {
+                        return None;
+                    }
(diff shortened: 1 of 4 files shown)


vicsn:
We can assume one IP address per validator.

node/bft/src/gateway.rs
```diff
@@ -1050,16 +1050,36 @@ impl<N: Network> Gateway<N> {
...
+                .filter_map(|(addr, peer)| {
+                    if !peer.is_trusted() {
+                        return None;
+                    }
+                    if let Peer::Candidate(c) = peer {
+                        if let Some(aleo_addr) = c.last_known_aleo_addr {
+                            if self.is_connected_address(aleo_addr) {
+                                return None;
+                            }
+                        }
```
ljedrz, line 1069: The more robust (and simpler) way to check whether there's already a peer connected with the given Aleo address is to use `Resolver::get_peer_ip_for_address` - the `Resolver` only ever holds the addresses of connected peers.

ljedrz (changes requested):
This seems like the exact kind of issue that would manifest in the way we've observed. My suggestion is to use a preexisting tool, but other than that, the approach is sound.

Issue #4140, opened 2026-03-04
[Proposal] Introduce additional node hardening measures


ljedrz:
Other than increasing the number of validators which will already happen organically, there are several additional measures we can introduce in order to harden the node further. Eventually this can give us confidence to disable firewalls.

- [ ] per-IP connection count limits 
- [ ] per-subnet connection count limits
- [x] more aggressive liveness checks, including R/W timeouts
- https://github.com/ProvableHQ/snarkOS/pull/4141
- [ ] asymmetrical handshakes (more lightweight for the responder, heavier for the initiator)
- [x] a new round of `snarkos-node-tcp` hardening tweaks
- https://github.com/ProvableHQ/snarkOS/pull/4141
- [x] raw TCP socket hardening
- https://github.com/ProvableHQ/snarkOS/pull/4143
- [ ] apply heuristics to disconnect from outlier peers: 
  - average traffic (too few/many msgs sent/receives compared to the average)
  - average handshake time (consistently slow)
  - number of connection attempts over a specific amount of time (suggesting choppy connection or DoS attempts)
- [ ] either dedicated block sync streams or message chunking for anything larger than 4MiB
- [ ] can connection attempts from IP addresses which previously succesfully connected be prioritized even during a DoS attempt?
- [ ] document the worst case DoS attack possible. Can we get to a state where, in the worst case, only connection attempts are impossible?

Issue #4224, opened 2026-04-22
[Bug] Sync speed regression


vicsn:
## 🐛 Bug Report

We recently observed a P2P sync speed regression: https://provablehq.github.io/snarkOS/dev/bench/

After merging: https://github.com/ProvableHQ/snarkOS/commit/2965cd6b8077cdd76cdcf8f8ed6e2730cef4e161#diff-77117c4c0dcf15f3243ecab8a8d03f035dd20930c6b1a4cc3398a5dcc66516ca

However, those changes primarily impacted BFT sync, so there's no good reason why client sync is impacted.

`node/sync/src/block_sync.rs` had +456-273 lines changed, maybe we can find the issue in there?

vicsn:
For both Kai and me, we can't reproduce the sync slowdown issue locally. It might only happen in the benchmark setting where the CPUs are much less powerful.

vicsn:
Closed by https://github.com/ProvableHQ/snarkOS/pull/4270

PR #4232, opened 2026-04-24
[Perf] Dedicated sync streams

ljedrz:
This PR is a draft implementation of dedicated sync streams listed in the [node hardening issue](https://github.com/ProvableHQ/snarkOS/issues/4140), as an alternative to the [general chunking proposal](https://github.com/ProvableHQ/snarkOS/pull/4055). Initially, it applies only to the validators, which will benefit from it the most.

The main reasons for the proposal are:
- increasing the performance of consensus (by reducing Head-of-Line blocking caused by large network messages)
- reducing DoS surface (by greatly reducing the maximum network message size)

Benefits over chunking:
- simpler non-sync network plumbing
- the potential to reduce the maximum network message size to a much greater degree

The general approach is as follows: the `Gateway`'s `Event` gains 2 new variants (in order to temporarily maintain backward compatibility via the existing `BlockRequest` and `BlockResponse`), `SyncRequest` and `SyncResponse`. When a node receives a `SyncRequest`, it responds with an address to a dedicated sync stream, and a short-lived access token that must be used in order to establish the connection. Once established, the responder sends `BlockResponse` messages (through the dedicated stream), and the existing `BlockSync` plumbing handles the rest. Once the maximum number of responses (if such a limit is desired) per a sync stream is sent, a new sync stream needs to be opened in order to receive more blocks.

The rough list of code changes, enumerated for simpler referencing if need be:
1. Some of the network messages are moved to `snarkos-node-network` to avoid circular dependencies (it will also make sense for the currently `Gateway`-only messages to reside there once this syncing is extended to non-validators).
2. New structs, `SyncToken` and `SyncResponse`, are introduced.
3. The `Event` is extended with 2 new variants, `SyncRequest` (holding a `BlockRequest`) and `SyncResponse` (which holds the new `SyncResponse` struct).
4. A `SyncStreams` object is introduced; it is essentially a node that requires no special peer handling or address resolution, and has a trivial handshake. Just like the `Gateway`, it contains a clone of sync-related `Sender`s, and the `LedgerService`. Using a node makes stream handling a lot simpler, and - compared to ad-hoc streams - reduces potential NAT/firewall issues (since only a single listener port is involved). The `Tcp` node is very lightweight, and so are its connections.
5. The `BlockSync` plumbing is adjusted to account for the new logic.

The current state of the PR: nodes can successfully establish dedicated sync streams and send/receive blocks, but I'm running into design details of the current `BlockSync` setup that are incompatible with the new approach; the problematic spots can be seen commented out in `block_sync.rs`. The issues I've identified thus far are:
- [ ] the sync requests are matched with addresses associated with `BlockLocators` (as opposed to the dedicated sync streams)
- [ ] the sync responses are currently distributed in a somewhat "fanout" fashion; instead, we should now also be prepared to send many `BlockResponse` (or even just `Block`) messages to single peers via a single stream; the expected block ranges should also be much larger, so as to minimize the number of `SyncRequest` messages that need to be handled by the `Gateway`

Once these are resolved, the related tests will also need some adjustment.

@kaimast please let me know if you have ideas on how the aforementioned `BlockSync` integration issues can be solved while maintaining backward compatibility, or suggest how this logic could be delegated elsewhere; feel free to commit to this PR if you'd like to integrate these changes with `BlockSync` in a way that's aligned with your design and future plans.

Open questions:
1. Do we want to "cycle" through several streams while syncing? It might be unnecessary, since at that point we're already past twofold authentication (the validator handshake + the access token). Since the sync streams are not used for anything else and are lightweight, I see little harm in it.
2. Should we limit the requestor to a single sync stream? This is a problem if we wanted to have high syncing performance while cycling streams (as we couldn't begin a new stream before concluding the existing one).
3. Do we want to limit the number of requested blocks? This needs to be weighed with the desired syncing redundancy factor and the performance implications it has for the providers. Note: dedicated sync streams are by design more robust than singular requests for blocks, so we may not need as much redundancy anymore.
4. The values for some of the `const`s.


node/network/src/sync_token.rs
```diff
@@ -0,0 +1,75 @@
...
+
+use rand::RngExt;
+use std::{
+    fmt,
+    io::{Read, Write},
+    ops::Deref,
+};
+
+#[derive(Clone, PartialEq, Eq, Hash)]
+pub struct SyncToken([u8; 32]);
```
vicsn, line 27: Can you document what this is and how it works? Are we exposed to worse MITM attacks compared to the Gateway handshake?

node/sync/src/node.rs
```diff
@@ -0,0 +1,378 @@
...
+impl<N: Network> Disconnect for SyncStreams<N> {
+    async fn handle_disconnect(&self, peer_addr: SocketAddr) {
+        self.active_streams.lock().remove(&peer_addr);
+    }
+}
+
+#[async_trait]
+impl<N: Network> OnConnect for SyncStreams<N> {
+    async fn on_connect(&self, peer_addr: SocketAddr) {
+        // Check if we're the ones who provide the sync.
```
vicsn, line 296: Can two peers sync from each other concurrently? Should we add a test for these edge cases?

vicsn (commented):
1./2. I would try and optimize for simplicity of the implementation. With the foundations in place, in the future we can more formally balance robustness and performance under various scenario's.

3./4. Can you take a first stab in a google doc for the limits we choose? We can try to allocate, say, a *rough* 10GB at any given time for all peers combined. Assuming current average blocksize if needed.

node/network/src/sync_token.rs
ljedrz: The token is sent in plaintext and could be observed or tampered with by an inline MITM, yes. However, it is strictly an anti-DoS mechanism to prevent unauthorized resource consumption on the validator, not a cryptographic session key. The integrity of the downloaded blocks is guaranteed by the hashes and signatures, not by the transport layer, so any MITM tampering would be instantly caught and rejected.

node/sync/src/node.rs
ljedrz: This behavior is controlled by the `BlockSync` logic, and remains unchanged. I don't see how such a scenario would be useful, so I'm pretty sure it is disallowed. As for the comment, this check is only required due to the modular nature of the `Tcp` plumbing - at the point of `OnConnect::on_connect`, we don't readily know the side of the connection (though this could be exposed by the `Tcp` if needed), so we can look up the applicable block request (which we would need anyway) in order to check it.

node/network/src/block_response.rs
```diff
@@ -128,12 +116,12 @@ impl<N: Network> DataBlocks<N> {
...
-        let candidate_end_height = 1 + self.last().map(|b| b.height()).unwrap_or(0);
+        // let candidate_start_height = self.first().map(|b| b.height()).unwrap_or(0);
+        // let candidate_end_height = 1 + self.last().map(|b| b.height()).unwrap_or(0);
         // Check that the range matches the block request.
-        if start_height != candidate_start_height || end_height != candidate_end_height {
-            bail!("Peer '{peer_ip}' sent an invalid block response (range does not match block request)")
-        }
+        // if start_height != candidate_start_height || end_height != candidate_end_height {
+        //     bail!("Peer '{peer_ip}' sent an invalid block response (range does not match block request)")
+        // }
```
ljedrz, line 124: if we are to reuse the existing `BlockRequest`, this check is no longer correct, as the request can span a much greater range of blocks than the responses

cli/src/commands/start.rs
```diff
@@ -155,6 +155,10 @@ pub struct Start {
     #[clap(long, requires = "validator")]
     pub bft: Option<SocketAddr>,
 
+    /// Set the IP address and port used for providing block synchronization.
```
kaimast, line 158: Should we add a comment here `Will be set to 0.0.0.0:6130 if not specified`?

cli/src/commands/start.rs
```diff
@@ -507,6 +511,13 @@ impl Start {
             self.node = Some(address);
         }
 
+        if self.sync_listener.is_none() {
+            let port = get_devnet_router_address_for_node(dev).port() + 2000; // 6130+
```
kaimast: Can we add a `get_devent_sync_address_for_node` method?

cli/src/commands/start.rs
```diff
@@ -846,7 +860,7 @@ impl Start {
 
         // Initialize the node.
         let node = match node_type {
-            NodeType::Validator => Node::new_validator(node_ip, self.bft, rest_ip, self.rest_rps, account, &trusted_peers, &trusted_validators, genesis, cdn, storage_mode, node_data_dir, self.trusted_peers_only, self.auto_db_checkpoints.clone(), dev_txs, self.dev, signal_handler.clone()).await,
+            NodeType::Validator => Node::new_validator(node_ip, sync_listener, self.bft, rest_ip, self.rest_rps, account, &trusted_peers, &trusted_validators, genesis, cdn, storage_mode, node_data_dir, self.trusted_peers_only, self.auto_db_checkpoints.clone(), dev_txs, self.dev, signal_handler.clone()).await,
```
kaimast, line 864: We should open an issue about improving how `Node` is created. The number of arguments is getting a little out of hand.
A builder pattern or a struct of arguments would be better at this point.

node/network/src/block_request.rs
```diff
@@ -38,20 +40,6 @@ impl BlockRequest {
     }
 }
 
-impl EventTrait for BlockRequest {
```
kaimast, line 41: Could we have it implement `Display` and then use the display trait downstream?

node/bft/events/src/lib.rs
```diff
@@ -90,6 +83,52 @@ pub trait EventTrait: ToBytes + FromBytes {
...
+        }
+        .into()
+    }
+}
+
+impl<N: Network> EventTrait for CertificateRequest<N> {
+    /// Returns the event name.
+    #[inline]
+    fn name(&self) -> Cow<'static, str> {
+        "CertificateRequest".into()
```
kaimast, line 120: Could this print the certificate ID as well? The round number would also help, but it is not included in the message (I think).

node/sync/src/helpers/sync_channel.rs
```diff
@@ -0,0 +1,122 @@
...
+
+use anyhow::{Result, anyhow};
+use std::net::SocketAddr;
+use tokio::sync::{mpsc, oneshot};
+
+const MAX_CHANNEL_SIZE: usize = 8192;
+
+#[derive(Clone, Debug)]
+pub struct SyncSender<N: Network> {
+    pub tx_block_sync_insert_block_response: mpsc::Sender<(
```
kaimast: Could some of these be private? There are helper functions for some of these that should be used instead of directly accessing them.

node/sync/src/node.rs
kaimast: I think it could happen if for some reason a block request gets delayed. However, it is very unlikely.

kaimast:
I left a bunch of nitpicky comments, but overall the PR looks great so far!

node/network/src/block_request.rs
ljedrz: totally, pretty sure it's an artifact of some old development plans that were ultimately abandoned

ljedrz: opened a dedicated issue: https://github.com/ProvableHQ/snarkOS/issues/4243

node/bft/events/src/lib.rs
ljedrz: [added](https://github.com/ProvableHQ/snarkOS/issues/4243#issuecomment-4350608258) to the dedicated `fmt::Display` issue

cli/src/commands/start.rs
ljedrz: created a dedicated issue: https://github.com/ProvableHQ/snarkOS/issues/4244

node/sync/src/helpers/sync_channel.rs
ljedrz: this is just a 1:1 move, but yes, it seems that 2 of these members can be made private right now

ljedrz:
@kaimast thanks for the comments so far; also highlighting the request from the description, as I wouldn't want to break the current setup due to its incompatibilities with the new design:
> please let me know if you have ideas on how the aforementioned `BlockSync` integration issues can be solved while maintaining backward compatibility, or suggest how this logic could be delegated elsewhere; feel free to commit to this PR if you'd like to integrate these changes with `BlockSync` in a way that's aligned with your design and future plans.

ljedrz:
Rebased to fix a conflict.

node/network/src/block_response.rs
kaimast: Wasn't the plan to change block requests to individual blocks?

ljedrz: yes, ultimately I'd want to do so, but I wanted to minimize the diff, at least for now; this would also be easier once the backward compatibility is no longer needed

kaimast:
Does this branch already work? It would be useful to see whether the performance regression noted in #4224 is also present here. 

I am not sure if you have permission to trigger the benchmark workflow. If you do, please trigger when the initial implementation is done.

ljedrz:
No, as I'm not sure how to plug these changes into `BlockSync` while maintaining backward compatibility - I was hoping you could suggest how to work around the locator address matching and range limits.

vicsn:
@ljedrz can you update the precise benefit of this approach in the README above after https://github.com/ProvableHQ/snarkOS/pull/4419 is merged

cbeck88:
I think the main thing is 4419 doesn't help the gateway, it only helps the router.

So it doesn't help for threat model of a validator node that goes Byzantine and tries to DOS the other nodes.

PR #4234, opened 2026-04-27
[Fix] Make the validators dump their block tree correctly

### LINKED ISSUE #4227: [Bug] The stored height is different than the one in the block tree
## 🐛 Bug Report

Over the past weeks, when restarting and upgrading **validators**, the following issue occurs:

```
2026-04-22T19:54:51.131252Z DEBUG snarkvm_synthesizer::vm: Loading deployments 11008-11010 (of 11010)...
2026-04-22T19:54:55.086523Z ERROR snarkos: Failed to start node — Failed to initialize the ledger — The stored height is different than the one in the block tree; please ensure that the cached block tree is valid or delete the 'block_tree' file from the ledger folder
2026-04-22T19:54:55.086555Z DEBUG snarkos: Stopping process with exitcode 1
```

Systemd service:
```
[Unit]
Description=snarkOS Node Service
After=network.target
Wants=network.target

StartLimitIntervalSec=500
StartLimitBurst=5

[Service]
Type=simple
User=root
Group=root
LimitMEMLOCK=infinity
Environment=RUST_BACKTRACE=1
Environment=RUST_LOG=trace
WorkingDirectory=/mnt/snarkos
ExecStart=/usr/bin/snarkos start --client \
  --network 1 \
  --nodisplay \
  --node 0.0.0.0:4130 \
  --rest 0.0.0.0:3030 \
  --verbosity 4 \
  --logfile /mnt/snarkos/log/snarkos.log \
  --ledger-storage /mnt/snarkos/ledger \
  --node-data-storage /mnt/snarkos/node-data \
  --auto-migrate-node-data \
  --auto-db-checkpoints=/mnt/snarkos/auto-db-checkpoints \
  --nocdn  \
  --metrics

ExecStop=/bin/kill -s INT ${MAINPID}
RestartSec=30
Restart=always
LimitNOFILE=65535
# StandardOutput=null

[Install]
WantedBy=multi-user.target
```

For clients the issue did not occur.

If finding the cause of the issue is not straightforward, I'd suggest we add more logging and do a few more restarts.

ljedrz:
The `Drop` impl of the `Ledger` is what causes the block tree to be cached on shutdown. As long as there are any "zombie tasks" which hold a reference to the ledger post-shutdown, that impl won't fire, forcing the node to fully rebuild the block tree on startup (or to report an invalid cached block tree if it's old, as it was reported in the linked issue).

This was debugged using `tokio-console`. I've also considered downgrading some of the `Ledger`-related `Arc`s to `Weak`, but sadly, that would result in a `Resultpocalypse`. That being said, it is perfectly fine to use `Arc` everywhere (like we do now), as long as we make sure that the persistent tasks that own them are terminated during shutdown.

Fixes https://github.com/ProvableHQ/snarkOS/issues/4227.

### DIFF
--- a/node/rest/src/lib.rs
+++ b/node/rest/src/lib.rs
@@ -131,6 +131,11 @@ impl<N: Network, C: ConsensusStorage<N>, R: Routing<N>> Rest<N, C, R> {
     pub const fn handles(&self) -> &Arc<Mutex<Vec<JoinHandle<()>>>> {
         &self.handles
     }
+
+    /// Shuts down the REST instance.
+    pub fn shut_down(&self) {
+        self.handles.lock().iter().for_each(|handle| handle.abort());
+    }
 }
 
 impl<N: Network, C: ConsensusStorage<N>, R: Routing<N>> Rest<N, C, R> {
--- a/node/sync/src/ping.rs
+++ b/node/sync/src/ping.rs
@@ -135,6 +135,10 @@ impl<N: Network> Ping<N> {
         let mut new_block = false;
 
         loop {
+            if router.ledger().is_stopped() {
+                break;
+            }
+
             // Do not hold the lock while waiting.
             let sleep_time = {
                 let mut inner = inner.lock();
--- a/node/sync/src/ping.rs
+++ b/node/sync/src/ping.rs
@@ -130,6 +130,12 @@ impl<N: Network> Ping<N> {
         self.notify.notify_one();
     }
 
+    /// Wake up the ping task so that its inner loop iterates and breaks.
+    /// It should only be called after the ledger has already been stopped.
+    pub fn stop(&self) {
+        self.notify.notify_one();
+    }
+
     /// Background task that periodically sends out new ping messages.
     async fn ping_task(inner: &Mutex<PingInner<N>>, router: &Router<N>, notify: &Notify) {
         let mut new_block = false;
--- a/node/sync/src/ping.rs
+++ b/node/sync/src/ping.rs
@@ -73,11 +73,11 @@ impl<N: Network> Ping<N> {
 
         {
             let inner = inner.clone();
-            let router = router.clone();
+            let router_ = router.clone();
             let notify = notify.clone();
 
-            tokio::spawn(async move {
-                Self::ping_task(&inner, &router, &notify).await;
+            router.spawn(async move {
+                Self::ping_task(&inner, &router_, &notify).await;
             });
         }
 
@@ -92,11 +92,11 @@ impl<N: Network> Ping<N> {
 
         {
             let inner = inner.clone();
-            let router = router.clone();
+            let router_ = router.clone();
             let notify = notify.clone();
 
-            tokio::spawn(async move {
-                Self::ping_task(&inner, &router, &notify).await;
+            router.spawn(async move {
+                Self::ping_task(&inner, &router_, &notify).await;
             });
         }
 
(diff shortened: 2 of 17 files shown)


node/sync/src/ping.rs
```diff
@@ -135,6 +135,10 @@ impl<N: Network> Ping<N> {
         let mut new_block = false;
 
         loop {
+            if router.ledger().is_stopped() {
+                break;
```
vicsn, line 139: Was this the source of the issue? Is there any ways to prevent us from creating new zombie tasks in the future? If not in the type-system, then perhaps we can check in our `.ci` scripts if on any shutdown the `block_tree` file has a recent timestamp...

ljedrz: At the end of the shutdown of the `Validator` object, I checked which tasks are still active, and this was one of them (the others were the ones related to the BFT sync, hence the additional call to shutdown in the `Primary`).

As for where this can be checked in the CI... yes, we could either check for the existence of the `block_tree` file - that would likely be the simplest solution.

vicsn: Another more general idea, is it possible/desirable to run tokio console on our `.ci` to spot issues?

ljedrz: not until it's a lot more stable; however, I see that the `Runtime` shutdown methods are now stable, I can look into just shutting it down on node shutdown (which would abort all the attached tasks)

vicsn: Sure if you have a more robust method that would be great. Otherwise let's add `block_tree` existence checks in `.ci` tests

ljedrz: Done; if there are **any** zombie tasks on shutdown, an `ERROR` log is emitted, and IIRC, this would break the `devnet.sh` script, providing us with an automatic CI check.

ljedrz:
I've updated the PR - now no zombie tasks are expected to exist at shutdown, and if any are found, an `ERROR` log is emitted.

vicsn (dismissed):
Awesome!

node/rest/src/lib.rs
```diff
@@ -131,6 +131,11 @@ impl<N: Network, C: ConsensusStorage<N>, R: Routing<N>> Rest<N, C, R> {
     pub const fn handles(&self) -> &Arc<Mutex<Vec<JoinHandle<()>>>> {
         &self.handles
     }
+
+    /// Shuts down the REST instance.
+    pub fn shut_down(&self) {
+        self.handles.lock().iter().for_each(|handle| handle.abort());
```
kaimast, line 137: It doesn't matter as much but using `drain(..)` would also drop the handles when calling shut down.

node/sync/src/ping.rs
```diff
@@ -130,11 +130,21 @@ impl<N: Network> Ping<N> {
         self.notify.notify_one();
     }
 
+    /// Wake up the ping task so that its inner loop iterates and breaks.
+    /// It should only be called after the ledger has already been stopped.
+    pub fn stop(&self) {
+        self.notify.notify_one();
```
kaimast: Can't we just abort the ping task here?

node/rest/src/lib.rs
ljedrz: ah yeah, probably a bit cleaner, but effectively it doesn't matter, as they all get dropped at the end of this scope

node/sync/src/ping.rs
ljedrz: good idea, I realized this task can be attached to the `Router`: https://github.com/ProvableHQ/snarkOS/pull/4234/commits/62caba2301dd83f73f04b63b1382ee5113195fa1

ljedrz:
@vicsn addressed Kai's comments

kaimast (approved):

PR #4235, opened 2026-04-28
Introduce hotswapping prod committee for dev committee

vicsn:
## Motivation

E2E test would be more robust if we manage to fully replicate our production state and environment.

One blocker to this end is that snarkOS currently tightly couples ledger state and consensus. More concretely: a dev committee is not authorized to produce new blocks on top of production ledger state. This PR changes this by supporting hotswapping of a ledger without too many code changes through the use of a `DEV_COMMITTEE_NUM_VALIDATORS` environment variable and a `--dev-on-prod` flag. The relevant codepaths are compile-guarded by the `test_network` feature. I didn't introduce a new compilation feature to keep our testing environment simple.

The new settings enable the use of a production genesis block and disable checks on certificates. In order to avoid mutilating snarkVM, the entire call to `check_next_block` is skipped.

## Test Plan

- [x] New CI script passes
- [x] Prerelease tests passed

### DIFF
--- a/.ci/test_dev_on_prod_restart.sh
+++ b/.ci/test_dev_on_prod_restart.sh
@@ -180,6 +180,12 @@ function copy_setup_ledger() {
   cp -r "$source" ledger-1
   cp -r "$source" ledger-2
   cp -r "$source" ledger-3
+
+  # Clear any state cached by a previous test run (proposal cache, dev committee state, etc.).
+  # The ledger we just copied is fresh, so any persisted dev state must be regenerated from it.
+  for i in $(seq 0 $((NUM_DEV_NODES - 1))); do
+    snarkos clean "--dev=${i}" "--network=0"
+  done
 }
 
 function start_dev_nodes() {
@@ -188,10 +194,10 @@ function start_dev_nodes() {
 
   for i in $(seq 0 $((NUM_DEV_NODES - 1))); do
     log "Starting dev node ${i}"
-    DEV_COMMITTEE_NUM_VALIDATORS=${NUM_DEV_NODES} snarkos start --nodisplay --validator --ledger-storage "ledger-${i}" --node-data-storage "node-data-${i}" --dev "${i}" \
+    snarkos start --nodisplay --validator --ledger-storage "ledger-${i}" --node-data-storage "node-data-${i}" --dev "${i}" \
       --no-dev-txs --nocdn --dev-num-validators "${NUM_DEV_NODES}" --verbosity 2 \
       --allow-external-peers --logfile "dev_logs/val-${i}.txt" --dev-on-prod &
-    DEV_PIDS[$i]=$!
+    DEV_PIDS[i]=$!
     sleep 1
   done
 }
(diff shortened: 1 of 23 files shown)


node/bft/ledger-service/src/ledger.rs
```diff
@@ -68,6 +68,10 @@ pub struct CoreLedgerService<N: Network, C: ConsensusStorage<N>> {
     latest_leader: Arc<RwLock<Option<(u64, Address<N>)>>>,
     stoppable: Arc<dyn Stoppable>,
     update_lock: Arc<Mutex<()>>,
+    #[cfg(feature = "test_network")]
+    dev_start_round: Option<u64>,
+    #[cfg(feature = "test_network")]
+    dev_num_validators: Option<String>,
```
ljedrz: why not immediately expect a `u16`?

node/bft/ledger-service/src/ledger.rs
```diff
@@ -136,10 +140,46 @@ impl<N: Network, C: ConsensusStorage<N>> CoreLedgerService<N, C> {
...
+            #[cfg(feature = "test_network")]
+            dev_start_round,
+            #[cfg(feature = "test_network")]
+            dev_num_validators,
+        }
+    }
+
+    /// Returns the deterministic dev committee for rounds at or after the hotswap start.
+    #[cfg(feature = "test_network")]
+    fn dev_committee_for_round(&self, round: u64) -> Result<Option<Committee<N>>> {
```
ljedrz: shouldn't this be created just once? `current_committee` is called routinely

node/bft/ledger-service/src/ledger.rs
```diff
@@ -247,6 +296,13 @@ impl<N: Network, C: ConsensusStorage<N>> LedgerService<N> for CoreLedgerService<
 
     /// Returns the committee lookback for the given round.
     fn get_committee_lookback_for_round(&self, round: u64) -> Result<Committee<N>> {
+        #[cfg(feature = "test_network")]
+        {
+            if let Some(dev_committee) = self.dev_committee_for_round(round)? {
```
ljedrz, line 386: shouldn't this only be triggered if `round >= dev_start_round`?

ljedrz (commented):
Left some initial comments.

node/bft/ledger-service/src/ledger.rs
vicsn: `dev_committee_for_round` checks that condition internally.

node/bft/ledger-service/src/ledger.rs
vicsn: 55cbb7f6e

node/bft/ledger-service/src/ledger.rs
vicsn: Yeah you're right. I didn't see a way of overriding `current_committee` at the right time without touching snarkVM. Added a new cached version here: 7ec762d8e

.ci/test_dev_on_prod_restart.sh
```diff
@@ -0,0 +1,239 @@
...
+SETUP_ADVANCE_BLOCKS=5
+DEV_ADVANCE_BLOCKS=10
+SETUP_MAX_WAIT=40
+DEV_MAX_WAIT=100
+
+BOOTSTRAP_PID=""
+declare -a DEV_PIDS=()
+SNARKOS_SETUP_BIN="snarkos"
+
+function get_height() {
```
kaimast: Can some of these be in `.ci/utils.sh`? There's already a `get_block_height` for example.

kaimast:
Looks good! It needs a rebase and formatting fixes, though.

node/bft/src/sync/mod.rs
kaimast, line 423: It might be good to add a note here.

```
// The block was already verified when it was added to the ledger, so we do not have to re-check its certificates here.
```

kaimast:
Can we eventually rename the `test_network` feature to `devnet`? It is a little confusing because we refer to it as devnet in the code and the `test_network` feature is unrelated to the test network.

.ci/test_dev_on_prod_restart.sh
vicsn: 608f506cf

node/bft/src/sync/mod.rs
vicsn: 599598a1c

.ci/test_dev_on_prod_restart.sh
vicsn: And made an issue for the future: https://github.com/ProvableHQ/snarkOS/issues/4262

kaimast (approved):

PR #4237, opened 2026-04-29
[Perf] REST enhancements

ljedrz:
It's been reported that nodes exposing their REST server were experiencing slowdowns. Based on the logs, the most likely culprit was the `/block/N` route. Upon closer inspection, it appears that we currently aren't using blocking tasks to perform the associated queries (which may be slow), and that we don't do any sort of caching, despite many of the requests being for the genesis block, the latest block, or a repeated height.

This PR introduces the following isolated changes:
1. b65d953680033ee799758dbc83983d55c3736b2f - time the requests (for all routes), and adjust the related logs; this makes the node produce 2 logs per request, as opposed to 1:
```
// old format
INFO snarkos_node_rest: Received 'GET /block/N' from '<IP:PORT>'
// new format
<TS> INFO REST{method=GET uri=/block/N addr=<IP:PORT>}: snarkos_node_rest: Received a request
<TS> INFO REST{method=GET uri=/block/N addr=<IP:PORT>}: snarkos_node_rest: Finished request in <DURATION>
```
2. 91d5f836a101c52f37a9e7977968733d16923c28 - introduce a block cache, and use it in `/block/latest` and `/block/{height or hash}`; the cache is keyed by the block hash (as all block queries ultimately use the hash), and the values are JSON (as block serialization can be costly)
3. 5fc1a9d13282dc67b6bee731a89daa9322a9d534 - reorders the REST-applicable layers so that they are applied in the desired order (the outermost/last ones are the first ones to be applied)

The most important change is the block cache, which both reduces the response time by an order of magnitude (milliseconds to microseconds), and reduces the strain on the async executor.

### DIFF
--- a/node/rest/Cargo.toml
+++ b/node/rest/Cargo.toml
@@ -61,6 +61,9 @@ workspace = true
 features = [ "parking_lot" ]
 optional = true
 
+[dependencies.lru]
+workspace = true
+
 [dependencies.once_cell]
 workspace = true
 
--- a/node/rest/src/lib.rs
+++ b/node/rest/src/lib.rs
@@ -51,9 +51,10 @@ use axum::{
 use axum_extra::response::ErasedJson;
 #[cfg(feature = "locktick")]
 use locktick::parking_lot::Mutex;
+use lru::LruCache;
 #[cfg(not(feature = "locktick"))]
 use parking_lot::Mutex;
-use std::{net::SocketAddr, sync::Arc, time::Duration};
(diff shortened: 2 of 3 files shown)


vicsn (approved):

kaimast:
> the latest block,

The latest block is cached by `snarkvm-ledger`, as well as the most recent blocks. 

Edit: This change is still useful as we cache the serialized blocks.

PR #4242, opened 2026-04-30
Fetch previous certificates concurrently

kaimast:


ljedrz (approved):
LGTM, just one question: should we cancel the other attempts if any of them fails? Because that's what `try_join_all` will do, as opposed to something more lenient, like `join_all`.

vicsn:
> LGTM, just one question: should we cancel the other attempts if any of them fails? Because that's what try_join_all will do, as opposed to something more lenient, like join_all.

Good question. I wonder if we risk having inconsistent state due to the potential early cancellation? For example, `sync_with_certificate_from_peer` could stop between `spawn_blocking!(storage.insert_certificate(...))` and `cb.add_new_certificate(...)`. So then the certificate is in storage but not in the DAG.

PR #4258, opened 2026-05-08
Add --onlycdn flag to only sync from CDN

vicsn:
## Motivation

When P2P sync experiences issues, this allows a node to purely use the CDN.

## Test Plan

- [ ] We should test this as tip on a synced testnet client

### DIFF
--- a/cli/src/commands/start.rs
+++ b/cli/src/commands/start.rs
@@ -265,6 +265,11 @@ pub struct Start {
     #[clap(long)]
     pub nocdn: bool,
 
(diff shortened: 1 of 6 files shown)


kaimast:
This still uses p2p/bft sync after the initial CDN sync is successful, right? Maybe something like `--force-cdn` or `--retry-cdn-sync` would explain the functionality better?

vicsn:
The goal was to only use CDN, but converting to draft because I'm not sure we'll actually want to promote sole reliance on the CDN

PR #4261, opened 2026-05-11
[Fix] Stop proposing when a node starts syncing again

kaimast:
The recent batch proposal improvements added a subtle bug where a node may repropose old batches after it starts syncing. This adds additional `is_syncing` checks, and tests to ensure this behavior works as expected.

Most of this PR is tests for the behavior. The main change in functionality are two new `if !primary.is_synced()` checks.


vicsn (approved):
LGTM, did this come up as a halting node during prerelease testing?

PR #4266, opened 2026-05-14
[Fix] Ensure `--dev-on-prod works` as expected

kaimast:
Companion PR for https://github.com/ProvableHQ/snarkVM/pull/3259


.circleci/config.yml
vicsn: Can we make this PR independent from the renaming of `test_network` to `devnet`? It is backwards incompatible and will likely eat up a lot of time.

cli/src/commands/start.rs
```diff
@@ -298,15 +298,17 @@ pub struct Start {
     #[clap(long, group = "dev_flags", requires = "dev_num_validators", default_value_t = false)]
     pub dev_on_prod: bool,
 
+    /// Override the round at which the dev committee becomes active.
+    ///
+    /// Only meaningful with `--dev-on-prod`. When unset, the round is derived from
+    /// `ledger.latest_round()` at first start and then persisted to disk; with this
+    /// flag set, every node agrees on the same round regardless of startup timing.
+    #[clap(long, group = "dev_flags", requires = "dev_on_prod")]
+    pub dev_hotswap_start_round: Option<u64>,
```
vicsn: Is there any way that we can run this without manually indicating an explicit hotswap start round - which is hard to infer from just a ledger file? Or do you want to read this out from the start_round which we write to disk...?

vicsn (changes requested):

.circleci/config.yml
kaimast: Done

vicsn (dismissed):
LGTM!

kaimast:
I don't have permission to merge anymore, but it should be good to go.

vicsn (dismissed):

cli/src/helpers/args.rs
```diff
@@ -42,8 +42,7 @@ pub(crate) fn prepare_endpoint(endpoint: Uri) -> Result<Uri> {
     }
 
     if parts.path_and_query.is_none() {
-        // An empty path is fine for the base URL.
-        parts.path_and_query = Some(uri::PathAndQuery::from_static(""));
+        parts.path_and_query = Some(uri::PathAndQuery::from_static("/"));
```
ljedrz, line 45: is this related?

vicsn: Nope: https://github.com/ProvableHQ/snarkOS/issues/4275

ljedrz (approved):

PR #4270, opened 2026-05-15
[Fix] Avoid banning peers that are still responding to block requests

kaimast:
### Motivation
Block sync bans peers that do not respond to a block request within a certain time. However, they may experience a backlog and still respond to other requests that we sent.  This issue became more noticeable after merging https://github.com/ProvableHQ/snarkOS/pull/4039 which lowered the request timeout from 10 minutes to 1 minute.

### Proposed Changes
With this change, `BlockSync` tracks the last time a peer responded to *any* block requests. If it does not respond to a particular block request within the timeout, we now only ban the peer if it has not responded to any other request either within the timeout.

The PR also removes a log message that is not informational and clutters the logs with many entries in the form of `TRACE snarkos_node_sync::block_sync: Failed to add block request, as block {n} exists in the requests map`

### Results

This change appears to fully resolve the sync speed degradation we saw after merging #4039.

### DIFF
--- a/node/rest/src/routes.rs
+++ b/node/rest/src/routes.rs
@@ -1269,7 +1269,7 @@ impl<N: Network, C: ConsensusStorage<N>, R: Routing<N>> Rest<N, C, R> {
         .map_err(|e| {
             let msg = e.to_string();
             if let Some(stripped) = msg.strip_prefix("404: ") {
-                RestError::not_found(anyhow!("{}", stripped))
+                RestError::not_found(anyhow!("{stripped}"))
             } else {
                 RestError::internal_server_error(e)
             }
--- a/node/sync/src/block_sync.rs
+++ b/node/sync/src/block_sync.rs
@@ -235,6 +235,12 @@ pub struct BlockSync<N: Network> {
     /// Tracks failed requests that need to be re-issued.
     failed_requests: Mutex<FailedRequests<N::BlockHash>>,
 
+    /// Tracks the last time each peer delivered a successful block response.
+    ///
+    /// Used in `handle_block_request_timeouts` to avoid banning peers that are actively
+    /// responding but cannot keep up with the request rate.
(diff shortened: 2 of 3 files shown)


node/rest/src/routes.rs
```diff
@@ -1269,7 +1269,7 @@ impl<N: Network, C: ConsensusStorage<N>, R: Routing<N>> Rest<N, C, R> {
         .map_err(|e| {
             let msg = e.to_string();
             if let Some(stripped) = msg.strip_prefix("404: ") {
-                RestError::not_found(anyhow!("{}", stripped))
```
kaimast, line 1272: Had to fix this clippy warning for CI to pass

node/sync/src/block_sync.rs
```diff
@@ -1419,9 +1445,11 @@ impl<N: Network> BlockSync<N> {
                 let timer_elapsed = now.duration_since(e.timestamp) > BLOCK_REQUEST_TIMEOUT;
                 // Determine if the request is complete.
                 let is_complete = e.sync_ips().is_empty() && e.response.is_some();
+                // If any assigned peer is still actively responding, the request is not stuck.
+                let has_responsive_peer = e.sync_ips().iter().any(|ip| responsive_peers.contains(ip));
 
                 // Determine if the request has timed out.
-                let is_timeout = timer_elapsed && !is_complete;
+                let is_timeout = timer_elapsed && !is_complete && !has_responsive_peer;
```
vicsn, line 1452: I guess a malicious peer which withholds a block reponse could use this to delay being banned for a very long time (in the worst case for around `max_total_requests * BLOCK_REQUEST_TIMEOUT` seconds). But for dealing with "honest but lazy" peers - which is the only scenario we want to protect against at this point in time, this seems sufficient.

For dealing with malicious peers, in the future we could track relative response performance to dynamically target the best responders.

vicsn (approved):

PR #4284, opened 2026-05-29
fix(node/bft): suppress benign message about certificate already exis…

### LINKED ISSUE #4283: [Bug] Benign log is reported as ERROR
## 🐛 Bug Report

On v4.7.0 60.000 ERRORs are printed per hour of the type:
```
Failed to fetch previous certificate — Certificate with this author for round 38161972 already exists in storage
Failed to fetch previous certificate — Certificate for round 38161972 already exists in storage (gc = 38161866)
```

kaimast:
Fixes #4283 by introducing a `CheckCertificateError` type.

### DIFF
--- a/node/bft/src/helpers/storage.rs
+++ b/node/bft/src/helpers/storage.rs
@@ -43,6 +43,33 @@ use std::{
     },
 };
 
+/// Errors returned by [`Storage::check_certificate`] (and therefore [`Storage::insert_certificate`]).
+///
+/// The `SameCertificate` and `SameAuthorAndRound` variants describe benign races: concurrent sync
+/// paths regularly try to insert the same certificate, and the loser of that race should treat its
+/// failure as a no-op (the certificate is in fact present in storage) rather than a hard error.
+#[derive(Debug, thiserror::Error)]
+pub enum CheckCertificateError {
+    #[error("Certificate round {round} already exists in storage (gc_round = {gc_round})")]
+    SameCertificate { round: u64, gc_round: u64 },
+    #[error("Certificate with this author in round {round} is already in storage (gc_round = {gc_round})")]
+    SameAuthorAndRound { round: u64, gc_round: u64 },
+    #[error("Certificate round {round} is at or below the GC round {gc_round}")]
+    RoundTooLow { round: u64, gc_round: u64 },
+    #[error(transparent)]
+    Other(#[from] anyhow::Error),
+}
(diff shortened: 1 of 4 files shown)


node/bft/src/primary.rs
```diff
@@ -1860,18 +1863,30 @@ impl<N: Network> Primary<N> {
             self.sync_with_batch_header_from_peer::<IS_SYNCING, false>(peer_ip, batch_header).await?;
 
         // Check if the certificate needs to be stored.
-        if !self.storage.contains_certificate(certificate.id()) {
```
kaimast, line 1863: This is now caught by `insert_certficate` and handled appropriately below.

node/bft/src/primary.rs
```diff
@@ -1860,18 +1863,30 @@ impl<N: Network> Primary<N> {
...
+        // Store the batch certificate.
+        let (storage, certificate_) = (self.storage.clone(), certificate.clone());
+        match tokio::task::spawn_blocking(move || {
+            storage.insert_certificate(certificate_, missing_transmissions, Default::default())
+        })
+        .await
+        {
+            Ok(Ok(_)) => {} // continue
+            Ok(Err(err)) if err.is_benign() => {
+                trace!("Skipping insertion for certificate - {err}");
```
kaimast, line 1875: Not sure if this should even be logged, but it can always be removed in the future

node/bft/src/helpers/storage.rs
```diff
@@ -618,27 +650,27 @@ impl<N: Network> Storage<N> {
 
         // Ensure the certificate ID does not already exist in storage.
         if self.contains_certificate(certificate.id()) {
-            bail!("Certificate for round {round} already exists in storage {gc_log}")
+            return Err(CheckCertificateError::SameCertificate { round, gc_round });
         }
 
         // Ensure the storage does not already contain a certificate for this author in this round.
         if self.contains_certificate_in_round_from(round, certificate.author()) {
```
kaimast, line 657: These subsequent checks happen without any locking. Otherwise, this second check would only trigger if there are two certificates with same author+round with different IDs. 

It would make sense to merge the two checks to at least do this part atomically, and have a dedicated error for the more severe case where two certificates with different IDs exist.

vicsn:
Superseded by https://github.com/ProvableHQ/snarkOS/pull/4479

PR #4293, opened 2026-06-02
Start consensus handlers only after CDN sync is complete

vicsn:
## Motivation

CDN sync fails when consensus starts up too quickly: `Failed to sync block(s) from the CDN — Failed to advance to block ab1ewv9fjmlgxr6l3fhv0tdqsn6xh3yjg285yghvncmnqv4fjmh9g9sdfngdz at height 18954553 — The given block is not the direct successor of the latest block`

We should only start consensus handlers after CDN sync is complete.

The base branch is in development. After approval we can consolidate it into mainnet.

## Test Plan

- [x] Manually try syncing from CDN with a validator and observe the logs
- [x] CI

### DIFF
--- a/node/consensus/src/lib.rs
+++ b/node/consensus/src/lib.rs
@@ -31,6 +31,8 @@ use snarkos_node_bft::{
     Primary,
     helpers::{
         ConsensusReceiver,
+        ConsensusSender,
+        PrimaryReceiver,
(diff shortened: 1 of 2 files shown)


ljedrz (approved):
LGTM!

raychu86 (approved):

PR #4298, opened 2026-06-03
Postrelease merge mainnet

vicsn:
## Motivation

^


ljedrz (dismissed):

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "snarkos"
-version = "4.7.1"
+version = "4.7.4"
```
ljedrz, line 3: just to double-check: this is `.4`, despite the members all being `.3`?

ljedrz (approved):
LGTM provided that the discrepancy I commented on is intentional.

Cargo.toml
vicsn: Yeah it's fine we didn't want to do another crates.io release but still bump the snarkOS reported /version

PR #4308, opened 2026-06-09
[Feat] Stricter fd limit checks

### LINKED ISSUE #4306: [Proposal] Improve reporting of open file limit issues
## 💥 Proposal

The past month 3 node operators experienced opaque halts/sync issues reporting as previous state root mismatch, whereas the real issue was a lack of open file descriptors.

Currently we only issue a warning at startup about a draft limit. We should:
- turn the startup check into a panic
- periodically see if we can open a new file and report an ERROR if it fails.

ljedrz:
Startup enforcement is trivial, just a switch from an existing log to a panic.

As for the fs check, this PR proposes a more advanced solution than the one outlined in the issue, because attempting to open a file:
- would only flip at 100% capacity (no room for recovery)
- consumes an fd to run the check, potentially being the one that breaks the node
- is racy: the probe might succeed, while a microsecond later, the node might fail

I tested it briefly locally, checking both the per-process (by adding connections) and per-system (by starting external sockets) value.

The `nix` dependency (needed for non-Linux UNIXes) already exists in our tree; `rlimit` is new, but it's small, and we're already using its only dependency (`libc`). Using `libc` directly would break the `deny(unsafe)` rule.

Closes https://github.com/ProvableHQ/snarkOS/issues/4306.

### DIFF
--- a/cli/src/helpers/fd_check.rs
+++ b/cli/src/helpers/fd_check.rs
@@ -110,6 +110,8 @@ pub fn system_fd_usage() -> std::io::Result<SystemFd> {
     let (cur_oid, max_oid) = ("kern.num_files", "kern.maxfiles");
     #[cfg(any(target_os = "openbsd", target_os = "netbsd"))]
     let (cur_oid, max_oid) = ("kern.nfiles", "kern.maxfiles");
+    #[cfg(not(any(target_os = "freebsd", target_os = "macos", target_os = "openbsd", target_os = "netbsd")))]
+    return Err(std::io::Error::new(std::io::ErrorKind::Unsupported, "system fd probe unsupported on this OS"));
 
     fn read(oid: &str) -> std::io::Result<u64> {
         let out = std::process::Command::new("sysctl").arg("-n").arg(oid).output()?;
--- a/cli/src/helpers/fd_check.rs
+++ b/cli/src/helpers/fd_check.rs
@@ -166,7 +166,7 @@ pub fn spawn_fd_monitor() {
                 Err(e) => error!(error = %e, "process fd probe failed"),
             }
 
-            // (2) whole-machine fds
+            // (2) whole-machine fds are allowed 5 percentage points more leeway.
             match system_fd_usage() {
                 Ok(s) => {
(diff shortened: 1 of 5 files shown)


vicsn (commented):
Compiling on macOS:

```
error[E0277]: the trait bound `i32: AsFd` is not satisfied
   --> cli/src/helpers/fd_check.rs:75:41
    |
75  |     let n = (0..max).filter(|&fd| fcntl(fd, FcntlArg::F_GETFD).is_ok()).count();
    |                                   ----- ^^ the trait `AsFd` is not implemented for `i32`
    |                                   |
    |                                   required by a bound introduced by this call
    |
    = help: the following other types implement trait `AsFd`:
              &T
              &mut T
              Arc<T>
              AsyncFd<T>
              BorrowedFd<'_>
              Box<T>
              ChildStderr
              ChildStdin
            and 47 others
note: required by a bound in `nix::fcntl::fcntl`
   --> /Users/victorsintnicolaas/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/nix-0.30.1/src/fcntl.rs:862:18
    |
862 | pub fn fcntl<Fd: std::os::fd::AsFd>(fd: Fd, arg: FcntlArg) -> Result<c_int> {
    |                  ^^^^^^^^^^^^^^^^^ required by this bound in `fcntl`
```

cli/src/helpers/fd_check.rs
```diff
@@ -0,0 +1,195 @@
...
+                                pct = format!("{pct:.1}%"),
+                                "node fd usage elevated"
+                            );
+                        }
+                    }
+                }
+                Err(e) => error!(error = %e, "process fd probe failed"),
+            }
+
+            // (2) whole-machine fds
```
vicsn:
```suggestion
            // (2) whole-machine fds are allowed 5 percentage points more leeway.
```

ljedrz:
@copilot are you able to resolve the aforementioned macOS build error in `count_open_fds`?

vicsn (approved):
No ERROR on my machine or in CI

PR #4316, opened 2026-06-11
[Tests] Relax the cert count check under test_network

### LINKED ISSUE #4274: [Bug] dev-on-prod-restart failed
## 🐛 Bug Report

Note that this was on commit 4b156672d before https://github.com/ProvableHQ/snarkOS/pull/4266 was merged, just making sure to document this.

https://app.circleci.com/pipelines/gh/ProvableHQ/snarkOS/3880/workflows/052b1ebc-76be-4ff4-aa26-d88447953274/jobs/80612

```
[ci-runner] 2026-05-26T05:53:33Z dev-network-first-run: timed out waiting for height advance (final height: 7, target: 17)
[ci-runner] 2026-05-26T05:53:33Z Error at line 175 while running: return 1
```

Seeing a repeated:
```
2026-05-26T05:53:31.874638Z  WARN snarkos_node_bft::primary: Cannot sign a batch at round 25 from '127.0.0.1:5003' — Too many previous certificates for round 25 (gc = 0)
```

ljedrz:
The check for the number of certificates vs. the number of committee members is too strict for `dev_on_prod`-style tests, as the `test_network` feature makes `get_committee_lookback_for_round` always return the dev committee, which is typically smaller than the prod one, easily failing the aforementioned check.

The fix is trivial: add an extra compile-time condition to the existing check. It skips the check for `test_network`, while being completely ignored in other builds.

Fixes https://github.com/ProvableHQ/snarkOS/issues/4274.

note: I haven't been able to reliably trigger the issue in local tests, but I've been unable to to so with this change in place.

### DIFF
--- a/node/bft/src/helpers/storage.rs
+++ b/node/bft/src/helpers/storage.rs
@@ -505,7 +505,12 @@ impl<N: Network> Storage<N> {
                 bail!("Missing certificates for the previous round {previous_round} in storage {gc_log}")
             }
             // Ensure the number of previous certificate IDs is at or below the number of committee members.
-            if batch_header.previous_certificate_ids().len() > previous_committee_lookback.num_members() {
+            // As for the `test_network` clause, a more robust approach would be to set the
+            // previous_committee_lookback appropriately when dev-on-prod is activated, or to only skip this
+            // check right at the hotswap height.
+            if !cfg!(feature = "test_network")
+                && batch_header.previous_certificate_ids().len() > previous_committee_lookback.num_members()
+            {
                 bail!("Too many previous certificates for round {round} {gc_log}")
             }
             // Initialize a set of the previous authors.


node/bft/src/helpers/storage.rs
```diff
@@ -505,7 +505,9 @@ impl<N: Network> Storage<N> {
                 bail!("Missing certificates for the previous round {previous_round} in storage {gc_log}")
             }
             // Ensure the number of previous certificate IDs is at or below the number of committee members.
-            if batch_header.previous_certificate_ids().len() > previous_committee_lookback.num_members() {
+            if !cfg!(feature = "test_network")
+                && batch_header.previous_certificate_ids().len() > previous_committee_lookback.num_members()
```
vicsn, line 512: Can you add a comment that a more robust approach would be to set the previous_committee_lookback appropriately when dev-on-prod is activated, or to only skip this check right at the hotswap height.

vicsn (changes requested):
Can you adjust .circleci/config.yml to run the dev-on-prod restart test on this branch?

ljedrz:
Done; I'll resolve the conflict once the CI is green and the PR is approved.

vicsn (dismissed):

ljedrz:
Rebased (since the diff is trivial), and left the CI config unchanged (as the extra job has already passed).

vicsn (approved):

PR #4330, opened 2026-06-28
Mainnet v4.8.x

vicsn:
## Motivation

Release PR merging testnet into mainnet and updating snarkVM rev to https://github.com/ProvableHQ/snarkVM/pull/3314

No crates.io release will be made yet in this PR, but we can do this retro-actively when certain applications require it.

- [x] sync test passed
- [ ] prerelease testing passed


node/bft/ledger-service/src/ledger.rs
```diff
@@ -270,6 +159,68 @@ impl<N: Network, C: ConsensusStorage<N>> CoreLedgerService<N, C> {
...
+    })
+}
+
+/// Persists the given `start_round` to `<node_data_dir>/dev-committee-state` if the file
+/// doesn't already exist. Called after [`Ledger::load_with_dev_committee`] so that the
+/// round chosen by snarkVM (its `latest_round()` default) survives restarts.
+#[cfg(feature = "test_network")]
+pub fn persist_dev_committee_start_round_if_unwritten(node_data_dir: &NodeDataDir, start_round: u64) -> Result<()> {
+    let path = node_data_dir.dev_committee_state_path();
+    if !path.exists() {
```
Antonio95, line 201: If the path does exist, should this match `start_round` against the contents of the file or even update them to that value? (right now, it does nothing).

node/tcp/src/helpers/connections.rs
```diff
@@ -171,3 +174,38 @@ pub(crate) fn create_connection_span(addr: SocketAddr, parent: &Span) -> Span {
     try_span!(Level::WARN);
     error_span!(parent: parent, "conn", addr = %addr)
 }
+
+/// Describes what triggered a disconnect, as delivered to [`Disconnect::handle_disconnect`].
```
Antonio95, line 178: Are there test cases triggering these connection reasons?

Antonio95 (approved):
This falls largely outside my domain, but I did go through all of it and did not see any obvious red flags. Left two minor comments.

node/rest/src/routes.rs
```diff
@@ -1134,6 +1203,134 @@ impl<N: Network, C: ConsensusStorage<N>, R: Routing<N>> Rest<N, C, R> {
...
+            Err(err) => return Err(RestError::unprocessable_entity(anyhow!("Invalid request body: {err}"))),
+        };
+
+        // Parse the inputs into `Value<N>`.
+        let inputs = parse_view_inputs::<N>(&raw_inputs)?;
+
+        // Evaluate the view function in a blocking task.
+        // The latest block's state is captured inside the task to minimise the window
+        // between state sampling and evaluation.
+        let (outputs, height) = match tokio::task::spawn_blocking(move || {
```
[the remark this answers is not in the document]
vicsn: The API is known to cost resources

PR #4340, opened 2026-07-07
Test history read fix

vicsn:
## Motivation

Companion to https://github.com/ProvableHQ/snarkVM/pull/3323


ljedrz (approved):
LGTM, even with the additional commit on snarkVM side (the newtype, which is irrelevant here).

PR #4344, opened 2026-07-07
Reject trailing bytes in BFT transmission payloads

eranrund:
## Motivation

BFT transmission validation previously accepted raw transaction and solution buffers that contained a valid serialized object followed by trailing bytes. The checksum was computed over the full raw buffer, but the payload was later canonicalized to `Data::Object`, stripping the trailing bytes and causing future checksum computations to differ for the same `TransmissionID`.

This PR adds strict deserialization for BFT transaction and solution transmissions so raw buffers must contain exactly one serialized object. Padded buffers are rejected before canonicalization.

## Test Plan

Unit tests were added for the new strict deserialization helpers. Aside from that we will be relying on the existing tests.

## Documentation

N/A

## Backwards compatibility

From what I can tell this is not expected to cause any backward compatibility issues.

### DIFF
--- a/node/bft/ledger-service/Cargo.toml
+++ b/node/bft/ledger-service/Cargo.toml
@@ -86,3 +86,7 @@ workspace = true
 workspace = true
 optional = true
 features = [ "log" ]
+
+[dev-dependencies.snarkvm]
+workspace = true
+features = [ "test-helpers" ]
--- a/node/bft/ledger-service/src/helpers.rs
+++ b/node/bft/ledger-service/src/helpers.rs
@@ -1,0 +1,171 @@
+// Copyright (c) 2019-2026 Provable Inc.
+// This file is part of the snarkOS library.
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at:
+
+// http://www.apache.org/licenses/LICENSE-2.0
+
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use snarkvm::{
+    ledger::{Transaction, narwhal::Data, puzzle::Solution},
+    prelude::{FromBytes, Network, Result},
+};
+
+use anyhow::ensure;
+use std::io::Cursor;
+
+/// Deserializes a transaction, requiring raw buffers to contain exactly one serialized transaction.
+///
+/// For buffered transmissions, this rejects buffers that exceed the network transaction size limit or leave trailing
+/// bytes after `Transaction::read_le` succeeds. Object transmissions are returned as-is.
+pub fn deserialize_transaction_strict<N: Network>(transaction: Data<Transaction<N>>) -> Result<Transaction<N>> {
+    match transaction {
+        Data::Object(transaction) => Ok(transaction),
+        Data::Buffer(bytes) => {
+            ensure!(
+                bytes.len() <= N::LATEST_MAX_TRANSACTION_SIZE(),
+                "Transaction exceeds maximum size - {} bytes > {} bytes",
+                bytes.len(),
+                N::LATEST_MAX_TRANSACTION_SIZE()
+            );
+
+            let bytes_len = u64::try_from(bytes.len())?;
+            let mut reader = Cursor::new(bytes.as_ref());
+            let transaction = Transaction::<N>::read_le(&mut reader)?;
+            ensure!(
+                reader.position() == bytes_len,
+                "Transaction buffer contains {} trailing bytes",
+                bytes_len.saturating_sub(reader.position())
+            );
+            Ok(transaction)
+        }
+    }
+}
+
+/// Deserializes a solution, requiring raw buffers to contain exactly one serialized solution.
+///
+/// For buffered transmissions, this rejects buffers that leave trailing bytes after `Solution::read_le` succeeds.
+/// Object transmissions are returned as-is.
+pub fn deserialize_solution_strict<N: Network>(solution: Data<Solution<N>>) -> Result<Solution<N>> {
+    match solution {
+        Data::Object(solution) => Ok(solution),
+        Data::Buffer(bytes) => {
+            let bytes_len = u64::try_from(bytes.len())?;
+            let mut reader = Cursor::new(bytes.as_ref());
+            let solution = Solution::<N>::read_le(&mut reader)?;
+            ensure!(
+                reader.position() == bytes_len,
+                "Solution buffer contains {} trailing bytes",
+                bytes_len.saturating_sub(reader.position())
+            );
+            Ok(solution)
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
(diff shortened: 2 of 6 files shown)


vicsn (changes requested):
Thx!

I think this may not be backwards compatible.

When e.g. only 50% of the network upgraded to this version, a transaction in a proposal may not be deserialized anymore, making it impossible for 66% of proposals to be certified. This would have to be guarded by the most recent ConsensusVersion.

Doing ConsensusVersion guarded checks during block creation is easy, doing it in snarkOS is extra tricky because different nodes may be at different block heights while they're proposing asynchonrously for different consensus rounds.

So what can be done is:
1. In `fn propose_batch` we stop including these at `CONSENSUS_HEIGHT(ConsensusVersion::whatever)`
2. In 'fn process_batch_propose` we stop signing at `CONSENSUS_HEIGHT(ConsensusVersion::whatever)+50`, because we can be sure all nodes reached that height by then

There is some prior art for this in the code history but we also pruned some of this ductapy code over time. I'd also welcome an abstracted function which handles activating logic in both paths at the two different block heights.

We have an `upgrade-test` in `merge-workflow` which may help to simulate this, but it's not practical to simulate every possible edge case transaction - CI would take forever.

--
The same problem as above may occur in the future if we ever change the max transaction size. One way to make that more robust, is to not pick the latest max transaction size, but:
- in the `fn propose_batch` flow use the one at `ConsensusVersion(current_height)`
- in the `fn process_batch_propose` flow allow the max size from `max(max_transaction_sizes[(ConsensusVersion(current_height-100))..]`

--
At this point this whole enhancement seems incredibly complex, so please do share your 2 cents on if it's worth it from a maintainability perspective. Maybe the attack vector is not worth closing at this point or there is a simpler approach which solves most of the issues.

eranrund:
Thank you for the detailed response @vicsn!

I want to separate two concerns: max transaction size, and trailing bytes / non-canonical transmission payloads.

## Max transaction size (the `N::LATEST_MAX_TRANSACTION_SIZE()` check)

I agree that if this value changes, then nodes running an old version could refuse transactions from newer upgraded nodes, or vice versa, depending on whether the value increased or decreased.

For example, if the value increases, old nodes would reject transactions that exceed their size limit but are accepted by upgraded nodes. Unless I am missing something, this behavior is not changed by this PR.

I am also not sure there is a way to fully address this for non-upgraded nodes, since they can only apply the logic they already know about. Either way, I think this is tangential to the checksum / trailing-bytes concern, and I think its a separate hardening discussion. I'm happy to dive deeper into it.

## Rejecting upon discovering trailing bytes
I am not convinced it will be backwards-incompatible (but certainly open to find out that it is!). Here's my thinking on it:

Normal unconfirmed transaction / solution ingress (e.g. client REST broadcasts) are deserialized before reaching BFT. Consensus sends typed objects to the primary as `Data::Object`, so original trailing bytes are not preserved for local BFT inclusion ([solutions](https://github.com/ProvableHQ/snarkOS/blob/d8125c9cb2e3ce22cb1a3b19c4385415d5c02e04/node/consensus/src/lib.rs#L361-L367), [transactions](https://github.com/ProvableHQ/snarkOS/blob/d8125c9cb2e3ce22cb1a3b19c4385415d5c02e04/node/consensus/src/lib.rs#L466-L477)).

The relevant BFT binary `Transmission` data path I traced is missing-transmission fetch. A peer proposal can reference a transmission the node does not have, causing the node to [sync from the batch header](https://github.com/ProvableHQ/snarkOS/blob/d8125c9cb2e3ce22cb1a3b19c4385415d5c02e04/node/bft/src/primary.rs#L958-L960), identify missing transmissions, and [fetch them from the peer](https://github.com/ProvableHQ/snarkOS/blob/d8125c9cb2e3ce22cb1a3b19c4385415d5c02e04/node/bft/src/primary.rs#L1948-L1995).

That fetch goes through [`worker.get_or_fetch_transmission`](https://github.com/ProvableHQ/snarkOS/blob/d8125c9cb2e3ce22cb1a3b19c4385415d5c02e04/node/bft/src/worker.rs#L220-L235). On a cache miss, the worker waits for a `TransmissionResponse`, and `finish_transmission_request` calls `ensure_transmission_is_well_formed` before completing the pending fetch ([response validation](https://github.com/ProvableHQ/snarkOS/blob/d8125c9cb2e3ce22cb1a3b19c4385415d5c02e04/node/bft/src/worker.rs#L552-L568)).

On an old node (running code before this PR), that first validation call can accept a padded `Data::Buffer` and mutate it into `Data::Object` ([transaction mutation](https://github.com/ProvableHQ/snarkOS/blob/d8125c9cb2e3ce22cb1a3b19c4385415d5c02e04/node/bft/ledger-service/src/ledger.rs#L390-L421), [solution mutation](https://github.com/ProvableHQ/snarkOS/blob/d8125c9cb2e3ce22cb1a3b19c4385415d5c02e04/node/bft/ledger-service/src/ledger.rs#L423-L447)). After that mutation, the checksum no longer matches the checksum in the `TransmissionID`, because the trailing bytes are not preserved by canonical object serialization.

However, `process_batch_propose_from_peer` re-runs `ensure_transmission_is_well_formed` on the returned `missing_transmissions` before `check_batch_header` and before signing ([second validation and return](https://github.com/ProvableHQ/snarkOS/blob/d8125c9cb2e3ce22cb1a3b19c4385415d5c02e04/node/bft/src/primary.rs#L962-L972), [signing happens later](https://github.com/ProvableHQ/snarkOS/blob/d8125c9cb2e3ce22cb1a3b19c4385415d5c02e04/node/bft/src/primary.rs#L996-L1002)). That second validation is what appears to prevent an old honest validator from signing a fresh batch proposal containing such a fetched transmission today.

The changes in this PR should result in rejecting earlier rather than changing an old honest validator from "would sign" to "would not sign". The change also removes reliance on the caller performing that second validation pass, which seems like useful defense in depth.

## A broader invariant

While thinking about this, I also had a thought that the broader invariant is canonical binary round-tripping:

`Transaction::read_le(bytes)?.to_bytes_le()? == bytes`

and similarly for solutions. Trailing bytes are one way to violate that invariant, but not the only possible way. Checking the full canonical round-trip invariant would require reserializing, which may have a measurable cost, especially for large transactions.

I haven't traced through all the binary serialization and deserialization paths so I am not sure if it is possible to have something such as, for example, deserializing a bool where any non-zero byte is stored as `true` in the struct, and then when serializing back we write `1`. If you think this might be a concern I am happy to spend some time reviewing this, it shouldn't be too difficult to look at all the various read_ methods with the help of AI.

vicsn (approved):
LGTM!

Agree with your analysis. :)

Can you make new issues for:
- fragility of latest_max_tx_size
- I'm pretty sure we deserialize/reserialize data too often throughout the code. Some parts of is_transmission_well_formed but also in snarkVM's check_transaction. With AI tracking down and removing duplicate checks for improved efficiency should be easy.

I'm not worried at the moment about the serialization roundtrip invariant.

eranrund:
> not pick the latest max transaction size, but:
> 
> * in the `fn propose_batch` flow use the one at `ConsensusVersion(current_height)`
> * in the `fn process_batch_propose` flow allow the max size from `max(max_transaction_sizes[(ConsensusVersion(current_height-100))..]`

Thank you!

I wrote https://github.com/ProvableHQ/snarkOS/issues/4387 and https://github.com/ProvableHQ/snarkOS/issues/4388

PR #4345, opened 2026-07-08
Allow more warns in CI

vicsn:
## Motivation

The alternative would be to downgrade `warn!("Received signature for an older batch {batch_id}");` to TRACE.


ljedrz (approved):
I'm guessing this is temporary? That's a healthy bump.

vicsn:
I think this issue may have been present for a while, so I'd deem it permanent. i.e. the warn limit is effectively useless for these two tests

PR #4346, opened 2026-07-08
[Fix] Abort the FD monitor task on shutdown

### LINKED ISSUE #4326: [Bug] merge-workflow tests fail due to live tasks
## 🐛 Bug Report

`upgrade-test` and `chaotic*` tests now fail several times in a row because of:
```
[ci-runner] 2026-06-16T11:58:44Z ❌ Test failed! Client #0 logs contain errors.
2026-06-16T11:56:41.464607Z ERROR There are still 1 live tasks
```

```
[ci-runner] 2026-06-16T11:58:44Z ❌ Test failed! Validator #0 logs contain errors.
2026-06-16T11:56:41.464607Z ERROR There are still 1 live tasks
```

One thought I had was to only check for ERROR logs up until the point that we start turning off all nodes. But this doesn't work because various tests restart nodes multiple times. 

We will need to track down the live task and kill it with kindness.

ljedrz:
It looks like a bit of a hack (the code checks against a `1` instead of `0` now), but the FD checker is a lower-layer resource, and we wouldn't gain anything by having its handle traverse all the `Node` creation APIs - or by pulling out the current shutdown plumbing.

I checked locally that both the `ERROR` is gone, and that the FD checker task gets explicitly aborted on shutdown (which already happens implicitly).

Fixes https://github.com/ProvableHQ/snarkOS/issues/4326.

### DIFF
--- a/node/src/traits.rs
+++ b/node/src/traits.rs
@@ -61,10 +61,15 @@ pub trait NodeInterface<N: Network>: Routing<N> {
         // Check if there are any stragglers left.
         if let Some(handle) = &handler.handle {
             let live_tasks = handle.metrics().num_alive_tasks();
-            // The 1 "extra" task comes from the FD monitor, which is aborted once
-            // wait_for_signals concludes its work.
-            if live_tasks != 1 {
-                error!("There are still {} live tasks", live_tasks - 1);
+            let expected_leftover = if cfg!(unix) {
+                // The 1 "extra" task comes from the FD monitor, which is aborted once
+                // wait_for_signals concludes its work.
+                1
+            } else {
+                0
+            };
+            if live_tasks > expected_leftover {
+                error!("There are {} surplus live tasks", live_tasks - expected_leftover);
             }
         }
     }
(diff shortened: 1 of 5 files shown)


node/src/traits.rs
```diff
@@ -61,8 +61,10 @@ pub trait NodeInterface<N: Network>: Routing<N> {
         // Check if there are any stragglers left.
         if let Some(handle) = &handler.handle {
             let live_tasks = handle.metrics().num_alive_tasks();
-            if live_tasks != 0 {
-                error!("There are still {live_tasks} live tasks");
+            // The 1 "extra" task comes from the FD monitor, which is aborted once
+            // wait_for_signals concludes its work.
+            if live_tasks != 1 {
```
vicsn: Is the fd-checker linux only? Shall we turn this into `<= 1` and use a `saturating_sub` below?

ljedrz: ah, good point, I forget there's other platforms :grin:; I'll send a patch shortly

ljedrz: done in https://github.com/ProvableHQ/snarkOS/pull/4346/commits/b322ecd9d089c36bf448dc358a37d2c5077b6290

vicsn (dismissed):

vicsn (approved):

Issue #4253, opened 2026-05-05
[Feature] Clean up transaction_spend_in_microcredits


vicsn:
## 🚀 Feature

After https://github.com/ProvableHQ/snarkVM/pull/3242 is merged, this logic is duplicate in snarkVM and snarkOS, it would be nicer if we can encapsulate it purely in snarkVM

eranrund:
Is the idea here to remove `transaction_spend_in_microcredits` in favor of a new helper we'll create in snarkVM to encapsulate https://github.com/ProvableHQ/snarkVM/blob/021669dd442f7ba7e1cec86bc2979226751ec1d0/synthesizer/src/vm/finalize.rs#L1106-L1124?

I know this is a fairly insignificant cleanup, I just used this issue as a simple introductory one to begin familiarizing myself with the repos :)

vicsn:
That would indeed be the idea! Just to keep things DRY. :)

eranrund:
This isn't actually done yet, only the matching piece in snarkVM. I'll put out a PR to use the new helper.

PR #4348, opened 2026-07-09
Use transaction compute spend helper

eranrund:
## Motivation

DRYing things up, see #4253 

## Test Plan

Relying on existing tests.

## Documentation

N/A

## Backwards compatibility

Logic should be identical so no backwards compatibility issues expected.

### DIFF
--- a/node/bft/ledger-service/src/ledger.rs
+++ b/node/bft/ledger-service/src/ledger.rs
@@ -43,10 +43,7 @@ use snarkvm::{
         bail,
         cfg_into_iter,
         consensus_config_value_by_version,
-        deploy_compute_cost_in_microcredits,
-        deployment_cost,
-        execute_compute_cost_in_microcredits,
-        execution_cost,
+        transaction_compute_spend_in_microcredits,
     },
 };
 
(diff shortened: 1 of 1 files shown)


node/bft/ledger-service/src/ledger.rs
```diff
@@ -543,32 +540,15 @@ impl<N: Network, C: ConsensusStorage<N>> LedgerService<N> for CoreLedgerService<
...
-                ensure!(
-                    compute_spend <= transaction_spend_limit,
-                    "Transaction '{id}' exceeds the transaction spend limit with compute_spend: '{compute_spend}'"
-                );
-                Ok(compute_spend)
-            }
-            Transaction::Execute(_, _, execution, _) => {
-                let (_, cost_details) = execution_cost(self.ledger.vm().process(), execution, consensus_version)?;
-                let compute_spend = execute_compute_cost_in_microcredits(cost_details, consensus_version);
-                if consensus_version >= ConsensusVersion::V11 {
```
mohammadfawaz, line 559: Is this consensus version gate still active with this refactor? Does `transaction_compute_spend_in_microcredits` exhibit the same behavior?

vicsn: Even if it isn't, it would be harmless to remove it as the network is already past that consensus version.

vicsn (approved):

PR #4354, opened 2026-07-28
[Fix/Feat] Asymmetrical Noise-XX handshake for the validators

ljedrz:
Replaces the gateway's challenge-response handshake with one built on the Noise XX pattern, where each side binds its Aleo identity to the session by signing the running Noise handshake hash. Identity is still the Aleo account key - the Noise static keys are per-connection and exist only to give the pattern a channel to bind to. The post-handshake traffic remains unencrypted.

The legacy handshake is still accepted, so validators can be upgraded one at a time. Guarded by `NOISE_HANDSHAKE_ACTIVATION_HEIGHT`.

Over 50% of the new code is tests, and much of the new plumbing can be reused if we want to utilize this handshake pattern with other node types as well.

Benefits:
- Relay/impersonation is closed. Each side signs its own session's handshake hash, and two sessions cannot share one, so a forwarded signature never verifies. Note: the relay would yield an eclipse, not a forgery - nothing attributable to the victim can be fabricated, and the attacker cannot speak for any other validator on that connection.
- The DoS asymmetry is inverted. The responder reads the initiator's cleartext hint and runs every cheap check - version, trusted, committee membership, already-connected - before deriving a single key. This makes the handshake initially more expensive for the initiator.
- Peer-pool mutation moved after the checks, so an unauthorized peer can no longer occupy a pool slot at all.
- Pre-authentication buffering drops 16x (1MiB -> 65535B) once legacy is retired.

As a small drive-by, there's also a hardening for the `EventCodec`'s encoder, in its [own commit](https://github.com/ProvableHQ/snarkOS/commit/2b5feeeb783d2d08497a4c6f77bb42536ee4ef15).

CC https://github.com/ProvableHQ/snarkOS/issues/4140


vicsn:
Can you enable the `merge-workflow` so we can make sure the `upgrade-test` succeeds? But before that tackle https://github.com/ProvableHQ/snarkOS/issues/4356 ?

node/bft/src/gateway.rs
```diff
@@ -118,6 +133,23 @@ const CONNECTION_ATTEMPTS_SINCE_SECS: i64 = 10;
...
+///
+/// Setting this height is not the end of the migration, only the middle of it. For as long as the
+/// responder still accepts the legacy handshake, two things remain reachable through it: the relay
+/// that the handshake binding exists to prevent, and the legacy handshake codec's 1 MiB frame limit -
+/// sixteen times what a Noise message may be, on a buffer an unauthenticated peer gets to size. Both
+/// are collected by the same step, which is dropping the legacy path rather than merely preferring
+/// the new one. Note also that the same relay is reachable through the router's handshake, which
+/// signs a byte-identical message with the same account key, so the gateway cannot be the last part
+/// of this to be converted.
+const NOISE_HANDSHAKE_ACTIVATION_HEIGHT: Option<u32> = None;
```
vicsn: Can we set this to `CONSENSUS_HEIGHT(ConsensusVersion::V19)` (regrettably a runtime value)

vicsn (changes requested):
Looks amazing!

It would be great if we can truly consider the issues closed with one PR (ignoring the pruning of dead code). Therefore:
- should we also adjust the gateway handshake done by the bootstrap nodes?
- either we:
  - keep `NOISE_HANDSHAKE_ACTIVATION_HEIGHT=V19` and introduce a new `NOISE_HANDSHAKE_DEACTIVATION_HEIGHT=V20`.
  - replace `NOISE_HANDSHAKE_ACTIVATION_HEIGHT=V19` to a `NOISE_HANDSHAKE_DEACTIVATION_HEIGHT=V19`. As soon as a validator has upgraded, they *attempt* a noise handshake. If it fails, they retry with the legacy handshake. At `NOISE_HANDSHAKE_DEACTIVATION_HEIGHT==V19` we assume all nodes upgraded and everyone is using noise.

ljedrz:
> should we also adjust the gateway handshake done by the bootstrap nodes?

done in https://github.com/ProvableHQ/snarkOS/pull/4354/commits/5fbacbab31cc8513c73ac9631e283dc9724a283b

> keep NOISE_HANDSHAKE_ACTIVATION_HEIGHT=V19 and introduce a new NOISE_HANDSHAKE_DEACTIVATION_HEIGHT=V20

I introduced 2 `const`s, one for activation of the new protocol, and one for deactivation of the legacy one; I'd apply `V19`, but the pinned snarkVM version still has `V18` as its maximum (so it wouldn't compile). It can be toggled once that's done.

ljedrz:
In the latest changes, the handshake message size was further restricted down to 1KiB, minimizing DoS exposure while still leaving ample room if we need to add any more fields in the future; extending it is backward-compatible, too.

node/bft/src/gateway.rs
```diff
@@ -1590,14 +1679,344 @@ async fn send_event<N: Network>(
...
+/// Concludes a Noise handshake, handing the stream back to the connection.
+///
+/// The stream deliberately goes back unframed. [`Reading`] builds a codec of its own, capped at the
+/// 256 MiB an event may be rather than the 64 KiB a Noise message may be, and takes full
+/// responsibility for the stream from here on. Handing it a bare stream is only sound because the
+/// session reads its messages exactly: anything the peer pipelined behind the last handshake message
+/// is still on the socket rather than in a buffer about to be dropped.
+fn finish_noise_handshake(noise: NoiseSession<&mut TcpStream>) {
+    // Note: the transport keys are discarded here, leaving the resulting connection unencrypted.
+    let _stream = noise.into_inner();
```
cbeck88, line 1723: i understand that discarding these transport keys is the only reasonable option in this PR, but for my understanding, what protections exist for the subsequent messages sent over the connection?

node/bft/events/src/helpers/codec.rs
```diff
@@ -53,13 +64,17 @@ impl<N: Network> Encoder<Event<N>> for EventCodec<N> {
...
+        // sink is fed several events before being flushed - and folding those into this event's frame
+        // would produce a single frame that no decoder can make sense of.
+        let buffered = dst.len();
         event
             .write_le(&mut dst.writer())
             // This error should never happen, the conversion is for greater compatibility.
             .map_err(|_| std::io::Error::new(std::io::ErrorKind::InvalidData, "serialization error"))?;
 
-        let serialized_event = dst.split_to(dst.len()).freeze();
+        let serialized_event = dst.split_off(buffered).freeze();
```
cbeck88, line 77: 👍

node/bft/src/gateway.rs
ljedrz: everything that contains consensus state carries its own proof, bound to something the receiver already knows independently - everything that's crypto-sensitive is cryptographically signed; an on-path attacker could read the traffic, and they could inject valid but useless messages to make a node do wasted verification work, bounded by the same rate limits and pending-request tracking that already bound a misbehaving authenticated peer; what they cannot do is forge anything that changes state, impersonate a validator, or replay the session elsewhere

node/bft/src/gateway.rs
```diff
@@ -118,6 +134,38 @@ const CONNECTION_ATTEMPTS_SINCE_SECS: i64 = 10;
...
+/// Only the initiator's choice is gated: a responder accepts either protocol as soon as this code
+/// ships, which is what allows validators to be upgraded one at a time. `None` means no switchover
+/// has been scheduled yet - except in development, where the Noise path is always taken so that
+/// devnets exercise it, and in tests, which pin the choice explicitly.
+///
+/// Setting this is not the end of the migration, only the middle of it; see
+/// [`LEGACY_HANDSHAKE_EXPIRY`].
+// TODO: this is meant to be `Some(ConsensusVersion::V19)`, which the pinned snarkVM does not define
+//  yet - its `ConsensusVersion` stops at `V18`. Set it once the dependency carries the variant.
+const NOISE_HANDSHAKE_ACTIVATION: Option<ConsensusVersion> = None;
```
vicsn: snarkVM staging should now have the required consensus versions

ljedrz:
It seems like that CI failure is just a re-insertion of a known cert into the DAG, which I believe could safely be downgraded in terms of its log level.

cbeck88 (dismissed):
looks good to me, thank you for doing this

node/bft/src/gateway.rs
```diff
@@ -118,6 +134,34 @@ const CONNECTION_ATTEMPTS_SINCE_SECS: i64 = 10;
...
+/// scheduled.
+///
+/// Only the initiator's choice is gated: a responder accepts either protocol as soon as this code
+/// ships, which is what allows validators to be upgraded one at a time. `None` means no switchover
+/// has been scheduled yet - except in development, where the Noise path is always taken so that
+/// devnets exercise it, and in tests, which pin the choice explicitly.
+///
+/// Setting this is not the end of the migration, only the middle of it; see
+/// [`LEGACY_HANDSHAKE_EXPIRY`].
+const NOISE_HANDSHAKE_ACTIVATION: Option<ConsensusVersion> = Some(ConsensusVersion::V19);
```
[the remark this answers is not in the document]
vicsn: Updated the PR README, we want the upgrade

ljedrz:
The `devnet-workflow` CI failure is just https://github.com/ProvableHQ/snarkOS/issues/4373.

vicsn:
Some Opus 5 findings I think are worth to close:

### 3. Medium — the initiator never checks it reached the validator it dialed

gateway.rs
Lines 1852-1854
if let Some(reason) = self.verify_peer_info(peer_addr, &peer_info, restrictions_id) {
    return Err(reason.into_connect_error(peer_addr));
}
verify_peer_info only requires the responder to be some committee member with a matching restrictions ID that isn't already connected. An attacker who can redirect A's dial of B to a different genuine validator C makes A register C's Aleo address under B's listener address in both peer_pool and the resolver — and because is_connected_address(C) then returns true, A's real connection to C gets refused as AlreadyConnectedToAleoAddress. That's a targeted eclipse with no key compromise.

This is pre-existing in the legacy handshake, but identity binding makes it cheap to close: the gateway generally knows the address it expects (Peer::last_known_aleo_addr, plus the validator set it dials from), so compare it to peer_info.address.

### 5. Low — the "no elliptic curve operations" claim is overstated
Parsing the hint decodes an Address. noise.rs:104 and gateway.rs:1918 shouldn't say "no elliptic curve operations at all".

vicsn (approved):

PR #4358, opened 2026-07-29
Remove thread pool starvation

vicsn:
## Motivation

A validator halted and/or stopped reporting metrics, printing this on repeat:
```
│2026-07-29T11:29:26.232703Z  INFO [locktick] checking for active lock guards                                                                           │
│2026-07-29T11:29:26.233098Z  INFO [locktick] /home/ubuntu/snarkOS/node/bft/src/bft.rs@602:53 (Lock): 39; 1 active; 18 waiting; avg d: 368.549535ms; avg│
│2026-07-29T11:29:26.233110Z  INFO [locktick] /home/ubuntu/snarkOS/node/bft/src/helpers/telemetry.rs@128:40 (Write): 5034; 0 active; 32 waiting; avg d: │
│2026-07-29T11:29:26.233113Z  INFO [locktick] /home/ubuntu/snarkOS/node/bft/src/helpers/telemetry.rs@178:36 (Read): 98; 1 active; 0 waiting; avg d: 1.54│
│2026-07-29T11:29:26.233116Z  INFO [locktick] /home/ubuntu/snarkOS/node/bft/src/helpers/telemetry.rs@179:36 (Read): 98; 1 active; 0 waiting; avg d: 1.48│
│2026-07-29T11:29:26.233118Z  INFO [locktick] /home/ubuntu/snarkOS/node/bft/src/helpers/telemetry.rs@180:38 (Read): 98; 1 active; 0 waiting; avg d: 1.41│
│2026-07-29T11:29:26.233122Z  INFO [locktick] /home/ubuntu/snarkOS/node/bft/src/primary.rs@475:69 (Write): 23; 1 active; 0 waiting; avg d: 114.098573ms;│
│2026-07-29T11:29:26.233126Z  INFO [locktick] /home/ubuntu/snarkOS/node/bft/src/sync/mod.rs@538:47 (Lock): 1; 1 active; 0 waiting; avg d: 0ns; avg w: 15│
│2026-07-29T11:29:26.233137Z  INFO [locktick] /home/ubuntu/snarkOS/node/bft/src/sync/mod.rs@375:34 (Lock): 2; 1 active; 0 waiting; avg d: 9.626086953s; │
│2026-07-29T11:29:26.233154Z DEBUG [locktick] finished the check in 430.4µs                                                                             │
```

Opus 5 believes:

The deadlock wasn't a lock-ordering cycle between two locks — it was the rayon thread pool starving itself through the telemetry locks.

## The cycle
Telemetry::update_participation_scores took read guards on all three trackers and then ran rayon parallel iterators while holding them:
```
let tracked_certificates = self.tracked_certificates.read();   // line 178
let validator_signatures = self.validator_signatures.read();   // line 179
let validator_certificates = self.validator_certificates.read(); // line 180
```
...
let signature_participation_scores: IndexMap<_, _> = cfg_iter!(&*validator_signatures)   // line 186
Meanwhile Telemetry::insert_subdag did the opposite: it fanned certificates out over rayon and acquired a write guard on the same lock inside each closure (insert_certificate, line 128). cfg_iter! expands to par_iter(), so both of these run on the single global pool, which is built with a fixed num_cpus workers in cli/src/commands/start.rs:

```
start.rs
Lines 993-997
rayon::ThreadPoolBuilder::new()
    .stack_size(8 * 1024 * 1024)
    .num_threads(num_rayon_cores_global)
    .build_global()
    .unwrap();
```

So: thread A is inside update_participation_scores, holding the three read guards, blocked waiting for its par_iter jobs to be picked up. Every worker in the pool is blocked in tracked_certificates.write() on behalf of a different, concurrent insert_subdag. Those writes can't be granted until A drops its read guards, and A can't proceed until a worker becomes free. Nothing breaks the tie, forever.

Your log says exactly this. All three read sites report 1 active simultaneously, so one thread is sitting past line 180 with all three guards. Line 128 reports 32 waiting, which is the whole pool parked on that one write lock — a worker count, not a coincidence.

The two concurrent insert_subdag calls come from the only two callers, and both appear wedged in the log: the BFT commit path at node/bft/src/bft.rs:804 (which is why bft.rs@602, the commit_lock, shows 1 active; 18 waiting) and the bootup/post-sync path at node/bft/src/sync/mod.rs:449 (which is why sync/mod.rs@375, the pending_blocks mutex, shows avg d: 9.6s, and why response_lock at sync/mod.rs@538 is still held). Everything else in that log, including propose_batch's latest_proposal_timestamp write at primary.rs@475, is downstream fallout rather than a cause.

## Why it surfaced now rather than always
It needs two insert_subdag calls overlapping, which happens when the sync path is replaying blocks at the same time as the BFT is committing. Note that telemetry is on by default for the node binary (root Cargo.toml:272), so this is live in production builds, not just a feature-gated experiment.

### DIFF
--- a/node/bft/src/helpers/telemetry.rs
+++ b/node/bft/src/helpers/telemetry.rs
@@ -18,7 +18,7 @@ use snarkvm::{
         committee::Committee,
         narwhal::{BatchCertificate, BatchHeader, Subdag},
     },
-    prelude::{Address, Field, Network, cfg_chunks, cfg_iter},
+    prelude::{Address, Field, Network, cfg_iter},
 };
 
 use indexmap::{IndexMap, IndexSet};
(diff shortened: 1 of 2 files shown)


ljedrz (approved):
LGTM 👍

PR #4360, opened 2026-07-30
[WIP] Gracefully handle nodes that now prune historical blocks

eranrund:
This is a companion PR to https://github.com/ProvableHQ/snarkVM/pull/3329

The main benefits of the smaller state, is that initial participation and recovery will be much faster, making the network more robust.

It contains two major changes, presented as individual commits that hopefully make it easier to review/reason about.

---

## Background

The [companion snarkVM PR](https://github.com/ProvableHQ/snarkVM/pull/3329) prunes authority data (`authority_map` + `certificate_map`) for blocks older than `AUTHORITY_RETENTION_BLOCKS` (100,000 blocks, roughly three days at 2.5s block times). Per review discussion there, peers do **not** advertise whether they keep full history. Instead the protocol contract becomes implicit and uniform: every peer serves `[tip − AUTHORITY_RETENTION_BLOCKS, tip]`, and anything older is fetched from the trusted CDN (automatic on startup) or from a ledger snapshot (manual).

That design needs two things from snarkOS, which today does neither: a node must be able to *decline* to serve blocks it no longer has without punishing the requester, and a node must avoid asking peers for blocks they cannot possibly still hold - and know what to do when nobody can serve it.

## Commit 1: `Don't treat get_blocks as failure and penalize requesting peer`

Today, when a node fails to load the blocks in a `BlockRequest`, the handler returns `false`/`Err` and the inbound dispatcher turns that into "peer sent an invalid block request", disconnecting the **requester** for a protocol violation. The failure is attributed to the wrong side: with pruning, an honest node would disconnect every peer that legitimately asks for a range it has pruned (and even without pruning, it's not the requesters fault that `get_blocks` failed...).

This commit makes an unavailable range a normal, benign outcome:

- **Serving side**: In the client router, validator router, and BFT gateway, a `get_blocks` failure now logs a warning and replies with an **empty `BlockResponse`** rather than failing the message. An empty response is expressible in the existing wire format (`DataBlocks` serializes a `u8` count), so there is no format change and no new message variant.
- **Requesting side**: `DataBlocks::ensure_response_is_well_formed` now accepts an empty response, documented as the peer's explicit "I cannot serve this range" signal. Ordering and range checks still apply to non-empty responses.
- `InsertBlockResponseError::EmptyBlockResponse` is now classified as benign. The existing machinery then does the right thing without further changes: the peer's outstanding requests are removed and marked for re-issue to other peers, so failover is immediate rather than waiting out the 60s request timeout.

**Compatibility.** No wire format change. Against an un-upgraded requester, an empty response is rejected and that peer disconnects the *server* and retries elsewhere - noisier, but assumed to be better than today, where the server disconnects an innocent requester. Un-upgraded servers never emit empty responses, so their behavior is unchanged.

## Commit 2: `Don't request blocks that peers have pruned, and shut down when no peer can serve them`

With commit 1 alone, a node far behind the network would request pruned ranges, receive empty responses, re-issue to the same peers, and churn indefinitely. This commit "teaches" the requester the retention contract and gives it an exit.

- **Retention-aware peer selection**: `find_sync_peers_inner` now skips peers whose assumed pruning floor lies above the next block we need. The floor is derived from the peer's advertised tip: `tip − (AUTHORITY_RETENTION_BLOCKS − RETENTION_SLACK)`. The slack (1,000 blocks, chosen arbitrarily) exists because an advertised tip is only a snapshot - by the time our request arrives the peer may have advanced and pruned correspondingly, so we cannot assume it serves exactly `tip − retention`.
- **Stuck detection**: `BlockSync` records when peers ahead of us exist but every one of them has pruned what we need, exposed as `sync_stuck_below_peer_floors()`. This state cannot resolve itself - peers prune further with every block they produce, so the gap only widens. It is deliberately distinguished from simply having no useful peers (not yet connected, isolated), which is transient and does not count.
- **Fatal shutdown**: After the condition persists for five minutes, the node logs how to recover and shuts down gracefully via the new `SignalHandler::stop_with_failure()`, which causes the process to exit non-zero so a supervisor restarts it - at which point the existing startup CDN sync closes the gap automatically. Restoring a snapshot is the manual alternative; nodes run with `--nocdn` must use one.

The choice of shutdown over in-process CDN fetching was deliberate. Mid-run CDN sync would require reworking the one-shot `CdnBlockSync` lifecycle, re-routing its ledger writes through the advancement lock it currently bypasses, and adding a pause/resume surface to `BlockSync` - and it would be difficult to cover validators, where a mid-run CDN jump would strand BFT/DAG state. Restarting into the existing, well-tested startup path achieves the same outcome. This is a thing we can improve upon if we decide to invest resources in it, but in reality I think we are not going to experience nodes being down for almost 3 days and then suddenly coming back to life and trying to sync. If a node fell that far behind, it might as well restart. This keeps the code and potential edgecases smaller and easier to reason about. Its also worth emphasizing that CDN sync takes place every time a node restarts, so this is standard practice.

## Misc notes

- **Devnet validation is still outstanding** (`.ci/test_devnet.sh`). We'll need to make the 100k prune block window configurable (maybe via a build env var), and we'd need a way to trigger a node falling far behind and aborting.
- **The stuck detector is inferred from peer claims.** A peer advertising a tip far above the real chain is unverifiable. In a progressing network this is inert: every real block gives an honest peer a servable tip and resets the timer, but during a network halt, when no honest peer is ahead, a lying peer could push idle nodes to exit. I am not convinced this is a problem since the network will (should?) never halt for any meaningful duration of time... if this is a concern, we might need hardening around this such as requiring a majority to choose the next tip, or choosing the minimum out of all peer-reprted tips that are ahead of us instead of choosing the maximum (which is what I think the code currently does).
- **REST still answers misleadingly for pruned blocks** (`GET /block/{h}` 404s; `GET /blocks` fails the whole range) - a separate follow-up. I'm considering using HTTP code 410 Gone (`Gone client error response status code indicates that the target resource is no longer available at the origin server`) to differentiate between a block that never existed and a block that used to exist but are no longer available due to pruning.

Another thing worth mentioning is that this shifts the entire network (after an upgrade) to assuming no node can actually provide post-prune-window block, even though some might (due to having the `history` feature enabled). That was a deliberate decision that favors simplicity over having each peer return different ranges it can fulfill ([explicitly suggested](https://github.com/ProvableHQ/snarkVM/issues/3230#issuecomment-5090974483) in the companion PR).

### DIFF
--- a/node/sync/src/block_sync.rs
+++ b/node/sync/src/block_sync.rs
@@ -26,7 +26,7 @@ use snarkos_node_sync_locators::{CHECKPOINT_INTERVAL, NUM_RECENT_BLOCKS};
 
 use snarkvm::{
     console::network::{ConsensusVersion, Network},
-    ledger::{Block, CheckBlockError},
+    ledger::{Block, CheckBlockError, store::AUTHORITY_RETENTION_BLOCKS},
     utilities::flatten_error,
 };
 
@@ -91,6 +91,21 @@ const MAX_BLOCK_REQUESTS: usize = 50; // 50 requests
 
 /// The maximum number of blocks tolerated before the primary is considered behind its peers.
 pub const MAX_BLOCKS_BEHIND: u32 = 1; // blocks
+
+/// A peer's advertised tip is only a snapshot: by the time our block request arrives, the peer may
+/// have advanced by a few blocks and pruned a corresponding number of old ones. We therefore do not
+/// assume a peer serves exactly `tip - AUTHORITY_RETENTION_BLOCKS`, but leave this many blocks of slack.
+const RETENTION_SLACK: u32 = 1_000;
+
+/// The duration after which a node that cannot request its needed blocks from any peer
+/// (see [`BlockSync::sync_stuck_below_peer_floors`]) should give up and shut down.
+pub const STUCK_SYNC_SHUTDOWN_THRESHOLD: Duration = Duration::from_secs(5 * 60);
+
+/// Returns the lowest block height the peer with the given tip is assumed to still serve,
+/// given that peers prune authority data older than `AUTHORITY_RETENTION_BLOCKS` blocks.
+fn assumed_peer_floor(peer_tip: u32) -> u32 {
+    peer_tip.saturating_sub(AUTHORITY_RETENTION_BLOCKS.saturating_sub(RETENTION_SLACK))
+}
 
 /// This is a dummy IP address that is used to represent the local node.
 /// Note: This here does not need to be a real IP address, but it must be unique/distinct from all other connections.
@@ -250,6 +265,10 @@ pub struct BlockSync<N: Network> {
     /// responding but cannot keep up with the request rate.
     last_response_at: Mutex<HashMap<SocketAddr, Instant>>,
 
+    /// Tracks since when the blocks this node needs are below the assumed pruning floor of every
+    /// connected peer, i.e. no peer can serve them and sync cannot progress.
+    unservable_since: Mutex<Option<Instant>>,
+
     /// Condition variable that wakes up waiting tasks when the node is synced.
     synced_notify: Notify,
 }
@@ -275,6 +294,7 @@ impl<N: Network> BlockSync<N> {
             failed_requests: Default::default(),
             last_response_at: Default::default(),
             synced_notify: Default::default(),
+            unservable_since: Default::default(),
         }
     }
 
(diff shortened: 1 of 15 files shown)


node/sync/src/block_sync.rs
```diff
@@ -92,6 +92,21 @@ const MAX_BLOCK_REQUESTS: usize = 50; // 50 requests
 /// The maximum number of blocks tolerated before the primary is considered behind its peers.
 pub const MAX_BLOCKS_BEHIND: u32 = 1; // blocks
 
+/// A peer's advertised tip is only a snapshot: by the time our block request arrives, the peer may
+/// have advanced by a few blocks and pruned a corresponding number of old ones. We therefore do not
+/// assume a peer serves exactly `tip - AUTHORITY_RETENTION_BLOCKS`, but leave this many blocks of slack.
+const RETENTION_SLACK: u32 = 1_000;
```
vicsn, line 98: I feel like this slack is already encompassed by the STUCK_SYNC_SHUTDOWN_THRESHOLD, we will keep trying for a while, and the slack is also insignificant compared to the size of AUTHORITY_RETENTION_BLOCKS. So I would remove but add that context above `fn assumed_peer_floor`

vicsn (commented):
Great to see!

> The choice of shutdown over in-process CDN fetching was deliberate.

I think I agree, but can you check that the current CDN-fetching is robust enough? It should not fail and switch to P2P sync on the first fetching error.


Some on AI findings:

> A single peer can force any client to shut down by advertising a fabricated tip

We should aim for the simplest fix for this. needed_blocks_below_all_peer_floors decides "no peer can serve me, this is unrecoverable" purely from locators.latest_locator_height(), which is an unverified peer claim:

if peer_tip > height {
    any_peer_ahead = true;
    if assumed_peer_floor(peer_tip) <= height.saturating_add(1) {
        return false;
    }
}
Block locators are only checked for structural validity — BlockLocators::ensure_is_valid requires checkpoints at multiples of CHECKPOINT_INTERVAL (10,000) and consistency between the last checkpoint and the last recent height. Nothing bounds the claimed height, and update_peer_locators explicitly documents that it does not cross-check against other peers. An attacker copies the honest checkpoint chain up to the real tip and appends ~10 fabricated checkpoints plus 100 fabricated recents.

The consequences chain together:

set_greatest_peer_height takes the max over all peers, and MAX_BLOCKS_BEHIND is 1, so one liar flips the node to Syncing — satisfying the !self.is_block_synced() guard.
The liar's assumed_peer_floor sits ~99,000 blocks above us, so it is excluded from candidate_locators; honest peers at exactly our height fail the peer_tip > current_height test. No candidates, no batches, no outstanding requests.
any_peer_ahead is true and no peer clears the floor test, so unservable_since is armed. After STUCK_SYNC_SHUTDOWN_THRESHOLD (5 min) the node calls stop_with_failure() and exits non-zero.
In a healthy network the timer resets whenever an honest peer advertises a tip strictly above ours, which happens every couple of seconds, so this is hard to sustain. The dangerous case is exactly the one you least want it in: a network halt or stall, when no honest peer is ever ahead. Then one connected peer can walk every client off the network in five minutes, and because the exit code is non-zero, supervisors restart into CDN sync and the attacker repeats it — a sustained crash loop. Clients accept inbound connections from anyone, so this needs no privileges. Validators are exposed too (they call prepare_block_requests through the BFT sync loop), but only to a Byzantine committee member, since gateway locators come from authorized validators.

The PR description acknowledges this ("a lying peer could push idle nodes to exit") and argues the network will never halt meaningfully. I don't think that's a safe assumption for a mechanism whose failure mode is mass shutdown. Concrete hardening options: require the stuck condition to be corroborated by a quorum or a majority of connected peers rather than any single one; derive the floor from the median or minimum of peer tips ahead of us instead of each peer's own claim; and require the node to have made no sync progress at all for the whole window, not merely that the flag stayed armed.


> Shutdown-as-recovery has no backoff and no bound on repeats

Indeed instead of trying and waiting for 5 minutes could we fail immediately and loudly when:
1. The node is far behind, and CDN sync failed.
2. The node is far behind, and have --nocdn turned on.

PR #4365, opened 2026-08-04
generate_ledger.sh: add missing network_name

eranrund:
## Motivation

I am trying to run some of the benchmark tests locally and ran into an issue with the `generate_ledger.sh` script not working, the change here appears to fix it and I was able to successfully generate the ledger zip file using it.

## Test Plan

Ran locally and saw that the output zip file got produced.

## Documentation

N/A

## Backwards compatibility

N/A


vicsn (approved):
LGTM!

PR #4368, opened 2026-08-04
bft: validate certificate responses before clearing pending requests

cbeck88:
## Motivation

this was marked as a TODO, and it is necessary because otherwise a maliciously crafted response that contains the certificate ID but invalid signatures would be broadcast by us to our peers.

## Test Plan

adds a test confirming that invalid certificate responses do not clear the pending request


vicsn (approved):
LGTM! This introduces a non-trivial but IMO likely only minor performance hit. In the future we could somehow cache that a certificate is already checked.

ljedrz (approved):

PR #4369, opened 2026-08-05
fix(bft): never move the signed-proposal cache backwards

cbeck88:
## Motivation

Fix a bft bug:

For correctness we must endorse at most one batch per author per round. The `signed_proposals` cache is the only record of what this primary has signed, so that guarantee is only as good as the cache.

The entry was overwritten whenever the cached round differed from the incoming round, in either direction. Handlers for the same author run concurrently, so one for an older round can be stalled while a handler for a newer round completes; the stalled handler would then roll the entry back and erase the record of the newer signature. A conflicting proposal for that round would be signed a second time, yielding two signatures over different batch IDs in one round -- enough to build conflicting certificates and fork the DAG.

Instead, only ever advance the cached round. For testability against this type of race, the update moves into `Primary::cache_signed_proposal`, which under a single write lock either caches the proposal or reports that it must not be signed.

## Test Plan

A unit test ensures that the race does not result in moving the cache backwards.

## Backwards compatibility

I believe that this is pure bug fix and is backwards compatible.

### DIFF
--- a/node/bft/src/primary.rs
+++ b/node/bft/src/primary.rs
@@ -1006,24 +1006,9 @@ impl<N: Network> Primary<N> {
         // Note: Due to the need to sync the batch header with the peer, it is possible
         // for the primary to receive the same 'BatchPropose' event again, whereby only
         // one instance of this handler should sign the batch. This check guarantees this.
-        match self.signed_proposals.write().0.entry(batch_author) {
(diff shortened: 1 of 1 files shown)


cbeck88:
rebased to resolve conflict in the test code

vicsn (approved):

PR #4370, opened 2026-08-05
fix(bft): clear the proposed batch under a single lock

cbeck88:
## Motivation

In some recent PRs, we attempted to fix race conditions that affect the BFT logic, where a condition is checked, and then acted on, but no lock ensures that by the time we act, the condition is still true. In some cases this led to a safety issue. We searched for additional instances of this pattern.

In the instance we found, `check_proposed_batch_for_expiration` read the proposed batch under one lock to decide whether it had expired, then took a second lock to clear it.

i don't know that this leads to a consensus safety issue, it's just the result of looking for other cases of the pattern "check and act with no lock covering both of them" in the bft code

## Test Plan

we added a unit test covering the behavior, although it's not quite as good as a regression test.

## Backwards compatibility

I believe this is completely backwards compatible.

### DIFF
--- a/node/bft/src/primary.rs
+++ b/node/bft/src/primary.rs
@@ -1622,21 +1622,31 @@ impl<N: Network> Primary<N> {
     }
 
     /// Checks if the proposed batch is expired, and clears the proposed batch if it has expired.
+    ///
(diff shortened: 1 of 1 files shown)


vicsn (approved):
Nice find

PR #4372, opened 2026-08-05
Google cloud config changes to get the benchmark GHA to run

eranrund:
With these changes we are now able to run the benchmark action: https://github.com/aleo-labs/snarkOS-test/actions/runs/31037254775


eranrund:
Oops, this should've gone to the fork.

PR #4375, opened 2026-08-06
perf(events): write event frames in place instead of copying twice

cbeck88:
## Motivation

While investigating a report that connections between peer validators reserve enormous buffers, we looked at how serialization of `Data::Object` works and noticed that the `Event` codec makes two needless copies of the data it is working on: once as a temporary byte buffer for the input, and once as the result of a `freeze` that splits data off the `BytesMut` we are supposed to be appending to.

This uses a different strategy that requires no temporary allocation. We write a placeholder for the frame's length prefix, serialize the event straight into the destination buffer using the trait functions available in this context, and then backpatch the length once it is known.

### Note on the non-empty-buffer bug

This PR originally also fixed a bug in the same routine: it assumed the `BytesMut` it was handed is empty, and would corrupt the stream if it was not. **That bug has since been fixed independently in #4354**, which landed while this PR was open — staging now captures `let buffered = dst.len()` and uses `split_off(buffered)` rather than `split_to(dst.len())`.

So the bugfix half of this PR is no longer needed, and what remains is the performance change. The rebase takes the approach here for the encoder, since it handles the non-empty buffer *and* avoids the copy, and keeps the proptest #4354 added (`events_encoded_into_a_shared_buffer_stay_separate`) in preference to the narrower fixed-input regression test this PR had added for the same property.

Note the same bug is **still present** in three other copies of this routine, which were not covered by #4354:

- `node/router/messages/src/helpers/codec.rs:62`
- `node/src/bootstrap_client/codec.rs:64` and `:80` (plus the `MessageOrEvent` encoder below them)

All three still use `dst.split_to(dst.len())`. That is worth fixing separately — it is reachable whenever a frame is left partially flushed, i.e. under write backpressure from a slow peer, and the router codec is on the client/prover path. Filing that as its own PR rather than growing this one.

## Test Plan

Unit test and regression test, plus a few criterion benchmarks added to understand the significance.

Verified on the rebased branch:

- `cargo test -p snarkos-node-bft-events` — 28 passed, including `event_roundtrip` and `events_encoded_into_a_shared_buffer_stay_separate`
- `cargo clippy -p snarkos-node-bft-events --all-targets -- -D warnings` — clean
- `cargo +nightly fmt -p snarkos-node-bft-events -- --check` — clean

Benchmarks can be run with `cargo bench -p snarkos-node-bft-events`. They separate the two costs of encoding an event: serializing the payload (paid only for `Data::Object`, and dominant when it applies) and copying the encoded bytes into the connection's write buffer (paid either way, and all that remains once the payload is a `Data::Buffer`).

## Backwards compatibility

No `ConsensusVersion` guard needed. The wire format is unchanged — the encoder writes the same little-endian `u32` length prefix that `LengthDelimitedCodec` was writing before, and the decoder is untouched. `FrameLength` and the explicit `length_field_type::<FrameLength>()` exist to keep the two sides from drifting now that the prefix is written by hand; `event_roundtrip` covers the agreement.

## Notes for the reviewer

- This has been rebased onto current staging now that #4354 has merged, as requested.
- Re: dropping `LengthDelimitedCodec` entirely — agreed the frame format is now defined encoder-side while the decoder still delegates, and that asymmetry is worth removing. I would prefer to do it separately: the decoder is the harder half (incremental framing across partial reads), and taking ownership of it should come with a proptest that feeds input a byte at a time, which is the property `LengthDelimitedCodec` currently provides for free.

### DIFF
--- a/node/bft/events/src/helpers/codec.rs
+++ b/node/bft/events/src/helpers/codec.rs
@@ -36,6 +36,13 @@ use tracing::*;
 const MAX_HANDSHAKE_SIZE: usize = 1024 * 1024; // 1 MiB
 /// The maximum size of an event that can be transmitted in the network.
 const MAX_EVENT_SIZE: usize = 256 * 1024 * 1024; // 256 MiB
+/// The integer type of a frame's length prefix.
+///
+/// The decoder is configured with this via `length_field_type`, and the encoder writes the prefix
+/// as one of these, so the two cannot disagree about the framing.
+type FrameLength = u32;
+/// The width of a frame's length prefix, in bytes.
+const LENGTH_FIELD_LEN: usize = size_of::<FrameLength>();
 
 /// The codec used to decode and encode network `Event`s.
 pub struct EventCodec<N: Network> {
@@ -54,7 +61,11 @@ impl<N: Network> EventCodec<N> {
 impl<N: Network> Default for EventCodec<N> {
     fn default() -> Self {
         Self {
-            codec: LengthDelimitedCodec::builder().max_frame_length(MAX_EVENT_SIZE).little_endian().new_codec(),
+            codec: LengthDelimitedCodec::builder()
+                .max_frame_length(MAX_EVENT_SIZE)
+                .length_field_type::<FrameLength>()
+                .little_endian()
+                .new_codec(),
             _phantom: Default::default(),
         }
     }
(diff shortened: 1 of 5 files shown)


cbeck88:
Later it was determined that this code was copy pasted:

  - node/bft/events/src/helpers/codec.rs:62 — fixed
  - node/router/messages/src/helpers/codec.rs:62 — not fixed
  - node/src/bootstrap_client/codec.rs:64 and :80 — not fixed (two encoders, and there's a third for MessageOrEvent just below)

node/bft/events/src/helpers/codec.rs
```diff
@@ -43,7 +50,11 @@ impl<N: Network> EventCodec<N> {
 impl<N: Network> Default for EventCodec<N> {
     fn default() -> Self {
         Self {
-            codec: LengthDelimitedCodec::builder().max_frame_length(MAX_EVENT_SIZE).little_endian().new_codec(),
+            codec: LengthDelimitedCodec::builder()
+                .max_frame_length(MAX_EVENT_SIZE)
+                .length_field_type::<FrameLength>()
+                .little_endian()
+                .new_codec(),
```
ljedrz, line 68: note: this doesn't change anything, as `u32` is the default value

node/bft/events/src/helpers/codec.rs
```diff
@@ -53,15 +64,41 @@ impl<N: Network> Encoder<Event<N>> for EventCodec<N> {
...
+        //
+        // The length is not known until the event has been serialized, so the alternative is to
+        // serialize the event elsewhere and copy the result in once its size is known. Writing the
+        // event straight into `dst` and filling in the length afterwards avoids that copy.
+        //
+        // Note that `dst` is the connection's write buffer and is not necessarily empty: it still
+        // holds any frames that have not been flushed yet. Everything below is therefore relative
+        // to where this frame starts, not to the start of the buffer.
+        let frame_offset = dst.len();
+        dst.extend_from_slice(&[0u8; LENGTH_FIELD_LEN]);
```
ljedrz, line 88:
```suggestion
        dst.put_u32_le(0);
```

node/bft/events/src/helpers/codec.rs
```diff
@@ -53,15 +64,41 @@ impl<N: Network> Encoder<Event<N>> for EventCodec<N> {
...
+        if frame_len > self.codec.max_frame_length() {
+            dst.truncate(frame_offset);
+            return Err(std::io::Error::new(std::io::ErrorKind::InvalidInput, "frame size too big"));
+        }
 
-        let serialized_event = dst.split_to(dst.len()).freeze();
+        // Fill in the length prefix, in the same little-endian `FrameLength` framing that
+        // `self.codec` is built to decode. The `event_roundtrip` test covers that agreement.
+        let frame_len = FrameLength::try_from(frame_len)
+            .map_err(|_| std::io::Error::new(std::io::ErrorKind::InvalidInput, "frame size too big"))?;
```
ljedrz, line 109: note: this cannot fail, as `FrameLength` is `u32` and `frame_len` was just checked against `self.codec.max_frame_length()`, which is `u32::MAX`

ljedrz (commented):
The perf change looks right, and the non-empty-buffer bug potential is real (already tackled in https://github.com/ProvableHQ/snarkOS/pull/4354 btw).

However, at this point we might as well drop the `LengthDelimitedCodec` altogether, as the frame format effectively ends up redefined on the `Encoder`'s side, with the `Decoder` still being bound to the `LengthDelimitedCodec` object.

Lastly, I'd like these changes to go in after the Noise-XX handshake, which is much larger. This PR will need to be rebased, but it shouldn't be too much hassle in that order.

node/bft/events/src/helpers/codec.rs
cbeck88: yes -- i just wanted it to be explicitly using the same type in both the encoder and decoder, so the reader doesn't need to know the defaults

not clear that it's necessary

node/bft/events/src/helpers/codec.rs
cbeck88: i think if we're trying to be parametric over FrameLength this is moving in the opposite direction

cbeck88: being parametric seems better from a defensive programming point of view, we don't need to optimize for number of symbols

node/bft/events/src/helpers/codec.rs
cbeck88: if you'd prefer we can panic here, but it's not ultimately that significant.
max_frame_length is actually 256 MB

cbeck88:
rebased

ljedrz (dismissed):
LGTM 👍

vicsn (dismissed):
LGTM, just a merge conflict left

vicsn (approved):

PR #4376, opened 2026-08-07
try to improve developer build and test cycles

cbeck88:
Some of the settings in our Cargo.toml and .cargo/config.toml make for a painful developer experience around build and test cycles.

1. For some reason, the test profile has "lto=thin".

I have never seen this in any project. Normally, LTO is only used in release mode, because recompiling after a code change is very difficult when lto has been used.

This seems to have been turned on as part of a larger commit by howard in 2020, with no comment on why.

https://github.com/ProvableHQ/snarkOS/commit/9a1e167174

The pattern that I have seen used successfully is, if some tests are too slow to run without optimizations, then they are gated on `cfg(debug_assertions)` in the code, and then they will only run when you invoke `cargo test --release`, which will build and run the tests in the release profile. Then CI will have one job that does `cargo test` and one that does `cargo test --release`.

The benefit of this is, `cargo test` can build and rebuild fast when you are changing the code. The tests that need optimizations to work can still get them, without forcing all test runs to always be optimized as if for release mode.

What I'd like to do is get rid of this `lto=thin` stuff as a part of `cargo test`. Then, if CI starts failing with tests that are too slow to pass, mark those to only run in `cargo test --release`, and also test that invocation in CI.

2. No effort has been made to enable the use of modern CPU instructions.

In C++ projects, it's extremely common to use `-march=native` when building and running tests. This flag tells the compiler to identify the CPU we are running on, and when generating code, allow the optimizer to use any instructions that are legal for that architecture.

In rust, unfortunately cargo doesn't really have an equivalent flag, or an easy way to run the only the tests using this. Your main option is to use -C target-arch=native as a RUSTFLAGS env value. However, RUSTFLAGS itself is somewhat fraught because there are many ways that env values can be clobbered, and they interact with the .cargo/config.toml file in somewhat unexpected ways.

Instead the default behavior of cargo is that when you run the tests on a typical x86-64 machine, it does not assume that sse2 or AVX2 is present. Even 7 years ago, it would assume that you might be targetting a CPU from 7 years before that.

For a lot of projects it doesn't truly matter, but for crypto projects that make heavy use of elliptic curves and finite field math etc., these vector instructions can become extremely important for performance, especially when using twisted edwards curves where these operations become part of serializing and deserializing curve points.

Putting skylake here is pretty safe because, the x86-64 machines that don't have those instructions are museum pieces today -- none of the developers or CI machines or cloud machines will fail to have these instructions. Actually we put this stuff in the mobilecoin .cargo/config.toml 7 years ago and we never experienced a crash because of it or got a report from someone saying that the program crashed with SIGILL (which is what happens if the CPU reads an instruction it doesn't support.)

https://github.com/mobilecoinfoundation/mobilecoin/blob/05cb699f8f4cc1bc21186392545820c5b38408db/.cargo/config.toml#L8

My hope is that, this will counteract / help pay for any reduction in overall performance of tests due to turning off LTO.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -2,7 +2,11 @@
 rustflags = [
     # LLD is generally faster and might be the default for Rust in the future.
     # See here: https://github.com/rust-lang/rust/issues/71515
-    "-C", "link-arg=-fuse-ld=lld",
+    #
+    # skylake CPU family includes most SSE and AVX2 instructions from about 10 years ago
+    # without this, optimization cannot use almost any vector instructions, unless
+    # you explicitly set target-cpu yourself when building.
+    "-C", "link-arg=-fuse-ld=lld", "-C", "target-cpu=skylake"
 ]
 
 [env]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -420,7 +420,7 @@ incremental = true
 
 [profile.test]
 opt-level = 2
-lto = "thin"
+lto = "off"
 incremental = true
 debug = true
 debug-assertions = true


vicsn (approved):
TIL where SIGILL comes from. We regularly have a random CI job failing because of it.

Can you also make the change for snarkVM?

vicsn:
CI runtimes: https://github.com/ProvableHQ/snarkOS/pull/4379

ljedrz (commented):
The change in flags is an improvement, but I think it would be best if we put `"target-cpu=native"` here, and apply `skylake` in CI jobs specifically.

As for `lto` settings, let's see what the CI says (though it will have been influenced by the extra flags too).

cbeck88:
> The change in flags is an improvement, but I think it would be best if we put "target-cpu=native" here, and apply skylake in CI jobs specifically.

The thing about putting target-cpu=native here is, nothing in this .cargo/config.toml is test specific. So that would start affecting any actual releases we build.

If you put target-cpu=native when you build binaries for production, then what build you can get vary based on what actual machine you build on, and if that doesn't match up with the person who actually runs it, then they get crashes.

What if we invert your suggestion -- in CI use RUSTFLAGS="-C target-cpu=native" and see if that works? I need to read the rules again for what cargo does if the same thing is set in RUSTFLAGS env and .cargo/config.toml. 

@ljedrz

ljedrz:
We don't provide binary releases though, do we? The users are instructed to run `cargo install`, which builds locally.

As for the CI, `skylake` is preferable, as machines may differ between runs.

cbeck88:
if someone is doing `cargo install` in a Dockerfile, and then shipping that docker image off somewhere, they can be bitten by `target-cpu=native` (maybe). if they are just doing exactly what you say, then it should work fine.

the best case for target-cpu=native is when you know that you are only going to be running the artifact on this specific machine, hence like testing scripts

cbeck88:
@vicsn 

the target-cpu=native was already done in snarkVM: https://github.com/ProvableHQ/snarkVM/commit/764617f121

i'm still going to open a PR there to turn off lto=thin

cbeck88:
> As for the CI, skylake is preferable, as machines may differ between runs.

It appears to me that, you guys have already been using cpu=native in CI in the snarkVM repo, and it doesn't appear to cause problems in your CI. that repo is providing libraries mainly though.

Let's start with skylake in this repo, because it creates a measurable improvement in some benchmarks I was running.

cbeck88:
> TIL where SIGILL comes from. We regularly have a random CI job failing because of it.
> 
> Can you also make the change for snarkVM?

@vicsn can you tell me more about this? Is it in snarkOS or snarkVM? I wonder if it could be related to the target-cpu=native stuff in snarkVM config toml.

PR #4377, opened 2026-08-07
Poll RocksDB internal metrics from a dedicated background task

vicsn:
## Summary

- Adds a dedicated RocksDB metrics polling task for validators and clients (every 15 seconds when the `metrics` feature is enabled).
- Updates the snarkVM pin to `79949e163437d61c4f4b2ee577bd709e5263d921` (snarkVM staging, includes merged [snarkVM #3296](https://github.com/ProvableHQ/snarkVM/pull/3296)).
- Enables `snarkvm/metrics` and `snarkvm/rocks` on the node `metrics` feature so `export_rocksdb_metrics()` is available at compile time.
- Rebases onto staging.

This is the snarkOS companion to [snarkVM #3296](https://github.com/ProvableHQ/snarkVM/pull/3296). When metrics are enabled, `spawn_rocksdb_metrics_polling()` runs as a tracked node background task and polls cheap in-memory RocksDB property counters. This is separate from the auto-checkpoint loop, so metrics export works even when `--auto-db-checkpoints` is not configured.

## Test plan

- [x] `cargo check -p snarkos-node --features metrics`
- [ ] Deploy with metrics enabled and confirm Prometheus gauges such as `snarkvm_rocksdb_compaction_pending` and `snarkvm_rocksdb_total_sst_files_size_bytes` update every ~15 s

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5183,8 +5183,7 @@ dependencies = [
 [[package]]
 name = "snarkvm"
 version = "4.9.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "dea37705e6ba5f11a8da0f61393ba14aa320fcaf9fb410dff835c396ba20283a"
+source = "git+https://github.com/ProvableHQ/snarkVM.git?rev=4c5509f089a49a01c66c166b8abfd9e54161fd7e#4c5509f089a49a01c66c166b8abfd9e54161fd7e"
 dependencies = [
  "anyhow",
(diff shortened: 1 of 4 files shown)


ljedrz (commented):
Not a huge fan of the callsite, but probably neither are you. That being said, it'll work, and it's a tiny diff, so LGTM.

Note: this shouldn't be merged until the snarkVM-side PR is in, so not approving yet.

ljedrz (approved):

PR #4378, opened 2026-08-07
Refactor/outbound bytes

cbeck88:
## Motivation

This is a series of changes to the way that the snarkOS validator handles outbound message queues, motivated by several reports. We have excerpted part of one of them, which itself mentions several individual issues.

> A snarkOS BFT validator sizes each peer's outbound message queue at `2 × MAX_GC_ROUNDS × LATEST_MAX_CERTIFICATES × MAX_TRANSMISSIONS_PER_BATCH = 400,000 slots`, and there is no per-peer rate limit on inbound `TransmissionRequest` events (the code carries an explicit TODO where that check belongs). A Byzantine committee validator learns valid transmission IDs from the public gossip it already receives, opens a connection on which it deliberately stops reading (TCP zero-window), and floods known-ID `TransmissionRequests`; each one makes the victim clone a full Transmission into that 400,000-slot queue, which the writer cannot drain because the 5-second flush timeout returns io::ErrorKind::TimedOut, which is not in the node's fatal-error set, so the connection is never torn down. 

Note that this also touches on a tracking issue: https://github.com/ProvableHQ/snarkOS/issues/4140

One of the first points of interest in this investigation is, the 400,000 slot queue does not hold simple bytes, but rather full deserialized objects. Instead the "codec" is used to actually serialize messages to the socket, and only at the last moment. 

(See `node/tcp/src/protocols/writing.rs`: the situation is actually worse. The messages are not even enqueued themselves, and to avoid a queue with parametric types, std::Any is used to box these types, type erasing everything about them except that they have a codec, and putting that in the queue, then downcasting it on the other side.)

However, this makes broadcasts very inefficient. What one would normally expect is, when I want to send one message to all of my peers, I serialize it to bytes once, then those bytes are enqueued to mpsc queues and eventually flushed to the socket. The `tokio::util::Bytes` object is designed exactly for this type of situation -- these bytes handles are ref counted, and can be cloned 40 times and put in 40 queues without copying the underlying data or introducing locks and contention.

In many programs, serializing before or after enqueuing, or serializing 40 times vs once, might not make a large performance impact. However, in crypto projects, it's common that the data contains elliptic curve points, especially more exotic ones like the twisted edwards curve, used here and in Ristretto and a number of other adjacent projects. Twisted edwards curve points generally are implemented as having a "compressed" and an "uncompressed" representation. The uncompressed representation is faster for actually using it, but when you need canonical bytes on the wire, you have to compress it. And unfortunately compressing it requires finite field arithmetic, so serialization of a large collection of these can be on the order of hundreds of microseconds.

So, we actually really want to be careful to try to serialize things once if possible. And outbound queues that contain large uncompressed things that are cloned and then serialized over and over again, should be rearranged so that serialization happens once upfront, and then tokio::Bytes are what is distributed to workers.

In node/tcp/src/protocol/writing.rs, we changed the implementation so that instead of `std::Any<Message>` (the type erased thing-with-codec), we simply put tokio Bytes in the queue. The broadcast loop gets the bytes first and reuses them for each peer. The unicast obtains bytes and then passes them.

Additionally, we make sure the timeout guard on writing covers both asynchronous operations -- writing and flushing, which addresses parts of the original report. This fix appears in `async fn write_to_stream`

It turns out that this broadcast loop is duplicated, however, in the gateway.rs code. This seems to be because we decided to implement outbound rate limiting -- we choose to rate limit ourselves if we are sending too many things to a peer (rather than the peer imposing a rate limit on us).

Therefore, we apply a very similar fix to this more complicated loop which includes the rate limit checks. (This also resolves a TODO).

To make it manageable, we introduce an enum which keeps track of, for a given Event, which rate limit category should apply to it. Then we abstract out the logic that enforces the rate limits. (This code is also a bit strange, in that, when we discover we have exceeded a rate limit, rather than skipping messages and moving on, we sleep for 10 milliseconds and check again. I'd be interested to discuss the rationale of this at some later time.)

One specific point for gateway is that, if `ErrorKind::TimedOut` is seen enough times, we tear down the peer connection.

Finally, we adjust the calculation for the gateway outbound queue size. The 400,000 slot size number came from using max_gc_rounds which was very conservative. Instead, we calculate based on the max fetch timeout -- if a message cannot be delivered from our outbound queue to the peer before that time, they will have dropped it anyways. This is compared with the max batch delay, which creates a lower bound on how many batches should have been proposed before timeouts occur. Within this number of rounds, it's still possible that every transmission and certificate is requested, but the total with this estimate is ~12,000 rather than 400,000, and the slots contain tokio::Bytes objects now, rather than Data::Object.

The change is organized as 4 commits, and there are really several ideas here, but they are related and share motivation.
It might be easier to review as a series of PRs instead. Let me know what you think.

Logical next steps include:
* The Router object, like the gateway, has its own bespoke multicast loop and per-peer rate limiting. We should apply the same changes there (or create an abstraction to avoid code duplication)

## Test Plan

Running e2e tests and running stress tests. The e2e tests are passing and it still needs stress test runs. This is still draft as I continue to test it, but I'm interested in feedback.

## Backwards compatibility

I believe this is backwards compatible and just makes the node more efficient, and more resilient against timeouts when sending to a peer. If the additional disconnects due to timeouts happen often, the worst case is that the peer just reconnects, but it needs testing to ensure that we are giving honest nodes enough slack.

### DIFF
--- a/node/bft/src/gateway.rs
+++ b/node/bft/src/gateway.rs
@@ -24,6 +24,7 @@ use crate::{
     helpers::{Cache, PrimarySender, Storage, SyncSender, WorkerSender, assign_to_worker},
     spawn_blocking,
 };
+use bytes::Bytes;
 use smol_str::SmolStr;
 use snarkos_account::Account;
 use snarkos_node_bft_events::{
@@ -117,6 +118,31 @@ const CONNECTION_ATTEMPTS_SINCE_SECS: i64 = 10;
 
 /// The amount of time an IP address is prohibited from connecting.
 const IP_BAN_TIME_IN_SECS: u64 = 300;
+
+/// Which of a peer's outbound budgets an event is charged against.
+///
+/// This is derived from the event before it is serialized, because serializing consumes the event
+/// and the resulting frame no longer says what kind of event it holds.
+#[derive(Clone, Copy)]
+enum OutboundLimit {
+    /// Certificate requests and responses.
+    Certificate,
+    /// Transmission requests and responses.
+    Transmission,
+    /// Everything else, charged against the peer's general event budget.
+    Event,
+}
+
+impl OutboundLimit {
+    /// Returns the budget the given event is charged against.
+    fn of<N: Network>(event: &Event<N>) -> Self {
+        match event {
+            Event::CertificateRequest(_) | Event::CertificateResponse(_) => Self::Certificate,
+            Event::TransmissionRequest(_) | Event::TransmissionResponse(_) => Self::Transmission,
+            _ => Self::Event,
+        }
+    }
+}
 
 /// Part of the Gateway API that deals with networking.
 /// This is a separate trait to allow for easier testing/mocking.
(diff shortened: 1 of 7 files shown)


ljedrz:
Note in advance: we shouldn't be modifying the `tcp` modules; we've already slightly diverged from the upstream it was vendored from, and it is much more polished - anything touching it would most likely be duplicating preexisting work.

Feel free to port from the [upstream](https://github.com/ljedrz/pea2pea), though, but this should be done in a dedicated PR.

cbeck88:
i didn't realize this was vendored, thank you for explaining

cbeck88:
i'm going to try to reduce the footprint of this to the minimum and avoid touching the node/tcp crate. however i think we should do a 1 line bugfix there.

will likely open a second PR and then close this

PR #4381, opened 2026-08-08
fix(tcp): disconnect peers that stop accepting writes

cbeck88:
## Motivation

Make outbound peer writing more resilient to timeouts / backpressure from the peer.

This extracts a minimal part of https://github.com/ProvableHQ/snarkOS/pull/4378

1. `write_to_stream` needs to put both the `feed` call and the `flush` call under the timeout.
2. Even when the timeout did fire, it produced io::ErrorKind::TimedOut, which was absent from `fatal_io_errors`, so the writer logged it and carried on. Such a peer therefore timed out every TIMEOUT forever and was never disconnected, pinning every message queued for it for as long as it kept the socket open.

(Note: after review, only part 1 was done in this PR, and 2 was done separately )

Per comments in #4378 we should not change node/tcp because it is essentially an old vendored version of pea2pea. However, I want to get these two bug fixes in because it's a very small change in terms of LOC and I don't know how I can prioritize updating the vendored version of pea2pea at this moment.

## Test Plan

Tests ensure that backpressure results in us dropping the connection, as intended.

## Backwards compatibility

I believe this is backwards compatible because it doesn't change the contents of any messages, it just helps nodes to avoid using excessive resources when the peer is not consuming messages. However, a stress test will help us determine if this causes regressions via frequent disconnects.

### DIFF
--- a/node/tcp/src/protocols/writing.rs
+++ b/node/tcp/src/protocols/writing.rs
@@ -57,9 +57,12 @@ where
         1024
     }
 
-    /// The maximum time (in milliseconds) allowed for a single message write to flush
-    /// to the underlying stream before the connection is considered dead.
-    const TIMEOUT: Duration = Duration::from_secs(5);
+    /// The maximum time allowed for a single message write, both encoding it and flushing it to
+    /// the underlying stream, before the connection is considered dead.
+    ///
+    /// This has to leave room for encoding the largest message we can produce, as blocks can
+    /// currently be quite large.
+    const TIMEOUT: Duration = Duration::from_secs(10);
 
     /// The type of the outbound messages; unless their serialization is expensive and the message
     /// is broadcasted (in which case it would get serialized multiple times), serialization should
@@ -187,12 +190,17 @@ impl<W: Writing> WritingInternal for W {
         message: Self::Message,
         writer: &mut FramedWrite<A, Self::Codec>,
     ) -> Result<usize, <Self::Codec as Encoder<Self::Message>>::Error> {
-        writer.feed(message).await?;
-        let len = writer.write_buffer().len();
-        // Guard against write starvation
-        match timeout(W::TIMEOUT, writer.flush()).await {
-            Ok(Ok(())) => Ok(len),
-            Ok(Err(e)) => Err(e),
+        // Guard against write starvation. `feed` is covered as well as `flush`, as `FramedWrite`
+        // flushes from within `feed` once its buffer is over the backpressure boundary, so a peer
+        // that has stopped reading blocks `feed` too.
+        let write = async {
+            writer.feed(message).await?;
+            let len = writer.write_buffer().len();
+            writer.flush().await?;
+            Ok(len)
+        };
+        match timeout(W::TIMEOUT, write).await {
+            Ok(result) => result,
             Err(_) => Err(io::Error::new(io::ErrorKind::TimedOut, "write timed out")),
         }
     }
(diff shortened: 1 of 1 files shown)


node/tcp/src/helpers/config.rs
```diff
@@ -82,7 +82,7 @@ impl Default for Config {
             listener_ip: default_ip(),
             desired_listening_port: None,
             allow_random_port: true,
-            fatal_io_errors: vec![ConnectionReset, ConnectionAborted, BrokenPipe, InvalidData, UnexpectedEof],
+            fatal_io_errors: vec![ConnectionReset, ConnectionAborted, BrokenPipe, InvalidData, UnexpectedEof, TimedOut],
```
ljedrz: the entire `fatal_io_errors` should be removed (can be done in a separate PR that could be merged first)

node/tcp/src/protocols/writing.rs
ljedrz: this needs to be increased in order to cover the encoding too

ljedrz (changes requested):
The fix is sound, but we need to bump the timeout and get rid of `fatal_io_errors` first.

node/tcp/src/helpers/config.rs
cbeck88: claude opus 5 suggests that removing `fatal_io_errors` may cause other problems:

> The argument is strong and this PR is itself the evidence: the allowlist was incomplete, TimedOut was missing, and the consequence was peers pinned indefinitely. Any such list will keep
  being incomplete. Deleting it removes a bug class rather than one bug.
>
>  But ErrorKind is doing double duty in this codebase — it carries both "the transport failed" and "the codec rejected a message," and those have different correct responses. Two concrete sites that are not currently fatal:
> ...
>   - node/router/messages/src/peer_response.rs:39 (write path) — InvalidInput for "Too many peers". That's our own encoding bug, and the connection is perfectly healthy: start_send is
  synchronous, so a failed encode puts no bytes on the wire. After removal we'd tear down a good connection because we built a bad message. That one looks straightforwardly wrong.

I'm still trying to make sense of this, but I wonder if we should not do the smallest change that will fix the bug we are currently working on (make TimedOut fatal) and investigate the fallout of removing `fatal_io_errors` completely in separate work, since that is a larger change.

node/tcp/src/protocols/writing.rs
cbeck88: imo, it should not take 5 seconds to encode a message

this thread should not be doing compute heavy work because it is an io thread and blocking here can stall the executor

in https://github.com/ProvableHQ/snarkOS/pull/4382 we are attempting to make sure that outbound messages are always serialized (converted to bytes) before they are queued and sent to codec.

if that lands first does this alleviate your concern?

ljedrz: This is just out of caution, note that we can currently technically produce quite large blocks; this will be void once we have dedicated sync streams or chunking.

node/tcp/src/helpers/config.rs
ljedrz: It should be quite easy; this setup made sense in the past, but these days, `TcpStream` read errors are all fatal, so it's no longer useful.

Smallest change is one thing, but we don't want to invest in an obsolete setup either.

ljedrz: That being said, I have the most context on this code, happy to introduce those changes tomorrow if you'd prefer.

cbeck88: thats fine with me

node/tcp/src/protocols/writing.rs
cbeck88: Relatedly, can you take a look at my comment on this issue and let me know if you agree? I would like us to get aligned on such a change as I'm proposing before any work happens. 

https://github.com/ProvableHQ/snarkOS/issues/4388#issuecomment-5244510536

I won't have time or expectation to do any implementation work until end of next week, but if there are good reasons not to do that it's helpful for me to find out sooner

cbeck88:
@ljedrz this is rebased and the timeout bumped to 10s

ljedrz (approved):
LGTM; we'll be able to decrease the timeout once the frame size is more bounded.

PR #4382, opened 2026-08-08
Fix/gateway outbound queues

cbeck88:
## Motivation

This is another redeveloped piece of https://github.com/ProvableHQ/snarkOS/pull/4378

The goal is to try to make outbound broadcast more efficient. There are two parts:

1. Instead of cloning a data object and enqueuing it for all of our peers, and then serializing it at the socket for each of our peers (doing any required field operations 40 times), instead do serialization once and convert to Bytes right before we enqueue it to each of the peers. The peers then are cloning the ref counted bytes handle, which reduces the memory footprint in the node significantly as well. This addresses a TODO comment that was already in the code.

2. Reduce the number of slots in the outbound queue per peer.

## Test Plan

This seems like a pretty safe change, since we should be greatly reducing the amount of CPU the node uses overall. The main question in my mind is whether the `tokio::spawn` just above our `serialize_payload` call needs to be a `spawn_blocking`, since the serialization can be CPU intensive.

I think that end to end testing and stress testing should help us detect any regression that could arise from this.

If the number of slots in the outbound queue is too small, we could drop messages. But for reasons stated in the code comments, I think the smaller number of slots should be adequate, since 400,000 represents a truly enormous queue, in terms of heap footprint.

## Backwards compatibility

I believe that this is backwards compatible because we're not changing any of the messages between nodes, we're reorganizing when it performs serialization -- immediately before fanning out for broadcast, not after fanning out. There was also a code comment that marked this as TODO

### DIFF
--- a/node/bft/events/src/lib.rs
+++ b/node/bft/events/src/lib.rs
@@ -343,15 +343,20 @@ pub mod prop_tests {
         Disconnect,
         DisconnectReason,
         Event,
+        ValidatorsRequest,
         batch_certified::prop_tests::any_batch_certified,
         batch_propose::prop_tests::any_batch_propose,
         batch_signature::prop_tests::any_batch_signature,
+        block_request::prop_tests::any_block_request,
+        block_response::prop_tests::any_block_response,
         certificate_request::prop_tests::any_certificate_request,
         certificate_response::prop_tests::any_certificate_response,
         challenge_request::prop_tests::any_challenge_request,
         challenge_response::prop_tests::any_challenge_response,
+        primary_ping::prop_tests::any_primary_ping,
         transmission_request::prop_tests::any_transmission_request,
         transmission_response::prop_tests::any_transmission_response,
+        validators_response::prop_tests::any_validators_response,
         worker_ping::prop_tests::any_worker_ping,
     };
     use snarkvm::{
@@ -402,11 +407,18 @@ pub mod prop_tests {
         .boxed()
     }
 
+    /// A strategy covering every [`Event`] variant.
+    ///
+    /// Keep this exhaustive. Several properties are asserted over "any event", and a variant that
+    /// is missing here is silently untested rather than failing -- which is how `BlockResponse` and
+    /// `PrimaryPing`, the two largest `Data`-carrying events, went uncovered.
     pub fn any_event() -> BoxedStrategy<Event<CurrentNetwork>> {
         prop_oneof![
             any_batch_certified().prop_map(Event::BatchCertified),
             any_batch_propose().prop_map(Event::BatchPropose),
             any_batch_signature().prop_map(Event::BatchSignature),
+            any_block_request().prop_map(Event::BlockRequest),
+            any_block_response().prop_map(Event::BlockResponse),
             any_certificate_request().prop_map(Event::CertificateRequest),
             any_certificate_response().prop_map(Event::CertificateResponse),
             any_challenge_request().prop_map(Event::ChallengeRequest),
@@ -428,8 +440,11 @@ pub mod prop_tests {
                 any::<Selector>()
             )
                 .prop_map(|(reasons, selector)| Event::Disconnect(Disconnect::from(selector.select(reasons)))),
+            any_primary_ping().prop_map(Event::PrimaryPing),
             any_transmission_request().prop_map(Event::TransmissionRequest),
             any_transmission_response().prop_map(Event::TransmissionResponse),
+            Just(ValidatorsRequest).prop_map(Event::ValidatorsRequest),
+            any_validators_response().prop_map(Event::ValidatorsResponse),
             any_worker_ping().prop_map(Event::WorkerPing)
         ]
         .boxed()
--- a/node/bft/events/src/lib.rs
+++ b/node/bft/events/src/lib.rs
@@ -119,9 +119,52 @@ impl<N: Network> From<DisconnectReason> for Event<N> {
     }
 }
 
+/// Replaces the payload with its serialized form, if it is not already serialized.
+fn serialize_payload<T: FromBytes + ToBytes + Send + 'static>(payload: &mut Data<T>) -> Result<()> {
+    if let Data::Object(object) = payload {
+        *payload = Data::Buffer(object.to_bytes_le()?.into());
+    }
+    Ok(())
+}
+
 impl<N: Network> Event<N> {
     /// The version of the event protocol; it can be incremented in order to force users to update.
     pub const VERSION: u32 = 10;
+
+    /// Serializes the event's [`Data`] payload, if it holds one that is not already serialized.
+    ///
+    /// [`Data`] defers serialization until the event is written to a stream, so an event that is
+    /// cloned for several peers is serialized once per peer, separately, inside each connection's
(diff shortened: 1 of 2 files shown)


node/bft/src/gateway.rs
```diff
@@ -1359,12 +1334,18 @@ impl<N: Network> Transport<N> for Gateway<N> {
     /// Broadcasts the given event to all connected peers.
     // TODO(ljedrz): the event should be checked for the presence of Data::Object, and
     // serialized in advance if it's there.
```
ljedrz: this comment can be removed now

node/bft/src/gateway.rs
```diff
@@ -1359,12 +1334,18 @@ impl<N: Network> Transport<N> for Gateway<N> {
...
-    fn broadcast(&self, event: Event<N>) {
+    fn broadcast(&self, mut event: Event<N>) {
         // Ensure there are connected peers.
         if self.number_of_connected_peers() > 0 {
             let self_ = self.clone();
             let connected_peers = self.connected_peers();
             tokio::spawn(async move {
+                // Serialize the event's payload once, rather than once per recipient.
+                // Each recipient then shares the bytes.
+                if let Err(err) = event.serialize_payload() {
```
ljedrz: this should be done within a blocking task, as serialization can be slow

node/bft/events/src/lib.rs
```diff
@@ -358,4 +401,33 @@ pub mod prop_tests {
         assert_eq!(original.id(), deserialized.id());
         assert_eq!(original.name(), deserialized.name());
     }
+
+    /// Serializing the payload ahead of time must be invisible on the wire, otherwise doing it
+    /// before a broadcast would change what peers receive.
+    #[proptest]
+    fn serialize_payload_preserves_the_encoding(#[strategy(any_event())] original: Event<CurrentNetwork>) {
```
ljedrz, line 466: `any_event` would now need to be extended (`::{PrimaryPing, BlockRequest, BlockResponse, ValidatorsRequest, ValidatorsResponse}`) in order to cover all the `Event`s of interest

ljedrz (changes requested):
Needs a blocking task, a test extension, and I think that the queue restrictions need to be aligned with the `max_cache_*` values in the `Gateway`.

node/bft/events/src/lib.rs
cbeck88: i'm working on this now, but BlockResponse is pretty slow because it requires generating proofs, serializing, deserializing. the test suite is going from 40s to ~220s after this is added.
just double checking that this is necessary, since it was omitted before this revision.

cbeck88: it seems like it's actually about the same speed after i ran the tests again

node/bft/src/gateway.rs
```diff
@@ -1261,7 +1291,32 @@ impl<N: Network> Transport<N> for Gateway<N> {
...
+        // the writer task -- a Tokio worker -- and inside the write timeout. For the large
+        // responses (`BlockResponse` in particular, which can carry `DataBlocks::MAXIMUM_NUMBER_
+        // OF_BLOCKS`) that is a substantial amount of compute on the reactor.
+        //
+        // The check keeps the hop off the common path: most events carry no payload, and a
+        // broadcast has already serialized its payload before fanning out here, so this is a no-op
+        // for every recipient after the first.
+        if event.has_unserialized_payload() {
+            let name = event.name();
+            event = match spawn_blocking!({
```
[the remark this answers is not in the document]
cbeck88: i added a change for this, but it does make the footprint of the PR larger. but it does seem like a good idea overall

node/bft/src/gateway.rs
```diff
@@ -515,6 +545,48 @@ impl<N: Network> Gateway<N> {
...
+    /// This is the admission control for [`Transport::send`], and it deliberately runs *before* the
+    /// payload is serialized. Serializing first would mean doing the expensive work for an event
+    /// that is about to be throttled -- or dropped outright, if the connection's outbound queue is
+    /// full by the time it is enqueued -- which lets a peer that requests large responses pile up
+    /// work in the blocking pool no matter how far behind its own queue is.
+    async fn throttle_outbound(&self, peer_ip: SocketAddr, event: &Event<N>) {
+        /// Waits until the given per-peer counter is back under its limit.
+        macro_rules! wait_until_under_limit {
+            ($cache_map:ident, $interval:expr, $limit:ident) => {{
+                while self.cache.$cache_map(peer_ip, $interval) > self.$limit() {
```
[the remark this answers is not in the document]
cbeck88: i agree, this should be fixed

node/bft/src/gateway.rs
```diff
@@ -1316,55 +1353,64 @@ impl<N: Network> Transport<N> for Gateway<N> {
...
     /// without waiting for the actual delivery; instead, the caller is provided with a [`oneshot::Receiver`]
     /// which can be used to determine when and whether the event has been delivered.
-    async fn send(&self, peer_ip: SocketAddr, event: Event<N>) -> Option<oneshot::Receiver<io::Result<()>>> {
-        macro_rules! send {
-            ($self:ident, $cache_map:ident, $interval:expr, $freq:ident) => {{
-                // Rate limit the number of certificate requests sent to the peer.
-                while $self.cache.$cache_map(peer_ip, $interval) > $self.$freq() {
-                    // Sleep for a short period of time to allow the cache to clear.
-                    tokio::time::sleep(Duration::from_millis(10)).await;
+    async fn send(&self, peer_ip: SocketAddr, mut event: Event<N>) -> Option<oneshot::Receiver<io::Result<()>>> {
```
ljedrz, line 1294:
```suggestion
    async fn send(&self, peer_ip: SocketAddr, event: Event<N>) -> Option<oneshot::Receiver<io::Result<()>>> {
```
nit: for a cleaner API, this `mut` could probably be removed in favor of a rebinding in the conditional serialization site (with an additional `else { event }`)

the same applies to `broadcast`

ljedrz (commented):
LGTM in general, I'd only request that you split off the final 2 commits into their dedicated PR, as they constitute a separate bugfix.

node/bft/src/gateway.rs
cbeck88: the `mut event` part is not part of the API, it is solely telling rust that we will mutate this variable within the body of the function

the API is exactly the same whether you have `event` or `mut event` -- the caller must pass an event by value which we take ownership of.

ljedrz: it's not an external API, yes, and the passing by value point is correct, it may just be considered odd that an object to be sent is mutable upfront; that's why I consider it a nit, it's personal preference

cbeck88: i agree the API is a bit wierd -- usually `fn send(...)` APIs would take a reference to the payload, so that you can do retries without copying the payload each time. but that wouldn't work here because we have to be able to mutate.

cbeck88:
@ljedrz i split the two fixes suggested by veria-dev into a separate branch `fix/outbound-rate-limits`, so this PR should be at the original scope now. we can make a PR for that into staging after this one

ljedrz (approved):

vicsn (approved):

PR #4383, opened 2026-08-08
fix: prevent BFT from certifying batches referencing unmaterializable aborted transmissions

eranrund:
## Motivation

BFT could certify and store a batch that references a transmission ID for which persistent BFT storage holds only an aborted entry and no retrievable bytes. Later, BFT commit cannot materialize those bytes and bails before producing the candidate block, temporarily stalling block production on the affected validator.

The root cause is that `StorageService::contains_transmission(id)` returns `true` for both concrete transmissions **and** aborted transmission IDs, while `get_transmission(id)` returns `None` for aborted IDs. Two paths turned this mismatch into a potential liveness issue:

- **Propose** (`Primary::propose_batch`): the storage-duplicate skip was bypassed for the *first* transmission in a proposal, so an already-aborted transmission sitting at the front of a worker's ready queue could be proposed as a normal transmission. The ledger check does not catch it, because the ledger does not store aborted transmissions.
- **Store** (`BFTPersistentStorage::find_missing_transmissions`): it used `contains_transmission` (true for aborted IDs), so caller-provided bytes for an already-aborted ID were dropped rather than persisted - leaving the certificate uncommittable. Note this also diverged from `BFTMemoryService`, which checks only the concrete map and persists the bytes.

This PR fixes both:

1. `propose_batch` now skips any transmission already recorded in storage (concrete or aborted), with no first-item exception. Proposing an empty batch is already valid and tested, so this is safe.
2. `BFTPersistentStorage::find_missing_transmissions` now checks only the concrete transmissions map, so caller-provided bytes for a previously-aborted ID are collected and persisted (matching `BFTMemoryService`).

## Test Plan

A unit test (`test_propose_batch_skips_aborted_first_transmission`) was added: it queues a worker with only aborted transmissions and asserts propose_batch skips all of them - including the first item, which previously bypassed the storage check and could be proposed.

## Documentation

N/A

## Backwards compatibility

No `ConsensusVersion` guard required. Neither change alters consensus messages, batch/certificate hashes, wire formats, or validation rules - it only affects local proposal selection.

### DIFF
--- a/node/bft/storage-service/Cargo.toml
+++ b/node/bft/storage-service/Cargo.toml
@@ -61,7 +61,7 @@ workspace = true
 
 [dev-dependencies.snarkvm]
 workspace = true
-features = [ "test", "test-helpers" ]
+features = [ "test-helpers" ]
 
 [dev-dependencies.tracing]
 workspace = true
--- a/node/bft/storage-service/Cargo.toml
+++ b/node/bft/storage-service/Cargo.toml
@@ -55,3 +55,13 @@ optional = true
 
 [dev-dependencies.bytes]
 workspace = true
+
+[dev-dependencies.parking_lot]
+workspace = true
+
+[dev-dependencies.snarkvm]
+workspace = true
+features = [ "test", "test-helpers" ]
+
+[dev-dependencies.tracing]
+workspace = true
--- a/node/bft/storage-service/src/persistent.rs
+++ b/node/bft/storage-service/src/persistent.rs
@@ -347,7 +347,7 @@ mod tests {
 
     use snarkvm::{
         console::network::MainnetV0,
-        ledger::narwhal::Data,
+        ledger::narwhal::{Data, batch_header, transmission},
         prelude::{Rng, TestRng, Uniform},
     };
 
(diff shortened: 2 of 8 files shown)


node/bft/storage-service/src/persistent.rs
```diff
@@ -132,8 +130,19 @@ impl<N: Network> StorageService<N> for BFTPersistentStorage<N> {
...
+            let is_stored = match self.transmissions.contains_key_confirmed(transmission_id) {
+                Ok(result) => result,
+                Err(error) => {
+                    error!("Failed to check if transmission ID exists in confirmed storage - {error}");
+                    false
+                }
+            };
             // If the transmission ID does not exist, ensure it was provided by the caller or aborted.
-            if !self.contains_transmission(*transmission_id) {
+            if !is_stored {
```
ljedrz: could we add a test for this behavior too?

eranrund: Unit tests added in 9f16ae4fc

ljedrz (commented):
So if I understand correctly, in case an aborted ID is referenced, instead of it being stored and delaying the validator, it should be rejected outright and fail loudly? If the in-memory storage already does this, is there perhaps a dedicated integration test we could repurpose for the persistent storage (in order to have better alignment between the two)?

Needs a rebase, too.

ljedrz (commented):
Thanks for the additional tests. My one remaining concern is that `contains_transmission` is still widely used and may return `true` for aborted txs elsewhere (e.g. in `node/bft/src/helpers/storage.rs#L863`), leading to a mismatch with the 2nd change (`persistent.rs`) - or rather making it effectively unreachable (due to having been skipped earlier on). Can you please check that callsite? We may also want to introduce the concrete check as its own trait method.

node/bft/storage-service/Cargo.toml
```diff
@@ -55,3 +55,13 @@ optional = true
 
 [dev-dependencies.bytes]
 workspace = true
+
+[dev-dependencies.parking_lot]
+workspace = true
+
+[dev-dependencies.snarkvm]
+workspace = true
+features = [ "test", "test-helpers" ]
```
ljedrz:
```suggestion
features = [ "test-helpers" ]
```
I think this will suffice?

PR #4384, opened 2026-08-10
[Feat] Make the connection stats ephemeral

ljedrz:
This PR removes the persistent `KnownPeers` in favor of the `Stats` (the same as those of `Tcp`), which live for as long as the associated object (in this case the connection).

There are several benefits to this approach:
- stats for multiple connections per IP can now be distinguished
- the values can be used more reliably for peer heuristics (CC https://github.com/ProvableHQ/snarkOS/issues/4140)
- no more lock contention for traffic accounting
- peers changing their hardware or internet connection won't be affected by past failures

If we're interested in tracking long-term peer reputation, this should be done at the `Peer` (i.e. higher) level, and based on their `Address`.


cbeck88 (approved):
this looks good to me, thaank you

Issue #4380, opened 2026-08-07
[Proposal] Review locking practices in snarkOS


cbeck88:
## 💥 Proposal

In many parts of the snarkOS primary (gateway, storage, telemetry), the most common synchronization primitive that I see is `RwLock<HashMap<..., ...>>`.

https://github.com/ProvableHQ/snarkOS/blob/9db149217b59991568b96be7cf223ce0633ca749/node/bft/src/helpers/cache.rs#L31

https://github.com/ProvableHQ/snarkOS/blob/9db149217b59991568b96be7cf223ce0633ca749/node/bft/src/helpers/telemetry.rs#L78

https://github.com/ProvableHQ/snarkOS/blob/9db149217b59991568b96be7cf223ce0633ca749/node/bft/src/helpers/pending.rs#L64

https://github.com/ProvableHQ/snarkOS/blob/9db149217b59991568b96be7cf223ce0633ca749/node/bft/src/gateway.rs#L200

https://github.com/ProvableHQ/snarkOS/blob/9db149217b59991568b96be7cf223ce0633ca749/node/bft/src/bft.rs#L66

For example:

Naively, `RwLock` seems very attractive especially when rapid prototyping, because if you don't really know whether the value will be heavily written or rarely written yet, the choice shouldn't be too terrible.

However, `RwLock` can also cause nasty problems, because it turns out that it's very difficult to design a generically good `RwLock`, and competing concerns cause it to behave in ways that aren't often expected.

1. Users of `RwLock` expect that it should satisfy eventual fairness. If a writer attempts to lock it, the implementation should guarantee that even if new readers show up and want to read concurrently with existing readers, they eventually have to get shut out so that the writer is guaranteed to get a turn.

If the implementation doesn't guarantee this, then it's possible that a writer tries to lock and then is blocked for an unbounded amount of time. If your program is broken because the writer just never gets a turn, it's not clear what you can do as a developer to fix it, so it's viewed as a defect in the `RwLock` implementation.

[...]

PR #4391, opened 2026-08-12
refactor(telemetry): replace shared locks with a single-owner worker

cbeck88:
## Motivation

Relates to #4380 (locking practices).

The idea is that this is low hanging fruit where we can eliminate locks from BFT and make it harder for BFT to deadlock.

Validator telemetry keeps its state in four `Arc<RwLock<..>>` fields shared between the BFT hot path (writers) and the heartbeat / metrics / REST paths (readers). The module carries a doc comment forbidding those locks from being acquired inside, or held across, a rayon parallel iterator:

> Note: none of the locks below may be acquired inside a parallel iterator, and none of them may be held while running one. The rayon thread pool has a fixed number of workers, so a parallel iterator that blocks on one of these locks can occupy every worker, which in turn prevents the lock holder's own parallel iterator from ever being scheduled.

That rule is enforced only by review -- the compiler can't catch it.

The participation scores are purely observational. They feed a debug log (`gateway.rs`, heartbeat), a set of metrics gauges (`node/consensus`), and a REST endpoint (`GET /{network}/validators/participation`). Nothing in consensus reads them. That makes it safe to take the state off the shared path entirely rather than trying to lock it more carefully.

## What changed

The module splits into three pieces:

- **`TelemetryState`** — the scoring algorithm. Plain owned data with `&mut self` methods; no locks, no interior mutability, no async. Directly unit-testable.
- **`TelemetryWorker`** — owns a `TelemetryState`, drains an mpsc queue in a single task, publishes snapshots to a `watch` channel.
- **`Telemetry`** — the cheap `Clone` handle everyone else holds: a queue sender and a snapshot receiver.

Effects:

1. The rayon hazard is gone by construction — there is no longer a lock for a parallel iterator to block on.
2. BFT can no longer block on telemetry. Enqueueing uses `try_reserve`; a full queue drops the update and bumps a counter (`num_dropped`).
3. Score recomputation is coalesced. It is `O(validators × rounds)` and previously ran once per `insert_subdag`; the worker now drains its whole queue and recomputes once per batch, which matters most during catch-up sync.
4. Signer-address recovery is only paid when there is somewhere to put the result. Recovering a signer address is a fixed-base scalar multiplication — measured at ~39 µs per signature — so a subdag is tens of milliseconds of CPU. `try_reserve` runs *before* that derivation, so a full queue costs a failed reservation rather than a wasted scalar-multiplication storm.

`insert_subdag` and `get_participation_scores` keep their exact signatures and stay synchronous, so all five call sites (bft, sync, gateway, consensus, rest) are untouched. Keeping reads synchronous is why scores are published over a `watch` channel rather than a request/response mailbox: `watch::Receiver::borrow()` does not require `.await`, and two of the call sites are non-async.

### Eventual consistency

Because back pressure causes us to drop updates, the scores can be temporarily wrong. They recover on their own, and the reason is a property of `TelemetryState` that this PR documents as an invariant: **every field is keyed by round, and none aggregates across rounds.** `validator_signatures` holds a per-round count per validator rather than a running total, so garbage collection can remove a round's entire contribution. A state that missed some rounds therefore becomes identical to one that saw them as soon as those rounds are pruned, which happens on its own as `gc_round` advances with consensus.

That property is load-bearing and easy to destroy — replacing the per-round counts with a running total would make every dropped update a permanent deficit — so it is stated in the module docs and on `TelemetryState`, and pinned by `test_scores_converge_after_dropped_updates`.

`Telemetry::is_complete()` makes it observable at runtime: the handle records the highest round any drop covered, the worker publishes the `gc_round` its scores were computed at, and `is_complete` reports whether the two have caught up.

## Test Plan

Six tests in the module, all passing:

- Four cover `TelemetryState` directly (insert, duplicate insert, score calculation, GC). These are ported from the previous lock-based tests and are now free of concurrency entirely.
- `test_scores_converge_after_dropped_updates` feeds two states the same certificates, with one missing a contiguous block of rounds, and asserts they **differ** before GC and are **exactly equal** after GC passes the gap.
- `test_worker_applies_updates` exercises the handle → queue → worker → watch path end to end, using a new `flush()` that acknowledges only after publication, and asserts the worker stops when the last handle is dropped.

Also verified: `cargo clippy -p snarkos-node-bft --tests -- -D warnings` clean both with and without `--features telemetry`; `cargo check` clean for `snarkos-node-consensus` and `snarkos-node-rest` with the feature enabled.

Not verified: the full test suite, and the build has not been re-run since rebasing onto current staging. The rebase was conflict-free and its net diff is byte-identical to the pre-rebase diff, and staging's recent commits touch only `node/network` and `node/tcp` while this PR touches only `node/bft`, so the risk is low — but CI is the real check.

## Backwards compatibility

No `ConsensusVersion` guard needed. This changes no messages, no wire formats, and no consensus behaviour — telemetry is observational and its consumers are a log line, metrics gauges, and a REST endpoint. The public API of `Telemetry` is unchanged apart from the additive `num_dropped()`, `is_complete()` and `flush()`.

The one behavioural change to be aware of: under sustained load the scores can now be computed over a partial history rather than blocking BFT to stay complete. `is_complete()` reports when that is the case, and the drop warning is rate limited to once per 10 seconds with a running total so the logs stay readable.

### DIFF
--- a/node/bft/src/gateway.rs
+++ b/node/bft/src/gateway.rs
@@ -14,7 +14,7 @@
 // limitations under the License.
 
 #[cfg(feature = "telemetry")]
-use crate::helpers::Telemetry;
+use crate::helpers::{Telemetry, TelemetryWorker};
 use crate::{
     CONTEXT,
     MAX_BATCH_DELAY,
@@ -198,8 +198,15 @@ pub struct InnerGateway<N: Network> {
     resolver: RwLock<Resolver<N>>,
     /// The collection of both candidate and connected peers.
     peer_pool: RwLock<HashMap<SocketAddr, Peer<N>>>,
+    /// The handle to the validator telemetry tracker.
     #[cfg(feature = "telemetry")]
     validator_telemetry: Telemetry<N>,
+    /// The telemetry worker, taken and spawned by `run`.
+    ///
+    /// This is a one-shot handoff from `new` to `run`, not a hot-path lock: the telemetry
+    /// state itself lives in the worker task, and is never shared.
+    #[cfg(feature = "telemetry")]
+    telemetry_worker: Mutex<Option<TelemetryWorker<N>>>,
     /// The primary sender.
     primary_sender: OnceCell<PrimarySender<N>>,
     /// The worker senders.
@@ -293,6 +300,10 @@ impl<N: Network> Gateway<N> {
         // some of the cached validators to trusted ones.
         initial_peers.extend(trusted_validators.iter().copied().map(|addr| (addr, Peer::new_candidate(addr, true))));
 
+        // Initialize the validator telemetry. The worker is spawned later, by `run`.
+        #[cfg(feature = "telemetry")]
+        let (validator_telemetry, telemetry_worker) = Telemetry::new();
+
         // Return the gateway.
         Ok(Self(Arc::new(InnerGateway {
             account,
@@ -303,7 +314,9 @@ impl<N: Network> Gateway<N> {
             resolver: Default::default(),
             peer_pool: RwLock::new(initial_peers),
             #[cfg(feature = "telemetry")]
-            validator_telemetry: Default::default(),
+            validator_telemetry,
+            #[cfg(feature = "telemetry")]
+            telemetry_worker: Mutex::new(Some(telemetry_worker)),
             primary_sender: Default::default(),
             worker_senders: Default::default(),
             sync_sender: Default::default(),
@@ -335,6 +348,13 @@ impl<N: Network> Gateway<N> {
         // If the sync sender was provided, set the sync sender.
         if let Some(sync_sender) = sync_sender {
             self.sync_sender.set(sync_sender).expect("Sync sender already set in gateway");
+        }
+
+        // Spawn the validator telemetry worker, which owns all telemetry state.
+        // It is registered in `handles`, so `shut_down` aborts it along with everything else.
+        #[cfg(feature = "telemetry")]
+        if let Some(telemetry_worker) = self.telemetry_worker.lock().take() {
+            self.spawn(telemetry_worker.run());
         }
 
         // Enable the TCP protocols.
--- a/node/bft/src/helpers/telemetry.rs
+++ b/node/bft/src/helpers/telemetry.rs
@@ -304,6 +304,7 @@ impl<N: Network> Telemetry<N> {
     /// all tracked state is keyed by round and pruned by round, the scores become correct
     /// again once garbage collection passes the highest round a drop touched, at which point
     /// this returns `true` permanently (until the next drop). See the module documentation.
+    #[cfg(test)]
     pub fn is_complete(&self) -> bool {
         // Read the published `gc_round` before `max_dropped_round`. Both only ever increase, so
         // in this order a drop recorded between the two reads is compared against the older
--- a/node/bft/src/helpers/telemetry.rs
+++ b/node/bft/src/helpers/telemetry.rs
@@ -36,6 +36,37 @@
 //! [`tokio::sync::watch`] channel. That channel does have an internal lock, but
 //! it is held only long enough to clone an `Arc`, there is exactly one writer,
 //! and no other lock is ever acquired underneath it, so no cycle can form.
+//!
+//! # Eventual consistency under dropped updates
+//!
+//! Because a full queue drops updates, the scores can be temporarily wrong. They
+//! are, however, guaranteed to become correct again, and the reason is a property
+//! of [`TelemetryState`] that must be preserved by anyone changing it:
+//!
+//! **Every piece of tracked state is keyed by round, and garbage collection prunes
+//! by round.** `tracked_certificates` is keyed by round; `validator_signatures`
+//! holds a per-round count for each validator rather than a running total; and
+//! `validator_certificates` holds a set of rounds. So the entire contribution of
+//! any given round can be, and is, removed wholesale by
+//! [`TelemetryState::garbage_collect_certificates`].
+//!
+//! A dropped subdag update means the rounds it covered are never inserted. Once
+//! `gc_round` advances past those rounds, a state that missed them and a state that
+//! saw them contain *exactly the same data*, because the only difference between
+//! them lived in rounds that have since been pruned. Since `gc_round` is derived
+//! from the anchor round of each committed subdag, it advances on its own as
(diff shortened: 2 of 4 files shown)


node/bft/src/gateway.rs
```diff
@@ -14,7 +14,7 @@
 // limitations under the License.
 
 #[cfg(feature = "telemetry")]
```
cbeck88, line 16: TODO: I think we should remove feature `telemetry` and just make it part of `metrics`, since it doesn't appear that we test those features independently so it will likely bit rot

node/bft/src/helpers/telemetry.rs
```diff
@@ -73,40 +173,313 @@ impl<N: Network> CertificateMetadata<N> {
...
+                        self.state.garbage_collect_certificates(gc_round);
+                        self.state.insert_certificate_metadata(&metadata);
+                        // Track the high-water mark rather than the last value seen. Updates
+                        // are applied in order today, but `is_complete` would silently start
+                        // reporting false negatives if a lower round were ever to follow.
+                        self.gc_round = self.gc_round.max(gc_round);
+                        recompute = true;
+                    }
+                    TelemetryUpdate::Certificate(metadata) => {
+                        self.state.insert_certificate_metadata(std::slice::from_ref(&*metadata));
```
eranrund, line 441: Flagging that this does not currently trigger a recompute. It doesn't appear that `insert_certificate` is called from non-test code, and I believe this was the behavior prior to this PR, but it might be worth mentioning this explicitly here.

node/bft/src/helpers/telemetry.rs
```diff
@@ -73,40 +173,313 @@ impl<N: Network> CertificateMetadata<N> {
...
+    /// again once garbage collection passes the highest round a drop touched, at which point
+    /// this returns `true` permanently (until the next drop). See the module documentation.
+    pub fn is_complete(&self) -> bool {
+        let max_dropped_round = self.max_dropped_round.load(Ordering::Relaxed);
+        // Nothing has ever been dropped.
+        if max_dropped_round == 0 {
+            return true;
+        }
+        // Every round a drop could have touched has since been garbage collected.
+        self.scores.borrow().gc_round >= max_dropped_round
```
eranrund: I think there may be a small race here.
We load `max_dropped_round` before snapshotting `gc_round`. If a drop at round `R` is recorded between those reads, while the worker publishes a `gc_round` that covers the previously observed max but not `R`, this can return `true` even though the new drop is still relevant. I think it would safer to snapshot `gc_round` first and then load `max_dropped_round`. That ordering can produce a transient false negative under concurrency, but not this false positive.

eranrund (dismissed):
Aside from some minor nits, I think this looks good assuming we are okay with one observable behavior change.

I think it's worth noting for the record: readers of the participation scores now lag producers slightly. Previously `insert_subdag` recomputed the scores synchronously on the commit path, so the [consensus metrics gauges](https://github.com/ProvableHQ/snarkOS/blob/cebf21dba7b0859a98b103d97ef46de2cd4775ba/node/consensus/src/lib.rs#L722-L729) and the gateway heartbeat log always reflected the just-committed block. With the worker model they read the last *published* snapshot, which may trail the tip by a bit while the worker catches up. This is harmless for observational data and consistent with the module's documented eventual-consistency model, but I think it is externally visible - e.g. if there are any dashboards that compare scores against block height. This can likely be remedied by exposing the published `gc_round` from the handle and exporting it as a gauge.

I think this is an overall small price to pay for the benefits of a simplified locking strategy and moving unnecessary computation off the consensus critical path.

Also noting that I am new to this codebase and potentially lack extra context on how telemetry/metrics are used so it is possible I am missing something and would love for @vicsn and/or @ljedrz to take a look when they can.

Thank you!

node/bft/src/gateway.rs
vicsn: Indeed there was no good reason to have it separate. May want to keep a stub around at the top level to not break people's builds.

node/bft/src/helpers/telemetry.rs
vicsn: A bit of lag seems fine

node/bft/src/gateway.rs
cbeck88: this change is #4406

node/bft/src/helpers/telemetry.rs
```diff
@@ -73,40 +173,339 @@ impl<N: Network> CertificateMetadata<N> {
...
+        scores_for_committee(&snapshot.scores, committee)
+    }
+
+    /// Returns `true` if the published scores reflect every update that was produced.
+    ///
+    /// This is `false` only while a dropped update can still be affecting the result. Because
+    /// all tracked state is keyed by round and pruned by round, the scores become correct
+    /// again once garbage collection passes the highest round a drop touched, at which point
+    /// this returns `true` permanently (until the next drop). See the module documentation.
+    pub fn is_complete(&self) -> bool {
```
ljedrz, line 308:
```suggestion
    /// this returns `true` permanently (until the next drop). See the module documentation.
    #[cfg(test)]
    pub fn is_complete(&self) -> bool {
```
I don't see it called elsewhere

node/bft/src/helpers/telemetry.rs
```diff
@@ -73,40 +173,339 @@ impl<N: Network> CertificateMetadata<N> {
...
+
+        // Sending on a reserved permit cannot fail.
+        permit.send(TelemetryUpdate::Subdag { gc_round, metadata });
+    }
+
+    /// Insert a certificate into the telemetry tracker.
+    ///
+    /// This does not cause the scores to be recomputed on its own; the next subdag picks the
+    /// certificate up. See the note on the certificate arm of [`TelemetryWorker::run`].
+    pub fn insert_certificate(&self, certificate: &BatchCertificate<N>) {
```
ljedrz, line 279:
```suggestion
    /// certificate up. See the note on the certificate arm of [`TelemetryWorker::run`].
    #[cfg(test)]
    pub fn insert_certificate(&self, certificate: &BatchCertificate<N>) {
```

ljedrz (commented):
LGTM in general, left comments on potentially dead APIs.

eranrund (dismissed):

vicsn (dismissed):
LGTM if Lukasz' comments are addressed

vicsn (approved):

PR #4396, opened 2026-08-14
Post-release merge from testnet back into staging

vicsn:
## Motivation

Also moves consensus-gated features back by 1.

snarkVM rev belongs to https://github.com/ProvableHQ/snarkVM/pull/3380


cbeck88 (approved):
thank you

PR #4399, opened 2026-08-19
fix(bft): enforce worker capacity limit at transmission insertion

cbeck88:
## Motivation

`process_transmission_id_from_ping` checks the ready-queue capacity (`num_transmissions() > MAX_TRANSMISSIONS_PER_WORKER`) before spawning an asynchronous fetch, but the insertion that follows did not re-check. Because the fetch is asynchronous, many concurrent pings can each observe a non-full ready queue before any of them insert, then collectively exceed the limit (a time-of-check-to-time-of-use race). A malicious validator can exploit this to inflate a peer's ready queue past `MAX_TRANSMISSIONS_PER_WORKER`, which in turn pushes `num_unconfirmed_transmissions` past `MAX_TRANSMISSIONS_TOLERANCE` and stalls admission of honest unconfirmed transactions/solutions into the BFT layer.

## The fix

Re-check the capacity limit atomically with the insertion (under a single write lock), but **only on the worker-ping fetch path**. The insertion logic is factored into a shared `insert_transmission_from_peer` helper with an `enforce_capacity_limit` flag:

- `process_transmission_from_ping` — the worker-ping path — enforces the limit, closing the TOCTOU.
- `process_transmission_from_peer` — batch-proposal materialization and all other callers — is left uncapped.

The limit is enforced only on the ping path because `process_unconfirmed_{solution,transaction}` and `insert_front`/`reinsert` must stay uncapped so real user submissions and primary re-queueing keep priority.

## Test Plan

- `test_process_transmission_from_ping_respects_capacity`: floods the worker-ping path and asserts the ready queue never grows beyond the per-worker limit.
- `test_process_transmission_from_peer_ignores_capacity`: asserts the batch-materialization path retains every transmission, even beyond the limit.

## Alternatives

The first commit took the simpler route of enforcing the cap directly inside `process_transmission_from_peer`, which was the only function performing the insertion. On review this turned out to be unsafe, because that function is shared by two callers, not just the ping path:

1. the async worker-ping fetch (the intended target), and
2. `Primary::insert_missing_transmissions_into_workers`, which materializes the transmissions of a peer's batch proposal **immediately before the node signs that batch**.

On path 2, an early return under capacity pressure silently dropped the transmission while `insert_missing_transmissions_into_workers` still returned `Ok(())`, so the node would sign and broadcast a `BatchSignature` for a batch whose transmissions it does not hold — the "certified batch referencing unmaterializable transmissions" failure this work is meant to avoid. It was also reachable without a Byzantine peer: since the unconfirmed solution/transaction paths are intentionally uncapped, `ready` can legitimately sit above `MAX_TRANSMISSIONS_PER_WORKER` under load, and while it does, every honest peer's proposal would have its transmissions dropped and get signed anyway.

Rather than propagate the drop as an error and have the node decline to sign (safer, but changes proposal-handling behaviour under load), we scoped the cap to the ping path via the split above. This matches the original intent — the batch path keeps priority and never drops ahead of signing — and is the smaller behavioural change.

## Backwards compatibility

I think this is backwards compatible, it just prevents us from exceeding the limit via the TOCTOU race.

### DIFF
--- a/node/bft/src/worker.rs
+++ b/node/bft/src/worker.rs
@@ -298,7 +298,7 @@ impl<N: Network> Worker<N> {
                         // Insert the transmission into the ready queue.
                         // Note: This method checks `contains_transmission` again, because by the time the transmission is fetched,
                         // it could have already been inserted into the ready queue.
-                        self_.process_transmission_from_peer(peer_ip, transmission_id, transmission);
+                        self_.process_transmission_from_ping(peer_ip, transmission_id, transmission);
                     }
                 }
                 // If the transmission was not fetched, then attempt to fetch it again.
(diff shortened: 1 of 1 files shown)


vicsn (approved):
LGTM, seems like a solid hardening

PR #4401, opened 2026-08-20
test(bft): cover the in-memory storage service, and fix a deadlock in it

cbeck88:
## Motivation

Part of a sweep over the workspace's least-tested crates. `BFTMemoryService` (206 lines) had no tests of its own, despite being the storage service every BFT test constructs — `worker.rs`, `bft.rs`, `sync/mod.rs`, and `helpers/storage.rs` all build their `Storage` on top of it.

That made it load-bearing test infrastructure whose own semantics were asserted nowhere. In particular its reference counting — a transmission is evicted only once the last certificate referencing it is removed — was never checked, so a regression there would surface as confusing failures in unrelated BFT tests rather than as a failure here.

Writing the tests surfaced a deadlock, which the second commit fixes.

## Commit 1 — the tests

- **The reference-counting lifecycle.** Two certificates referencing one transmission (as happens when the primary processes batches from several peers carrying the same transaction); dropping one reference keeps it, dropping the last evicts it.
- **The same lifecycle for aborted transmission IDs**, which are tracked in a separate map with the same semantics.
- **The aborted/stored asymmetry** — an aborted ID reports `contains_transmission == true` but `get_transmission == None`.
- **First writer wins.** When a second certificate arrives carrying a different payload under an ID already in storage, the stored transmission is kept and only the certificate ID is recorded.
- **Removal of an unrelated certificate** does not evict the transmission, and removal against an empty service is a no-op.
- **All three outcomes of `find_missing_transmissions`**: it returns only what storage lacks (dropping ids the caller offered but storage already holds), accepts an aborted declaration without returning it, and rejects a declaration that is neither provided nor aborted — the check that stops a peer from having a certificate accepted while withholding the transmissions it commits to.

## Commit 2 — the deadlock fix

**The bug.** In the vacant-entry arm of `insert_transmissions`, when a declared transmission ID was neither provided in `missing_transmissions` nor aborted, the code called `self.contains_transmission(transmission_id)` while already holding the write guard on `self.transmissions`. That method takes a read guard on the same lock. `parking_lot::RwLock` is not reentrant, so the thread parked forever.

**How it is reached.** `find_missing_transmissions` and `insert_transmissions` take separate locks, with no barrier between them. `Storage::insert_certificate` calls the first (via `check_certificate`) and then the second. An ID that the first call saw in storage — and therefore correctly omitted from `missing_transmissions` — can be evicted by a concurrent `remove_transmissions` before the second call runs. The second call then finds a vacant entry for an ID that is neither provided nor aborted, and parks.

The `contains_transmission` call appears to have been there to handle exactly this race benignly, by re-checking whether the entry was known after all. As written it deadlocked instead.

**The fix.** Answer the question from the guards already in scope. Reaching the vacant arm establishes that `self.transmissions` does not contain the ID, so `contains_transmission` reduces to a lookup in the aborted map — whose write guard is already held a few lines above. The check becomes:

```rust
if !aborted_transmission_ids.contains(&transmission_id)
    && !aborted_transmission_ids_lock.contains_key(&transmission_id)
```

This is an exact equivalence, not a weakening: `contains_transmission(id)` is `transmissions.contains_key(id) || aborted.contains_key(id)`, and the first disjunct is known to be `false` in this arm. Observable behavior is unchanged — an undeliverable transmission is logged and skipped — it just no longer deadlocks.

**Scope.** Only `BFTMemoryService` was affected, so the impact was on test infrastructure rather than on running nodes: production wires `BFTPersistentStorage` (`node/consensus/src/lib.rs:150`). The persistent backend has the same-shaped code at `persistent.rs:183`, but its `contains_transmission` reads RocksDB rather than the `cache_transmissions` mutex it is holding, so it does not deadlock and is left alone.

The repro from commit 1 is no longer `#[ignore]`d. It now asserts that the call returns and that no phantom entry is left behind, and it runs on its own thread with a 10s watchdog so that a regression fails on the timeout rather than hanging the test binary.

## Test Plan

```
cargo test -p snarkos-node-bft-storage-service --features memory,persistent
```

14 passing, 0 ignored. The feature set matches what CI already runs for this member.

I confirmed the regression test actually guards the fix by reverting the one-line change and re-running it — it fails on the watchdog timeout, then passes again with the fix restored.

Also verified:

- `cargo clippy -p snarkos-node-bft-storage-service --features memory,persistent --all-targets`
- `cargo check -p snarkos-node-bft-storage-service --features memory,persistent,locktick --all-targets`, since the lock type is feature-swapped
- `cargo test -p snarkos-node-bft --lib helpers::storage` — the 8 tests that exercise this service through `Storage`, all passing

## Documentation

None needed.

## Backwards compatibility

No behavior change; nothing to guard by `ConsensusVersion`.

### DIFF
--- a/node/bft/storage-service/src/memory.rs
+++ b/node/bft/storage-service/src/memory.rs
@@ -204,3 +204,307 @@ impl<N: Network> StorageService<N> for BFTMemoryService<N> {
         self.transmissions.read().clone().into_iter().collect()
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    use snarkvm::{
+        console::network::MainnetV0,
+        ledger::narwhal::Data,
+        prelude::{PrivateKey, Rng, TestRng, Uniform},
+    };
+
+    use bytes::Bytes;
+    use std::{sync::mpsc, time::Duration};
+
+    type CurrentNetwork = MainnetV0;
+
+    fn sample_transmission_id(rng: &mut TestRng) -> TransmissionID<CurrentNetwork> {
+        TransmissionID::Transaction(
+            <CurrentNetwork as Network>::TransactionID::from(Field::rand(rng)),
+            <CurrentNetwork as Network>::TransmissionChecksum::from(rng.random::<u128>()),
+        )
+    }
+
+    fn sample_transmission(payload: &[u8]) -> Transmission<CurrentNetwork> {
+        Transmission::Transaction(Data::Buffer(Bytes::from(payload.to_vec())))
+    }
+
+    /// Builds a batch header declaring the given transmission IDs.
+    fn sample_batch_header(
+        transmission_ids: &[TransmissionID<CurrentNetwork>],
+        rng: &mut TestRng,
+    ) -> BatchHeader<CurrentNetwork> {
+        let private_key = PrivateKey::<CurrentNetwork>::new(rng).unwrap();
+        // Round 1 is the only round that may carry no previous certificate IDs.
+        BatchHeader::new(
+            &private_key,
+            1,
+            0,
+            Field::rand(rng),
+            transmission_ids.iter().copied().collect(),
+            Default::default(),
+            rng,
+        )
+        .unwrap()
+    }
+
+    #[test]
+    fn a_new_service_holds_nothing() {
+        let rng = &mut TestRng::default();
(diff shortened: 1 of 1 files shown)


vicsn (approved):
Nice catch!

PR #4412, opened 2026-08-26
[Fix] Don't certify batches that reference aborted transactions

ljedrz:
An alternative to #4383, building on some of its changes. Same bug, different fix.

BFT storage knows two kinds of transmission ID:
- ones it holds a transmission for
- ones it only recorded as aborted, where it knows the ID but has no bytes to hand back

`contains_transmission` returns true for both, while `get_transmission` returns `Some` only for the first. That was the only containment query available, so callers that needed to tell the two apart could not. A few things went wrong as a result.

1. The primary could propose a transmission that nobody can produce.
In `propose_batch`, the storage check was skipped while the proposal was still empty - an exception meant to avoid proposing an empty batch. So if the first ID drained from the worker queue was one that storage knows only as aborted, it was proposed as if it were a real transmission. The ledger check just above it catches that case only once the aborting block has reached the ledger; while that block is still pending, BFT storage is the only place that knows, and its check was the one being skipped.

2. Persistent storage discarded transmissions it was handed.
`BFTPersistentStorage::find_missing_transmissions` used `contains_transmission`, which is true for an aborted ID, so it treated the ID as already taken care of and never pulled the bytes out of the caller's map. The certificate was accepted, but the transmission never reached storage - `get_transmission` still returned `None`, so the certificate could not be committed. `BFTMemoryService` did not have this problem, because it checked only its map of concrete transmissions.

3. In-memory storage rejected certificates it should have accepted.
The other side of the same coin. Checking only the concrete map, `BFTMemoryService::find_missing_transmissions` then demanded that any remaining ID be either provided by the caller or declared aborted by the caller. Neither ever happens for an already-aborted ID: `fetch_missing_transmissions` does not fetch bytes for an ID storage already knows, and the primary always passes an empty aborted set. So a peer's certificate referencing such an ID was rejected outright.

4. A reference to an already-aborted ID was never counted.
`insert_transmissions` credits the certificate to the transmissions map for the IDs it stores, and to the aborted map only for the IDs in the aborted set it was given. An ID that storage already knows as aborted, and that the caller did not re-declare, is credited to neither - yet `remove_transmissions` later decrements both maps for every ID in the certificate. The two sides are therefore asymmetric: once the certificate that first recorded the abort is garbage collected, the entry's count drops to zero and the entry is deleted, even though a second certificate still references that ID. The node then no longer knows the ID at all, and goes looking for bytes that no peer has.

5. Block sync left those IDs unaccounted for.
`sync_certificate_with_block` skipped any ID that `contains_transmission` reported, so an already-aborted one never reached the classification below it and ended up in neither the transmission map nor the aborted set. Same hole as (4), reached from the sync path instead of the primary.

The fix splits the one query into three: `contains_transmission` (either kind, unchanged), `contains_retrievable_transmission` (storage holds a transmission), and `contains_aborted_transmission` (storage recorded an abort). Each caller now asks what it actually means. find_missing_transmissions becomes one default trait method shared by both services, instead of two copies that had drifted apart.

What differs from #4383:

- This PR adds two queries instead of one, so a caller can also ask whether an ID was aborted.
- Here, `find_missing_transmissions` accepts an ID that storage already knows as aborted, instead of rejecting it.
- The reference counting is fixed in `insert_certificate`, which covers the primary's call sites too, in addition to the block sync path.
- On block sync, an aborted ID goes through the normal classification instead of being skipped, so the block's bytes are used if it happens to carry them.
- The shared tests also cover the three queries and the already-aborted cases.
- Refuses to sign a peer's batch proposal that references an ID the node knows only as aborted.

Each new test was checked to fail with its fix reverted.

### DIFF
--- a/node/bft/src/helpers/storage.rs
+++ b/node/bft/src/helpers/storage.rs
@@ -882,8 +882,13 @@ impl<N: Network> Storage<N> {
             if missing_transmissions.contains_key(transmission_id) {
                 continue;
             }
-            // If the transmission ID exists in storage, skip it.
-            if self.contains_transmission(*transmission_id) {
+            // If the transmission can be retrieved from storage, skip it.
+            //
+            // Note that an ID recorded as aborted by an earlier certificate is deliberately not
+            // skipped here: storage holds no transmission for it, so it has to go through the
+            // classification below - which either recovers the transmission from the block, or
+            // declares the ID as aborted for this certificate as well.
+            if self.contains_retrievable_transmission(*transmission_id) {
                 continue;
             }
             // Retrieve the transmission.
@@ -896,12 +901,17 @@ impl<N: Network> Storage<N> {
                     //
                     // Aborted transmissions only appear in the aborted set of the first block that contains them,
                     // for subsequent blocks, we can check that `contains_transmission` is true to determine if the transmission was aborted.
+                    //
+                    // Note that the ledger only answers for the blocks it already holds: the block that
+                    // aborted the ID may still be awaiting its availability check in `Sync::pending_blocks`,
+                    // so storage is consulted as well - it recorded the abort when that block was processed.
                     if let Some(solution) = block.get_solution(solution_id) {
                         missing_transmissions.insert(*transmission_id, (*solution).into());
                     } else if let Ok(Some(solution)) = self.ledger.get_solution(solution_id) {
                         missing_transmissions.insert(*transmission_id, solution.into());
                     } else if aborted_solutions.contains(solution_id)
                         || self.ledger.contains_transmission(transmission_id).unwrap_or(false)
+                        || self.contains_aborted_transmission(*transmission_id)
                     {
                         aborted_transmissions.insert(*transmission_id);
                     } else {
@@ -915,12 +925,17 @@ impl<N: Network> Storage<N> {
                     //
                     // Aborted solutions only appear in the aborted set of the first block that contains them,
                     // for subsequent blocks, we can check that `contains_transmission` is true to determine if the transaction was aborted.
+                    //
+                    // Note that the ledger only answers for the blocks it already holds: the block that
+                    // aborted the ID may still be awaiting its availability check in `Sync::pending_blocks`,
+                    // so storage is consulted as well - it recorded the abort when that block was processed.
                     if let Some(transaction) = unconfirmed_transactions.get(transaction_id) {
                         missing_transmissions.insert(*transmission_id, transaction.clone().into());
                     } else if let Ok(Some(transaction)) = self.ledger.get_unconfirmed_transaction(*transaction_id) {
                         missing_transmissions.insert(*transmission_id, transaction.into());
                     } else if aborted_transactions.contains(transaction_id)
                         || self.ledger.contains_transmission(transmission_id).unwrap_or(false)
+                        || self.contains_aborted_transmission(*transmission_id)
                     {
                         aborted_transmissions.insert(*transmission_id);
                     } else {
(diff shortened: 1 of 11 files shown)


node/bft/storage-service/src/traits.rs
```diff
@@ -25,21 +25,75 @@ use std::{
...
+                match transmissions.remove(transmission_id) {
+                    // Append the transmission if it exists.
+                    Some(transmission) => {
+                        missing_transmissions.insert(*transmission_id, transmission);
+                    }
+                    // If the transmission does not exist, it must be aborted: either the caller
+                    // declares it as such, or storage already recorded it as aborted for an earlier
+                    // certificate - in which case there are no bytes to be had from anyone.
+                    None => {
+                        if !aborted_transmissions.contains(transmission_id)
```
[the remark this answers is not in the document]
ljedrz: This is resolved with https://github.com/ProvableHQ/snarkOS/pull/4412/commits/ee52fae0b68164b58c821a99bfde29c41371aebc.

node/bft/src/helpers/storage.rs
```diff
@@ -1250,6 +1297,147 @@ pub(crate) mod tests {
...
+            .sync_certificate_with_block(&block, certificate_2.clone(), &Default::default(), false)
+            .expect("syncing a certificate referencing a previously-aborted transmission must succeed");
+        assert!(storage.contains_certificate(certificate_2.id()));
+
+        // The ID must have been recorded as aborted for the second certificate too, so dropping the
+        // first one must not take the aborted entry with it.
+        assert!(storage.remove_certificate(certificate_1.id()));
+        assert!(storage.contains_transmission(transmission_id));
+        assert!(!storage.contains_retrievable_transmission(transmission_id));
+    }
```
eranrund, line 1378: Stricter assertions, similar to what the `test_insert_certificate_counts_a_previously_aborted_transmission` test does.

```suggestion

        // The second certificate was its last holder: dropping it must drop the aborted entry.                                        
        assert!(storage.remove_certificate(certificate_2.id()));                                                                       
        assert!(!storage.contains_transmission(transmission_id));   
    }
```

node/bft/src/primary.rs
```diff
@@ -986,6 +989,24 @@ impl<N: Network> Primary<N> {
...
+        // Note: this is deliberately stricter than the certificate and block sync paths, which have
+        // to keep accepting such IDs - see `StorageService::find_missing_transmissions`. It is also a
+        // judgment made from local sync state, so an honest peer that has not seen the aborting block
+        // yet can trip it; skip the proposal rather than treat the peer as malicious.
+        if let Some(transmission_id) =
+            batch_header.transmission_ids().iter().find(|id| self.storage.contains_aborted_transmission(**id))
+        {
+            debug!(
+                "Primary is safely skipping a batch proposal from '{peer_ip}' - {}",
+                format!("it references the aborted transmission '{}'", fmt_id(transmission_id)).dimmed()
```
eranrund: Nit: Using `format!` here seems weird, is this due to a line length lint or something like that?

node/bft/storage-service/src/lib.rs
```diff
@@ -28,3 +28,7 @@ pub use persistent::*;
 
 pub mod traits;
 pub use traits::*;
+
+// The expectations shared by every service, which need every service to assert them against.
+#[cfg(all(test, feature = "memory", feature = "persistent"))]
```
eranrund: This might be suboptimal since only one of the two features might be enabled. We can make it more permissive.
```suggestion
#[cfg(all(test, any(feature = "memory", feature = "persistent")))]
```

node/bft/storage-service/src/tests.rs
```diff
@@ -0,0 +1,256 @@
...
+// limitations under the License.
+
+//! The expectations that every [`StorageService`] implementation has to meet.
+//!
+//! These are the queries whose answers the rest of the BFT reasons about, and
+//! `find_missing_transmissions` is now shared by every implementation, so they are asserted here
+//! once and instantiated per service rather than restated in each one. Behavior specific to a single
+//! service - the persistent one's cache, the in-memory one's locking - stays in its own module.
+
+use crate::{StorageService, memory::BFTMemoryService, persistent::BFTPersistentStorage};
```
eranrund: If the other cfg change is accepted.

```suggestion
use crate::StorageService;                                                                                                             
#[cfg(feature = "memory")]                                                                                                             
use crate::memory::BFTMemoryService;                                                                                                   
#[cfg(feature = "persistent")]
```

node/bft/storage-service/src/tests.rs
```diff
@@ -0,0 +1,256 @@
...
+            #[test]
+            fn find_missing_transmissions_rejects_an_unprovided_undeclared_id() {
+                check_find_missing_transmissions_rejects_an_unprovided_undeclared_id(&$service);
+            }
+        }
+    };
+}
+
+storage_service_tests!(memory, BFTMemoryService::<CurrentNetwork>::new());
+storage_service_tests!(persistent, BFTPersistentStorage::<CurrentNetwork>::open(StorageMode::new_test(None)).unwrap());
```
eranrund, line 553: If the other related changes are accepted
```suggestion
#[cfg(feature = "memory")]                                                                                                             
storage_service_tests!(memory, BFTMemoryService::<CurrentNetwork>::new());
#[cfg(feature = "persistent")]                                                                                                         
storage_service_tests!(persistent, BFTPersistentStorage::<CurrentNetwork>::open(StorageMode::new_test(None)).unwrap());
```

node/bft/src/primary.rs
```diff
@@ -986,6 +989,24 @@ impl<N: Network> Primary<N> {
         // Inserts the missing transmissions into the workers.
         self.insert_missing_transmissions_into_workers(peer_ip, missing_transmissions.into_iter())?;
 
+        // Do not sign a proposal referencing a transmission this node knows only as aborted: storage
```
eranrund: Consider moving this before `self.sync_with_batch_header_from_peer`, I think there's no need to bother syncing from the peer if we are ultimately going to discard this batch header.

node/bft/src/helpers/storage.rs
```diff
@@ -692,13 +711,26 @@ impl<N: Network> Storage<N> {
...
+        // nor the aborted entry - and removing the certificate that first recorded the abort would
+        // drop the entry while this certificate still refers to it.
+        for transmission_id in certificate.transmission_ids() {
+            if !missing_transmissions.contains_key(transmission_id)
+                && !self.contains_retrievable_transmission(*transmission_id)
+                && self.contains_aborted_transmission(*transmission_id)
+            {
+                aborted_transmissions.insert(*transmission_id);
+            }
+        }
```
eranrund: Should we be concerned about a race between this check and the insert? 

An alternative is to drop this loop and handle it inside `insert_transmissions` in `BFTMemoryService` and `BFTPersistentStorage`. Both already reach the "vacant, not provided, but aborted-in-storage" branch and only suppress the error there; inserting the ID into `aborted_transmission_ids` at that point lets the existing aborted loop count the certificate, and the check runs under the locks those functions already hold (which `remove_transmissions` also takes). It also covers callers that go straight to `insert_certificate_atomic` (in case those exist or will in the future).

Both impls then need to stay in sync, so a shared test in `tests.rs` would be worth adding.

node/bft/storage-service/src/memory.rs
eranrund, line 187: Possibly out of scope for this PR but a lot of the test here can be moved into the shared tests file since they only call the trait methods and we want identical behavior between the two implementations.

eranrund: Example of a test that we might want to do on both implementations.

eranrund (commented):
I was in the middle of reviewing when @cbeck88 posted his analysis so I will need to give that some more thought... I also only understand limited portions of it right now.

node/bft/storage-service/src/lib.rs
ljedrz: Good intuition, but this is actually the way CI tests that crate:
```
  node-bft-storage-service:
    executor: rust-docker
    resource_class: << pipeline.parameters.medium >>
    steps:
      - run_serial:
          workspace_member: node/bft/storage-service
          cache_key: node-bft-storage-service
          flags: --features memory,persistent
```
These features may be compiled together, which is practical for test purposes.

node/bft/src/primary.rs
ljedrz: IIRC, this is needed for the call to `dimmed`

node/bft/src/primary.rs
ljedrz: good point, though as an optimization, I'd defer that to a separate PR

node/bft/src/primary.rs
eranrund: Oh right, I missed that.

node/bft/storage-service/src/lib.rs
eranrund: I agree they may be compiled together, but I can also see only one being enabled when someone is running tests.
Switching to `any` will increase the chances of tests running, which I think is a good outcome. `all` is stricter and I don't think it is either needed or has any advantage over `any`.

node/bft/src/primary.rs
```diff
@@ -2009,8 +2014,15 @@ impl<N: Network> Primary<N> {
...
-            // If the transmission does not exist in storage, proceed to fetch the transmission.
-            if !self.storage.contains_transmission(*transmission_id) {
+            // If the transmission cannot be retrieved from storage, proceed to fetch it.
+            //
+            // Note that an ID storage knows only as aborted is deliberately fetched too: storage
+            // holds no payload for it, so treating the aborted marker as "already have it" would
+            // accept a certificate committing to a transmission this node can never materialize.
+            // A batch is an availability claim by its author, so ask for the bytes rather than
+            // excuse them - the author can serve them, and a certificate carries `2f + 1` such
+            // claims, so an honest peer holds them even if this one does not.
```
eranrund, line 2024: Nit: This implies that we might fetch from a different peer, but we don't actually have logic to do that.

node/bft/storage-service/src/traits.rs
```diff
@@ -25,23 +25,88 @@ use std::{
...
 
-    /// Takes a certificate and its transmissions, and returns the subset of transmissions that
-    /// did not yet exists in the storage.
+    /// Takes a batch header and the transmissions provided for it, and returns the subset of those
+    /// transmissions that the storage does not already hold.
+    ///
+    /// A transmission that cannot be retrieved from storage has to be provided by the caller, or be
+    /// aborted - either as declared by the caller, or as already recorded in storage, in which case
+    /// there are no bytes to be had from anyone. Anything else is an error: a peer does not get to
+    /// have a certificate accepted while withholding a transmission that it commits to.
```
eranrund, line 61: Nit - I think this can be phrased a bit better, mostly to remove the statement `in which case there are no bytes to be had from anyone`, since I don't think there's a guarantee the transmission isnt available from any of the peers.


```suggestion
    /// A transmission that cannot be retrieved from this storage must either be provided by the caller
    /// or be classified as aborted, whether declared by the caller or already recorded in storage.
    ///
    /// An aborted marker only means that this storage has no payload for that entry; it does not imply
    /// that a peer cannot provide the bytes.
```

eranrund (dismissed):
Thank you for taking this on @ljedrz - aside from the merge conflict this looks good as far as I can tell.

eranrund (approved):

brunoffranca (approved):

PR #4413, opened 2026-08-26
fix(router): reject oversized messages before buffering their body

cbeck88:
# Motivation

Extend the check_size mechanism to improve node hardening (see also #4140 ).

Previously, the "check_size" function was used to reject transactions on the wire that are larger than maximum possible transaction size before they reach the node.

[`node/router/messages/src/helpers/codec.rs#L72-L91`](https://github.com/ProvableHQ/snarkOS/blob/77308d1890c155cbefb0a14571feb7d2d436e26b/node/router/messages/src/helpers/codec.rs#L72-L91):
```rust
  let bytes = match self.codec.decode(source)? { ... };   // <- 128 MiB buffering
  Self::Item::check_size(&bytes)?;                          // <- gate, operates on raw bytes
  match Message::read_le(reader) { ... }                    // <- never reached if check_size rejects
```

If the size of the payload is larger than a valid message of that type, we can bail without going to deserialization and avoid spending CPU on that. However, we already allocated a 128 MiB buffer to hold the message, when ideally we would have bailed earlier, because 128 MiB allocation is still risky.

Therefore:

Move the check earlier: peek the 4-byte length prefix and 2-byte message ID (at most 6 bytes, arriving effectively immediately regardless of the declared length) directly off the incoming buffer, and reject before waiting for the rest of the body. Crucially, also stop *reserving* the declared length while waiting — otherwise a peer that has sent six bytes still gets an allocation of whatever size it named. The read buffer now grows with what the peer has actually sent.

This replaces `check_size` — which ran after the full frame was buffered, and only special-cased one message type — with a per-ID size table.

## What is capped

| Message | Cap |
| --- | --- |
| `BlockResponse` | 128 MiB (the general frame ceiling) |
| `Ping` | 16 MiB |
| `UnconfirmedTransaction` | `LATEST_MAX_TRANSACTION_SIZE` + 39 bytes of envelope |
| everything else | 8 KiB |

Before this change the tightest constraint enforced was 128 MiB, except for `UnconfirmedTransaction`; everything else fell through to the general ceiling.

`BlockResponse` stays at the ceiling — it's the one message that is legitimately large, and tightening it needs the requester's own pending-request state, which this framing layer doesn't have. #4140's "dedicated block sync streams or message chunking" bullet is the real fix there.

`Ping` is the one capped message whose size is not fixed: its block locators grow with chain height, up to `u32::MAX / CHECKPOINT_INTERVAL` checkpoints at 36 bytes each. 16 MiB is that arithmetic, asserted at compile time. It is far looser than any realistic height needs, but tightening it is a decision about the maximum chain height the wire format carries, not a framing change.

## Exhaustiveness

The per-ID table is keyed on a `MessageId` enum rather than a bare `u16`, and `Message::id`, `Message::read_le` and `MessageId::max_size` are all matches with no fallback arm. Adding a message variant without giving it an ID, or an ID without giving it a size, is a compile error. (`MessageId::from_u16` is the one direction the compiler can't check — a `u16` can hold values no variant claims — so a test covers it.)

## Not disconnecting honest peers

The risk in this change isn't a cap being too loose, it's a cap being too tight. There are compile-time assertions that the caps clear the worst case an honest node can produce for `PeerResponse` (255 IPv6 peers with heights: 6,124 bytes) and `Ping`, plus tests that check the largest honest message of each capped type against the real serializer.

## Test Plan

Unit and property testing of the codec: the size gate's boundary is pinned per ID; decoding is exercised at every chunk boundary of a real message; each capped type is tested from both sides of its cap.

End to end testing in deployed networks under load.

## Documentation

N/A

## Backwards compatibility

Honest traffic does not exceed these limits, and the tests above pin that for each capped type.

One case gets *more* permissive: `UnconfirmedTransaction` was capped at `LATEST_MAX_TRANSACTION_SIZE`, but the length prefix counts the whole frame, which is 39 bytes larger than the transaction it carries. A transaction of exactly the maximum size is valid — `ledger-service` and the REST endpoint both accept one — so `staging` today disconnects the peer relaying it. That off-by-39 is fixed here.

### DIFF
--- a/node/router/messages/src/helpers/codec.rs
+++ b/node/router/messages/src/helpers/codec.rs
@@ -65,18 +65,62 @@ impl<N: Network> Encoder<Message<N>> for MessageCodec<N> {
     }
 }
 
+/// The width, in bytes, of the length prefix on the wire. Matches `LengthDelimitedCodec`'s
+/// default (`.little_endian()` is the only builder option `MessageCodec` sets besides
+/// `max_frame_length`, so the rest of the format - a 4-byte length field, not itself included in
+/// the count, immediately followed by that many payload bytes - is `LengthDelimitedCodec`'s
+/// default framing).
+const LENGTH_PREFIX_SIZE: usize = 4;
+
+/// The width, in bytes, of the message ID that leads every payload (see `Message::id`/`ToBytes`).
+const ID_SIZE: usize = 2;
+
 impl<N: Network> Decoder for MessageCodec<N> {
     type Error = std::io::Error;
     type Item = Message<N>;
 
     fn decode(&mut self, source: &mut BytesMut) -> Result<Option<Self::Item>, Self::Error> {
-        // Decode a frame containing bytes belonging to a message.
-        let bytes = match self.codec.decode(source)? {
-            Some(bytes) => bytes,
-            None => return Ok(None),
-        };
+        // Wait for the length prefix to fully arrive.
+        if source.len() < LENGTH_PREFIX_SIZE {
+            return Ok(None);
+        }
+        let declared_len = u32::from_le_bytes(source[..LENGTH_PREFIX_SIZE].try_into().unwrap()) as usize;
 
-        Self::Item::check_size(&bytes)?;
+        // Reject a frame that couldn't be valid for any message type, without waiting for the
+        // rest of a peer-declared, possibly enormous, body to arrive.
+        if declared_len > self.codec.max_frame_length() {
+            return Err(std::io::Error::new(std::io::ErrorKind::InvalidData, "frame is too large"));
+        }
+
+        // Wait for the message ID - which leads the payload - to arrive. This is at most
+        // `LENGTH_PREFIX_SIZE + ID_SIZE` bytes, arriving effectively immediately regardless of how
+        // large the declared frame is, since it's the front of whatever the peer sends first.
+        if source.len() < LENGTH_PREFIX_SIZE + ID_SIZE {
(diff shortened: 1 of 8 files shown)


ljedrz (changes requested):
The early rejection on the `Decoder` side is practical and worth keeping - but what actually matters is dropping the `reserve`, which neither #4055 nor #4232 (one of which we want regardless) would fix, since both keep a length-delimited codec with a large ceiling.

I'd rather we keep `LengthDelimitedCodec` and only hand it the buffer once the full frame has arrived - that would kill the amplification angle in just a few lines, leaves `encode` untouched, and needs no per-ID table. Same fix applies to `EventCodec` (which is capped at 256 MiB).

The off-by-39 is accurate and worth its own PR.

The `Ping` specifically should rather have a much more restrictive number of checkpoint, as it's a message that's sent all the time, and keeping it small would be a measurable benefit (also in its own PR).

One correction: frames declaring fewer than 2 bytes don't park the connection on `staging` - `LengthDelimitedCodec` rejects at `len < 2`. That hazard is introduced by the hand-rolled decoding.

If we do want to fully drop the underlying codec, we should do a full fuzzing run before merging it. Alternatively, we can keep it, but hand it the buffer only after the frame has actually arrived, which would simplify the encoding.

All in all, there are some worthwhile changes in this PR, but we should tackle them individually, and try to reduce the scope of changes, and lean on well-tested plumbing (like `LengthDelimitedCodec`) when possible.

PR #4416, opened 2026-08-27
fix(router): account for the frame envelope in the transaction size cap

cbeck88:
## Motivation

Split out from #4413 per review

`check_size`'s special case for `UnconfirmedTransaction` (id 12) rejects the message if the whole frame exceeds `LATEST_MAX_TRANSACTION_SIZE`. But the frame is larger than the transaction it carries - message ID, transaction ID, and `Data`'s own version byte and length prefix - 39 bytes in total (2 + 32 + 5). A transaction of exactly the maximum size is valid (`ledger-service` and the REST endpoint both accept one), so on `staging` today the peer relaying it gets disconnected for relaying a transaction the network itself considers acceptable.

Fix: check the frame length against `LATEST_MAX_TRANSACTION_SIZE` plus that 39-byte envelope, not the cap alone.

## Test Plan

- `overly_large_unconfirmed_transaction`'s strategy generated exactly `LATEST_MAX_TRANSACTION_SIZE` junk bytes, which after this fix is a legitimate frame size, not an oversized one - bumped it to `+1` so it still tests a genuinely-too-large frame.
- Added `max_size_unconfirmed_transaction_is_accepted`, which is the actual regression test: a transaction body of exactly `LATEST_MAX_TRANSACTION_SIZE` now round-trips through the codec instead of being rejected.
- Full `snarkos-node-router-messages` suite (16 tests) and a `snarkos-node` compile check both pass.

## Backwards compatibility

Strictly more permissive: the only behavior change is that a previously-rejected, legitimately-sized transaction is now accepted.

### DIFF
--- a/node/router/messages/src/unconfirmed_transaction.rs
+++ b/node/router/messages/src/unconfirmed_transaction.rs
@@ -29,10 +29,12 @@ pub struct UnconfirmedTransaction<N: Network> {
 }
 
 /// The bytes an `UnconfirmedTransaction` frame carries in addition to the transaction itself:
-/// the message ID that precedes every message payload, `transaction_id`, and `Data`'s own
-/// version byte and length prefix around `transaction`. Kept next to `write_le` below, which is
-/// what actually defines these three numbers - changing that encoding without updating this
-/// would silently reintroduce the mismatch `check_size` was fixed to avoid.
+/// the message ID that precedes every message payload (written by `Message::write_le` in
+/// lib.rs, not here), `transaction_id`, and `Data`'s own version byte and length prefix around
+/// `transaction` - the latter two defined by `write_le` below. Changing either encoding without
+/// updating this would silently reintroduce the mismatch `check_size` was fixed to avoid; the
+/// `unconfirmed_transaction_overhead_matches_the_real_encoding` test pins it against the actual
+/// serializer rather than trusting this arithmetic on its own.
 pub const UNCONFIRMED_TRANSACTION_OVERHEAD: usize = 2 // the message ID
     + 32 // `transaction_id`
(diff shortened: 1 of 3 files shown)


node/router/messages/src/unconfirmed_transaction.rs
```diff
@@ -28,6 +28,17 @@ pub struct UnconfirmedTransaction<N: Network> {
...
+/// The bytes an `UnconfirmedTransaction` frame carries in addition to the transaction itself:
+/// the message ID that precedes every message payload (written by `Message::write_le` in
+/// lib.rs, not here), `transaction_id`, and `Data`'s own version byte and length prefix around
+/// `transaction` - the latter two defined by `write_le` below. Changing either encoding without
+/// updating this would silently reintroduce the mismatch `check_size` was fixed to avoid; the
+/// `unconfirmed_transaction_overhead_matches_the_real_encoding` test pins it against the actual
+/// serializer rather than trusting this arithmetic on its own.
+pub const UNCONFIRMED_TRANSACTION_OVERHEAD: usize = 2 // the message ID
+    + 32 // `transaction_id`
+    + 5; // `Data`'s own version byte and length prefix, around `transaction`
```
ljedrz: these values are unlikely to ever change, though it would be better to reference them directly, as long as we have all the necessary imports in place

ljedrz (dismissed):
LGTM, left a small comment.

node/router/messages/src/unconfirmed_transaction.rs
cbeck88: it doesn't appear that they are exported from snarkVM, but i declared the required constants and left a TODO

snarkworld (approved):
👍

PR #4417, opened 2026-08-27
fix(router): cap the size of a Ping message

cbeck88:
## Motivation

Ping had no message-specific size check at all, unlike UnconfirmedTransaction - it fell through to the general 128 MiB frame ceiling. Unlike every other message type, Ping has no fixed shape: its block_locators grow with chain height, up to NUM_RECENT_BLOCKS recent entries plus MAX_CHECKPOINTS checkpoints (BlockLocators::read_le's own ceiling), each a (u32, BlockHash) pair.

* Add a Ping-specific cap of 16 MiB, derived from that arithmetic (the "real" worst case is ~14.75 MiB)
* Use it in `fn check_size`

Exports MAX_CHECKPOINTS from the locators crate, which was private, so this cap can be derived from it instead of restating the bound.

## Test Plan

Pinned by a compile-time assertion so it can't silently drift below what the wire format actually permits. This is the theoretical ceiling, not a tight number.

It might be that there's a better bound we can justify using the maximum ping size, which would give better node hardening. Getting it down to 768kb would match the maximum size of an unconfirmed transaction.

## Documentation

N/A

## Backwards compatibility

This is backwards compatible.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5444,6 +5444,7 @@ name = "snarkos-node-router-messages"
 version = "4.9.0"
 dependencies = [
  "bytes",
+ "indexmap 2.14.0",
  "proptest",
  "snarkos-node-bft-events",
  "snarkos-node-network",
--- a/node/router/messages/Cargo.toml
+++ b/node/router/messages/Cargo.toml
@@ -45,6 +45,9 @@ features = [ "codec" ]
 [dependencies.tracing]
 workspace = true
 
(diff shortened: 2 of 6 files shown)


ljedrz (dismissed):
LGTM; I'd still restrict the number of checkpoints further, but that's not my call to make. This is purely an engineering hardening, and doesn't alter the status quo.

snarkworld (approved):
LGTM

PR #4418, opened 2026-08-27
fix clippy & cargo audit failing

cbeck88:
## Summary

`check-clippy` has been failing on `staging`'s own tip since #4406 landed
```
error: variant `Certificate` is never constructed
   --> node/bft/src/helpers/telemetry.rs:185:5
    |
178 | enum TelemetryUpdate<N: Network> {
    |      --------------- variant in this enum
...
185 |     Certificate(Box<CertificateMetadata<N>>),
    |     ^^^^^^^^^^^
```

This is because the variant is only used by tests now, so we mark it allow ignored.

---

Also, chacha20 0.10.1 was yanked, and we need to bump to 0.10.2.

## Fix

Add `#[cfg_attr(not(test), allow(dead_code))]` directly on the variant, scoped to exactly the case that's a false positive.

## Test plan
- [x] `cargo clippy -p snarkos-node-bft --lib -- -D warnings` passes (previously failed with the error above)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -681,9 +681,9 @@ dependencies = [
 
 [[package]]
 name = "chacha20"
-version = "0.10.1"
+version = "0.10.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d524456ba66e72eb8b115ff89e01e497f8e6d11d78b70b1aa13c0fbd97540a81"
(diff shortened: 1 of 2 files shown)


ljedrz (approved):
It's already folded into some other PRs, but there should be no issue if it's merged sooner individually.

PR #4419, opened 2026-08-27
feat(router): cap each connection's frame length at its allowed types' max

cbeck88:
## Summary

Redevelops #4414 around a simpler idea after review:

1. A codec may permit only a subset of the possible message IDs to be decoded - the router codec knows it's only supposed to get certain messages, the gateway codec a different set, and so on.
2. The codec's maximum supported length is computed from that set of allowed messages, so excluding the largest messages reduces the memory the codec needs when handling untrusted input.
3. The codec itself also rejects message types that aren't allowed, rather than only bounding their size.

With this redevelop, we continue using `LengthDelimitedCodec` as requested - so it should stay compatible with either of the two PRs Lukasz pointed to in review (#4055, #4232).

For a validator's router connections specifically:

```
┌────────────────────┬─────────┬─────────────────┐
│                    │  limit  │ who can trigger │
├────────────────────┼─────────┼─────────────────┤
│ before this stack  │ 128 MiB │ any P2P peer    │
├────────────────────┼─────────┼─────────────────┤
│ after this PR      │  16 MiB │ any P2P peer    │
└────────────────────┴─────────┴─────────────────┘
```

`Client` and `Prover` router connections are unchanged (they legitimately expect `BlockResponse`, so they keep the general 128 MiB ceiling); `node/bft/events` and `node/src/bootstrap_client` are separate codecs, also unchanged - out of scope here.

The 16 MiB here is `Ping`'s own cap (the largest of what remains once `BlockResponse` is excluded) - if that can be tightened without breaking compatibility, this ceiling would shrink further, to roughly 1 MiB (max size of unconfirmed transaction). That's follow-up work, not part of this PR.

## What changed

- **`MessageId`**: a new enum mirroring `Message`'s wire IDs, with `max_size::<N>()` giving each type's worst-case frame size. `BlockResponse` stays at the general 128 MiB ceiling (legitimately large, bounded by block count rather than byte size); `Ping` and `UnconfirmedTransaction` keep their existing dedicated caps from the earlier PRs in this stack; everything else gets a new `MAX_SMALL_MESSAGE_SIZE` (8 KiB) cap.
- **`Message::check_size`** now looks up `MessageId::max_size` instead of special-casing two IDs by hand - so `PuzzleResponse` and `UnconfirmedSolution` are finally capped too (previously unbounded up to the general 128 MiB ceiling - this was one of the independently-reported findings this whole stack traces back to).
- **`MessageCodec::for_allowed_ids` / `excluding`**: build a codec whose `max_frame_length` is the max over a given set of allowed IDs, rather than always the general ceiling. `excluding` also rejects any frame whose ID is excluded outright, the moment the ID is visible - not just an oversized one. A validator's router connections now use `excluding(&[MessageId::BlockResponse])` (validators sync blocks exclusively over the committee-gated BFT gateway, never issuing a `BlockRequest` over the router, so a `BlockResponse` there is never legitimate). This closes the 128 MiB allocation an unsolicited `BlockResponse` could previously force, and gives a clearer rejection than relying on the later, semantic-level `Validator::block_response` handler alone.
- **`Message::id()`** now returns `MessageId` instead of a raw `u16`. Combined with `MessageId::max_size`, both matches are now exhaustive with no fallback arm, so a new `Message` variant forces a `MessageId` variant - and, in turn, a cap - before the crate compiles.

## Test plan
- [x] `cargo test -p snarkos-node-router-messages --all-targets` - 33 tests, including a full sweep of `MessageId::all()` pinning every type's size-gate boundary exactly, and dedicated tests for `excluding` (ceiling shrinks; an excluded type is rejected even when small and well-formed; an unrelated type still decodes normally).
- [x] `cargo +nightly-2026-04-02 fmt --check` on the touched crates.
- [x] Full workspace build (`cargo build -p snarkos`) and `cargo clippy -p snarkos-node-router-messages -p snarkos --all-targets -- -D warnings`.

Based on #4417 (Ping cap), which is based on #4416 (transaction frame overhead) and #4418 (an unrelated clippy hotfix it needed to build on).

## Backwards compatibility

Excluding `BlockResponse` from a validator's router codec is not a behavior change for any honest peer: [`Validator::block_response`](https://github.com/ProvableHQ/snarkOS/blob/126536511b390eb48d11dfe8041552fe09bbff60/node/src/validator/router.rs#L222-L230) already unconditionally rejects a `BlockResponse` received over the router today, regardless of this PR:

```rust
fn block_response(
    &self,
    peer_ip: SocketAddr,
    _blocks: Vec<Block<N>>,
    _latest_consensus_version: Option<ConsensusVersion>,
) -> bool {
    warn!("Received a block response through P2P, not BFT, from {peer_ip}");
    false
}
```

No honest peer sends a validator a `BlockResponse` over the router in the first place - a validator never issues a `BlockRequest` there (it syncs blocks exclusively over the committee-gated BFT gateway), so nothing it receives is ever a solicited response. This PR just moves the rejection earlier - by ID, before the frame is even sized or deserialized - rather than changing what ultimately gets accepted.

### DIFF
--- a/node/router/messages/src/helpers/codec.rs
+++ b/node/router/messages/src/helpers/codec.rs
@@ -13,7 +13,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use crate::Message;
+use crate::{MAX_SMALL_MESSAGE_SIZE, Message, MessageId};
 use snarkvm::prelude::{FromBytes, Network, ToBytes};
 
 use ::bytes::{Buf, BufMut, BytesMut};
@@ -22,9 +22,6 @@ use tokio_util::codec::{Decoder, Encoder, LengthDelimitedCodec};
 
 /// The maximum size of a message that can be transmitted during the handshake.
 const MAXIMUM_HANDSHAKE_MESSAGE_SIZE: usize = 1024 * 1024; // 1 MiB
-
-/// The maximum size of a message that can be transmitted in the network.
-pub(crate) const MAXIMUM_MESSAGE_SIZE: usize = 128 * 1024 * 1024; // 128 MiB
 
 /// The codec used to decode and encode network `Message`s.
 pub struct MessageCodec<N: Network> {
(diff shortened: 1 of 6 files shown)


ljedrz (dismissed):
Looks reasonable to me; most of the diff is new tests, and the new boilerplate is pretty straightforward.

snarkworld (approved):
👍

PR #4432, opened 2026-09-02
Chore/bump rust 1.96

cbeck88:
* Bump snarkVM to latest staging
* Bump rust to 1.96 in toolchain and MSRV (otherwise snarkVM doesn't build)
* `clippy fix --workspace`
* Manually ignore the mutable key error after reviewing the possible impact

### DIFF
--- a/node/bft/src/primary.rs
+++ b/node/bft/src/primary.rs
@@ -2048,6 +2048,7 @@ impl<N: Network> Primary<N> {
         }
 
         // Fetch the missing previous certificates.
+        #[allow(clippy::mutable_key_type)]
         let missing_previous_certificates =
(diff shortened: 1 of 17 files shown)


node/bft/src/primary.rs
```diff
@@ -2051,6 +2048,7 @@ impl<N: Network> Primary<N> {
         }
 
         // Fetch the missing previous certificates.
+        #[allow(clippy::mutable_key_type)]
```
cbeck88, line 2049: see commit message -- this is happening because we added a OnceLock cache on the `BatchCertificate` so that we don't recompute signers over and over again. But it doesn't affect the integrity of hash set here because that cache doesn't participate in Eq, PartialEq, or Hash implementations.

vicsn (dismissed):

vicsn (approved):

Issue #4431, opened 2026-09-01
[Bug] BFT gateway disconnects a peer the moment our own outbound queue to it is full


cbeck88:
Split out from #4394 (closed). One of several follow-ups from that PR's closing discussion.

## The behaviour

`Gateway::send_inner` disconnects a peer whenever `unicast` returns an error:

```rust
let result = self.unicast(peer_addr, event);
// If the event was unable to be sent, disconnect.
if let Err(err) = &result {
    warn!("{CONTEXT} Failed to send '{name}' to '{peer_ip}': {err:?}");
    debug!("{CONTEXT} Disconnecting from '{peer_ip}' (unable to send)");
    self.disconnect(peer_ip);
}
```

https://github.com/ProvableHQ/snarkOS/blob/2a3b2c24ce366698b0fc4c73cd544bb6bd6f223a/node/bft/src/gateway.rs#L623-L641

The error this reacts to is not a socket error. `unicast` does a non-blocking `try_send` into a
bounded per-peer channel, and collapses *every* `try_send` failure into `io::ErrorKind::Other`:

https://github.com/ProvableHQ/snarkOS/blob/2a3b2c24ce366698b0fc4c73cd544bb6bd6f223a/node/tcp/src/protocols/writing.rs#L120-L133

The information is lost twice over. First in `unicast`: `TrySendError::Full` — ordinary
backpressure, the queue is simply at capacity — and `TrySendError::Closed` — the writer task is
gone, the peer really is finished — arrive as the same `Other`. Then in `send_inner`, which
disconnects on *any* `Err` and so does not distinguish `Other` from the `NotConnected` and
`Unsupported` cases `unicast` does still report separately
([writing.rs#L114-L119](https://github.com/ProvableHQ/snarkOS/blob/2a3b2c24ce366698b0fc4c73cd544bb6bd6f223a/node/tcp/src/protocols/writing.rs#L114-L119)). A full queue and a broken peer
take the same path.

The outbound queue is `per_connection_queue_depth()` slots
([gateway.rs#L147-L151](https://github.com/ProvableHQ/snarkOS/blob/2a3b2c24ce366698b0fc4c73cd544bb6bd6f223a/node/bft/src/gateway.rs#L147-L151)) — on mainnet today
`2 * 3 * 40 * (50 + 1)` = 12,240 — so this fires when a peer is 12,240 events behind, which is a
load condition, not a fault.

## Why it is wrong

[...]

Issue #4429, opened 2026-09-01
[Proposal] Improvements around gateway's per-peer outbound rate limiter


cbeck88:
Split out from #4394 (closed). One of several follow-ups from that PR's closing discussion.
#4394 attempted to fix internals of this mechanism. We decided to take a step back and
investigate whether this mechanism is actually needed or if it could be simplified instead.

## What it is

Every outbound gateway event passes through a self-imposed rate limit: we count what we have sent
to a peer in a recent window and, if we are over our own ceiling, sleep 10ms and re-check, in a
loop, before sending.

```rust
macro_rules! send {
    ($self:ident, $cache_map:ident, $interval:expr, $freq:ident) => {{
        // Rate limit the number of certificate requests sent to the peer.
        while $self.cache.$cache_map(peer_ip, $interval) > $self.$freq() {
            // Sleep for a short period of time to allow the cache to clear.
            tokio::time::sleep(Duration::from_millis(10)).await;
        }
        // Send the event to the peer.
        $self.send_inner(peer_ip, event)
    }};
}
```

https://github.com/ProvableHQ/snarkOS/blob/2a3b2c24ce366698b0fc4c73cd544bb6bd6f223a/node/bft/src/gateway.rs#L1395-L1405

Note this is not the recipient limiting us — we limit ourselves before sending.

## The thresholds are decoupled from the queue they feed

All values below are mainnet, non-test, at the latest consensus version, and were re-derived from
the source rather than taken from the PR discussion:

| quantity | derivation | value |
|---|---|---|
| `QUEUE_WINDOW_ROUNDS` | `MAX_FETCH_TIMEOUT / MAX_BATCH_DELAY` = 7500ms / 2500ms | 3 |
| `per_connection_queue_depth()` | `2 × 3 × LATEST_MAX_CERTIFICATES(40) × (MAX_TRANSMISSIONS_PER_BATCH(50) + 1)` | **12,240** |
| `max_cache_certificates()` | `2 × MAX_GC_ROUNDS(100) × max_committee_size(40)` | 8,000 |
| `max_cache_transmissions()` | `8,000 × MAX_TRANSMISSIONS_PER_BATCH(50)` | **400,000** |
| `max_cache_events()` | `= max_cache_transmissions()` | **400,000** |
| `CACHE_EVENTS_INTERVAL` / `CACHE_REQUESTS_INTERVAL` | `MAX_BATCH_DELAY.as_secs()` | **2 s** |

[...]

Issue #4428, opened 2026-09-01
[Bug] BFT proposal sends outlive the round they were spawned for, with no bound and no cancellation


cbeck88:
Split out from #4394 (closed). One of several follow-ups from that PR's closing discussion.

## The concern

`propose_batch` takes a write lock for its whole body, so the primary is only ever *proposing* for
one round at a time:

```rust
async fn propose_batch(&self) -> Result<bool> {
    // Ensure there are not concurrent executions of this function.
    let mut lock_guard = self.latest_proposal_timestamp.write().await;
```
https://github.com/ProvableHQ/snarkOS/blob/2a3b2c24ce366698b0fc4c73cd544bb6bd6f223a/node/bft/src/primary.rs#L470-L475

That lock scopes the *decision* to propose. It does not scope the *sends* that decision launches.
Every send `propose_batch` starts is handed to a detached `tokio::spawn` whose `JoinHandle` is
dropped on the spot, so all of them outlive both the lock and the round. When the primary moves on
to round R+1, nothing tells the sends still in flight for round R to stop.

## How a send outlives its round

`Gateway::send` is `async` and parks before it queues anything, in the outbound rate limiter's
poll-and-sleep loop. The loop has no deadline — it sleeps 10ms and rechecks, indefinitely, for as
long as the peer stays over the limit:

```rust
while $self.cache.$cache_map(peer_ip, $interval) > $self.$freq() {
    tokio::time::sleep(Duration::from_millis(10)).await;
}
$self.send_inner(peer_ip, event)
```
https://github.com/ProvableHQ/snarkOS/blob/2a3b2c24ce366698b0fc4c73cd544bb6bd6f223a/node/bft/src/gateway.rs#L1395-L1405

So a task spawned for round R can sit parked across an arbitrary number of subsequent rounds and
then emit a `BatchPropose` for R once it finally clears. Nothing in the task rechecks the round it
was spawned for, because nothing can: the round is captured only to be printed in a log line.

Both proposal send paths have this shape.

**1. The initial broadcast.** `propose_batch` publishes the new header via `Gateway::broadcast`:
https://github.com/ProvableHQ/snarkOS/blob/2a3b2c24ce366698b0fc4c73cd544bb6bd6f223a/node/bft/src/primary.rs#L813

`broadcast` is not `async` — it spawns one detached task that walks the connected peers and
`await`s `send` for each *sequentially*:
[...]

PR #4435, opened 2026-09-04
feat(bft): instrument the gateway's outbound queue and rate limiter

### LINKED ISSUE #4427: [Proposal]  Instrument gateway outbound queue depth and rate-limiter engagement
Split out from #4394 (closed). Requested during that PR's review by @vicsn and @ljedrz, and a
prerequisite for the other follow-ups from it.

## Why

Review of #4394 stalled on a question nobody could answer from the code: do we ever actually hit
these limits? @ljedrz: *"I'm not sure if we've ever hit these limits before - the values seem like
worst-case scenario, not strict fine-tunes. I agree with @vicsn that additional instrumentation
(even temporary) would improve our understanding of the scale of the issue and the fix."*

The derivation says we do not — the outbound rate limiter permits 400,000 sends to a peer per
2-second window into a queue holding 12,240 slots, so the queue should fill and the peer should be
disconnected long before the limiter engages (see the rate-limiter issue for the full arithmetic).
But that is a derivation, not a measurement, and the decisions that depend on it are significant:
whether to delete the rate limiter, and how aggressively to retry instead of disconnecting.

## What to measure

Per peer, on the gateway:

- **Outbound queue occupancy** — current depth and high-water mark against
  `per_connection_queue_depth()` (12,240 on mainnet today). This is the number that tells us
  whether `send_inner`'s disconnect-on-full is a live path or a theoretical one.
- **`send_inner` disconnects attributable to a full queue**, counted separately from other
  disconnect causes. Today `unicast` collapses `TrySendError::Full` and every other failure into
  `io::ErrorKind::Other`, so these are not currently distinguishable — separating them is part of
  the work.
  https://github.com/ProvableHQ/snarkOS/blob/2a3b2c24ce366698b0fc4c73cd544bb6bd6f223a/node/tcp/src/protocols/writing.rs#L120-L133
- **Rate-limiter engagement** — how often the `send!` macro's `while` loop iterates at all, and for
  how long, broken down by the four limit classes (`max_cache_certificates`,
  `max_cache_transmissions`, `max_cache_events`, and the block-request path).
- **Outstanding spawned send tasks**, as a gauge. Currently unmeasurable, since the `JoinHandle`s
  are dropped at every call site (see the spawn/cancellation issue) — a counter incremented at
  spawn and decremented on completion would do.
- **`Writing::TIMEOUT` write-timeout disconnects** (Mechanism B), so the two disconnect causes can
  be told apart in production data.

## Notes

- Temporary instrumentation is acceptable per the review discussion, but most likely it's preferable
  to have permanent metrics for this unless there is some serious performance degradation for some reason.
- The existing `metrics` feature in `node/bft` is the natural home for the counters.
- This is deliberately scoped to measurement only — no behaviour change. It should land before the
  rate-limiter decision so that decision is made on data.

cbeck88:
## Motivation

Closes #4427.

Review of #4394 stalled on a question nobody could answer from the code: do we ever actually hit the outbound rate limits, and is `send_inner`'s disconnect-on-full-queue a live path or a theoretical one? The derivation says the queue fills long before the limiter engages -- 400,000 permitted sends per 2-second window into a queue holding 12,240 slots -- but that is a derivation, not a measurement, and two significant decisions rest on it: whether to delete the rate limiter (#4429), and whether to retry instead of disconnecting (#4431).

This is measurement only; no send, disconnect or rate-limiting behaviour changes. It should land before those decisions so they are made on data.

* Outbound queue occupancy. `Writing` gains `outbound_queue_occupancy` and `outbound_queue_occupancies`, which read depth and capacity straight off the per-connection `mpsc::Sender`. The gateway's existing 1 Hz metrics loop samples every peer into a histogram and publishes the current maximum and the queue capacity, so the ratio against `per_connection_queue_depth()` is a PromQL expression. A queue that fills and drains between two ticks would be invisible to a sampler, so the high-water mark is instead maintained by `send_inner` on every successful send, where it costs one relaxed `fetch_max`.
* Queue-full disconnects. `unicast` collapsed `TrySendError::Full` and every other failure into `io::ErrorKind::Other`, so backpressure disconnects could not be told apart from any other cause. `Full` now maps to `WouldBlock` and `Closed` to `BrokenPipe`, and `send_inner` counts failures by reason. No caller matched on the old kind. The queue-full warning also now names the queue depth, since that case is finally distinguishable in the logs.
* Rate-limiter engagement. The `send!` macro takes a label for its limit class and counts sends that enter the retry loop at all, iterations of that loop, and the total wait, broken down by `certificates`, `transmissions`, `events` and `block_request` -- the last shares the general limit but is counted apart, as it is the path the sync layer drives. The accounting is a drop guard rather than a pair of statements around the loop: these sends run in spawned tasks whose join handles are dropped at the call site (#4428), so they can be cancelled at any await point, and a cancelled send would otherwise leak the gauge and lose its wait.
* Outstanding send tasks. A gauge held for the lifetime of each `Transport::send` call. Nearly every call site spawns one task per event, so this is the count the issue asks for, without having to instrument roughly fifteen spawn sites individually and keep them in sync.
* Write-timeout disconnects. The `node/tcp` writer task now separates `Writing::TIMEOUT` expiry from other write failures, so Mechanism B can be told apart from Mechanism A in production data. `snarkos-node-bft`'s `metrics` feature now enables `snarkos-node-tcp/metrics` so the two arrive together.

Test plan: `full_outbound_queue_is_reported_as_would_block` covers the one externally visible change, that a full queue is now reported as `WouldBlock` and that occupancy tracks it. `cargo test -p snarkos-node-tcp --lib` and `cargo test -p snarkos-node-bft --lib` pass, and clippy is clean for the touched crates both with and without the `metrics` feature.

## Test Plan

This is non-functional change, but we can test it using a local devnet

## Backwards compatibility

This doesn't change any existing metrics, and none of the existing metrics overlap this

### DIFF
--- a/node/bft/src/gateway.rs
+++ b/node/bft/src/gateway.rs
@@ -228,6 +228,9 @@ pub struct InnerGateway<N: Network> {
     resolver: RwLock<Resolver<N>>,
     /// The collection of both candidate and connected peers.
     peer_pool: RwLock<HashMap<SocketAddr, Peer<N>>>,
+    /// The deepest per-connection outbound queue seen since the node started, across all peers.
+    #[cfg(feature = "metrics")]
+    outbound_queue_high_water_mark: std::sync::atomic::AtomicUsize,
     /// The handle to the validator telemetry tracker.
     #[cfg(feature = "metrics")]
     validator_telemetry: Telemetry<N>,
@@ -343,6 +346,8 @@ impl<N: Network> Gateway<N> {
             cache: Default::default(),
             resolver: Default::default(),
             peer_pool: RwLock::new(initial_peers),
+            #[cfg(feature = "metrics")]
+            outbound_queue_high_water_mark: Default::default(),
             #[cfg(feature = "metrics")]
             validator_telemetry,
             #[cfg(feature = "metrics")]
@@ -593,6 +598,35 @@ impl<N: Network> Gateway<N> {
         if let Some(count) = self.number_of_connecting_peers() {
             metrics::gauge(metrics::bft::CONNECTING, count as f64);
         }
+        self.update_outbound_queue_metrics();
+    }
+
+    /// Samples the occupancy of every per-connection outbound queue.
+    ///
+    /// The high-water mark this publishes is maintained by [`Self::send_inner`], not by this
+    /// sampler: a queue that fills and drains between two ticks is invisible here.
+    #[cfg(feature = "metrics")]
+    fn update_outbound_queue_metrics(&self) {
+        let mut deepest = 0;
+        for (_peer_addr, depth, _capacity) in self.outbound_queue_occupancies() {
+            metrics::histogram(metrics::bft::GATEWAY_OUTBOUND_QUEUE_DEPTH, depth as f64);
+            deepest = deepest.max(depth);
+        }
+        metrics::gauge(metrics::bft::GATEWAY_OUTBOUND_QUEUE_DEPTH_MAX, deepest as f64);
+        metrics::gauge(metrics::bft::GATEWAY_OUTBOUND_QUEUE_CAPACITY, per_connection_queue_depth::<N>() as f64);
+        metrics::gauge(
+            metrics::bft::GATEWAY_OUTBOUND_QUEUE_HIGH_WATER_MARK,
+            self.outbound_queue_high_water_mark.load(std::sync::atomic::Ordering::Relaxed) as f64,
+        );
+    }
+
+    /// Raises the outbound queue high-water mark to the current depth of the given peer's queue,
+    /// if that is a new maximum.
+    #[cfg(feature = "metrics")]
+    fn record_outbound_queue_depth(&self, peer_addr: SocketAddr) {
+        if let Some((depth, _capacity)) = self.outbound_queue_occupancy(peer_addr) {
+            self.outbound_queue_high_water_mark.fetch_max(depth, std::sync::atomic::Ordering::Relaxed);
+        }
     }
 
     /// Inserts the given peer into the connected peers. This is only used in testing.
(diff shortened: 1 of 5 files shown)


node/bft/src/gateway.rs
```diff
@@ -593,6 +598,35 @@ impl<N: Network> Gateway<N> {
...
+    fn update_outbound_queue_metrics(&self) {
+        let mut deepest = 0;
+        for (_peer_addr, depth, _capacity) in self.outbound_queue_occupancies() {
+            metrics::histogram(metrics::bft::GATEWAY_OUTBOUND_QUEUE_DEPTH, depth as f64);
+            deepest = deepest.max(depth);
+        }
+        metrics::gauge(metrics::bft::GATEWAY_OUTBOUND_QUEUE_DEPTH_MAX, deepest as f64);
+        metrics::gauge(metrics::bft::GATEWAY_OUTBOUND_QUEUE_CAPACITY, per_connection_queue_depth::<N>() as f64);
+        metrics::gauge(
+            metrics::bft::GATEWAY_OUTBOUND_QUEUE_HIGH_WATER_MARK,
```
vicsn, line 618: I recall a suggestion from an infra colleague that we should try and offer base metrics from which Grafana may be able to derive other metrics (like the maximum observed). So the high water mark could be removed.

vicsn (changes requested):
Mostly looks good to me, one small nit.

Low priority thoughts, feel free to ignore: More metrics seems great in general. The only potential downside I can think of is that we may pay for it in infrastructure costs. At low granularity it's fine, at high granularity it may be an annoyance. AI is telling me `GATEWAY_RATE_LIMIT_DELAY` is the most expensive and we can pass `set_quantiles(&[0.5, 0.99])` when building the recorder to take every histogram from 9 series to 4.

Also, the person who typically maintains our snarkos-incident-channel dashboards is on leave, so it may take a while before we can use this. Let me know if you're up for editing the reference Grafana dashboards.

PR #4438, opened 2026-09-07
fix(sync): Enable block syncing redundancy

brunoffranca:
## Summary

This PR enables block-sync redundancy in production (previously disabled outside tests) and reworks the sync logic to support competing/conflicting block responses per height. Instead of assuming a single canonical block per height, `BlockSync` now tracks all distinct candidate blocks reported by peers and lets both the ledger-advance and BFT-sync paths try each candidate in turn, falling back to the next one if a candidate fails to apply (e.g. it lost a fork). The peer-vouching threshold used to accept a height/hash is also made adaptive to the number of currently known sync peers, so nodes with fewer peers than `REDUNDANCY_FACTOR` can still sync.

## Changes

- Set `REDUNDANCY_FACTOR = 3` unconditionally (was 1 in non-test builds).
- Rewrote construct_request to compute an adaptive `peer_threshold = min(REDUNDANCY_FACTOR, sync_peers.len())` and pick the most-vouched-for `(hash, previous_hash)` pair via frequency counting, replacing the old dishonest-peer/extra-redundancy logic.
- Added/updated unit tests covering insufficient redundancy and multi-peer request construction, and adjusted existing tests to register `REDUNDANCY_FACTOR` peers per scenario.
- `OutstandingRequest.response (Option<Block>)` replaced with `response_map: HashMap<BlockHash, Block>`, and `peek_next_block` now returns `Option<Vec<Block>>` of all reported candidates for a height instead of a single block.
- `node/sync/src/block_sync.rs`'s ledger-advance loop and `node/bft/src/sync/mod.rs`'s BFT-sync loop iterate over candidate blocks per height, accepting the first one that applies successfully.

## Test Plan

The redundancy changes are covered by 5 unit tests: 4 existing tests broke when `REDUNDANCY_FACTOR` was raised to 3 and were updated to register the corresponding number of peers per scenario, and 1 new test was added covering `construct_request`'s adaptive peer-threshold/frequency-counting logic. All 5 pass locally.

## Backwards compatibility

This PR doesn't touch any public interfaces. On the sync side, the changes only affect how this node evaluates and picks among the candidate blocks it receives while syncing — not how it responds to other peers' block requests, so this node keeps serving requests to peers on older versions the same way as before. Since nothing here changes consensus-critical validation rules or the block/request wire format, I don't think this needs to be gated behind a `ConsensusVersion`.

## Notes

- This is not a proper solution but rather a quick fix. The more resilient way to solve this would be to not stop the consensus participation when the node enters syncing mode.
- The adaptive peer threshold in `construct_request` is a compromise. If we don't adapt the threshold, then every sync update has more certainty, but then we won't be able to sync if we are connected to fewer peers than the redundancy factor.
- It was also discussed temporarily banning peers that provide invalid blocks or fail to respond. I decided that it would be better to do in a separate PR, if we decide to do it.

### DIFF
--- a/node/sync/src/block_sync.rs
+++ b/node/sync/src/block_sync.rs
@@ -101,9 +101,9 @@ type FailedRequests<H> = BTreeMap<u32, (Option<H>, Option<H>)>;
 struct OutstandingRequest<N: Network> {
     request: SyncRequest<N>,
     timestamp: Instant,
-    /// The corresponding response (if any).
-    /// This is guaranteed to be Some if sync_ips for the given request are empty.
-    response: Option<Block<N>>,
+    /// The candidate blocks received for this request, keyed by block hash.
+    /// This is guaranteed to be non-empty if sync_ips for the given request are empty.
+    response_map: HashMap<N::BlockHash, Block<N>>,
 }
 
 /// Information about a block request (used for the REST API).
@@ -146,8 +146,6 @@ pub enum InsertBlockResponseError<N: Network> {
         "The previous block hash in candidate block {height} from '{peer_ip}' is incorrect: expected {expected}, but got {actual}"
     )]
     InvalidPreviousBlockHash { height: u32, peer_ip: SocketAddr, expected: N::BlockHash, actual: N::BlockHash },
-    #[error("Candidate block {height} from '{peer_ip}' is malformed")]
-    MalformedBlock { height: u32, peer_ip: SocketAddr },
     #[error("The sync pool did not request block {height} from '{peer_ip}'")]
     WrongSyncPeer { height: u32, peer_ip: SocketAddr },
     #[error("{}", flatten_error(.0))]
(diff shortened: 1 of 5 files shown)


node/sync/src/block_sync.rs
```diff
@@ -66,10 +66,6 @@ use metrics::BlockSyncMetrics;
 
 // The redundancy factor decreases the possibility of a malicious peers sending us an invalid block locator
 // by requiring multiple peers to advertise the same (prefix of) block locators.
-// However, we do not use this in production yet.
-#[cfg(not(test))]
-pub const REDUNDANCY_FACTOR: usize = 1;
-#[cfg(test)]
```
cbeck88, line 72: i am not sure we should actually change this from 1 in prodcution, it isn't well tested and when i looked at it before i wasn't sure if it would actually work. i'll try to remember what my concerns were

historically aleo was proof of work when it started and transition to this narwhal / proof of stake model after. some of these kind of constants may have only made sense in the proof of work and may now be legacy / not really changeable anymore

vicsn: Agreed that it is very sensitive code, but I've asked Bruno to focus his full attention on making this more robust along the way. It's been a long known painpoint and he should have all the means to propose the right adjustment. He just told me that he still has to run some extended testing before marking this ready for review.

PR #4440, opened 2026-09-08
perf(bft): isolate block construction on a dedicated rayon pool

vicsn:
## Summary

- Adds a process-wide block-prep `rayon::ThreadPool` (50% of logical cores, 8 MiB stacks) and stores `Arc<ThreadPool>` on `CoreLedgerService` / `LedgerUpdate`.
- Runs `prepare_advance_to_next_quorum_block`, `check_next_block`, and `check_block_content` inside `pool.install(...)` so nested snarkVM `par_iter` / `execute_with_max_available_threads` inherit that pool instead of saturating the global one.
- Leaves proposal-time work (`check_transaction_basic`, `rayon::spawn_fifo` deserialization, etc.) on the process-wide pool, with the goal of reducing latency variance while a block is being constructed.

@niklaslong @ljedrz — would you be up for running this on a DPS client to evaluate the impact on latency variance?

A GitHub Actions benchmark run is being dispatched on this branch (`workflow_dispatch` of `.github/workflows/benchmarks.yml`). Results stay in the workflow summary and are not pushed to `gh-pages`.

## Test plan

- [x] `cargo +nightly fmt`
- [x] `cargo test -p snarkos-node-bft-ledger-service --features ledger,ledger-write --lib block_prep_pool`
- [x] `cargo test -p snarkos-node-bft --lib`
- [ ] GitHub Actions snarkOS benchmarks on this branch
- [ ] DPS client run to compare proposal / commit latency variance vs staging
- [ ] Optional before merge: `.ci/test_devnet.sh` (not run locally)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5254,6 +5254,7 @@ dependencies = [
  "async-trait",
  "indexmap 2.14.0",
  "locktick",
+ "num_cpus",
  "parking_lot",
  "rand 0.10.2",
  "rayon",
--- a/node/bft/ledger-service/Cargo.toml
+++ b/node/bft/ledger-service/Cargo.toml
@@ -18,7 +18,7 @@ edition = "2024"
 
 [features]
 default = [ ]
-ledger = [ "parking_lot", "rand", "rayon", "tokio", "tracing" ]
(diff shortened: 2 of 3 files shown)


ljedrz (approved):
LGTM

PR #4441, opened 2026-09-08
build: remove lld from cargo config

brunoffranca:
## Summary

Removes the lld linker configuration from the cargo build settings. No longer necessary since #4432 because lld became the default in Rust 1.90.0.

https://blog.rust-lang.org/2025/09/18/Rust-1.90.0/#lld-is-now-the-default-linker-on-x86-64-unknown-linux-gnu


cbeck88 (approved):
Looks good, I didn't know this became the default. Thank you

PR #4442, opened 2026-09-09
feat(metrics): publish HELP text for every metric, enforced at compile time

cbeck88:
## Motivation

No snarkOS metric has ever had a description. `metrics-exporter-prometheus` writes a `# HELP`
line only for names registered through `describe_*`, and we call none of those, so every scrape
we have ever produced carries `# TYPE` and nothing else.

The reason this matters more than it used to: **`HELP` text is documentation that travels with
the data.** It is served by Prometheus' own metadata API, and it is what Grafana's metric browser
shows. When you point an agent at a Prometheus instance and ask it what a node is doing, it can
of course also read this repository — but that means locating the emission site for each name and
reconstructing the meaning from the surrounding code, one metric at a time, before it can answer
anything. A description in the exposition collapses those hops into the scrape it has already
done. The context arrives with the data instead of having to be assembled from source.

Our own metric set has a pair in it today where that is the difference between a right and a
wrong answer:

```
snarkos_consensus_prepare_advance_secs
snarkos_consensus_prepare_advance_to_next_quorum_block_latency_secs
```

One is recorded on every block. The other was never wired up and has always been empty. Nothing
in the names says which, and finding out means going and reading two different call sites. After
this change the first reads *"Seconds spent in prepare_advance_to_next_quorum_block, i.e.
constructing the block."* and the second reads *"Never recorded. Superseded by
`snarkos_consensus_prepare_advance_secs`."*

That is also why the descriptions are mandatory rather than best-effort below. A metric set where
most things are described is one where an undescribed name reads as an oversight rather than as a
signal, which puts the reader back in the source for every name it is unsure about.

## What this does

`names.rs` becomes a single `declare_metrics!` block that generates both the name constants and
the registration table from one source. Each entry's doc comment becomes its `HELP` text:

```rust
declare_metrics! {
    pub mod bft {
        /// The number of validators this node is connected to, excluding bootstrap clients.
        Gauge CONNECTED = "snarkos_bft_connected_total";
    }
}
```

The doc comment is not optional. A declaration without one fails to compile, naming the metric:

```
error[E0080]: evaluation panicked: the metric `UNDOCUMENTED_ON_PURPOSE`
has no doc comment, so it has no HELP text to publish
```

The Rust doc and the Prometheus description are now the same string, so they cannot drift apart,
and a new metric cannot reach the exposition undescribed.

Resulting output:

```
# HELP snarkos_bft_connected_total The number of validators this node is connected to, excluding bootstrap clients.
# TYPE snarkos_bft_connected_total gauge
snarkos_bft_connected_total 0
```

### Kinds

A metric declares itself as `Counter`, `Gauge` or `Histogram`, or as `LabeledCounter`,
`LabeledGauge` or `LabeledHistogram` when it is only ever recorded with labels. Both get a
description, since `HELP` belongs to the metric name rather than to a series. Only the unlabelled
ones get a series created at startup: a labelled metric's series are defined by its label values,
so an unlabelled one would be an artefact that nothing ever updates.

### Drift this removes

`COUNTER_NAMES`, `GAUGE_NAMES` and `HISTOGRAM_NAMES` were maintained by hand alongside the
constants. Five metrics that are recorded in the code had never been added to them, so they were
absent from the exposition until first recorded:

```
snarkos_consensus_prepare_advance_secs
snarkos_consensus_check_next_block_secs
snarkos_consensus_advance_to_next_block_secs
snarkos_consensus_validator_participation_gc_round
snarkos_consensus_validator_participation_dropped_total
```

They are present from startup now. Unifying the declaration is what makes this class of mistake
impossible rather than merely fixed.

### Four metrics that are never recorded

Writing an honest description for everything surfaced four names that measure nothing:

- `snarkos_router_restricted_total` — the restricted-peers concept no longer exists in the router
  at all. The in-repo Grafana dashboard still plots it.
- Three consensus `*_latency_secs` histograms — never wired up, and superseded by the `_secs`
  metrics listed above.

Their descriptions say so rather than describing behaviour they do not have. They are left
registered so that this change removes nothing from the exposition; deleting them, and the
Grafana panel, is a follow-up.

## Test Plan

Two new tests in `node/metrics`:

- `metric_declarations_are_well_formed` checks the invariants the exposition format depends on:
  every description non-empty, single-line and unique, every name `snarkos_`-prefixed.
- `descriptions_reach_the_prometheus_exposition` builds the real exporter, runs the registration
  through a local recorder, and asserts the exact `# HELP` line for every pre-registered metric.
  This covers the whole path from doc comment to scrape output, which is otherwise only exercised
  by running a node and reading its metrics endpoint.

Clippy is clean for `snarkos-node-metrics` and its dependents with the `metrics` feature.

## Backwards compatibility

Additive. Pre-registered metrics go from 40 to 45; nothing is removed from the exposition, and no
metric changes name, kind or value. Dashboards and alerts are unaffected.

## Note on ordering

#4435 also touches `names.rs`. Whichever of the two lands second needs its metrics moved into the
`declare_metrics!` form — mechanical, and the compiler names anything missed.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01BhN4F74hAQ4Ewe7wAyJYGW

### DIFF
--- a/node/metrics/src/lib.rs
+++ b/node/metrics/src/lib.rs
@@ -77,19 +77,37 @@ pub fn initialize_metrics(ip: Option<SocketAddr>) {
     // Register the snarkVM metrics.
     snarkvm::metrics::register_metrics();
 
-    // Register the metrics so they exist on init.
-    for name in crate::names::GAUGE_NAMES {
-        register_gauge(name);
-    }
-    for name in crate::names::COUNTER_NAMES {
-        register_counter(name);
-    }
-    for name in crate::names::HISTOGRAM_NAMES {
-        register_histogram(name);
-    }
+    describe_and_register_metrics();
 
     // Set the build information metric
     set_build_info();
+}
+
+/// Publishes every declared metric's `HELP` text, and creates a series for each metric that is not
+/// labelled, so that it is present in the exposition before it is first recorded.
+///
+/// This acts on whichever recorder is currently installed.
+fn describe_and_register_metrics() {
+    for metric in crate::names::ALL_METRICS {
+        let description = metric.description.trim();
+        match metric.kind {
+            MetricKind::Counter => {
+                describe_counter(metric.name, description);
+                register_counter(metric.name);
+            }
+            MetricKind::Gauge => {
+                describe_gauge(metric.name, description);
+                register_gauge(metric.name);
+            }
+            MetricKind::Histogram => {
+                describe_histogram(metric.name, description);
+                register_histogram(metric.name);
+            }
+            MetricKind::LabeledCounter => describe_counter(metric.name, description),
+            MetricKind::LabeledGauge => describe_gauge(metric.name, description),
+            MetricKind::LabeledHistogram => describe_histogram(metric.name, description),
+        }
+    }
 }
 
 /// Stops the task running the metrics exporter, if one was started.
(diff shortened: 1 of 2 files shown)


ljedrz (commented):
I don't have strong feelings about this - more docs are fine, but the additional "plumbing" might be a bit of an overkill in case something needs changing in the future; that being said, I haven't worked on the metrics that much, so I'll let others decide.

PR #4444, opened 2026-09-09
fix(rest): raise body limit for current max transaction size

vicsn:
## Summary
- Raise the REST request body cap from 1.5MiB (`2 * 768 * 1024`, based on the old 768KB transaction limit) to `2 * LATEST_MAX_TRANSACTION_SIZE` (currently 4,608,000 bytes).
- Valid deployments and upgrades can already be up to 2,304,000 bytes, and their JSON representation can be larger, so the old cap rejected them before the binary size check.

## Backwards compatibility
The new limit is strictly higher than the previous one, and only impacts behaviour local to a node.


mohammadfawaz (approved):
Awesome, thank you! 🚀

PR #4445, opened 2026-09-09
build: improve dev compile times

brunoffranca:
## Summary

Changes the dev profile configuration to improve compile times. Also removes profile configuration that is redundant.

## Changes

- Changed the dev profile optimization level to 0. Improved compile time (full build) from 3m24s to 2m32s.
- Changed the dev profile debug info to line tables only. Higher levels of debug info are only useful for interactive debuggers. Improved compile time (full build) from 2m32s to 2m05s.
- Overall improved compile times for full builds by 39%. (measured with a single run though)
- Removed all configuration that was equal to the default profile values already. https://doc.rust-lang.org/cargo/reference/profiles.html#default-profiles

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -404,9 +404,7 @@ version = "0.9"
 version = "2"
 
 [profile.release]
-opt-level = 3
 lto = "thin"
-incremental = false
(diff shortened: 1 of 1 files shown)


cbeck88 (approved):
thank you, this simplifies things

Issue #4443, opened 2026-09-09
v1 REST error middleware renders every failure as "Failed to convert error", hiding rate limiting behind what looks like missing data


cbeck88:
## Summary

`v1_error_middleware` (`node/rest/src/lib.rs`, ~L374) collapses every non-success response into the literal string `Failed to convert error` whenever the body does not deserialize as `SerializedRestError`. A rate-limit rejection, a genuine "no value at this height", and an internal fault are then indistinguishable to a caller — all arriving as HTTP 500 with the same six words.

## How it bites

The default is `--rest-rps 10`. Sampling a mapping's history across a range of heights with a straightforward loop:

```bash
for h in 30000 40000 50000 60000 70000; do
  curl -s ".../program/credits.aleo/mapping/bonded/$ADDR/history/$h"
done
```

exceeds that after roughly eight requests, and every subsequent response is `Failed to convert error`. Because the limiter refills, the failures are not even contiguous — a later request can succeed while an earlier one failed.

That produces a result which looks exactly like a data problem: an apparent per-key "ceiling" in the history, at a different height for each key, that **moves between runs over identical, immutable data**.

We hit this while validating a history repair against archive fixtures, and it cost about an hour: the pattern was initially read as evidence of a storage-layer fault in the historical floor-seek path, and two synthetic reproductions of that non-existent bug had to pass before the real cause was found. Passing `--rest-rps 1000` makes every height return correctly.

An operator debugging genuine history trouble would follow the same false trail, and with more at stake.

## Why the message is empty

The fallback is reached when `serde_json::from_slice::<SerializedRestError>(&bytes)` fails — that is, whenever a non-success response was produced by something that does not emit that shape, such as the rate-limiting layer or a panic. So precisely the cases where a caller most needs to know what happened are the ones where the middleware discards it.

## Suggestions

- Preserve the upstream status code rather than mapping everything to 500. A 429 is self-explanatory in a way that 500 is not, and callers can back off correctly.
- When the body cannot be deserialized, include the original status in the fallback text (`"Failed to convert error (HTTP 429)"`) rather than dropping it.
- Consider whether the history endpoints should distinguish "no value at this height" from "this node does not have history for this height", which is a real distinction today for a node that has not been rebuilt and would become more important under any history-horizon design (see ProvableHQ/snarkVM#3418).

## Related

- ProvableHQ/snarkVM#3418 — history does not record deletions, so "no value" is already not trustworthy at the storage layer.

Issue #4337, opened 2026-07-03
[Bug] REST API returns 500 instead of 404 for blocks not yet synced (causes tip-race failures)


qnodeinfinite:
## 🐛 Bug Report

When a client requests a block by height and the node does not have that block yet, the REST API returns **HTTP 500 (internal_server_error)** instead of **HTTP 404 (not_found)**.

This happens on two routes:

- `GET /{network}/block/{height}` returns `{"message":"Failed to get a block's hash","error_type":"internal_server_error","chain":["Missing block hash for block N"]}`
- `GET /{network}/blocks?start=N&end=N+1` returns `{"message":"Block N does not exist in storage","error_type":"internal_server_error"}`

**Why this matters:** each node writes new blocks to its own database a few hundred milliseconds apart from its peers. When multiple nodes run behind a load balancer, a client that polls `/block/height/latest` and immediately fetches the new height can land on a node that is a few milliseconds behind. That node answers 500. Load balancers and monitoring treat backend 5xx as a node failure, so healthy nodes get pulled from rotation and clients report "periodic outages" for what is really just normal block propagation. In our production measurements, roughly 15-20% of fetches issued immediately after a height change hit this. A 404 would be accurate ("I don't have that block yet") and would not trip health checks.

**Where it comes from** (v4.8.0, `node/rest/src/routes.rs`):

In `get_block`, the height-to-hash lookup passes its error straight through:

```rust
let hash = if let Ok(height) = height_or_hash.parse::<u32>() {
    rest.ledger.get_hash(height).with_context(|| "Failed to get a block's hash")?
}
```

In `get_blocks`, each height in the range does the same:

```rust
.map(|height| rest.ledger.get_block(height))
.collect::<Result<Vec<_>, _>>()?;
```

Any error that isn't given a specific status falls into the default in `node/rest/src/helpers/error.rs`, which maps it to 500:

```rust
impl From<anyhow::Error> for RestError {
    fn from(err: anyhow::Error) -> Self {
        // Default to 500 Internal Server Error
        Self::InternalServerError(err)
    }
}
```

The fix pattern already exists a few lines later in the same `get_block` function: a failed hash-to-block lookup correctly returns `RestError::not_found(...)`. The height-to-hash miss and the `get_blocks` per-height miss just never got the same treatment.

## Steps to Reproduce

**Single node, deterministic** (ask for a height one past the tip):

```bash
BASE="http://localhost:3030"   # any snarkOS node with the REST server enabled
[...]

PR #4450, opened 2026-09-10
fix(rest): report missing resources as 404, and make v1 errors identifiable

### LINKED ISSUE #4097: [Bug] Some REST enpoints return `200` with `null` in their body, instead of `404`
## 🐛 Bug Report

Most of our endpoints correctly returns a `404 Not Found` error, when you query an unknown object, e.g., a block that does not exist on the node.
But there are some, such as the `stateRoot` endpoints, that return `200 OK` with `null` as their body. This is inconsistent and harder to parse.

We should verify that all endpoints return the correct error codes. However, this might require another REST version bump to not break existing behavior.

cbeck88:
## Motivation

Third in the stack, on top of #4448. That PR built a route-level test harness and used it on the three new range routes; this one points the same harness at the pre-existing REST surface, and fixes what that turned up.

Two real bugs, both of which had gone unnoticed because nothing exercised a route end to end:

**`map_missing_resource_error` only inspected the outermost message.** Callers attach context with `with_context`, which is what `to_string` then reports, so a marker added by the ledger ends up one level down where the check never looked. `/block/{height}` was the visible case: `get_block` wraps the ledger's "Missing block hash for block {h}" in "Failed to get a block's hash", so a height the node did not have came back as a 500 rather than a 404 -- the same failure #4337 was meant to have fixed. It now walks the whole chain. `/height/{hash}` and `/program/{id}` had the same symptom for a simpler reason, never calling the mapper at all, and are wired through it.

**The v1 error middleware discarded the status it replaced** (#4443). Any non-success response whose body is not a `SerializedRestError` became the literal string "Failed to convert error". Since v1 maps every error to 500, a rate-limited caller had no way to learn that -- a `--rest-rps` rejection, missing data and an internal fault were the same six words. The fallback now keeps the original body when there is one and names the status, so the rate limit case reads `Too Many Requests! Wait for 0s (HTTP 429)`. The status stays 500, so the v1 contract itself is unchanged.

One small refactor to make the above testable: `spawn_server`'s router construction moves into `build_versioned_router`, so the version prefixes can be exercised without binding a port.

I deliberately did **not** fix #4097 here. That issue notes that making the `200 null` endpoints return 404 needs a version bump, and it asks for verification of what each endpoint actually returns. So there is a test pinning `/stateRoot/{height}` -> `200 null`, commented as recording the inconsistency rather than endorsing it. If someone changes it, that test fails and the decision gets made deliberately.

## Test Plan

`cargo test -p snarkos-node-rest --lib` -- 32 tests, up from 13 before this stack. New coverage beyond the fixes above:

- an unknown route, and an unknown network prefix, on all three prefixes
- the wrong HTTP method on a known route
- rate limiting, as a 429 on `/v2` and as an identifiable 500 on `/v1`
- the same route set being served on all three prefixes, differing only in how errors are represented
- routes requiring consensus reporting 503 when a node has none
- the latest-block routes against a genesis-only ledger

Worth mentioning that I probed each of these before asserting anything, and it corrected two assumptions. I expected an unknown route to produce `500 "Failed to convert error"`; it is actually a clean empty 404, because unmatched paths are answered by the outer router, above where the v1 middleware is layered. Writing those tests from reasoning alone would have encoded the wrong behavior.

## Documentation

No in-repo REST reference exists to update, as noted on #4447; the handlers' doc comments are the documentation. The `/v1` vs `/v2` error-shape caveat that this PR's tests pin is now also written on the three new routes' doc comments in #4447.

## Backwards compatibility

No `ConsensusVersion` guard is needed -- this is REST only, and touches no consensus path.

Two behavior changes worth reviewing explicitly:

- On `/v2`, `/block/{height}`, `/height/{hash}` and `/program/{id}` now return 404 instead of 500 for a resource the node does not have. That is a status code change, but in the direction #4337 already established as correct, and 500 was never a meaningful answer for a height that simply has not synced.
- On the default and `/v1` prefixes, those cases remain 500, because the v1 middleware maps every error status to 500 regardless. So **v1 clients see no status change at all** from this PR. What they do see is a more informative error body in the cases that previously read "Failed to convert error".

### DIFF
--- a/node/rest/src/lib.rs
+++ b/node/rest/src/lib.rs
@@ -338,10 +338,12 @@ impl<N: Network, C: ConsensusStorage<N>, R: Routing<N>> Rest<N, C, R> {
             .layer(trace_layer)
     }
 
-    async fn spawn_server(&mut self, rest_ip: SocketAddr, rest_rps: u32) -> Result<()> {
-        // Log the REST rate limit per IP.
-        debug!("REST rate limit per IP - {rest_rps} RPS");
-
+    /// Builds the router served by `spawn_server`: the routes under the default, `/v1` and `/v2`
+    /// prefixes, with the v1 error middleware applied to the first two.
+    ///
+    /// This is separate from `spawn_server` so that the version prefixes can be exercised in tests
+    /// without binding a port.
+    fn build_versioned_router(&self, rest_rps: u32) -> axum::Router {
         // Add the v1 API as default and under "/v1".
         let default_router = axum::Router::new().nest(
             &format!("/{}", N::SHORT_NAME),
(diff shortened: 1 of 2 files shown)


vicsn (approved):
LGTM! CC @dms120 @lempira can you confirm if "On /v2, /block/{height}, /height/{hash} and /program/{id} now return 404 instead of 500" won't cause any issues for explorer/wallet?

PR #4453, opened 2026-09-11
feat: add subdag/DAG-density metrics

brunoffranca:
## Motivation

`ProvableHQ/snarkos-stress-testing` [PR #460](https://github.com/ProvableHQ/snarkos-stress-testing/pull/460) added consensus-metrics tooling that measures block times, latencies, connectivity, and DAG structure across test scenarios (baseline, network delay, validator offline). Everything it measures is already exposed by snarkOS as a Prometheus metric except subdag/DAG-density data (certificates per round, signatures per certificate, previous-round certificate references, rounds per committed block) — for that, the stress-testing script has to fetch and parse raw blocks over the REST API instead of scraping Prometheus like everything else. This PR closes that gap by exporting the missing data natively.

## Changes

- Added four new Prometheus histograms in the BFT commit path (`node/bft/src/bft.rs`, `commit_leader_certificate`): `snarkos_bft_subdag_certificates_per_round`, `snarkos_bft_subdag_certificate_signatures`, `snarkos_bft_subdag_certificate_previous_refs` (DAG-density proxy), and `snarkos_bft_subdag_rounds_per_block`.
- Registered the new metric names in `node/metrics/src/names.rs` (`bft` module + `HISTOGRAM_NAMES`) so they pre-register and appear on `/metrics` before the first observation.
- Reused data already computed in the commit path for an existing debug log (`subdag_metadata`) and existing certificate accessors (`signatures()`, `previous_certificate_ids()`) rather than recomputing anything.
- All emissions are gated behind `#[cfg(feature = "metrics")]`, consistent with existing metrics in this file.

## Test Plan

- `cargo check -p snarkos-node-bft` with and without `--features metrics`, both compile.
- `cargo test -p snarkos-node-bft --features metrics --lib bft::`, all 15 existing BFT commit-path tests pass.
- Ran a local 4-validator devnet and scraped each validator's `/metrics` endpoint after 11+ committed blocks. All four new series appear as Prometheus summaries with plausible values. Observation counts increase in lockstep across all 4 validators as blocks commit, confirming the metrics are recorded consistently network-wide, not just locally.

## Documentation

No user-facing docs to update. Internal metrics reference docs (if any list all `snarkos_*` metric names) should add these four.

## Backwards compatibility

Purely additive: new metric names/observations only, no changes to consensus logic, message formats, or ledger data. No `ConsensusVersion` gate needed.

### DIFF
--- a/node/bft/src/bft.rs
+++ b/node/bft/src/bft.rs
@@ -751,25 +751,6 @@ impl<N: Network> BFT<N> {
             // Retrieve metadata about the subdag.
             let subdag_metadata = subdag.iter().map(|(round, c)| (*round, c.len())).collect::<Vec<_>>();
 
-            // Update subdag/DAG-density metrics.
-            #[cfg(feature = "metrics")]
-            {
-                metrics::histogram(metrics::bft::SUBDAG_ROUNDS_PER_BLOCK, subdag_metadata.len() as f64);
-                for (_, certs_in_round) in &subdag_metadata {
-                    metrics::histogram(metrics::bft::SUBDAG_CERTIFICATES_PER_ROUND, *certs_in_round as f64);
-                }
-                for certificate in commit_subdag.values().flatten() {
-                    metrics::histogram(
-                        metrics::bft::SUBDAG_CERTIFICATE_SIGNATURES,
-                        certificate.signatures().len() as f64,
-                    );
-                    metrics::histogram(
-                        metrics::bft::SUBDAG_CERTIFICATE_PREVIOUS_REFS,
-                        certificate.previous_certificate_ids().len() as f64,
-                    );
-                }
-            }
-
             // Ensure the subdag anchor round matches the leader round.
             ensure!(
                 anchor_round == leader_round,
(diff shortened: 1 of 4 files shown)


node/bft/src/bft.rs
```diff
@@ -778,6 +780,25 @@ impl<N: Network> BFT<N> {
                         return Ok(());
                     }
                 }
+
+                // Update subdag/DAG-density metrics.
+                #[cfg(feature = "metrics")]
+                {
+                    metrics::histogram(metrics::bft::SUBDAG_ROUNDS_PER_BLOCK, subdag_metadata.len() as f64);
```
vicsn, line 787: Would total number of subdag certificates also be useful? Not sure if that can be inferred currently

vicsn (approved):

node/bft/src/bft.rs
brunoffranca: Do you mean the total number of certificates per block or the total number of certificates observed so far/since genesis? If it's the first, then it already exists in COMMITTED_CERTIFICATES. If it's the second, maybe it can be calculated but doesn't seem that useful.

vicsn: The first, ok! Merging

PR #4454, opened 2026-09-14
fix(metrics): standardize metric units, names, and histogram buckets

brunoffranca:
## Motivation

Standardizes the metrics module: consistent units (seconds instead of mixed ms/secs), consistent naming, tuned histogram buckets, and a bug fix for commit-latency recording.

## Changes

- Fix `COMMIT_ROUNDS_LATENCY` being recorded even when the round doesn't advance (`node/bft/src/bft.rs`), by moving the timer read inside the block that only runs on an actual round advance.
- Convert `BLOCK_LAG` from milliseconds to seconds for consistency with other latency metrics, and rename the old `*_LATENCY` consensus histogram constants to `*_SECS`.
- Remove unused metric name constants in `node/metrics/src/names.rs`.
- Set explicit Prometheus histogram buckets for `_secs`-suffixed metrics, `TRANSMISSION_LATENCY`, and `snarkos_bft_subdag_*` count metrics, instead of relying on the exporter's default buckets.

## Test Plan

Built with `--features metrics` and confirmed the exporter initializes without panicking (bucket matchers apply cleanly). Manually reviewed exported metric names/units for the renamed and rescaled histograms.

## Documentation

None needed; metric names are internal Prometheus identifiers not covered by user-facing docs.

## Backwards compatibility

Metric name and unit changes (`BLOCK_LAG` ms to secs, renamed consensus latency constants) are breaking for any external dashboards/alerts querying the old names/units. Not consensus-relevant, so no `ConsensusVersion` guard is needed.

### DIFF
--- a/node/bft/src/bft.rs
+++ b/node/bft/src/bft.rs
@@ -240,17 +240,6 @@ impl<N: Network> PrimaryCallback<N> for BFT<N> {
             false => self.is_leader_quorum_or_nonleaders_available(current_round),
         };
 
-        #[cfg(feature = "metrics")]
-        {
-            let start = self.leader_certificate_timer.load(Ordering::SeqCst);
-            // Only log if the timer was set, otherwise we get a time difference since the EPOCH.
-            if start > 0 {
-                let end = now();
-                let elapsed = std::time::Duration::from_secs((end - start) as u64);
-                metrics::histogram(metrics::bft::COMMIT_ROUNDS_LATENCY, elapsed.as_secs_f64());
-            }
-        }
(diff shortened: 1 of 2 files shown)


vicsn (approved):
My understanding is that the increased number of buckets would increase our grafana costs somewhat. But if that is a concern we may have other ways to reduce the grafana bill

brunoffranca:
If that's a worry, we can reduce the number of buckets. Although I wouldn't get rid of them completely as it's quite useful to know our p99s.

PR #4458, opened 2026-09-16
Serve the `history` routes from the Provable historical API (`--history-compat-mode`)

cbeck88:
## Motivation

The `history` cargo feature served historical mapping values from tables the node maintained itself. Those tables never recorded deletions and mixed incompatible height encodings (ProvableHQ/snarkVM#3418, ProvableHQ/snarkVM#3408), so what they served was wrong: for `credits.aleo/unbonding`, a claimed unbond is reported as still pending at every later height. That is the query Coinbase runs against Quicknode's history node today. The feature is being removed from snarkVM (schema-version + migration PR, to follow) and this PR removes it from snarkOS.

Operators who exposed those routes need to keep them up while the feature goes away. Rather than a separate shim binary, this adds a runtime flag, **`--history-compat-mode [URL]`**: the node registers the same routes and produces the same response bodies, but answers them from the Provable historical staking API (`https://{network}.historical-staking.provable.com` by default), which records a **full snapshot** of five `credits.aleo` staking mappings at every block since genesis. A full snapshot represents a deleted key by its absence, so this mode is correct in exactly the case the tables were not. Quicknode takes the release and adds the flag; Coinbase changes nothing and starts getting correct `unbonding` answers.

This is a stop-gap until a proper archival-node mode exists, and it makes Provable's historical API load-bearing for anyone on the flag — that instance (and its storage) has to stay up until the replacement can serve the same queries.

What compat mode serves:

| Route | Response |
|---|---|
| `GET /program/{id}/mapping/{name}/{key}/history/{height}` | value plaintext string, or `null` if the key is absent at that height / the height is above the node's — same body as before. Only `credits.aleo` × `{bonded, delegated, metadata, unbonding, withdraw}` (what the upstream records); anything else → 404 naming what's supported. |
| `GET …/mapping/{name}/history/{height}?keys=…` | same, one upstream request for all keys |
| `GET /staking/rewards/{address}/{height}` | `[validator, reward, new_stake]` as `history-staking-rewards` returned it; `new_stake` joined from the `bonded` snapshot at the same height |
| `POST /program/{id}/view/{function}/{height}` | always 404 with an explanation — staking snapshots aren't enough state to evaluate a view at a past height. Kept so old clients get an answer rather than a routing miss. |

Behaviour worth knowing:
- While the node is syncing the routes answer 503 (same guard as the other height-bound routes); a fresh compat node must sync before its `null`s mean anything.
- The upstream reports a block it has no snapshot of as a **500** whose body names the missing file (verified live), not a 404; that's mapped to 404 here and the test stub mirrors the real behaviour.
- Snapshots are cached per height with single-flight fetching that shares the outcome (success or failure), so an indexer fanning out over addresses at one height costs one upstream request. Upstream requests are bounded to 8 in flight, 20 s each, 64 MiB bodies.
- `history-staking-rewards` is left in place for now (snarkVM is turning it into a runtime setting); its handler is registered only when compat mode is off.
- The comments in `node/rest/src/history_compat.rs` and its handlers are deliberately much heavier than the repo guideline and record the literal requests/responses observed from the upstream; the module says why. This code is temporary and proxies another service, so the facts it depends on are written next to it.

## Test Plan

- `cargo test -p snarkos-node-rest` — 50 passed. Includes: snapshot-name and single-flight logic (shared success, shared failure, cancellation leaves nothing behind); every compat route end to end through the real axum router against a stub upstream serving responses captured from the live API at block 1,000,000 (values, `null` cases, 404s incl. the upstream-500 case, the staking-reward join, routes absent without the flag).
- `cargo test -p snarkos-cli history_compat_mode_flag` — flag parsing: absent / bare / with URL, invalid URLs refused at startup, `--dev` requires an explicit URL, conflicts with `--norest`.
- `cargo clippy -p snarkos-node-rest -p snarkos-node -p snarkos-cli --all-targets -- -D warnings`, with and without `history-staking-rewards`: clean.
- Verified against the live API: response shapes, sizes (`bonded` 31 KB at 1M), the 500-for-missing behaviour, and that `bonded@h == bonded@(h-1) + reward@h` (the join the staking-rewards route relies on).
- **Not yet done:** a live node run with `--history-compat-mode` against the real upstream; one manual `curl` of `/mainnet/program/credits.aleo/mapping/unbonding/<addr>/history/<h>` on such a node before handing to Quicknode.

## Documentation

- README's `snarkos start --help` listing gains `--history-compat-mode`.
- The `history` cargo feature is gone; `cargo install --features history` now fails. Release notes should say so and point at the flag.
- AleoNet/welcome, if it documents the `/history/` endpoints, should note they now require `--history-compat-mode` and cover only the five `credits.aleo` staking mappings.

## Backwards compatibility

No consensus or storage change; nothing needs a `ConsensusVersion` guard. REST API: the `/history/` and `/staking/rewards` routes keep their paths and bodies under the flag; without it they are absent, as they were without the cargo feature. `Rest::start`, `Node::new_validator/new_client`, `Validator::new`, `Client::new` gain one positional parameter (`history_api_url: Option<String>`).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01VD8HcYE7QnBCZF3YeN5jjt


cbeck88:
Pushed a second commit (a0ea7b328): an ignored test that sweeps the **live** upstream API and checks every assumption the compat routes make about its responses, at ~160 heights.

```
cargo test -p snarkos-node-rest live_upstream_sweep -- --ignored --nocapture
```

Per height it fetches all six snapshots and asserts:

- every key and value parses as `Plaintext`/`Value` **and reprints identically** — the lookup by `Plaintext::to_string` depends on the upstream's keys being canonical;
- each mapping's values have the shape the handlers use: `bonded` is `{validator: address, microcredits: u64}`, `unbonding` is `{microcredits: u64, height: u32}`, `delegated` u64, `withdraw` address, `metadata` exactly the two known keys with u32 values;
- cross-mapping consistency: every validator named in `bonded` is a `delegated` key; every `unbonding` staker has a `withdraw` entry;
- the join `get_staking_reward_compat` relies on: `stakingrewards` has exactly the `bonded` key set, each row names the validator the staker is bonded to, and `bonded@h == bonded@(h-1) + reward` for all but a handful of stakers (those who bonded/unbonded in block h);
- a height far above the tip is reported as *missing* (404), not as an outage (503).

Heights: blocks 1–3, 10, 100, 1000, 65535/65536/65537, both sides of every mainnet consensus-version boundary, tip−10/−11/−50/−1000, plus a seeded random sample (`HISTORY_SWEEP_RANDOM`, default 100; `HISTORY_SWEEP_SEED`). `HISTORY_API_URL` and `HISTORY_SWEEP_TIP` override the upstream and the tip lookup.

Results today: seed 0 / 164 heights and seed 7 / 264 heights, **0 failures**, ~15–25 s each (~1,000 requests, bounded to 8 in flight by the same semaphore the node uses).

One upstream fact it surfaced, now recorded in the module doc: **block 0 is not recorded** — snapshots start at block 1 (the first run tripped on joining height 1 with `bonded@0`).

Not in CI, by design — it needs the network and hits a third-party service; the intent is to run it before a release and whenever the upstream is suspected of having changed shape.

node/rest/src/routes.rs
```diff
@@ -1567,6 +1489,169 @@ impl<N: Network, C: ConsensusStorage<N>, R: Routing<N>> Rest<N, C, R> {
...
+    /// needs to be consulted for it:
+    ///
+    /// - `Err(503)` while this node is still syncing. Its own height is then meaningless as a
+    ///   bound, and answering `null` for a height it has not reached would read as "the key was
+    ///   absent" -- the wrong answer for, say, an unbond that the upstream knows is pending.
+    /// - `Ok(false)` for a height above this node's: the block may not exist yet, so it is
+    ///   answered with `null` without asking the upstream, as the removed feature did.
+    /// - `Ok(true)` otherwise: the upstream has (or should have) a snapshot.
+    fn history_compat_height_available(&self, height: u32) -> Result<bool, RestError> {
+        if !self.routing.is_within_sync_leniency() {
```
vicsn: If we want to maintain the same behaviour as the existing historical staking endpoint, I think we should get rid of the sync leniency check. It's for querying historic data.

vicsn (dismissed):
LGTM except for the comment about `is_within_sync_leniency`. 

Do you want to rebase on top of `testnet`?

Can't wait for us to get rid of this patchwork logic again in the future but I see it'll make our life easier now :)

node/rest/src/routes.rs
ljedrz: Won't this expression resolve to `false` throughout CDN sync? `is_within_sync_leniency` is documented to return `true` if no block locators have been received thus far (which is a guaranteed state until the CDN sync is concluded). This would cause this method to never throw a `503` while syncing with the CDN.

cbeck88: Both right — and the guard was treating a symptom. The compat handlers had carried over the old feature's `height > current_block_height → null` short-circuit, which made this node's own height (meaningless in compat mode) gate the answer; a syncing node answered `null` for every height above its own, and the sync-leniency check was my attempt to paper over that.

Removed both in a87fcb714: the routes always ask the upstream, which knows its own extent. A syncing node now answers correctly from minute one — verified on a devnet: correct answers for heights 409/410 (pending / claimed unbond) while the node itself was at height 194.

The one visible delta from the old feature: a height the upstream has no snapshot of (above its tip, which trails the network by ~5 blocks; or block 0) is a 404 "no snapshot recorded" rather than `null` — `null` from this route means "the key wasn't in the mapping at that height", which isn't known for such a height.

cbeck88: Both right — and the guard was treating a symptom. The compat handlers had carried over the old feature's `height > current_block_height → null` short-circuit, which made this node's own height (meaningless in compat mode) gate the answer; a syncing node answered `null` for every height above its own, and the sync-leniency check was my attempt to paper over that.

Removed both in a87fcb714: the routes always ask the upstream, which knows its own extent. A syncing node now answers correctly from minute one — verified on a devnet: correct answers for heights 409/410 (pending / claimed unbond) while the node itself was at height 194.

The one visible delta from the old feature: a height the upstream has no snapshot of (above its tip, which trails the network by ~5 blocks; or block 0) is a 404 "no snapshot recorded" rather than `null` — `null` from this route means "the key wasn't in the mapping at that height", which isn't known for such a height.

vicsn (dismissed):

snarkworld (dismissed):

cbeck88:
## Verification beyond the unit tests

Three additional checks were run on this branch. The scripts and the recorded run are shared in Slack.

### 1. Live-upstream sweep (`live_upstream_sweep`, ignored test in the PR)

Runs against the real `mainnet.historical-staking.provable.com` and checks, at ~160 heights (blocks 1–3 and other early ones, both sides of every mainnet consensus-version boundary, the blocks just behind the tip, and a seeded random sample), every assumption the compat routes make: keys and values are canonical plaintext strings, each mapping's values have the expected shape, `stakingrewards` has exactly the `bonded` key set and `bonded@h == bonded@(h-1) + reward`, and a height above the tip is reported as missing. Two seeds / 264 heights: 0 failures. Details in [this comment](#issuecomment-5691042148).

### 2. Devnet end-to-end against ground truth

**Method.** A 4-validator devnet on one machine (`--dev`, `test_network` features so the consensus ladder settles by block ~30), with three clients following it:

| client | binary | role |
|---|---|---|
| **A** | snarkOS v4.10.0 `--features history,history-staking-rewards` | the removed feature, bugs included — what a history node serves today |
| **B** | snarkOS v4.10.0 built against `data-snarkVM` (`historical-rewards`) | writes the per-block JSON snapshots that back the Provable historical API |
| **C** | this PR, `--history-compat-mode http://127.0.0.1:8000` | the thing under test |
| stub | `stub_upstream.py` serving B's snapshot directory | stands in for the Provable API: same paths, same 500 body for a missing block |

A delegator D is driven through `transfer_public` → `bond_public` (landed at **h=37**) → `unbond_public` (**h=46**) → 360 blocks → `claim_unbond_public` (**h=410**). The scenario's transactions define ground truth: which keys must be present in `bonded` / `unbonding` / `withdraw` / the staking-rewards table at each height. `compare.py` then queries **A** and **C** at the heights around each step and checks three things: C matches the truth; A and C agree byte-for-byte wherever the old feature was right; A is wrong exactly where the removed feature's missing-deletion bug predicts.

**Result — `ALL CHECKS PASSED`:**

| query | compat (C) vs. truth | old node (A) vs. compat (C) |
|---|---|---|
| `bonded[D]` | correct at every height | byte-identical through h=45; **A stale from h=46 on** (the unbond) |
| `unbonding[D]` | correct | identical through h=409; **A stale from h=410 on** (the claim) — the Coinbase case |
| `withdraw[D]` | correct | identical through h=409; A stale from h=410 |
| `/staking/rewards/D` | correct | **byte-identical at every height** (exact-keyed table; compat reproduces `[validator, reward, new_stake]` exactly) |
| validator key, all mappings + rewards | — | byte-identical at every height |
| batch `?keys=`, unsupported mapping, view-at-height, height above the upstream's tip | as designed (404s on `/v2`) | — |

Raw per-height table in `compare.log` alongside the scripts. So: compat mode is right where the old feature was wrong, and produces the same bytes everywhere the old feature was right.

Also on the same devnet, after a6fcdd0b0: **C restarted from genesis** answered correctly for h=409/410 while it was itself at height 194 — the node's own sync state no longer affects these routes.

### 3. In-place upgrade and load

- **Upgrade rehearsal:** stopped A, started this PR's binary on A's ledger directory (which still holds the old feature's tables) with the flag → opened, synced, correct answers, no errors. This is what an operator upgrading a history node will do.
- **Burst:** 600 concurrent requests for one height → 2 upstream fetches (single-flight cache), all 200; 300 distinct heights at once → 300 fetches in 1.0 s, ≤ 8 in flight at the upstream (the node's bound), all 200.

One thing worth knowing when reading raw responses: the default and `/v1` REST prefixes map every error to 500 (pre-existing snarkOS behaviour, identical on the old node); real status codes are on `/v2`.

vicsn (approved):

snarkworld (approved):

PR #4459, opened 2026-09-17
Allow operators to limit REST verification concurrency at startup

vicsn:
## Motivation

REST deploy, execute, and solution verification currently always run at the protocol maxima, which were determined just based on OOM prevention, not efficient CPU utilization. . These flags let an operator reduce work stealing and potentially increase throughput; they default to the maxima and reject larger values.

## Test Plan

- [x] `cargo test -p snarkos-cli --lib rest_verification_limits`
- [x] Start a client/validator with the new flags at 1 and confirm REST still accepts deploys, executions, and solutions

## Documentation

If operator CLI flags are documented on [AleoNet/welcome](https://github.com/AleoNet/welcome), add `--num-verifying-deploys`, `--num-verifying-executions`, and `--num-verifying-solutions`.

## Backwards compatibility

Defaults match the existing protocol maxima, so behavior is unchanged unless the flags are set. No `ConsensusVersion` is needed: this is operator configuration, not consensus.


cbeck88:
Main finding is, the semaphores are improperly implemented -- they block using `.acquire().await`. So this never actually produces a 429 when there are too many requests, it just hangs the responses and continues to consume server resources.

https://github.com/ProvableHQ/snarkOS/blob/2a4fc8cf685d97cfa8d32652536f108579824cea/node/rest/src/routes.rs#L1198-L1208

https://github.com/ProvableHQ/snarkOS/blob/2a4fc8cf685d97cfa8d32652536f108579824cea/node/rest/src/routes.rs#L1287-L1292

We should change these to use `try_acquire` so that it actually produces a 429 error immediately if we are over the limit.

cbeck88 (changes requested):
we shouldn't block on acquiring these semaphores

node/rest/src/routes.rs
```diff
@@ -1196,17 +1196,12 @@ impl<N: Network, C: ConsensusStorage<N>, R: Routing<N>> Rest<N, C, R> {
         let check_transaction = check_transaction.check_transaction.unwrap_or(false);
 
         if check_transaction {
-            // Select the semaphore based on the transaction type.
-            let (slot, err_msg) = if tx.is_execute() {
-                (rest.num_verifying_executions.acquire().await, "Too many execution verifications in progress")
+            let _verification_slot = if tx.is_execute() {
```
brunoffranca, line 1199:
```suggestion
            if tx.is_execute() {
```
nit

node/rest/src/routes.rs
```diff
@@ -1285,11 +1280,7 @@ impl<N: Network, C: ConsensusStorage<N>, R: Routing<N>> Rest<N, C, R> {
         }
 
         if check_solution {
-            // Try to acquire a slot.
-            let slot = rest.num_verifying_solutions.acquire().await;
-            if slot.is_err() {
-                return Err(RestError::too_many_requests(anyhow!("Too many solution verifications in progress")));
-            }
+            let _verification_slot = rest.verification_slots.solutions.acquire().await?;
```
brunoffranca, line 1283:
```suggestion
             rest.verification_slots.solutions.acquire().await?;
```
same nit

vicsn:
@cbeck88 great point. At the risk of overengineering, what do you think of dc54c35ef which ensures capacity has some extra leeway compared to the parallel verification limit.

cbeck88 (dismissed):
it looks good to me now, thank you

i think the semaphore for queued and concurrent probably makes sense, and cannot hurt anything

vicsn:
@copilot please fix the merge conflicts in this pull request.

brunoffranca (approved):

Issue #4282, opened 2026-05-28
[Bug] Validators seem to propose old batches


vicsn:
## 🐛 Bug Report

On mainnet v4.6.4 we often see the warning (20 times per hour on a single validator): "Cannot sign a batch at round A from 'X:5000' — Peer (Y proposed a batch for a previous round (A), latest signed round: B"

However, under normal conditions, a validator should not be rebroadcasting old proposals. We should figure out where the race condition or logic bug is coming from

kaimast:
It may be caused by the proposal logic being stuck at waiting for signatures even when the BFT round advances.

vicsn:
NOTE: we recently merged a couple fixes to proposal/signing atomicity, so before attempting a fix we should check whether this is still an issue in CI or production logs

PR #4461, opened 2026-09-17
fix(bft): decouple consensus participation from peer-reported sync status

brunoffranca:
## Motivation

`is_synced()` was used to gate proposing, signing, and certifying batches, but it is computed entirely from `greatest_peer_height`, a number derived from peer locator messages. Any peer (including a freshly spun-up malicious one, since banning is trivial to route around) can claim to be far ahead and permanently flip a validator to `Unsynced`/`Syncing`, silencing it. Enough of those and the network cannot reach quorum and consensus halts.

The BFT layer already has a complete, purely local notion of readiness that does not depend on anything a peer reports: `Storage::get_certificate_authors_for_round(round)` combined with `Committee::is_quorum_threshold_reached(authors)`. Certificates are signed by the committee, so unlike a claimed block height this signal cannot be forged. The peer-height gates layered on top of it were redundant at best and an attacker's lever at worst.

Out of scope and unchanged: the one-time CDN-bootstrap gate that defers `start_consensus_handlers()`, `node/sync`'s tip/`find_sync_peers()` computation, router-level gossip gating, and `Consensus`'s use of `is_synced()` for log-level selection.

Fixes PRO2-6

## Changes

- Removed every `is_synced()` / `wait_for_synced_if_syncing()` gate from the consensus path in `primary.rs` and `proposal_task.rs`: the proposal task stages, the round-increment loop, `try_increment_to_the_next_round`, the proposed-batch / batch-signature / certified-batch handlers, and the primary-ping and worker-ping handlers. The existing local quorum checks and `Storage::check_batch_header` / `check_certificate` remain the gate. This also drops the now-pointless `IS_SYNCING` const generic from `sync_with_certificate_from_peer` / `sync_with_batch_header_from_peer`.
- Made `Storage`'s round/height pointers race-free: `sync_height_with_block`, `sync_round_with_block`, and `update_current_round` now use `fetch_max` instead of check-then-act, so the BFT path and the sync-applied-block path can only ever advance them. `increment_to_next_round` now tolerates a concurrent writer having already moved past its computed round and returns the true storage round.
- Added exponential backoff (from `CREATE_BATCH_INTERVAL` up to `MAX_BATCH_DELAY`) to the Stage 2 proposal retry loop, which previously held a 250ms cadence indefinitely on a stalled round, taking the proposal lock and reading the committee lookback each time.
- Deleted the dead code this leaves behind: `wait_for_synced`/`wait_for_synced_if_syncing` and `synced_notify` in `BlockSync`, `has_pending_responses`, three unused `Storage` round lookups, the `DummyCommunicationService` test helper and its `test-helpers` feature, and the sync test-only setters.
- Housekeeping: `.gitignore` patterns anchored to the repo root and `cargo audit` lockfile bumps (`rustls` 0.23.43 -> 0.23.45, `windows-sys` 0.52 -> 0.59).

## Test Plan

- `cargo test -p snarkos-node-bft` and a full workspace build (no dead-code warnings from the removed helpers).
- New `test_concurrent_round_and_height_updates_never_regress` in `storage.rs`: three barrier-synchronized threads drive `increment_to_next_round` and `sync_round_with_block`/`sync_height_with_block` (the latter in descending order, so a stale write lands after a fresher one) while an observer samples both values; neither may regress and both must converge to the max proposed.
- New `test_propose_batch_without_previous_round_quorum` in `primary.rs`: with the round pointer advanced but only one round-1 certificate in storage, the primary must refuse to propose. This asserts the local quorum check is a real gate now that `is_synced()` no longer backstops it.
- Tests that only existed to assert the peer-height gate were removed (`test_batch_propose_from_peer_when_not_synced`, the two `proposal_task` sync-pausing tests); the remaining proposal-task tests are unchanged apart from dropping the `is_synced` stubs.
- CircleCI devnet, chaotic-minority-reset, chaotic-majority-reset, and upgrade tests are enabled on this branch.

## Documentation

No user-facing API or CLI changes. Any internal/operator docs describing `is_synced()` as a precondition for consensus participation should be updated to say that participation is gated on local DAG quorum instead, and that peer-reported height now only drives block fetching.

## Backwards compatibility

No wire format, message, or ledger state transition changes, so no `ConsensusVersion` guard is needed. The change is node-local: a patched validator participates in rounds it previously would have sat out, which is strictly more liveness, and is what an honest, locally-ready validator was always allowed to do. Patched and unpatched nodes interoperate on the same network.

One item to clean up before merge: the `.circleci/config.yml` branch filters still list `bruno/decouple`.


node/bft/src/primary/proposal_task.rs
```diff
@@ -164,27 +149,30 @@ impl<N: Network> ProposalTask<N> {
...
-                return false;
-            }
-
             if attempt > 1 {
-                sleep(CREATE_BATCH_INTERVAL).await;
+                // The round only advances on an external event (a certificate or an applied
+                // block), so a round that is not progressing would otherwise hold the 250ms
+                // retry cadence indefinitely, each attempt taking the proposal lock and reading
+                // the committee lookback from the ledger.
+                sleep(backoff).await;
```
cbeck88, line 174: this was a pre-existing bug, but what i'd suggest is,

(1) we should check if the current round changed after this sleep completes, as well as before
(2) ideally we wake up and return false immediately if the round changed, but that requires more changes elsewhere so that we can be notified

node/bft/src/primary/proposal_task.rs
```diff
@@ -548,166 +495,4 @@ mod tests {
         // Stage 3 may make additional rebroadcast calls after success, so use >.
         assert!(propose_count.load(Ordering::SeqCst) > RETRIES, "expected at least {} total attempts", RETRIES + 1);
     }
-
-    /// While the node is syncing, Stage 3 must not rebroadcast the proposed batch — its previous
-    /// certificates may be stale and peers will not sign it. Once sync completes, rebroadcast
-    /// should resume.
-    #[test_log::test(tokio::test)]
-    async fn test_proposal_task_pauses_rebroadcast_while_syncing() {
```
cbeck88, line 556: why are we removing these tests?

node/bft/src/primary/proposal_task.rs
```diff
@@ -134,13 +126,6 @@ impl<N: Network> ProposalTask<N> {
                 return false;
             }
 
-            // A node cannot propose while it is syncing.
-            if let Some(fut) = primary.wait_for_synced_if_syncing() {
-                fut.await;
-                // Re-check round after sync completes.
-                continue;
-            }
```
cbeck88, line 142: why is it okay to remove this?

node/consensus/src/lib.rs
```diff
@@ -609,6 +609,18 @@ impl<N: Network> Consensus<N> {
...
+                return Ok(false);
+            }
+            Err(CheckBlockError::InvalidHeight { .. }) => {
+                debug!("The ledger advanced while we were constructing the next block");
+                return Ok(false);
+            }
+            Err(CheckBlockError::InvalidRound { new, previous }) => {
+                debug!("The subDAG round is too low. Expected >{previous}, got {new}");
+                return Ok(false);
+            }
```
cbeck88, line 623: interesting that this function didn't match the documented contract... i wonder why it was working before

cbeck88:
i'll read some more in an hour or two

node/bft/src/primary/proposal_task.rs
brunoffranca: It needs to be removed since we want the validator to propose independently of the sync mechanism. But it's okay for three reasons:
- The Primary isn't even started until the initial CDN sync is finished.
- We never propose until storage catches up to the round we are proposing anyway. So we can't propose invalid blocks.
- Even if somehow we propose an invalid or duplicate block, the other validators would not sign it. So we wouldn't be able to commit it to the DAG and it would just be dropped.

node/bft/src/primary/proposal_task.rs
brunoffranca: They tested the previous "not proposing while we are syncing" behavior. So are not needed anymore.

node/bft/src/primary/proposal_task.rs
brunoffranca: Seems reasonable, my solution was a bit hacky. I'll do it tomorrow.

node/consensus/src/lib.rs
brunoffranca: We didn't really have two paths calling this at the same time before. We would either advance the ledger from the peer sync or from the consensus. Now that both can happen at the same time, those error paths started happening. Only caused some overzealous logging though, not problematic.

cbeck88: AI had the following finding at the previous revision:

```
  node/consensus/src/lib.rs:612-623 converts BlockAlreadyExists / InvalidHeight / InvalidRound from Err to Ok(false), which changes what BFT does at bft.rs:769. Previously Ok(Err(..)) logged
  error! and returned early, skipping the DAG commit; now it falls through, logs info!("Committing a subDAG with anchor round …") and runs dag_write.commit(..) with no block produced. For
  those three variants the block genuinely does already exist, so committing the DAG is defensible and matches the pre-existing arms at 647-655 — but the info! claims a commit that never
  reached the ledger, and reinsert_transmissions still fires on that path, pushing already-committed transmissions back into the mempool.
```

maybe dropping the already-committed transmissions is a good change?
assuming this is correct

node/bft/src/helpers/storage.rs
```diff
@@ -219,20 +219,39 @@ impl<N: Network> Storage<N> {
...
         Ok(next_round)
     }
 
     /// Updates the storage to the next round.
+    ///
+    /// This is called concurrently from two independent paths: the BFT round-certification path
+    /// (`increment_to_next_round`) and the sync-applied-block path (`sync_round_with_block`).
+    /// `fetch_max` ensures the round only ever advances, regardless of interleaving, instead of a
+    /// plain store letting a stale writer regress it.
     fn update_current_round(&self, next_round: u64) {
```
cbeck88, line 253: suggest maybe return the new max, and the caller can discover if they somehow are behind now?
or maybe its for later work

node/bft/src/helpers/storage.rs
```diff
@@ -219,20 +219,39 @@ impl<N: Network> Storage<N> {
...
+        // (now stale) round update.
+        ensure!(
+            next_round >= gc_round,
+            "The next round {next_round} is behind the current GC round {gc_round}, likely because a concurrent sync advanced past it"
+        );
+
+        // Storage may already be ahead of `next_round` if a concurrent sync-applied round update
+        // landed in between; return the true storage round rather than the stale `next_round`.
+        if storage_round > next_round {
+            return Ok(storage_round);
```
cbeck88, line 239: AI flagged the following:

when we get to this line we already wrote the metric above around line 180:

```
        #[cfg(feature = "metrics")]
        metrics::gauge(metrics::bft::LAST_STORED_ROUND, next_round as f64);
```

so even though we will return `Ok(storage_round)` on this branch, we recorded `next_round` as the LAST_STORED_ROUND. maybe we should move that to after this branch

node/bft/src/helpers/storage.rs
brunoffranca: I could, but there's no caller right now that cares about that information. At most we would log it.

node/bft/src/helpers/storage.rs
brunoffranca: Good find. It's a wrong metric, yes.

node/bft/src/primary/proposal_task.rs
brunoffranca: Fixed by c183bf1c.

node/consensus/src/lib.rs
brunoffranca: This took some investigation, I was not sure as well. There are two potential issues here:

1. The block already exists (or the ledger moved on) and then the transmissions get reinserted in the mempool at (consensus/src/lib.rs:582).
2. The block already exists but then BFT still tries to commit the subDAG anyway (bft/src/bft.rs:805).

On the first one, repeated transmissions fail to be reinserted. The path is:

- reinsert_transmission (consensus/lib.rs:807) sends each one to the primary.
- The primary passes it to Worker::process_unconfirmed_{transaction,solution} (worker.rs:434/470).
- The first thing those functions do is check self.contains_transmission(id) (worker.rs:193). That check looks at the ready queue, the proposed batch, storage and the ledger, and returns Ok(false) if the transmission is in any of them.
- reinsert_transmissions treats Ok(false) as a debug!("... Already exists.") (consensus/lib.rs:784).

Note that only the case "BlockAlreadyExists" justified dropping the transmissions, "InvalidHeight" and "InvalidRound" could apply to blocks that had new transmissions.

On the second one, dag_write.commit (bft.rs:814) is idempotent. The info log is misleading though and the commit call may be wasted work (it cleans up the DAG, which is good. remember again that BlockAlreadyExists is only one of the failure modes). So I'll fix the logging.

brunoffranca: Happy to remove the call to write_dag.commit too if you think that's unnecessary. I'm on the fence.

brunoffranca: a4ebee3c

node/bft/src/helpers/storage.rs
brunoffranca: 010b621c

node/bft/src/helpers/storage.rs
cbeck88: There's some tickets related to propose batch happening for rounds that already closed

#4282 is a report of it happening 
#4428 is a plausible mechanism

I feel like I saw another way it could happen reading this code , but I don't want to sidetrack you, so if you think it's for later work lmk

node/bft/src/helpers/storage.rs
```diff
@@ -212,27 +212,46 @@ impl<N: Network> Storage<N> {
         // Update the storage to the next round.
         self.update_current_round(next_round);
 
-        #[cfg(feature = "metrics")]
-        metrics::gauge(metrics::bft::LAST_STORED_ROUND, next_round as f64);
-
         // Retrieve the storage round.
         let storage_round = self.current_round();
```
cbeck88: I guess relevant to my other comment: we could instead of this have

let storage_round = self.update_round(next_round);

And it might be a little simpler,  like, the race you are asserting against couldn't happen

cbeck88 (dismissed):

cbeck88:
Overall it looks good to me

node/bft/src/helpers/storage.rs
brunoffranca: Yes, I would prefer doing those in a separate PR. Let's keep this one more self-contained. I'll assign myself to those issues.

node/bft/src/helpers/storage.rs
brunoffranca: 9c949387

brunoffranca:
@cbeck88 I had to fix one more logging issue in the last commit, one the CI jobs was failing because of it. It was basically the same issue as a81c417

brunoffranca:
@cbeck88 @vicsn FYI 80e916e

PR #4460, opened 2026-09-17
[Release] Testnet v4.10.1

snarkworld:
## Motivation

Testnet point release v4.10.1, carrying #4458 (`--history-compat-mode`), which replaces the removed `history` cargo feature with a runtime flag that serves the same `/history/` and `/staking/rewards` routes from the Provable historical staking API.

- Bumps all snarkOS crate versions and `.cargo/release-version` to `4.10.1`. The snarkVM dependency stays at the `test_net_v4100` rev (`7b032dac9`, snarkVM 4.10.0); snarkVM is unchanged in this release.
- Patches `rustls` past RUSTSEC-2026-0285 and moves off the yanked `wnaf` 0.14.0, so `check-cargo-audit` is green again on this branch.
- Enables extended CI for this branch.

No consensus, storage, or snarkVM change. Operators who built with `--features history` must switch to `--history-compat-mode`; `cargo install --features history` no longer works.

## Test Plan

- [x] `cargo clippy --workspace --all-targets --all-features -- -D warnings` and rustfmt pass locally (pre-commit hook)
- [x] `cargo audit` clean locally
- [x] Extended CI on this branch: all six workflows pass functionally. Two devnet jobs fail only their final log scan, on messages that also fail on `staging`:
  - `chaotic-minority-reset-test`: "There are 2 surplus live tasks" (same as #4456 and #4409).
  - `chaotic-majority-reset-test`: "Failed to garbage collect certificates — Attempted to decrease GC round" during a reset. Same message failed staging pipelines 4236 and 4248; #4458 does not touch `node/bft`, `node/sync` or `node/consensus`.
- [ ] Live node run with `--history-compat-mode` against the real upstream, and a manual `curl` of an `unbonding` history query (the two items left open in #4458)

PR #4462, opened 2026-09-17
[Release] Mainnet v4.10.1

snarkworld:
## Motivation

Mainnet point release v4.10.1, the mainnet counterpart of `testnet-v4.10.1` (#4460). snarkOS only; snarkVM is unchanged and stays pinned to crates.io `snarkvm = "=4.10.0"`.

- Merged `testnet` into `mainnet`. The only conflict was the extended-CI branch filter, resolved to `main_net_v4101`. Everything else auto-merged: all crates at `4.10.1`, `.cargo/release-version` `v4.10.1`, `rustls` 0.23.45 and `wnaf` 0.14.1 already carried from #4460.
- Carries #4458: the `history` cargo feature is removed and replaced by the `--history-compat-mode` runtime flag, which serves the `/history/` and `/staking/rewards` routes from the Provable historical staking API. This is the change Quicknode needs on mainnet so `credits.aleo/unbonding` history queries return correct results for claimed unbonds.

No consensus, storage, or activation-height change. The mainnet `ConsensusVersion::V20` height of `22_175_000` from v4.10.0 is unaffected.

## Test Plan

- [x] `cargo clippy --workspace --all-targets --all-features -- -D warnings` and rustfmt pass locally (pre-commit hook) against crates.io snarkvm 4.10.0
- [x] `cargo audit` clean locally
- [x] Extended CI on this branch. First run: snarkos, windows, devnet, node and cargo workflows pass (cargo-audit included); `upgrade-test` and `dev-on-prod-restart-test` pass. Two devnet jobs failed:
  - `chaotic-majority-reset-test`: network reached the target height, then the log scan caught "Failed to garbage collect certificates — Attempted to decrease GC round from 36 to 34". Same message failed staging pipelines 4236 and 4248 and the testnet 4.10.1 PR (#4460); nothing in this release touches `node/bft`.
  - `chaotic-minority-reset-test`: did not run. Its setup step got an HTTP 500 from github.com while downloading; needs a rerun.
- [ ] `testnet-v4.10.1` validated with `--history-compat-mode` against the testnet upstream before this merges


cbeck88 (approved):
looks good to me

vicsn (approved):
Failed CI is not urgent, documented it in a new issue: https://github.com/ProvableHQ/snarkOS/issues/4466

The Veria finding is indeed low priority because: 1. REST is only exposed via a few clients, not validators 2. this compat feature is temporary

PR #4463, opened 2026-09-18
Expose snarkVM's state-root skip as a snarkOS build feature

vicsn:
## Motivation

snarkVM #3439 adds a compile-time `dev_skip_state_root_check` feature so `check_transaction` can admit executions and deployments whose `global_state_root` is not in the local block store. This PR forwards that flag from snarkOS (`--features dev_skip_state_root_check`) for local and test builds, matching `73a2938cb0`.

Depends on ProvableHQ/snarkVM#3439 being present in the pinned snarkVM rev.

## Test Plan

- [ ] `cargo check` (default features; skip remains off)
- [ ] After snarkVM #3439 is on the pinned rev: `cargo check --features dev_skip_state_root_check`

## Documentation

No docs changes. Opt-in compile-time feature for local/dev builds.

## Backwards compatibility

Default path is unchanged. Enabling the feature is a build-time choice and is not a consensus change.


vicsn:
We can also add another run-time flag to ensure this feature is only set in combination with `test_network`

vicsn:
https://github.com/ProvableHQ/snarkVM/pull/3439#issuecomment-5791475527

PR #4464, opened 2026-09-18
perf(cli): size the tokio pools from measured demand

vicsn:
## Motivation

I revived our end-to-end load test setup https://github.com/ProvableHQ/snarkos-stress-testing/pull/517, which is now in a state that we can run it before every major release to assure blocktime and TPS do not regress. This PR (and https://github.com/ProvableHQ/snarkVM/pull/3438 ) are the first changes driven by it.

Running it against 5- and 30-validator devnets on 30-core hosts, with per-thread CPU accounting, shows that **neither rayon nor tokio is the bottleneck** — not at any load we were able to generate. Across three windows the pools never came close to their width:

| Pool | Busiest observation |
| --- | --- |
| rayon | 23% of its thread count |
| tokio workers | 0.63 cores of 30 |
| tokio blocking | 53 threads concurrently busy, against a cap of 512 |
| rocksdb | every counter flat zero |

Time spent waiting for a thread in the blocking pool was **0.1 ms at p99**. Confirmation latency variance, which is what prompted the investigation, turned out to be a queueing and admission-policy problem rather than a resource one, and is not addressed here.

So this is a small cleanup, **acceptable but not urgent**. It removes provisioning that the measurements show does nothing, and it is safe to defer or drop if it complicates anything else in flight.

## The change

Worker threads move from `2 * num_cores` to `num_cores`. The comment justifying the surplus said the node is I/O-bound, but that argues for more *tasks*, not more *workers* — an async worker multiplexes many I/O tasks, and every CPU-bound path already leaves the runtime through `spawn_blocking!` and fans out onto rayon. On a 30-core box the 60 workers were doing 0.63 cores of work, while giving the OS 30 more threads to schedule against rayon's 30.

The blocking pool cap moves from 512 to 256. This is deliberately a smaller cut than it first appeared to warrant. The pool averaged 69 threads with 66 idle, but blocking threads linger after their task completes, so idle counts understate demand. Pairing the two counters sample by sample puts peak concurrent blocking work at **134 threads, p99 at 43**. A cap near that peak would convert bursts into queueing ahead of rayon, so 256 keeps close to 2x headroom over the highest concurrency observed while still bounding how many callers can pile into a rayon pool one-thread-per-core wide.

Rayon is unchanged.

## Verification

Re-ran the same devnet shape after the change. Both failure modes the change could have introduced did not occur:

| Check | Before | After |
| --- | --- | --- |
| Global queue depth (would show worker starvation) | 0.000 mean, max 0 | 0.007 mean, max 1 |
| Blocking wait p99 | 0.001s | 0.001s |
| Peak concurrent blocking threads | 18 | 53 (of the new 256 cap) |

One prediction did not hold, worth recording: I expected halving the workers to reduce park and no-op churn. It did not — parks went 227 → 245/s and no-ops 97 → 104/s. Since those are summed across workers, flat totals with half the workers means each worker now parks about twice as often. The benefit is 30 fewer threads competing with rayon, not less scheduler work.

Throughput was unchanged at ~5.9 tx/s, as expected for a change to pools that were never the constraint.

`cargo clippy --workspace --all-targets --all-features -- -D warnings` passes.

## Note for the reviewer

The instrumentation these numbers come from is #4451, which is still open. I removed a reference to `snarkos_cpu_blocking_wait_secs` from the code comment so this PR does not depend on that one landing first — the metric is still named above and in the commit message as provenance. Once #4451 is in, that metric is the regression detector for the 512 → 256 cap: it sat at 0.1 ms p99 under the old cap and should not move.


cbeck88 (approved):
sounds good to me, thank you

PR #4470, opened 2026-09-21
ci: migrate from cargo-audit to cargo-deny

brunoffranca:
## Motivation

`cargo-audit` only checks for known vulnerability advisories. It doesn't
verify dependency license compatibility or restrict which registries/git
sources dependencies may come from. `cargo-deny` covers all three (advisories,
licenses, sources) plus dependency-graph hygiene (banned/duplicate crates,
wildcard version requirements) in a single tool and config file, giving us
license compliance and supply-chain source pinning we didn't have before.

## Changes

- Add `deny.toml` at the repo root: migrates the existing RUSTSEC ignore list from `.cargo/audit.toml`, adds an Apache-2.0-compatible license allowlist, restricts dependency sources to crates.io plus the `snarkvm` git remote, and enables (non-blocking) duplicate-version and wildcard-dependency reporting.
- Remove `.cargo/audit.toml` (superseded by `deny.toml`).
- Replace the `check-cargo-audit` CircleCI job with `check-cargo-deny`, running `cargo deny check` instead of `cargo audit -D warnings`.
- Add `cargo deny check` to the local pre-commit hook (`scripts/lint.sh`), guarded behind a `command -v cargo-deny` check so it degrades to a warning if not installed locally.
- Bump `rustls` 0.23.43 -> 0.23.45 to resolve RUSTSEC-2026-0285 (TLS 1.3 handshake message validation bug), surfaced by the new advisory check and not covered by the old ignore list.

## Test Plan

- `cargo deny check` (advisories, bans, licenses, sources) passes cleanly against the current dependency tree.
- `./scripts/lint.sh` run locally to confirm the new cargo-deny step executes.

## Documentation

No user-facing docs need updating. Internal contributor docs mentioning `cargo audit` (if any exist outside this repo, e.g. onboarding wikis) should be updated to reference `cargo deny check`.

## Backwards compatibility

None. This only affects CI/local tooling for dependency auditing, not any consensus- or protocol-level behavior. No `ConsensusVersion` gating is needed.


.circleci/config.yml
```diff
@@ -820,7 +818,7 @@ workflows:
       - check-fmt
       - check-clippy
       - check-unused-dependencies
-      - check-cargo-audit
```
cbeck88, line 823: this creates a lot of red builds that people just merge over -- i think it is running much too frequently right now. can we make it so that it onyl runs nightly or something?

also turning on dependabot would make it easier to actually fix the problems when some random dependency is yanked

cbeck88: otherwise lgtm

vicsn: Dependabot was turned on against canary/testnet/mainnet branches last week for security issues (not minor warnings).

I agree we can leave cargo deny also just as a prerelease and not a PR or merge workflow

brunoffranca: I disagree. Most times you just run `cargo update -p` and it's done. On rare cases, the dependency itself has no security updates for some time or the license is incompatible. Only finding about those cases on pre-release means some PRs end up being reopen.
I agree that people just merge over it, but that's mostly because they don't want to wait for the entire CI again. I did add cargo deny to the pre-commit hook, I believe it's a better solution. The commit fails right away, you update the dependency, it's no more work than rustfmt or clippy.

cbeck88: okay fine with me

cbeck88 (dismissed):

.circleci/config.yml
cbeck88: I think the reason I don't like this so much is

* rusftmt and clippy are warnings about your code that you are working on right now.
* cargo deny is usually not something about your code. it has nothing to do with what you are working on right now.

it's like, e.g. the author of `crypto-digest` religiously yanks old patch versions whenever they publish a new one, even if its semver compatible and there was no real problem. or the problem only affects an obscure architecture that we don't build for anyways. maybe its trivial to update, but now every engineer who is working on snarkOS at this moment has to context switch and evaluate that in order to commit. 

sometimes, upgrading is not trivial and you have to actually spend time evaluating what changed and why they decided to break their API. but not every single developer working on snarkOS needs to stop what they are doing and evaluate this in parallel (which is what commit hook would do.)

my thinking is that usually, the thing that was yanked isn't actually that important, and usually the old version will work fine. the most efficient thing for the team is if we just run cargo deny once before we plan a release, then fix any problems quickly on staging, then branch for release.

that's just my opinion, willing to be wrong and willing to try it in the pre-commit hook

brunoffranca: You can commit without running the pre-commit hook (`git commit --no-verify`, or `w` in lazygit). I do it all the time when I don't want to bother to fix lint errors and just want to do it later. We can also move this to a pre-push hook instead of pre-commit.

Anyway, it seems there's no objection to the change to cargo deny, the main point is about when to run it, which already applied to cargo audit. I see 3 options:
1. We have it in the pre-commit hook, which is this PR.
2. We run it only in the PR CI, which is the status quo.
3. We run it on a pre-push hook. I think this is strictly better than number 2. It's quite annoying to push something to realize only after the CI runs that you forgot to run cargo audit.
4. We run it only on release CI.

@vicsn You're the owner of the snarkOS/VM repos, pick one of the options so we avoid bike-shedding on this.

vicsn: I have the same hesitations as Chris, but happy to try it out. In the land of AI I'm never manually making commits anymore anyway.

vicsn (dismissed):

vicsn (approved):

PR #4471, opened 2026-09-22
feat(rest): serve GET /blocks from the block cache

vicsn:
## Motivation

`GET /<network>/blocks?start={start_height}&end={end_height}` reloaded and reserialized every block in the range. A block already serialized for `GET /block/{height}` or `GET /block/latest` sits in `block_cache` and was ignored.

The range now loads hashes through the same lookup as `get_block_hashes`, serves any hash already in `block_cache`, and reads, serializes, and caches only the misses. The JSON array matches a pretty-printed `Vec` of those blocks, so a cache hit and a cache miss return the same body.

## Test Plan

- [x] `cargo test -p snarkos-node-rest --lib -- pretty_json blocks_ block_hashes block_transactions an_empty_range a_range_past`
- [x] `cargo +nightly-2026-04-02 fmt`
- [ ] Confirm a warm `GET /blocks` range on a node that has already served those heights via `GET /block/{height}` skips the ledger read for the cached hashes

## Documentation

No docs change. The route, query parameters, and response JSON are unchanged.

## Backwards compatibility

No consensus change, and nothing needs a `ConsensusVersion` guard. Callers receive the same pretty-printed JSON array as before.


node/rest/src/routes.rs
```diff
@@ -337,32 +337,93 @@ impl<N: Network, C: ConsensusStorage<N>, R: Routing<N>> Rest<N, C, R> {
...
+                            rest.ledger.try_get_block_by_hash(&hash).map_err(map_missing_resource_error)?.ok_or_else(
+                                || RestError::not_found(anyhow!("Block {height} does not exist in storage")),
+                            )?;
+                        Ok((index, hash, ErasedJson::pretty(block)))
+                    })
+                    .collect::<Result<Vec<_>, RestError>>()?;
+
+                let mut cache = rest.block_cache.lock();
+                for (index, hash, json_block) in loaded {
+                    cache.put(hash, json_block.clone());
```
cbeck88: i wonder if it's a good idea to allow anyone hitting the rest API to add things to the cache and evict from the cache?

also, imagine i'm calling GET /blocks in a loop and trying to get a large range by making a series of small requests. any block that i've already requested, is not likely to be requested again soon, because i don't need it, so putting it in the cache is counterproductive.

cbeck88: i agree with pulling from the cache first

cbeck88: the block cache seems most important for consensus / bft participants where, if one of my peers requests a block from me, it's safe to assume the others will.

so maybe REST api should only read the BFT block cache but not write it?

cbeck88: also there's an argument that we shouldn't do this at all because it gives a way for REST to block BFT

if no one is holding the lock for a long time then it's probably fine?

cbeck88 (changes requested):
i question the wisdom of putting blocks in the cache in response to REST queries -- it seems like the block cache is most important for BFT and so only requests coming from BFT should be allowed to put things in the cache.

willing to be persuaded otherwise

vicsn:
You're right, this is a bad design.

For BFT we already have a `block_cache` in snarkVM of just the most recent blocks. 

This REST `block_cache` caches the JSON serialized version. Pushing a change so we only write the most recent blocks to it.
 
Separately, we're working on improvements to our indexing pipeline where a slipstream plugin (recently removed) or [rocksdb](https://github.com/ProvableHQ/snarkVM/pull/3002) sidecar can stream data to subscribers.

